PR #7, opened 2019-10-21
Fuzz capability sets

ekse:
This PR fixes a panic where we would attempt to initialize a vector with a negative size when the provided length was too small.

It also adds a fuzzer for capability sets.


vaffeine:
Looks good. Thank you for fixing it.

Issue #14, opened 2020-01-14
Unbounded allocation in ProductInfo::from_buffer


ekse:
fuzz_pdu panics due to a very large allocation in ProductInfo::from_buffer. `company_name_len` and `product_id_len` are read from the server and used directly as an allocation size.

```rust
        let company_name_len = stream.read_u32::<LittleEndian>()?;
        if company_name_len < 2 {
            return Err(ServerLicenseError::InvalidProductInfoStringField);
        }

        let mut company_name = vec![0u8; company_name_len as usize];
```

Stacktrace

```
==17238== ERROR: libFuzzer: out-of-memory (malloc(3239510016))

 #14 0x561b4aef201c in alloc::vec::from_elem::h68c60dd992591487 /rustc/85976442558bf2d09cec3aa49c9c9ba86fb15c1f/src/liballoc/vec.rs:1722:4
    #15 0x561b4adb252c in _$LT$ironrdp..rdp..server_license..server_license_request..ProductInfo$u20$as$u20$ironrdp..PduParsing$GT$::from_buffer::heb80a67c194ab28f /home/sduquette/Devolutions/Wayk/IronRDP/ironrdp/fuzz/<::alloc::macros::vec macros>:1:33
    #16 0x561b4adb82b2 in _$LT$ironrdp..rdp..server_license..server_license_request..ServerLicenseRequest$u20$as$u20$ironrdp..PduParsing$GT$::from_buffer::hab00f9ddb1418f5d /home/sduquette/Devolutions/Wayk/IronRDP/ironrdp/src/rdp/server_license/server_license_request.rs:159:27
    #17 0x561b4acf02ea in rust_fuzzer_test_input /home/sduquette/Devolutions/Wayk/IronRDP/ironrdp/fuzz/fuzz_targets/fuzz_pdu.rs:25:12
```

ekse:
We should size limits on `company_name` and `product_id`.

ekse:
There is another unbounded allocation in ServerSecurityData::from_buffer.

https://github.com/Devolutions/IronRDP/blob/master/ironrdp/src/gcc/security_data.rs#L106

ekse:
Another unbounded allocation in ServerLicenseRequest::from_buffer.

https://github.com/Devolutions/IronRDP/blob/master/ironrdp/src/rdp/server_license/server_license_request.rs#L179

ekse:
Another one in MonitorLayoutPdu::from_buffer.

[...]

PR #17, opened 2020-01-16
Fix fuzzers

Fotosmile:
Added allocations bounds according to the fuzzers results (fixed issue #14 )


ironrdp/src/gcc/security_data.rs
```diff
@@ -95,11 +95,17 @@ impl PduParsing for ServerSecurityData {
             } else {
                 let server_random_len = buffer.read_u32::<LittleEndian>()?;
                 if server_random_len != SERVER_RANDOM_LEN as u32 {
-                    return Err(SecurityDataError::InvalidServerRandomLen);
+                    return Err(SecurityDataError::InvalidServerRandomLen(server_random_len));
                 }
 
                 let server_cert_len = buffer.read_u32::<LittleEndian>()?;
 
+                if server_cert_len > 1024 {
```
ekse: nit: define constants for these values

Fotosmile: Thank you! Fixed.

PR #29, opened 2020-08-31
Preconnection PDU parsing improvements

pacmancoder:
This PR introduces a few improvements to the Preconnection PDU parsing, which were made in the scope of Devolutions Jet JWT routing feature implementation.

Changes:
- Changed InvalidDataLength errors to io::Error(ErrorKind::Eof,...) where more data is required to parse the packet. Thiss will make Preconnection PDU parsing error handling consistent with other PDU parsers. Without this change, it is hard to tell if PDU parsing is actually failed or buffer just had not enough data
- Moved around preconnection PDU header check code to eliminate case, when we could read first 4 bytes of the packet and immediately use it as required PDU data buffer size to allocate, which can cause some nasty out-of-memory errors

### DIFF
--- a/ironrdp/Cargo.toml
+++ b/ironrdp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ironrdp"
-version = "0.4.0"
+version = "0.4.1"
 edition = "2018"
 readme = "../README.md"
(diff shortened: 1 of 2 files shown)


awakecoding (approved):
LGTM

PR #31, opened 2020-09-11
Fix decoding for invalid SystemDate

bcortier-devolutions:
As specified in [RDP protocol doc](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/526ed635-d7a9-4d3c-bbe1-4e3fb17585f4)
we should be able to ignore invalid date encoding.

### DIFF
--- a/ironrdp/src/rdp/client_info.rs
+++ b/ironrdp/src/rdp/client_info.rs
@@ -356,10 +356,10 @@ impl PduParsing for ExtendedClientOptionalInfo {
 pub struct TimezoneInfo {
     pub bias: u32,
     pub standard_name: String,
-    pub standard_date: SystemTime,
+    pub standard_date: Option<SystemTime>,
     pub standard_bias: u32,
     pub daylight_name: String,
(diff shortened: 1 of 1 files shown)


ironrdp/src/rdp/client_info.rs
```diff
@@ -430,52 +430,74 @@ impl PduParsing for TimezoneInfo {
...
 #[derive(Debug, Clone, PartialEq)]
-pub struct SystemTime {
-    pub month: Month,
-    pub day_of_week: DayOfWeek,
-    pub day: DayOfWeekOccurrence,
-    pub hour: u16,
-    pub minute: u16,
-    pub second: u16,
-    pub milliseconds: u16,
+pub enum SystemTime {
```
awakecoding, line 433: wouldn't it be better to use Option<SystemTime>? it looks a bit odd to have an enum that contains a subenum or whatever this is called, one of which is valid and the other being invalid.

PR #32, opened 2020-10-19
Jet-related IronRDP bugfix

pacmancoder:
This PR fixes the following issues:
- Fixed incorrect FastPath packets fixing (Fixes FreeRDP connection issues with /gfx enabled)
- Fixed issue with IronRDP crash when connecting  from Win7 (caused by Invalid connection request buffer size calculation logic)
- Fixed ironrdp_client connection issues when connecting to Windows Server 2019
- Applied cargo fmt & clippy

Fixes: WAYK-2183


awakecoding (approved):
LGTM

Issue #34, opened 2022-01-11
[ERROR] Fast-Path error: Bitmap error: Invalid first row size, must be 0


dheater:
# Reproduction
## Setup

Deploy a Windows server 2022 machine in OpenStack and verify that you can establish an RDP connection with a known good RDP client.

Build and run latest IronRDP client from the command line.

Connect to the server using the IronRDP client

## Run

Run IronRDP from the command line like this:

cargo run – 192.168.10.10:3389 --username username --password ********

The connection will terminate

[ERROR] Fast-Path error: Bitmap error: Invalid first row size, must be 0, but got: 17197

You should find an error similar to above at the end of `ironrdp_client.log`

dheater:
Instead of panic, seems like we should handle this error in a way that allows the client to make a best effort to continue.

Is non-zero first row now valid? (update to spec?)

awakecoding:
@dheater I agree we shouldn't panic an catch a protocol error to gracefully disconnect, but in this case, this is more likely an unrecoverable error. Make sure that the RDP server is configure for RDP 7.1 RemoteFX, otherwise you will end up getting RDP6 GDI messages, and that part of the specification is partially implemented in IronRDP. RDP 7/8 is actually a lot easier to get working than the old stuff, I would suggest you make progress without getting involved with the RDP6 GDI messages (yet).

dheater:
>  Make sure that the RDP server is configure for RDP 7.1 RemoteFX

I haven't found where I can configure the server protocol version (MS native server).  Any tips?

CBenoit:
Is the panic triggered at the IronRDP library level, or is it triggered by the IronRDP demonstration client? The latter is an issue as we don't want to library to cause the user application to abort, but the former is probably intentional (simply stop execution on unrecoverable error).

dheater:
The error is bubbling up from the library. I dropped a PR with my workaround: https://github.com/Devolutions/IronRDP/pull/35

awakecoding:
As reference, here are the similar code paths in FreeRDP. We should make sure to skip messages that we don't yet support, such that we can just throw a warning in the logs instead of dropping the connection:

https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/fastpath.c#L367
https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/fastpath.c#L264
https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/update.c#L201
[...]

PR #35, opened 2022-01-11
Slk 11764

dheater:
Propagate non-zero first row size not zero error for bitmaps from the library instead of panic'ing

Issue #34

### DIFF
--- a/ironrdp/src/basic_output/bitmap.rs
+++ b/ironrdp/src/basic_output/bitmap.rs
@@ -97,7 +97,13 @@ impl<'a> PduBufferParsing<'a> for BitmapData<'a> {
         }
 
         let compressed_data_header = if !compression_flags.contains(Compression::NOT_COMPRESSED) {
-            Some(CompressedDataHeader::from_buffer_consume(buffer)?)
+            let consumed_buffer = CompressedDataHeader::from_buffer_consume(buffer);
+            match consumed_buffer {
+                Ok(buffer) => Some(buffer),
+                Err(error) => {
+                    return Err(error);
+                },
+            }
         } else {
             None
         };
@@ -174,7 +180,10 @@ impl<'a> PduBufferParsing<'a> for CompressedDataHeader {
     fn from_buffer_consume(buffer: &mut &[u8]) -> Result<Self, Self::Error> {
         let size = buffer.read_u16::<LittleEndian>()?;
         if size != FIRST_ROW_SIZE_VALUE {
-            return Err(BitmapError::InvalidFirstRowSize(size as usize));
+            return Err(BitmapError::InvalidFirstRowSize {
+                actual: size as usize,
+                expected: FIRST_ROW_SIZE_VALUE as usize,
+            });
         }
 
         let main_body_size = buffer.read_u16::<LittleEndian>()?;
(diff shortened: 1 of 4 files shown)


ironrdp/src/basic_output/bitmap.rs
```diff
@@ -97,7 +97,13 @@ impl<'a> PduBufferParsing<'a> for BitmapData<'a> {
...
 
         let compressed_data_header = if !compression_flags.contains(Compression::NOT_COMPRESSED) {
-            Some(CompressedDataHeader::from_buffer_consume(buffer)?)
+            let consumed_buffer = CompressedDataHeader::from_buffer_consume(buffer);
+            match consumed_buffer {
+                Ok(buffer) => Some(buffer),
+                Err(error) => {
+                    return Err(error);
+                },
+            }
```
CBenoit, line 106: I think this is strictly equivalent to using the `?` operator as before. Am I missing something?

ironrdp/src/basic_output/fast_path.rs
```diff
@@ -177,8 +177,14 @@ impl<'a> FastPathUpdate<'a> {
                 Ok(Self::SurfaceCommands(commands))
             }
             UpdateCode::Bitmap => {
-                let bitmap = Bitmap::from_buffer_consume(buffer)?;
-                Ok(Self::Bitmap(bitmap))
+                let bitmap = Bitmap::from_buffer_consume(buffer);
+                match bitmap {
+                    Ok(bitmap) => Ok(Self::Bitmap(bitmap)),
+                    Err(error) => {
+                        println!("Invalid bitmap: {:?}", error);
```
CBenoit, line 184: We generally don't want libraries to `println!`

CBenoit (commented):
Thank you!

Regarding issue https://github.com/Devolutions/IronRDP/issues/34

> Instead of panic, seems like we should handle this error in a way that allows the client to make a best effort to continue.

I don't see where the code was panicking

ironrdp/src/basic_output/fast_path.rs
dheater: Agree.  I forgot that one.

CBenoit: Then I think it's good to go with
```rust
let bitmap = Bitmap::from_buffer_consume(buffer).map_err(|_| FastPathError::UnsupportedFastPathUpdate(code))?;
```
if the idea is just to override the Bitmap error :slightly_smiling_face:

ironrdp/src/basic_output/bitmap.rs
dheater: You're correct. This can just be allowed to propagate up.

dheater: Reverted changes

ironrdp/src/basic_output/fast_path.rs
dheater: Yep. `map_err` is cleaner!

Change to propagating as `FastPathError::BitmapError` and handling it in the client.

CBenoit (approved):
This looks good to me! (@awakecoding if you have any comment, otherwise it's good to merge)

PR #39, opened 2022-04-15
ARC-49: use new type for RDP version instead of enum

CBenoit:
We don't actually need strictness of enums for this.
In fact, in this case, a new type pattern is more robust
because we are trying to allow as much as we can.

This prevents issues such as the one described in this PR:
https://github.com/Devolutions/IronRDP/pull/30


awakecoding (approved):
LGTM

PR #42, opened 2022-05-27
Update sspi-rs version and pass SPN to sspi

RRRadicalEdward:


ironrdp_client/src/connection_sequence.rs
```diff
@@ -129,14 +130,21 @@ pub fn process_cred_ssp(
...
     credentials: sspi::AuthIdentity,
     server_public_key: Vec<u8>,
+    routing_addr: &SocketAddr,
 ) -> Result<(), RdpError> {
     let mut transport = TsRequestTransport::default();
 
+    let destination_host = lookup_addr(&routing_addr.ip())
+        .map_err(|err| RdpError::UserInfoError(format!("unable to query destination host name: {:?}", err)))?;
+    // sspi-rs expects null-terminated string
+    let service_principal_name = format!("TERMSRV/{}\0", destination_host);
```
CBenoit, line 140: Maybe we should consider a more "Rusty API" here. For example, if the string is not null-terminated, just push a `\0` at the end in the internal code of sspi-rs so that you don't need to care about that in user code :slightly_smiling_face:

RRRadicalEdward: Make sense. I don't mind fixing this, but I need to issue a testing request for QA. Wouldn't you mind if I fix this issue later alongside with fixing bugs QA will find?

CBenoit: To elaborate further, since it is correct for Rust strings to have any number of null terminator characters (`\0`) at any point and it is not treated specially at all including at the end position:
- `CString` / `&CStr` should be used instead of `String` / `&str` when C semantic applies
- If a Rust string can be used for most operations, then you can use it for convenience and use `CString` at a later point when appropriate
- I'm okay with doing more allocations because of this
- Above, I said to push the null terminator at the end of a `String` in required, but given the previous points I don't think it is the best option. You may consider patterns such as `"abc\0"`, `s.push('\0')`, `s.len() - 1` as code smells (like in "error prone").

Using the appropriate type at the appropriate place will help ensure correctness and express intents more clearly :slightly_smiling_face:

CBenoit (approved):
LGTM

PR #44, opened 2022-10-06
Migrating to clap 4.0

msabansal:
Migrated the client app to use clap 4 which enables some code cleanup.


msabansal:
Don't remember why I did this :) but was likely due to some experimentation i was doing a few months back.

msabansal:
This might be a minor breaking change as by default some options are renamed replacing _ with -.  For example: hybrid_ex -> hybrid-ex

CBenoit (approved):
Thank you! This looks good to me 💯

PR #47, opened 2022-10-26
Adding some missing message parser

msabansal:
Added some parsers to parse graphic messages from the gfx channel. In addition added some support to send fastpath input messages.


awakecoding (approved):
LGTM, thanks!

PR #48, opened 2022-11-03
Adding helper method for async

msabansal:
This is part 2 of a 3 part PR with the intent to add a minimal GUI on top of IronRDP. 

Soliciting feedback if this is the right approach for moving ahead. The problem with IronRDP in the current state is that the connection stream is synchronous. This makes it hard to send events from client side while waiting for a message from the RDP server. In the prototype i created for the interactive client i had a hack [here](https://github.com/msabansal/IronRDP/blob/gui/ironrdp_client/src/utils/cloneable_stream.rs) which would allow for the data to be send while waiting for messages. The proper way to do this in my opinion is to make the client asynchronous. 

This PR contains the minimal set of changes to enable that. The basic premise being we asynchronously wait for a full X224 frame to arrive before processing the frame synchronously.

### DIFF
--- a/ironrdp_client/src/main.rs
+++ b/ironrdp_client/src/main.rs
@@ -1,12 +1,10 @@
 mod config;
 
-use std::io::{self, Write};
-use std::net::TcpStream;
+use std::io::{self};
 
 use ironrdp_client::{process_active_stage, process_connection_sequence, RdpError, UpgradedStream};
 use log::error;
-#[cfg(all(feature = "native-tls", not(feature = "rustls")))]
-use native_tls::{HandshakeError, TlsConnector};
+use tokio::{io::AsyncWriteExt, net::TcpStream};
 use x509_parser::prelude::{FromDer, X509Certificate};
 
 use self::config::Config;
@@ -35,11 +33,12 @@ mod danger {
     }
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
     let config = Config::parse_args();
     setup_logging(config.log_file.as_str()).expect("failed to initialize logging");
 
-    let exit_code = match run(config) {
+    let exit_code = match run(config).await {
         Ok(_) => {
             println!("RDP successfully finished");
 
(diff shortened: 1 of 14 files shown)


awakecoding:
@CBenoit can you take a look and provide your comments?

CBenoit (commented):
Hi!

Facilitating wrapping into futures the way you did is a good thing in my opinion. I think we can even consider providing stuff like Encoder/Decoder implementations in a separate crate. The way I see it, we should be able to have a fully "synchronous" codebase for all the low level stuff while having an easy way to work with it in an async codebase if necessary.

`tungstenite` is a good example of this:
- https://github.com/snapview/tungstenite-rs
- https://github.com/snapview/tokio-tungstenite

I was thinking that maybe we could provide an async-agnostic state machine to share some logic between async and sync, but I’m not sure yet what’s the best way for that.
FYI, our company is holding a hackaton for a few days next week and I plan on looking a bit into all that. 🙂

msabansal:
Thanks for getting back on this. I had multiple takes on this trying to figure out the simplest path forward. Moving to a resumable state machine is the best option if we can get there. Looking forward to your take on this.

CBenoit:
Hi @msabansal! I could use some of the improvements you made in this PR during the hackaton. Would you be comfortable with me merging this PR as-is? Clean up could be done in a separate PR 🙂

ironrdp_client/src/main.rs
```diff
@@ -82,54 +81,48 @@ fn setup_logging(log_file: &str) -> Result<(), fern::InitError> {
...
     };
 
     #[cfg(feature = "rustls")]
     let mut tls_stream = {
-        let client_config = rustls::client::ClientConfig::builder()
+        let mut client_config = rustls::client::ClientConfig::builder()
             .with_safe_defaults()
             .with_custom_certificate_verifier(std::sync::Arc::new(danger::NoCertificateVerification))
             .with_no_client_auth();
+        client_config.key_log = std::sync::Arc::new(rustls::KeyLogFile::new());
```
msabansal, line 118: Wanted to highlight this particular change. It makes packet inspection with wireshark extremely easy. We can set up the SSLKEYLOGFILE environment variable to have rustls write to this file. After that we have to point wireshark to the keylog file and it can decrypt the RDP packets. It makes debugging RDP issues easier.

CBenoit: Neat!
FYI @awakecoding

CBenoit (approved):
Thank you!

PR #52, opened 2022-11-16
IronRDP GUI hackaton

CBenoit:
cc @ngirot-devolutions @zachariaellaham @msabansal

# IronRDP client hackaton report (2022-11)

## Demonstration

https://user-images.githubusercontent.com/3809077/202049929-76f42471-aeb0-41da-9118-0dc6ea491bd2.mp4

## Work done

- Re-usable components for both a native and a web-based client were built.

- A native IronRDP client was built using tauri (frontend is web-based using `Angular` and `Material` frameworks).

- We had to use a WebSocket instead of relying on Tauri’s JSON-RPC protocol for backend-frontend communication.
  This is a stopgap until Tauri provides a more efficient RPC protocol.

- The `ironrdp_client` crate was broken down and simplified into two crates: `ironrdp-session` and `ironrdp-cli`.
  The first should be a portable component providing higher-level abstraction to RDP protocol (async state machine, etc).

- `ironrdp-session` code was changed to use `futures-io`’s `AsyncRead` and `AsyncWrite` IO traits. These traits are more portable
  than `tokio`’s ones. Indeed, `tokio` do not provide its IO traits outside of the tokio crate itself. This is an issue
  because we don’t want to bring the `tokio` runtime in our dependencies when targeting a web browser.
  The implementation is less efficient with `futures-io`’s traits (additional copies are incurred), but more portable.
  Again, this is a stopgap that we can revisit later.

- Work to make the RDP client run in the web browser was started.
  This however requires further work: either WASM-based TLS (double TLS approach) or Devolutions Gateway extension.

## Future work

- Remove hard dependency on the hostname crate in sspi crate (obviously not WASM-compatible)

- Remove hard dependency on reqwest in sspi crate (should be fine at first, reqwest appears to be WASM-compatible)

- Abstract further `ironrdp-session` by using a non-async state machine (so we don’t need to rely on `futures-io` traits on all clients)

- Remove requirement for WASM-based TLS by pushing forward the Devolutions Gateway extension


CBenoit:
@msabansal as you can see, this is starting to get in shape! :)

With this PR, we now have a minimal web-based GUI for IronRDP using Tauri. However, you are welcomed to contribute your `glow`-based front-end as an alternative. The rationale for choosing Tauri is to share the front end code between the native application and the future in-browser application that we plan to build soon.

awakecoding (approved):
Looks very good to me!

msabansal:
This is amazing work. Congratulations to everyone involved in the hackathon 👏 

Truly amazing work. Made my day.

msabansal:
@CBenoit I am unable to build the tauri client. Can you please take a look. One of the issues is the dependency on @Devolutions/icons package. Other than that some top level build instructions will be super helpful.

CBenoit:
Oh, apparently we used private NPM packages.
Pinging @ngirot-devolutions

ngirot-devolutions:
Oops! Sorry, I'll try to check it out tonight!

PR #55, opened 2022-12-07
Added support for a Glutin based GUI client.

msabansal:
This PR adds a glutin based GUI client to ironrdp. 

1. The code is split into modules and is additive in nature. All the code is hidden behind the gui feature flag. 
2. An additional utility is provided to dump and replay the gfx pipeline data. This helps in fixing rendering issues without connecting to an RDP server.
3. Some test data and a sample script is added to the repo for quick regression testing in case a change is made to the UI 

Known limitations:
- [ ] Keyboard and mouse input work but the support is very rudimentary and needs lots of improvements.
- [ ] Multi monitor support needs improvement.
- [ ] Drag and drop across monitors of different DPI causes issues.
- [ ] Unfortunately a lot of unsafe use because of how glutin works.


msabansal:
@CBenoit @awakecoding Read for review. PTAL. I can add more details if needed.

CBenoit (commented):
Thank you @msabansal!

I have a few requests:
- Do you think you could extract all the GUI stuff from `ironrdp-cli` (which is intended to be a command-line application) and put that in a separate crate? I suggest something like `ironrdp-client`. For now, feel fry to duplicate code from `ironrdp-cli` if necessary (I already have some pieces to refactor with the Tauri client).
- For uniformity, do you think you could rename `replay_client` into `ironrdp-replay-client`?
- Move `scripts/runtests.ps1` to the replay client folder since those are used together.
- Move the sample data into the replay client folder.
- I understand the crate `ui_core` is common glutin code for the replay client and the actual client. What about renaming it into `ironrdp-glutin`?

Question: in fact, should we have a single new crate: `ironrdp-client` also containing the replay client?

Out of interest, what is the motivation to use `glutin` instead of something like [wgpu](https://github.com/gfx-rs/wgpu)?

Also, it looks to me that some rendering work is duplicated. A complete image is still being rendered in memory (see `image` in `process_active_stage`). Maybe a `GfxHandler` could be used instead. However, I feel like message passing should not be required in order to retrieve the result in a sequential way. I’m not sure yet what’s a good approach here though.

msabansal:
> Thank you @msabansal!
> 
> I have a few requests:
> 
> * Do you think you could extract all the GUI stuff from `ironrdp-cli` (which is intended to be a command-line application) and put that in a separate crate? I suggest something like `ironrdp-client`. For now, feel fry to duplicate code from `ironrdp-cli` if necessary (I already have some pieces to refactor with the Tauri client).
> * For uniformity, do you think you could rename `replay_client` into `ironrdp-replay-client`?
> * Move `scripts/runtests.ps1` to the replay client folder since those are used together.
> * Move the sample data into the replay client folder.
> * I understand the crate `ui_core` is common glutin code for the replay client and the actual client. What about renaming it into `ironrdp-glutin`?
> 
> Question: in fact, should we have a single new crate: `ironrdp-client` also containing the replay client?
> 
> Out of interest, what is the motivation to use `glutin` instead of something like [wgpu](https://github.com/gfx-rs/wgpu)?
> 
> Also, it looks to me that some rendering work is duplicated. A complete image is still being rendered in memory (see `image` in `process_active_stage`). Maybe a `GfxHandler` could be used instead. However, I feel like message passing should not be required in order to retrieve the result in a sequential way. I’m not sure yet what’s a good approach here though.

Good points. Will incorporate this feedback.

Regarding glutin and wgpu all this work was done in January 2022 and i dont remember the exact context. I might have built some prototype at the time and didnt find it mature enough. Dont remember exactly.

ironrdp-session/src/active_session/x224.rs
```diff
@@ -16,8 +16,14 @@ use crate::transport::{
...
-const RDP8_GRAPHICS_PIPELINE_NAME: &str = "Microsoft::Windows::RDS::Graphics";
-const RDP8_DISPLAY_PIPELINE_NAME: &str = "Microsoft::Windows::RDS::DisplayControl";
+pub use self::gfx::GfxHandler;
+
+pub const RDP8_GRAPHICS_PIPELINE_NAME: &str = "Microsoft::Windows::RDS::Graphics";
+pub const RDP8_DISPLAY_PIPELINE_NAME: &str = "Microsoft::Windows::RDS::DisplayControl";
+
+pub trait DvcHandlerFactory {
+    fn graphics_handler(&self) -> Option<Box<dyn GfxHandler + Send>>;
+}
```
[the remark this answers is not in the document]
msabansal: I debated about removing the factory and adding the graphics handler directly. I wanted this to be more extensible initially but then i wanted to pause investment and let the design evolve in future when the need arises.
I can change it to just the handler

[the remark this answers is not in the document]
CBenoit: Yes! :slightly_smiling_face: 
I would prefer keeping a simpler API at first (handler only) and make it evolve if necessary later

CBenoit:
> Regarding glutin and wgpu all this work was done in January 2022 and i dont remember the exact context. I might have built some prototype at the time and didnt find it mature enough. Dont remember exactly.

I understand, no problem

> > Question: in fact, should we have a single new crate: `ironrdp-client` also containing the replay client?
> > Out of interest, what is the motivation to use `glutin` instead of something like [wgpu](https://github.com/gfx-rs/wgpu)?

I thought about this overnight and I lean in favor of keeping the crates separated for now, what do you think?

msabansal:
> > Regarding glutin and wgpu all this work was done in January 2022 and i dont remember the exact context. I might have built some prototype at the time and didnt find it mature enough. Dont remember exactly.
> 
> I understand, no problem
> 
> > > Question: in fact, should we have a single new crate: `ironrdp-client` also containing the replay client?
> > > Out of interest, what is the motivation to use `glutin` instead of something like [wgpu](https://github.com/gfx-rs/wgpu)?
> 
> I thought about this overnight and I lean in favor of keeping the crates separated for now, what do you think?

I have no major concerns with that. Keeping them separate has its benefits. It can always be refactored later on depending on how things evolve.

ui_core/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod draw;
+pub type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
+pub type Result<T> = std::result::Result<T, Error>;
+pub mod renderer;
+mod surface;
```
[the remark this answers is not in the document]
msabansal: fixed

ironrdp-session/src/active_session/x224.rs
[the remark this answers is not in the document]
msabansal: accepted. Fixed

Cargo.toml
```diff
@@ -2,5 +2,8 @@
 members = [
     "ironrdp",
     "ironrdp-session",
-    "ironrdp-cli"
+    "ironrdp-ui-client",
```
CBenoit, line 5: What about naming it `ironrdp-client`? I feel that `-ui` is a bit too much (I would expect a GUI by default from an RDP client)

Cargo.toml
```diff
@@ -2,5 +2,8 @@
...
+    "ironrdp-cli",
+    "ironrdp-renderer",
+    "ironrdp-replay-cli",
```
CBenoit, line 8: What about naming it `ironrdp-replay-client`? I feel the `-cli` suffix is not necessary

ironrdp-cli/src/gui/input.rs
```diff
@@ -0,0 +1,99 @@
+use std::sync::mpsc::Receiver;
```
CBenoit, line 1: This module appears to be unused now in `ironrdp-cli`. Do you think you could remove it?

ironrdp-cli/src/gui.rs
```diff
@@ -0,0 +1,154 @@
+use std::fmt::Debug;
```
CBenoit, line 1: This code appears to be now unused in `ironrdp-cli` as well.

ironrdp-session/src/lib.rs
```diff
@@ -38,4 +42,5 @@ pub struct InputConfig {
     pub global_channel_name: String,
     pub user_channel_name: String,
     pub graphics_config: Option<GraphicsConfig>,
+    pub gfx_dump_file: Option<PathBuf>,
```
CBenoit, line 45: This sounds like a downstream option that should be handled in `ironrdp-client`’s `Arg` struct directly rather than `ironrdp-session`’s `InputConfig`. What do you think?

CBenoit (commented):
Thank you! A few new comments below 🙂

Cargo.toml
msabansal: agreed.

Cargo.toml
msabansal: I think since there is a cli only client it is important to differentiate a little. No major concerns. Maybe this can be done in future if we find it necessary

ironrdp-cli/src/gui/input.rs
msabansal: i checked out master for the ironrdp-cli folder and didnt realize it would not remove existing file. Fixed

ironrdp-replay-cli/data/README.md
```diff
@@ -0,0 +1,3 @@
+Data generated using a custom parser and data file from [Here](https://github.com/microsoft/WindowsProtocolTestSuites/tree/main/TestSuites/RDP/Client/src/TestSuite/RDPEGFX/H264TestData)
```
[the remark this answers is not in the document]
msabansal: renamed to test_data folder

ironrdp-session/src/lib.rs
msabansal: Agreed. Missed this in refactoring.

ironrdp-cli/src/gui.rs
msabansal: fixed

ironrdp-ui-client/README.md
```diff
@@ -0,0 +1,8 @@
+### GUI client
```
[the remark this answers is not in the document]
msabansal: fixed

Cargo.toml
CBenoit: Okay! In that case, could you add a `g` (ironrdp-gui-client) so that it’s clear we are talking about a graphical interface?

msabansal: fixed.

CBenoit (approved):
Awesome! Thank you 🙂

PR #56, opened 2022-12-15
Change Angular client by Svelte Client

ngirot-devolutions:
We need to merge this before implementing web version. The client is already usable almost like the Angular client. 
I will make a lot of improvement in the next weeks.


CBenoit (approved):
Thank you!

PR #77, opened 2023-02-24
fix(ironrdp-session, ironrdp. ironrp-cli, ironrdp-gui-client): fix bulding with native-tls

RRRadicalEdward:
Building with `native-tls` is essential cause rustls doesn't allow to connect to Windows 7,8.1 without adding enhanced SSL cipher suites


CBenoit (approved):
Thank you!

PR #82, opened 2023-02-28
fix: do not returns an error when compression is used

CBenoit:
Sometimes the FastPathUpdatePdu header has the
COMPRESSION_USED bit flipped and it returns an error. It should not.


awakecoding (approved):

ngirot-devolutions (approved):
LGTM

PR #94, opened 2023-03-17
feat: softbuffer-based frontend

CBenoit:
Adds a portable RDP client without GPU acceleration using softbuffer and winit for windowing.

Issue: ARC-140


awakecoding (approved):
woohoo!

PR #101, opened 2023-03-29
refactor: re-organize workspace

CBenoit:
- Remove Tauri client.

  It was useful when initially prototyping the Web Component, but now it’s more of a maintenance burden. It’s notably not convenient to check in CI. Since we now have another native alternative that does not require any GPU (`ironrdp-client`), it’s probably a good time to part ways.

- Move wasm package into the workspace.

- Move Rust crates into a `crates` subfolder.

- Introduce `xtask` crate for free-form automation.


thenextman (approved):

CBenoit:
Rebased on master

Issue #119, opened 2023-05-08
RDP implementation library as a separate crate


dheater:
I would like to use `IronRDP` as a library implementing the RDP protocol without also consuming the clients.

This would enable me to implement an independent client.

dheater:
I am currently using `ironrdp = { path = "../../IronRDP" }`, but I would like to be able to publish this client and point to the IronRDP crate (once its updates are published).

CBenoit:
Hi!

> I would like to use `IronRDP` as a library implementing the RDP protocol without also consuming the clients.

Could you clarify what you mean here? IronRDP is already [broke down into numerous crates](https://github.com/Devolutions/IronRDP#architecture-work-in-progress). I intend to push this further and reduce the number of non vital dependencies from foundational crates in the future.

> I am currently using `ironrdp = { path = "../../IronRDP" }`, but I would like to be able to publish this client and point to the IronRDP crate (once its updates are published).

There is some API design work remaining before we can publish the crates, but it will happen. I’m considering publish a few candidate releases soonish. Have you a deadline or something similar so maybe I can adjust to it?

dheater:
> Could you clarify what you mean here? IronRDP is already [broke down into numerous crates](https://github.com/Devolutions/IronRDP#architecture-work-in-progress).

Ah I think I understand know from the architecture page. I think I want the `ironrdp` meta-crate.

I am writing a Tauri client (referencing the one that used to be in this repo), so I only want the "protocol" stuff and not the clients. 

I don't have any deadline. This is pretty much a "toy" project for now and I can continue referencing my local copy for development. 

Thanks!

dheater:
i think what through me off is the layout of the workspace. Specifically that `lib` (the meta-crate) is not under `crates` but the components of lib and the clients are.  I was expecting something more like this: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

CBenoit:
> i think what through me off is the layout of the workspace. Specifically that `lib` (the meta-crate) is not under `crates` but the components of lib and the clients are. I was expecting something more like this: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

I see… I agree the current layout is not the most obvious in practice. I’m actually thinking about moving everything under `crates` with the full crate name.

CBenoit:
@dheater FYI, I tried to clarify things in this PR: https://github.com/Devolutions/IronRDP/pull/123

dheater:
[...]

PR #123, opened 2023-05-09
refactor: clarify project architecture

CBenoit:
The current layout of the workspace [is not ideal](https://github.com/Devolutions/IronRDP/issues/119#issuecomment-1540165162).

I made [mistakes that other did before me](https://matklad.github.io/2021/08/22/large-rust-workspaces.html#Smaller-Tips):

> Make the root of the workspace a virtual manifest. It might
> be tempting to put the main crate into the root, but that
> pollutes the root with src/, requires passing --workspace to
> every Cargo command, and adds an exception to an otherwise
> consistent structure.

> Don’t succumb to the temptation to strip common prefix
> from folder names. If each crate is named exactly as the
> folder it lives in, navigation and renames become easier.
> Cargo.tomls of reverse dependencies mention both the folder
> and the crate name, it’s useful when they are exactly the
> same.

This PR is attempting to further clarify the architecture of the project.


ARCHITECTURE.md
```diff
@@ -0,0 +1,78 @@
...
+
+We use GitHub action and our workflows simply run `cargo xtask`.
+The expectation is that, if `cargo xtask ci` passes locally, the CI will be green as well.
+
+**Architecture Invariant**: `cargo xtask ci` and CI workflow must be logically equivalents. It must
+be the case that a successful `cargo xtask ci` run implies a successful CI workflow run and vice versa.
+
+### Testing
+
+TODO
```
pacmancoder, line 78: A few thoughts what points we should add here about testing:
- Not allow huge binary chunks to be defined directly in code files for testing (`*.rs`) and instead require them to be placed in separate test asset files and included via `include_bytes!`/`include_str!`
- Maybe we should add a link to `fuzz/README.md` ?
- Maybe suggest using the `expect_test` crate for comparing some structured data (e.g. error results or parsed PDUs) - it is both easy to create such tests and maintain them (something changed in PDU struct/enum representation - no problem, just run tests with `UPDATE_EXPECT=1` and tests will be fixed automatically, its pretty powerful tool
- Also I saw we use `rstest` for reducing code duplication when testing multiple tests cases, we could suggest here its usage when similar cases arise

CBenoit: Thank you for the suggestion! I updated ARCHITECTURE.md to include this :)

PR #124, opened 2023-05-09
ci: separate workflow for fuzzing using corpus

CBenoit:
Fetching the corpus in the main CI worflow is a problem because it will fail when the azure storage is modified during the download:

> ERROR: The specified blob does not exist.
> RequestId:76c40723-901e-0098-38b9-8215c9000000
> Time:2023-05-09T21:03:10.8992962Z
> ErrorCode:BlobNotFound

This occurs when there are multiple runs at the same time. This happens for example when two pull requests are opened. This will cause CI to go red for no good reason.

With this patch, main CI will simply run the fuzz targets without using the corpus, as a sanity check. Then, another workflow will run at 03:12 AM UTC on Sunday, Monday, and Saturday using the corpus from azure storage.

Future work: make the scheduled run fuzz longer and in parallel.

cc @pacmancoder

### DIFF
--- a/xtask/src/tasks.rs
+++ b/xtask/src/tasks.rs
@@ -4,6 +4,8 @@ use xshell::{cmd, Shell};
 use crate::section::Section;
 
 const CARGO: &str = env!("CARGO");
+const CARGO_FUZZ_VERSION: &str = "0.11.2";
+const LOCAL_CARGO_ROOT: &str = "./target/local_root/";
 const WASM_PACKAGES: &[&str] = &["ironrdp-web"];
 const FUZZ_TARGETS: &[&str] = &["pdu_decoding", "rle_decompression", "bitmap_stream"];
 
@@ -38,8 +40,6 @@ pub fn check_lints(sh: &Shell) -> anyhow::Result<()> {
 pub fn check_wasm(sh: &Shell) -> anyhow::Result<()> {
     let _s = Section::new("WASM-CHECK");
 
-    cmd!(sh, "rustup target add wasm32-unknown-unknown").run()?;
-
     for package in WASM_PACKAGES {
         println!("Check {package}");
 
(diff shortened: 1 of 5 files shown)


xtask/src/tasks.rs
```diff
@@ -121,6 +126,26 @@ pub fn fuzz_corpus_push(sh: &Shell) -> anyhow::Result<()> {
     Ok(())
 }
 
+pub fn fuzz_install(sh: &Shell) -> anyhow::Result<()> {
+    let _s = Section::new("FUZZ-INSTALL");
+
+    let cargo_fuzz_path: std::path::PathBuf = [LOCAL_CARGO_ROOT, "bin", "cargo-fuzz"].iter().collect();
```
pacmancoder, line 132: FYI: I had a similar idea for installing `typos` in the local root, and then caching compiled tool binaries, where the cache key will be the hash of a separate `xtask.toml` config file, containing the required tool versions to be installed. e.g.
```toml
[tool.typos]
version = "1.14.9"

[tool.cargo-fuzz]
version = "0.11.2"
```
xtask will search for the following:
- Search for **globally** installed tool on the machine (>= required_version)
- Search **local_root** for already built(or cached) binaries (>=required_version, additional local_root's binary version check required if developer made branch switch/rebase/etc. and `target` was not cleaned prior to xtask invocation)
- If all of the above failed - **clean** local_root's tool directory and **build** brand new version via `cargo install`

I'll prepare PR this week (maybe next) which will generalise tool installation/cache logic

pacmancoder (approved):
LGTM, I have encountered this issue a few times too, good that it will be fixed now!

xtask/src/tasks.rs
CBenoit: Sounds very good to me! Looking forward to it :+1:

awakecoding (approved):

PR #125, opened 2023-05-10
Improve Fuzz worflow

CBenoit:
- Fix initial fuzz workflow
- Extend CLI options for xtask fuzz action
- Update list of supported codecs in README.md
- Improve Fuzz workflow to run targets longer and in parallel


CBenoit:
Sample run: https://github.com/Devolutions/IronRDP/actions/runs/4939817833

awakecoding (approved):

PR #128, opened 2023-05-11
Minor fixes

ibeckermayer:
correcting small bugs after workspace restructuring


awakecoding (approved):

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -84,9 +84,6 @@ export class WasmBridgeService {
 
     connect(username: string, password: string, hostname: string, gatewayAddress: string, domain: string, authToken: string): Observable<NewSessionInfo> {
         const sessionBuilder = SessionBuilder.new();
-        sessionBuilder.gateway_address(gatewayAddress);
-        sessionBuilder.hostname(hostname);
-        sessionBuilder.domain(domain);
```
CBenoit, line 89: Oh yeah… we have tickets internally regarding some API changes. I thought this code was living in the repository of Devolutions Server and forgot to update.
This needs to be changed to:
```rust
        sessionBuilder.proxy_address(gatewayAddress);
        sessionBuilder.destination(hostname);
        sessionBuilder.server_domain(domain);
```

CBenoit (commented):
Thank you for that!

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
ibeckermayer: Got it, thanks. I'm now running into this in the console:

image

which I've traced down to being caused by this line:

https://github.com/Devolutions/IronRDP/blob/e4849a62f93552271183b808bfdc85067be9469c/crates/ironrdp-web/src/session.rs#L354-L357

Makes some sense that the browser would complain about trying to get `std::env::current_dir()`.

CBenoit: As you say! I just realized `wasm-pack` was crashing with a segfault, I was probably testing an older wasm module when I updated this code. `npm run dev-all` is not reporting any error.

CBenoit (approved):
I applied some additional fixes, rebased on master and force-pushed. It should be good to go.

PR #131, opened 2023-05-13
MCS + GCC client data block fixes, color depth fixes, RemoteFX fixes

awakecoding:


crates/ironrdp-connector/src/connection.rs
```diff
@@ -782,24 +782,22 @@ impl Sequence for ClientConnector {
...
+        _ => panic!("Unsupported color depth: {}", max_color_depth),
+    };
 
     ClientGccBlocks {
         core: ClientCoreData {
             version: RdpVersion::V5_PLUS,
             desktop_width: config.desktop_size.width,
             desktop_height: config.desktop_size.height,
-            color_depth: ColorDepth::Bpp4, // ignored because we use the optional core data below
+            color_depth: ColorDepth::Bpp8, // ignored because we use the optional core data below
```
CBenoit, line 800: Curiosity: is there any meaning to this change since it’s ignored?

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -22,11 +22,11 @@ impl Rectangle {
...
 
     pub fn width(&self) -> u16 {
-        self.right - self.left + 1
+        self.right - self.left
     }
 
     pub fn height(&self) -> u16 {
-        self.bottom - self.top + 1
+        self.bottom - self.top
     }
```
CBenoit, line 30: Did you try running the web client? IIRC, `width()` and `height()` are used when sending partial image updates, and this change may break that.

Also, my understanding is that an update rectangle such that: `left = 0`, `top = 0`, `right = 63` and `bottom = 63`, means we have a chunk of 64x64 pixels, hence the `width()` and `height()` should probably return 64, not 63.
I suspect the actual problem lies in the consumer code, the rfx module itself. I would check where `width()` and `height()` are called and see if there is occurrence of such code:
```rust
let some_rectangle = …;
let a_reconstructed_rectangle = Rectangle {
    left: some_rectangle.left,
    top: some_rectangle.top,
    right: some_rectangle.left + some_rectangle.width(), // <-- bad
    bottom: some_rectangle.top + some_rectangle.height(), // <-- bad
};
```
Seeing the change you made in the `clipping_rectangles`, that’s likely what’s happening somewhere.

crates/ironrdp-connector/src/connection.rs
awakecoding: it's defined as ignored by the spec, but even in those cases it's better to stick to the same values sent by the original client, just in case (that's what FreeRDP and mstsc send). Even if they are truly ignored, it helps removing noise when trying to compare packets for differences that should matter.

crates/ironrdp-pdu/src/geometry.rs
awakecoding: I had to fix failing unit tests as a result of my change, but I'd rather standardize our definition of rectangles to the one where right == width, bottom == height, and then do the +1, -1 when serializing/deserializing from packets that use a different definition. it appears to be working fine for me with and without RemoteFX, no black lines

CBenoit: > I had to fix failing unit tests as a result of my change, but I'd rather standardize our definition of rectangles to the one where right == width, bottom == height, and then do the +1, -1 when serializing/deserializing from packets that use a different definition.

Sound good then.

> It appears to be working fine for me with and without RemoteFX, no black lines

Sorry for being persistent, but that’s what you get when running the web client? Can you make sure you’re not running on an old build by removing the `crates/ironrdp-web/pkg` folder and building again from scratch? You’ll not get black lines but rather visual artifacts (one row and line missing from the graphical update).

CBenoit: Also, you need a _small_ graphical update: https://github.com/Devolutions/IronRDP/blob/8857c3c25ea40bd2a2f4d850f67b16b2306393c9/crates/ironrdp-web/src/image.rs#L19

CBenoit: https://github.com/Devolutions/IronRDP/blob/8857c3c25ea40bd2a2f4d850f67b16b2306393c9/crates/ironrdp-web/src/image.rs#L78

https://github.com/Devolutions/IronRDP/blob/8857c3c25ea40bd2a2f4d850f67b16b2306393c9/crates/ironrdp-web/src/image.rs#L34

https://github.com/Devolutions/IronRDP/blob/8857c3c25ea40bd2a2f4d850f67b16b2306393c9/crates/ironrdp-web/src/image.rs#L35

awakecoding: @CBenoit I only test with the command-line client, I don't have the web client set up for development

CBenoit: Okay, no problem I’ll handle that.

CBenoit:
RLE and RDP6 32bpp was broken, so, for now, I rectified the `Rectangle` struct as discussed on slack.
No black lines in sight, but there is still a small problem with RemoteFX: I notice some artifacts suggesting a off-by-one somewhere.
At least, it’s less broken. I will look into this further when I get a chance.

I updated, modified and force-pushed the commits.

CBenoit (approved):

PR #143, opened 2023-05-18
Small tweaks for the teleport integration

ibeckermayer:
Changes read_pdu to return BytesMut rather than Bytes and makes ironrdp_tls and ironrdp_tokio publicly available via feature flags.

### DIFF
--- a/crates/ironrdp/Cargo.toml
+++ b/crates/ironrdp/Cargo.toml
@@ -22,8 +22,6 @@ connector = ["dep:ironrdp-connector"]
 session = ["dep:ironrdp-session"]
 graphics = ["dep:ironrdp-graphics"]
 input = ["dep:ironrdp-input"]
-tokio = ["dep:ironrdp-tokio"]
-tls = ["dep:ironrdp-tls"]
 
 [dependencies]
 ironrdp-pdu = { workspace = true, optional = true }
@@ -31,5 +29,3 @@ ironrdp-connector = { workspace = true, optional = true }
 ironrdp-session = { workspace = true, optional = true }
 ironrdp-graphics = { workspace = true, optional = true }
 ironrdp-input = { workspace = true, optional = true }
-ironrdp-tokio = { workspace = true, optional = true }
-ironrdp-tls = { workspace = true, features = ["rustls"], optional = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1502,8 +1502,6 @@ dependencies = [
  "ironrdp-input",
  "ironrdp-pdu",
  "ironrdp-session",
- "ironrdp-tls",
- "ironrdp-tokio",
 ]
 
 [[package]]
--- a/crates/ironrdp/src/lib.rs
+++ b/crates/ironrdp/src/lib.rs
@@ -12,7 +12,3 @@ pub use ironrdp_input as input;
 pub use ironrdp_pdu as pdu;
 #[cfg(feature = "session")]
 pub use ironrdp_session as session;
-#[cfg(feature = "tls")]
-pub use ironrdp_tls as tls;
-#[cfg(feature = "tokio")]
-pub use ironrdp_tokio as tokio;


CBenoit (commented):
Changing `read_pdu` to return `BytesMut` sounds good to me!
However, I’m not sure how I feel about re-exporting `ironrdp_tokio` and `ironrdp_tls` from `ironrdp`. I suggest we only export core tier crates for now.

CBenoit:
(Oops, I thought I was clicking on "allow CI runs")

crates/ironrdp/Cargo.toml
```diff
@@ -15,17 +15,21 @@ categories.workspace = true
...
+ironrdp-tokio = { workspace = true, optional = true }
+ironrdp-tls = { workspace = true, features = ["rustls"], optional = true }
+
 [features]
 default = ["pdu", "connector", "session"]
 pdu = ["dep:ironrdp-pdu"]
 connector = ["dep:ironrdp-connector"]
 session = ["dep:ironrdp-session"]
 graphics = ["dep:ironrdp-graphics"]
 input = ["dep:ironrdp-input"]
```
CBenoit, line 33: Curiosity/Nit: is there a specific reason for moving the `[[dependencies]]` section before the `[[features]]` one?
[Most](https://github.com/snapview/tokio-tungstenite/blob/dfe10bb2a84110e221c0b60bbdd4ac9bd1a92885/Cargo.toml#L19) projects I know are keeping feature list [before](https://github.com/tokio-rs/tokio/blob/c84d0a14b189c45da8f5e963fd3a83790ec92f8e/tokio/Cargo.toml#L24), like we currently do

ibeckermayer: My thinking was that because the "dep:<crate-name>" features are being generated by the optional dependencies, it made sense for dependencies to go before features. I wasn't aware there was a common convention though, let's just stick with that.

ead3fe6060fb22df28f2867acdf25fcaebf8aa2b

ibeckermayer:
> However, I’m not sure how I feel about re-exporting ironrdp_tokio and ironrdp_tls from ironrdp. I suggest we only export core tier crates for now.

The reason I exported those is so that I could use pieces of them in the Teleport integration:

https://github.com/gravitational/teleport/blob/5c161497ada2e2b50a5588fa17386dbf067aa17f/lib/srv/desktop/rdp/rdpclient/src/lib.rs#L358-L465

That section is more or less mimicking the connection sequence you have in `ironrdp-client`

https://github.com/Devolutions/IronRDP/blob/ead3fe6060fb22df28f2867acdf25fcaebf8aa2b/crates/ironrdp-client/src/rdp.rs#L82-L119

Perhaps ideally that sequence is moved into one of the existing "core tier" crates for my reuse?

ibeckermayer:
Sure, I can do that. Incidentally, I notice something odd about this setup:

If I have

```toml
ironrdp-tokio = { git = "https://github.com/Devolutions/IronRDP" }
```

things work as expected, I can use things from `ironrdp_tokio`. However if I try to use my local repo like

```toml
ironrdp-tokio = { path = "/Users/ibeckermayer/Devolutions/IronRDP" }
```

I get the following error

```
Error: `cargo metadata` exited with an error: error: failed to get `ironrdp-tokio` as a dependency of package `rdp-client v0.1.0 (/Users/ibeckermayer/gravitational/teleport-2/lib/srv/desktop/rdp/rdpclient)`

Caused by:
  failed to load source for dependency `ironrdp-tokio`

Caused by:
  Unable to update /Users/ibeckermayer/Devolutions/IronRDP

Caused by:
  found a virtual manifest at `/Users/ibeckermayer/Devolutions/IronRDP/Cargo.toml` instead of a package manifest
```

The odd thing being, of course, that the Cargo.toml currently at https://github.com/Devolutions/IronRDP is also a ["virtual manifest" ](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace). Do you understand why these would behave differently?

I can fix it by changing the entry to

```toml
ironrdp-tokio = { path = "/Users/ibeckermayer/Devolutions/IronRDP/crates/ironrdp-tokio" }
```

I just find it odd behavior and am curious if you understand what causes the discrepancy.

CBenoit (approved):

CBenoit:
Thank you!

> I can fix it by changing the entry to
> 
> ```toml
> ironrdp-tokio = { path = "/Users/ibeckermayer/Devolutions/IronRDP/crates/ironrdp-tokio" }
> ```
> 
> I just find it odd behavior and am curious if you understand what causes the discrepancy.

Indeed, you really need to specify the path containing the package manifest and not the virtual manifest in this case. I don’t know the history behind, but it’s likely an "accidental" discrepancy. You _can_ reference the package manifest when using local paths, but you can’t when using a git _remote_ (`https://github.com/Devolutions/IronRDP(.git)` is a valid remote, but `https://github.com/Devolutions/IronRDP/tree/master/crates/ironrdp-tokio` is not). In this case, cargo have _no choice_ but to clone the whole repository and then rely on the virtual manifest to find the right package. I’m pretty sure `path` _could_ also handle virtual manifests the same way, but it’s likely that `path` predates `git` and no one did the work for that (yet?).

PR #145, opened 2023-05-19
CI improvements

CBenoit:
# Extended checks (feat Web Client)

- Check lints on Windows as well

- Check full SvelteKit-based Web Client compilation to ensure we don’t accidentally break it on master

# CI optimization

## Before

Without cache
image

With cache
image

Cache usage
image

## After grouping tests, lints and WASM check in a single "checks" job in order to share and cache and local compilation artifacts + switching to `Swatinem/rust-cache`

**Note that in this case, we also check lints on Windows as well as SvelteKit-based Web Client compilation.**

Without cache
image

With cache
image

Cache usage
image

It’s arguably slightly faster and cache usage is greatly reduced, especially given that additional checks are performed.

# Coverage report

Ported as-is from IronVNC repository.

Adds a new workflow reporting code coverage when opening a pull request.

The idea is to help find blind spots in our test suites and inform where testing efforts should go, but there is no hard requirement related to the numbers. We are allowed to merge pull requests regardless of the coverage percentage, and even if the coverage workflow fails.

Past coverage data is stored in a separate branch (cov-data). It gets updated when the workflow run against master.


awakecoding (approved):

thenextman:
Awesome improvement @CBenoit 🚀

PR #146, opened 2023-05-23
Fixed invalid YCoCg color conversion

### LINKED ISSUE #141: Cold colors when using RDP6 32bpp
RDP6 32bpp (with lossy compression):
image

RDP6 RLE 16bpp (without lossy compression):
image

I suspect an invalid bit shift somewhere

pacmancoder:
Color conversion (YCoCg -> RGB) was invalid and introduced an unneeded division by 2.
Fixed code and tests.

This mistake happened because I implemented color conversion was defined  as in RDP specification:
```
 |R|   |1   1/2   -1/2|   |Y |
 |G| = |1    0     1/2| * |Co|
 |B|   |1  -1/2   -1/2|   |Cg|
```

However, CLL bit shift logic was implemented the same way as it is in FreeRDP, which already takes division into account by reducing color shift value by 1.

Closes #141


CBenoit (commented):
Thank you, this makes sense!

However, I see you changed the `From<YCoCg>` trait implementation for `Rgb`.
I would expect this conversion to work in the general case, that is, I would expect this to implement the following conversion:
image

My understanding is that in the rdp6 bitmap module, the `YCoCg` struct, at some point, is holding values that are not _exactly_ the right YCoCg values for the resulting color. Maybe it makes sense locally, but in that case, I would suggest we keep the special conversion code in the same module, with an explicit name and maybe a comment explaining what’s happening.

What do you think?

pacmancoder:
@CBenoit I agree, local conversion code makes sense. I'll remove `struct YCoCg` from `color_conversion.rs` (It is not used anywhere else) and place the conversion code in `bitmap_stream.rs` as a function that encapsulates color loss shift and conversion.

```rust
fn shifted_ycocg_to_rgb(y: u8, co: u8, cg: u8) -> Rgb
```

I'll make that change an update PR

awakecoding:
In FreeRDP, the [YCoCg to RGB conversion function accepts a shift parameter](https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/primitives/prim_YCoCg.c#L50), such that it can do the shift at the same time as the conversion.

pacmancoder:
Updated review

PR #151, opened 2023-06-13
fix(pdu): NegoRequestData is optional

elmarco:


awakecoding:
@elmarco can you elaborate on the fix, and why checking for a minimum length of 15 specifically is the right way to handle this? there are variable-length fields before, can you explain it based on the specs? https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/18a27ef9-6f9a-4501-b000-94b1fe3c2c10

elmarco:
> @elmarco can you elaborate on the fix, and why checking for a minimum length of 15 specifically is the right way to handle this? there are variable-length fields before, can you explain it based on the specs? https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/18a27ef9-6f9a-4501-b000-94b1fe3c2c10

To me, the spec doesn't enforce one of the field routingToken or cookie. I noticed the issue while trying to decode a ConnectionRequest from FreeRDP. Fwiw, freerdp has a similar length check: https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/nego.c#L896

The 15 value comes from min(strlen("Cookie: mstshash="), strlen("Cookie: mstsc=")) + strlen(crlf) = 15, that is the minimum amount of data required to parse any of the two fields.

crates/ironrdp-pdu/src/nego.rs
```diff
@@ -105,6 +105,11 @@ impl NegoRequestData {
     }
 
     pub fn read(src: &mut ReadCursor<'_>) -> PduResult<Option<Self>> {
+        // minimum length for token is 15
+        if src.len() < 15 {
+            return Ok(None);
+        }
+
```
CBenoit, line 112: There is already a check here: https://github.com/Devolutions/IronRDP/blob/b98443b407187e9ab06bcdaf45270bd8a2443f75/crates/ironrdp-pdu/src/nego.rs#L372
I’m unsure how this is failing, but I would prefer we address it at this place instead of duplicating the check.

CBenoit (changes requested):
Hi! Thank you for the contribution!

Do you think you could isolate the packet causing the failure so we can add a test for it?

CBenoit:
(Note: I believe "Coverage" workflow will work if you rebase on current master)

crates/ironrdp-pdu/src/nego.rs
elmarco: This will return an Err NotEnoughBytes. We could eventually catch the error each time, but this is not much better.

CBenoit: Good point! What about changing it to:
```rust
        // minimum length is length of prefix + crlf
        if src.len() < prefix.len() + 2 {
            return Ok(None);
        }
```

elmarco: hmm, I guess you mean min(Cookie::PREFIX.len(), RoutingToken::PREFIX.len())...

CBenoit: At this place:
https://github.com/Devolutions/IronRDP/blob/b98443b407187e9ab06bcdaf45270bd8a2443f75/crates/ironrdp-pdu/src/nego.rs#L372
`prefix` will be either `Cookie::PREFIX` or `RoutingToken::PREFIX` depending on the caller:
https://github.com/Devolutions/IronRDP/blob/f209c9a77b834cf1f902084749e5c7316164a816/crates/ironrdp-pdu/src/nego.rs#L136
https://github.com/Devolutions/IronRDP/blob/f209c9a77b834cf1f902084749e5c7316164a816/crates/ironrdp-pdu/src/nego.rs#L155

elmarco: ok

CBenoit (approved):
Thank you!

PR #158, opened 2023-07-05
feat(blocking): initial implementation

CBenoit:
Adds a blocking I/O higher-level abstraction library: ironrdp-blocking.
It’s equivalent to ironrdp-async, but uses blocking I/O traits from standard library.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -46,6 +46,7 @@ a feature flag called `alloc` must exist to enable its use.
 Higher level libraries and binaries built on top of the core tier.
 Guidelines and constraints are relaxed to some extent.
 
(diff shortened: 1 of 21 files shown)


pacmancoder:
@CBenoit Is this review still in draft or I could start review on it?

CBenoit:
> @CBenoit Is this review still in draft or I could start review on it?

I’m waiting for @ibeckermayer’s feedback before removing the draft status, but it’s probably fine to start reviewing now as I don’t expect things to go wrong in a spectacular manner (it’s based on ironrdp-async, which we already use successfully)

pacmancoder (approved):
Looks good! 👍

PR #160, opened 2023-07-08
[DRAFT] [WIP] Cursor messages support

pacmancoder:


crates/ironrdp-pdu/src/basic_output/fast_path.rs
```diff
@@ -289,6 +333,9 @@ pub enum FastPathError {
     UnsupportedFastPathUpdate(UpdateCode),
     #[error("Invalid RDP Share Data Header: {0}")]
     InvalidShareDataHeader(String),
+    // TODO(@pacmancoder): Temporary solution
+    #[error(transparent)]
+    Pdu(#[from] PduError),
```
pacmancoder, line 338: @CBenoit A quick question:
This module is still using legacy parsing traits, so we have a few possibilities here:
- Refactor the fast path error handling and switch to `PduError` instead of the legacy module-specific `FastPathError` (I lean to this solution the most, just a bit more effort)
- Use legacy parsing traits and keep fast path module API the same
- Encapsulate `PduError` inside legacy error as I did temporarily here (I'll add `Box` around the error for good measure if we go with this plan)

What we should do in such cases? I want to hear your opinion on that

CBenoit: I’m all in refactoring away from legacy traits as we continue moving forward with new features. :+1: 

Of course, it’s also fine to continue using the legacy parsing traits for now if this is too much work all at once.

crates/ironrdp-glutin-renderer/src/surface.rs
```diff
@@ -238,12 +238,12 @@ impl Surface {
...
-                    left: 0,
-                    right: self.width - 1,
-                    bottom: self.height - 1,
-                }
+                Rectangle::new(
+                    0,
+                    0,
+                    self.width - 1,
+                    self.height - 1,
+                )
```
CBenoit, line 246: I generally prefer struct initialization syntax when possible, that is when there is no private field and no code to run when creating the object. I prefer that for readability and flexibility:
- since all fields are named, there is no need to find the function definition in order to understand which value is what, and it’s easier to spot places where the wrong value is put into the wrong field (i.e.: putting height value into width)
- it’s possible to use the struct update syntax, etc

I understand you had to change the definition of `Rectangle` and that’s why this also had to change.

(Rest in the next comment)

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -3,34 +3,118 @@ use std::io;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 
-/// An **inclusive** rectangle.
-///
-/// This struct is defined as an **inclusive** rectangle.
-/// That is, the pixel at coordinate (right, bottom) is included in the rectangle.
+pub trait RectangleImpl {
```
CBenoit, line 6: (Along the next comment, I suggest to call this trait `Rectangle`)

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -3,34 +3,118 @@ use std::io;
...
+
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Rectangle {
+pub struct BaseRectangle<I: RectangleImpl> {
     pub left: u16,
     pub top: u16,
     pub right: u16,
     pub bottom: u16,
+    _phantom_data: std::marker::PhantomData<I>,
 }
```
CBenoit, line 24: Instead of having a "BaseRectangle" with a PhantomData field, I feel we should have two structs, `InclusiveRectangle` and `ExclusiveRectangle`, both implementing the `Rectangle` trait so we can still define operations that can be applied to both in a generic way as necessity arise.

The `Rectangle` would defines some of the operations currently defined on `BaseRectangle` (`height`, `width`, …).
Other methods such as `union` may be either free standing functions taking generic `Rectangle` parameters (e.g.: `union(r1, r2)`), or defined on the `Rectangle` trait with a default implementation, or better, on a `RectangleExt` trait automatically implemented for any `Rectangle`.

CBenoit (commented):
Pre-review of the draft, about the rectangles. Tell me what you think!

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: Yep, I agree, I did not like that change too, I'll revisit that

pacmancoder: @CBenoit I updated the approach to `Rectangle`, please take a look -- what do you think about it?

CBenoit: I actually like it better! I’ll resolve this comment and add new ones for simplicity

crates/ironrdp-pdu/src/rdp/vc/dvc/gfx/graphics_messages/avc_messages.rs
```diff
@@ -6,7 +6,7 @@ use bitflags::bitflags;
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 
 use super::GraphicsMessagesError;
-use crate::geometry::Rectangle;
+use crate::geometry::InclusiveRectangle as Rectangle;
```
CBenoit, line 9: I feel we should use the actual name of the struct instead of `Rectangle`, to avoid confusion with the trait (and confusion between inclusive and exclusive rectangles also). What do you think?

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -206,10 +254,10 @@ impl<'de, I: RectangleImpl> PduDecode<'de> for BaseRectangle<I> {
...
+    use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+    use std::io;
 
-    pub type Rectangle = InclusiveRectangle;
-
-    impl Rectangle {
+    impl InclusiveRectangle {
         // TODO: clarify code related to rectangles (inclusive vs exclusive bounds, …)
         // See for instance:
         // https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/84a3d4d2-5523-4e49-9a48-33952c559485
```
CBenoit: I guess this TODO will be addressed by your changes?

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -1,161 +1,210 @@
...
 
-impl RectangleImpl for InclusiveRectangleImpl {
-    const WIDTH_VALIDATION_ERROR: &'static str = "width cannot be negative (inclusive rectangle)";
-    const HEIGHT_VALIDATION_ERROR: &'static str = "height cannot be negative (inclusive rectangle)";
+pub struct BaseRectangle {
+    left: u16,
+    top: u16,
+    right: u16,
+    bottom: u16,
+}
```
CBenoit, line 84: I understand this struct exists so blanket implementations can work with this concrete struct instead of the generic `Self`. Is that correct?
I think it’s smart. I see two potential alternatives:
- replace `to_base` and `from_base` by a `fn new(left: u16, top: u16, width: u16, height: u16)`, returning an `Option<Self>` or `Result<Self>` (e.g.: inclusive rectangles can’t have a width or height of 0, because it implies `bottom - top + 1 = 0 ⇔ bottom = top - 1 ⇒ top < bottom` which is an impossible rectangle). It’s generally possible to write a blanket function based solely on `left`, `top`, `width` and `height` (we can get an exclusive `right` bound by computing `left + width`.  From there, we can build a `Self` again using the `new` method from the `Rectangle` trait.
- seal and hide `to_base` and `from_base` along the `BaseRectangle`, ideally it should not appear to be part of the the public API as it is an implementation detail (we are mostly there). Maybe you can have a look at this and see if you come up with something better: https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust

To be honest, I don’t know what is the best here, so feel free to choose or to ignore. What do you think?

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: Yes, PR is still in draft sadly, my refactoring broke a bit more of a code than expected

crates/ironrdp-pdu/src/rdp/vc/dvc/gfx/graphics_messages/avc_messages.rs
pacmancoder: Yea, I planned to rename this, this is temporary to reduce the current PR diff

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: Oh! I didn't know that public traits could have trait bound which is not publicly exported! I like this idea, I agree that `to_base` and `from_base` in the public interface are ugly

crates/ironrdp-pdu/src/geometry.rs
CBenoit: No problem!

crates/ironrdp-pdu/src/rdp/vc/dvc/gfx/graphics_messages/avc_messages.rs
CBenoit: I understand, makes sense!

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: Well, this happens to be a rabbit hole... To properly refactor this, a huge amount of code (`rectangle_processing.rs`) should be rewritten to use generic `Rectangle` instead of concrete `InclusiveRectangle` and validated/tested properly, because currently `rectangle_processing.rs` is a huge 1600 LoC file with mixed code for both exclusive and inclusive rectangles...

I propose to leave this as is (ExplicitRectangle used as InclusiveRectangle for operations) but create an issue because It looks like is way out of scope currently :(

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: moved to `to_base` and `from_base` to sealed trait

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -1,90 +1,202 @@
...
+    pub left: u16,
+    pub top: u16,
+    pub right: u16,
+    pub bottom: u16,
+}
 
-/// An **inclusive** rectangle.
-///
-/// This struct is defined as an **inclusive** rectangle.
-/// That is, the pixel at coordinate (right, bottom) is included in the rectangle.
```
CBenoit, line 9: Let’s keep this doc comment for both `InclusiveRectangle` and `ExclusiveRectangle` (changing it a bit for `ExclusiveRectangle` since it’s exclusive)

crates/ironrdp-pdu/Cargo.toml
```diff
@@ -22,6 +22,7 @@ alloc = ["ironrdp-error/alloc"]
 
 [dependencies]
 bitflags = "2"
+const_format = "0.2"
```
CBenoit, line 25: In which way this new dependency is used inside `ironrdp-pdu`? I couldn’t find by "ctrl-f"ing the commit diff (I didn’t review the whole PR yet)
Since this is bringing a procedural macro, its use should be very compelling as it’s desirable to keep core crates build fast. Hopefully, it doesn’t appear to requires the `syn` crate depending on the feature flags we enable, so it might be acceptable depending on the use case. Ultimately I would like core crates to have no dependency on the `syn` crate, which is one of the most heavy dependency out there. This will allow `cargo` to build all the IronRDP core crates before or in parallel of `syn` even if it is used by some higher level crate.

CBenoit (commented):
Great job! It's evident that your changes are improving things and moving the codebase in the right direction. I’ll review further later today.

crates/ironrdp-pdu/src/geometry.rs
CBenoit: Sure, let’s keep all that as a future improvement! Feel free to create an issue on GitHub with the "improvement" label 👍

crates/ironrdp-pdu/Cargo.toml
pacmancoder: Ah... I just forgot to remove it, I initially planned to use it but abandoned that idea, thanks!

crates/ironrdp-pdu/src/geometry.rs
pacmancoder: Sure! Thinks, it just slipped through the cracks while I was investigating a better approach to handle rectangles!

CBenoit (approved):
Great work, LGTM!
Minor comments below.

Since I’m leaving for vacation and there is no major concern, I’m approving this PR immediately.
Feel free to address the comments and merge it yourself.
We generally squash all PRs and make sure the commit message is following the [conventional commit specification](https://www.conventionalcommits.org/en/v1.0.0/).
In this case, it would be something like "feat(pdu): cursor messages parsing". You can modify the message and body when squashing in GitHub UI.

PR #161, opened 2023-07-12
Attempt to create a blocking client using crossbeam

ibeckermayer:
whose channels can be waited on in a `select!` macro. This allows us to closely imitate the behavior of the ironrdp-client crate, but with a blocking client.

Currently this is not working, the client immediately fails like:

```
Connection error: [read frame by hint] custom error, caused by: not enough bytes
```


crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,270 @@
...
+#[derive(Debug)]
+pub enum RdpInputEvent {
+    Resize {
+        width: u16,
+        height: u16,
+    },
+    FastPath(SmallVec<[FastPathInputEvent; 2]>),
+    Close,
+    /// DevNull is a special event that is used to check if the input event channel is closed.
+    DevNull,
```
ibeckermayer, line 31: This is a hack, but I believe does the trick of replacing

https://github.com/Devolutions/IronRDP/blob/b742a3faee30f0ca7a1c259dcedbfc14af14768c/crates/ironrdp-client/src/gui.rs#L236-L238

since `crossbeam::channel::Sender`s don't have a corresponding `is_closed` method.

crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+}
+
+enum RdpControlFlow {
+    ReconnectWithNewSize { width: u16, height: u16 },
+    TerminatedGracefully,
+}
+
+type UpgradedFramed = ironrdp_blocking::Framed<native_tls::TlsStream<TcpStream>>;
```
ibeckermayer, line 85: I went with `native_tls::TlsStream` for now, surely it can also be done with `rustls`.

crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,270 @@
...
+    let server_public_key = cert
+        .tbs_certificate
+        .subject_public_key_info
+        .subject_public_key
+        .as_bytes()
+        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "subject public key BIT STRING is not aligned"))?
+        .to_owned();
+
+    Ok(server_public_key)
+}
```
ibeckermayer, line 156: This should all probably ultimately be moved to `ironrdp-tls`

crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,269 @@
...
+        PixelFormat::RgbA32,
+        connection_result.desktop_size.width,
+        connection_result.desktop_size.height,
+    );
+
+    let mut active_stage = ActiveStage::new(connection_result, None);
+
+    let (pdu_tx, pdu_rx) = crossbeam::channel::unbounded();
+    let framed = Arc::new(std::sync::Mutex::new(framed));
+    let framed_clone = framed.clone();
```
ibeckermayer, line 174: Lets framed stream be shared across threads

crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    std::thread::spawn(move || loop {
+        let frame = framed.lock().unwrap().read_pdu();
+        pdu_tx.send(frame).unwrap();
+    });
```
ibeckermayer, line 179: We can only `crossbeam::channel::select!` on a `crossbeam::channel`, so we create a new thread here which reads frames in a loop and sends them to the `pdu_rx` channel, which allows us to more or less imitate the logic in `ironrdp-client`

crates/ironrdp-client-blocking/src/rdp.rs
```diff
@@ -0,0 +1,269 @@
...
+                            .to_buffer(&mut frame)
+                            .map_err(|e| session::custom_err!("FastPathInput encode", e))?;
+
+                            framed_clone.lock().unwrap().write_all(&frame).map_err(|e| session::custom_err!("write FastPathInput PDU", e))?;
+                    }
+                    RdpInputEvent::Close => {
+                        // TODO: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/27915739-8f77-487e-9927-55008af7fd68
+                        break 'outer;
+                    }
+                    RdpInputEvent::DevNull => {}
```
ibeckermayer, line 262: Ignore `DevNull` events, these are just `is_closed` checking whether `input_event_receiver` has been dropped.

https://github.com/Devolutions/IronRDP/blob/c4ae04c73bd90e2dcb35b840dbc3ec11eeadd2bf/crates/ironrdp-client-blocking/src/gui.rs#L251-L257

crates/ironrdp-client-blocking/src/rdp.rs
ibeckermayer: This thread probably won't exit gracefully, I think it will panic once the `select!` loop below exits. That can obviously be fixed.

crates/ironrdp-client-blocking/src/rdp.rs
CBenoit: Sounds good!

crates/ironrdp-client-blocking/src/rdp.rs
CBenoit: Definitely! Will look into that later

PR #164, opened 2023-07-13
fix(connector): client should also join user channel id

mihneabuz:
According to [this](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/c2bb6f47-779c-42d4-98f3-d7afa2a4906b), during the channel connection phase, the client is supposed to also join the user channel id.
FreeRDP already does this.

This isn't the cleanest fix, but it's simple.


CBenoit (approved):
Oh, that’s right. This is looking good to me, thank you again!

PR #166, opened 2023-07-17
fix(pdu): ensure_enough before trying to parse tpkt header

mihneabuz:
Hi, I've been running into an error with `read_pdu` when it's trying to decode the tpkt header of a x224 packet but the internal buffer doesn't yet contain the 4 bytes of the header.

Adding this `ensure_enough` fixes it, and a similar check is made for fast path packets too.


CBenoit (approved):
Looks good to me! Thank you! :smile:

PR #167, opened 2023-07-25
Feature ironrdp-server

mihneabuz:
Hi, I'm working on implementing an [RDP server](https://gitlab.com/mihneabuz/qemu-display/-/tree/feature-rdp-server/qemu-rdp) using IronRDP, and I've extracted all the RDP-specific code into this crate.

My goal was to provide an idiomatic rust API for creating a simple RDP server.

So far, I have implemented the following:
- connection sequence (only supporting SSL security)
- receiving input events (both slow and fast path)
- sending bitmap display updates with RDP 6.0 compression

Right now, I am working on cleaning up the code and improving the efficiency by reducing the amount of copying, especially for server to client updates.

Would you be interested in merging this?
I'm open to any suggestions to ensure this feature meets the project's standards.


crates/ironrdp-server/README.md
```diff
@@ -0,0 +1,3 @@
+# IronRDP Server
+
+Utilities for implementing a RDP server
```
CBenoit, line 3: nit: other `README`s are not yet very elaborated either, but it would be nice to give a small overview here

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+            update_code: self.code,
+            compression_flags: None,
+            compression_type: None,
+            data,
+        };
+
+        let header = FastPathHeader {
+            flags: EncryptionFlags::empty(),
+            data_length: update.size(),
+            forced_long_length: true,
```
CBenoit, line 80: Is it really required? Maybe add a TODO or NOTE comment indicating the intention is simply to imitate the official Microsoft RDP server (I would guess?)

CBenoit (commented):
Really excited to see this happening!

Any input on the RDP6 bitmap encoder? @pacmancoder 

I’m not yet done reviewing, but this is already looking quite promising. Here are my initial comments.

EDIT: I turned the PR into a Draft since this is still WIP / preview

crates/ironrdp-server/README.md
mihneabuz: Agreed, I'll try to add some docs too :)

crates/ironrdp-server/src/encoder/mod.rs
mihneabuz: It's not, I read the code wrong and forgot about it, whoops

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+    display::{DisplayUpdate, RdpServerDisplay},
+    encoder::UpdateEncoder,
+    handler::RdpServerInputHandler,
+};
+
+pub struct RdpServer<H, D> {
+    opts: RdpServerOptions,
+    handler: H,
+    display: D,
+}
```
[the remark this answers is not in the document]
mihneabuz: I usually try to avoid dynamic dispatch if possible, especially since there should only be a single server instance in the codebase. But you're right, I haven't considered compilation time, and performance shouldn't be an issue here.

I agree, I'll box them.

mihneabuz:
A quick note on RDP6, for a lot of the encoding logic in `ironrdp-graphics` I used  `Iterator<Item = u8>` as input.
RDP specifies bitmaps are sent in bottom-to-top order, but (I think) most modern uses would have the bitmap data in top-to-bottom order. By using iterators, we can avoid copying and keep a nice API.

I'm curious if you have any other ideas here.

crates/ironrdp-server/src/server.rs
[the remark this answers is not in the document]
CBenoit: Thank you! IMO your point is absolutely valid as well: there will be only one monomorphized version, so at least there is no code size issue. But when it ends up in downstream code I prefer to be more careful… I would use static dispatch without too much hesitation if it’s an implementation detail (all monomorphized code is directly inside the same codegen unit), and of course for "small and shallow" generic code.
I take inspiration from bevy: it relies heavily on dynamic dispatching to achieve its simple(r) API and amazing iteration cycle speed
Keeping all this in mind, don’t hesitate to challenge this approach when you think it’s not appropriate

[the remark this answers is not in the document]
mihneabuz: Yeah, I never thought about the codegen unit issue until you pointed it out, but now it makes sens. 
Thanks for the tips! I haven't really worked on a lib before, this feedback is very useful to me :)

crates/ironrdp-pdu/src/basic_output/bitmap/rdp6.rs
```diff
@@ -134,6 +134,31 @@ impl<'a> PduEncode for BitmapStream<'a> {
     }
 }
 
+impl<'a> BitmapStream<'a> {
+    pub fn encode_header(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
```
pacmancoder, line 138: I think it is better to implement `PduEncode` trait here instead -- `BitmapStream` borrow their inner buffers instead of owning them, so there should be only minimal overhead when constructing `BitmapStream` with inner fields and actually encoding it via `ironrdp_pdu:;encode_cursor`. We will eventually migrate all pdu-related code to PduEncode/PduDecode`.

What do you think @mihneabuz ?

(Same comment for `BitmapUpdateData::encode_header`)

crates/ironrdp-graphics/src/rdp6/rle.rs
```diff
@@ -159,6 +166,185 @@ pub fn decompress_8bpp_plane(
...
+    fn encode_segment(&self, mut raw: &[u8], run: usize, dst: &mut WriteCursor<'_>) -> usize {
+        let mut extra_bytes = 0;
+
+        while raw.len() > 15 {
+            extra_bytes += self.encode_segment(&raw[0..15], 0, dst);
+            raw = &raw[15..];
+        }
+
+        let control = ((raw.len() as u8) << 4) + std::cmp::min(run, 15) as u8;
+        dst.write_u8(control);
```
pacmancoder, line 292: We need to ensure that the destination cursor has enough bytes here via `ensure_size`, because any `write_*` or `read_*` methods on cursor will panic if there is not enough space

crates/ironrdp-graphics/src/rdp6/rle.rs
```diff
@@ -159,6 +166,185 @@ pub fn decompress_8bpp_plane(
...
+        let mut written = 0;
+
+        while run >= 16 {
+            let current = std::cmp::min(run, MAX_DECODED_SEGMENT_SIZE) as u8;
+
+            let c_raw_bytes = std::cmp::min(current / 16, 2);
+            let n_run_length = current - c_raw_bytes * 16;
+
+            let control = (n_run_length << 4) + c_raw_bytes;
+            dst.write_u8(control);
```
pacmancoder, line 313: `ensure_size!` macro is needed

crates/ironrdp-server/src/encoder/bitmap.rs
```diff
@@ -0,0 +1,111 @@
...
+                }
+
+                PixelOrder::TopToBottom => {
+                    let bytes_per_pixel = bitmap.format.bytes_per_pixel() as usize;
+                    let pixels = chunk
+                        .chunks(row_len as usize)
+                        .rev()
+                        .flat_map(|row| row.chunks(bytes_per_pixel));
+
+                    Self::encode_iter(encoder, bitmap.format, pixels, self.buffer.as_mut_slice())
```
pacmancoder, line 54: Here we could have a problem - if the chunk size is bigger than the pre-allocated buffer, this code will throw an error (currently panic, as `ensure_size` is missing some places).

One of the possible solutions to this issue:
- We could have some method that will calculate the exact or at least the maximum amount of bytes required to encode the PDU. then we call it prior to the actual encoding step and resize the buffer accordingly.

@CBenoit what are your thoughts on this?

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+pub struct UpdateEncoder {
+    buffer: Vec<u8>,
+    bitmap: BitmapEncoder,
+}
+
+impl UpdateEncoder {
+    pub fn new() -> Self {
+        Self {
+            buffer: vec![0; 8192 * 8192],
```
pacmancoder, line 23: Same comment as in `crates/ironrdp-server/src/encoder/bitmap.rs`, we need to think of a way to make encode buffer size not hard coded.

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        Ok(())
+    }
+
+    async fn client_loop<S>(&mut self, mut framed: Framed<S>) -> Result<()>
+    where
+        S: FramedWrite + FramedRead,
+    {
+        let mut buffer = vec![0u8; 8192 * 8192];
```
pacmancoder, line 90: Same comment about hardcoded buffer size as above

pacmancoder (commented):
Great work! It is always inspiring to see such great contributions to IronRDP! 👍 

I have a few comments/suggestions:
- It is better to get rid of hardcoded buffer sizes for encoding PDUs, maybe @CBenoit will also have some alternative suggestions for handling this issue.
- Also, as I could see, `encode_header` methods could be removed in favor of `PduEncode` implementation.

crates/ironrdp-server/src/encoder/bitmap.rs
CBenoit: Agreed!

That being said, currently it looks like `buffer` is always assigned a `Vec` with a capacity for `u16::MAX` elements (initially all filled with `0`s):
https://github.com/Devolutions/IronRDP/pull/167/files/37954589ab655b8a570d8de3872f2b1f789b78ed#diff-fb555a3444a021d9c141746de60538f9d8c408b6952f069372ab3b59d957fdfbR22
Is it normal? If we could find a smaller upper-bound this would help with memory usage I imagine

CBenoit:
> A quick note on RDP6, for a lot of the encoding logic in `ironrdp-graphics` I used `Iterator<Item = u8>` as input. RDP specifies bitmaps are sent in bottom-to-top order, but (I think) most modern uses would have the bitmap data in top-to-bottom order. By using iterators, we can avoid copying and keep a nice API.
> 
> I'm curious if you have any other ideas here.

This approach is looking smart enough to me. :+1: 
I’ll let you know if I think of something else!

crates/ironrdp-server/src/encoder/bitmap.rs
mihneabuz: I am looking for a way to remove the need for this buffer entirely. If I can't find a clean way to do it, I will make the buffer resizable.

crates/ironrdp-pdu/src/basic_output/bitmap/rdp6.rs
mihneabuz: The purpose for these methods is to reduce the amount of intermediary buffers and copying.
Take this function for example: https://github.com/Devolutions/IronRDP/pull/167/files/37954589ab655b8a570d8de3872f2b1f789b78ed#diff-abb6fe01531dc5aacd020389e7c572eab4257beae3708a7910a6b4577578b1abR143
In order to use `ironrdp_pdu::encode_cursor`, I would need to have another auxiliary buffer to hold the compressed data, borrow it into the `BitmapStream`, and then copy it into the destination buffer.

With this method, I can just encode the header of the `BitmapStream` and then write the compressed data directly into the destination buffer.

I agree that the code is a little weird/ugly, but this seemed like a good tradeoff for me.
Alternatively, I could try to split these Pdus into a separate header and data parts and implement `PduEncode/PduDecode` for them.

Do you have any other ideas? @pacmancoder @CBenoit

crates/ironrdp-server/src/server.rs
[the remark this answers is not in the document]
mihneabuz: Done

crates/ironrdp-graphics/src/rdp6/rle.rs
mihneabuz: done

crates/ironrdp-graphics/src/rdp6/rle.rs
mihneabuz: done

crates/ironrdp-server/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+categories.workspace = true
+
+[lib]
+doctest = false
+test = false
+
+[dependencies]
+anyhow = "1.0"
+bytes = "1"
+rustls = "0.21"
```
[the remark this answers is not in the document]
mihneabuz: removed rustls from Cargo.toml

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+
+            unexpected => {
+                eprintln!("unexpected mcs message {:?}", ironrdp_pdu::name(&unexpected));
+            }
+        }
+
+        Ok(false)
+    }
+
+    async fn handle_input_even(&mut self, input: InputEventPdu) {
```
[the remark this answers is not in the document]
mihneabuz: Fixed

crates/ironrdp-server/src/encoder/mod.rs
mihneabuz: changed to `false`

crates/ironrdp-pdu/src/basic_output/bitmap/rdp6.rs
CBenoit: Yes I see what you mean. Indeed, you could create another structure called `BitmapStreamHeader` that is taking care of encoding / decoding only the header part and `BitmapStream` itself could be refactored to reuse this struct in its `encode` and `decode` methods. This way you don’t need to reference an empty slice `&[]`.

Also, it seems that the `ColorPlanes` enum could be changed a bit to accommodate this refactor.

Currently it’s defined as:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ColorPlanes<'a> {
    Argb {
        data: &'a [u8],
    },
    AYCoCg {
        color_loss_level: u8,
        use_chroma_subsampling: bool,
        data: &'a [u8],
    },
}
```

But we could remove the common field `data` and rename it so we get `ColorPlaneDefinition`:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ColorPlaneDefinition {
    Argb,
    AYCoCg {
        color_loss_level: u8,
        use_chroma_subsampling: bool,
    },
}
```

And ultimately get something looking like:
```rust
pub struct BitmapStreamHeader {
    pub enable_rle_compression: bool,
    pub use_alpha: bool,
    pub color_plane_definition: ColorPlaneDefinition,
}

pub struct BitmapStream<'a> {
    pub header: BitmapStreamHeader,
    pub color_planes: &'a [u8], // <- this is where we moved the `data` field of the old enum `ColorPlanes`
}
```

What do you think? @pacmancoder @CBenoit

CBenoit: Yes I see what you mean. Indeed, you could create another structure called `BitmapStreamHeader` that is taking care of encoding / decoding only the header part and `BitmapStream` itself could be refactored to reuse this struct in its `encode` and `decode` methods. This way you don’t need to reference an empty slice `&[]`.

Also, it seems that the `ColorPlanes` enum could be changed a bit to accommodate this refactor.

Currently it’s defined as:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ColorPlanes<'a> {
    Argb {
        data: &'a [u8],
    },
    AYCoCg {
        color_loss_level: u8,
        use_chroma_subsampling: bool,
        data: &'a [u8],
    },
}
```

But we could remove the common field `data` and rename it so we get `ColorPlaneDefinition`:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ColorPlaneDefinition {
    Argb,
    AYCoCg {
        color_loss_level: u8,
        use_chroma_subsampling: bool,
    },
}
```

And ultimately get something looking like:
```rust
pub struct BitmapStreamHeader {
    pub enable_rle_compression: bool,
    pub use_alpha: bool,
    pub color_plane_definition: ColorPlaneDefinition,
}

pub struct BitmapStream<'a> {
    pub header: BitmapStreamHeader,
    pub color_planes: &'a [u8], // <- this is where we moved the `data` field of the old enum `ColorPlanes`
}
```

What do you think? @pacmancoder @mihneabuz

mihneabuz: This is exactly what I had in mind!
As for [`BitmapUpdateData<'a>`](https://github.com/Devolutions/IronRDP/pull/167/files#diff-fa35c6d358b1347196b93a6d496da9869f8c68835d113b41357b1bd3d9614098R56), I think adding a header part is not ideal, as it would have to always be in sync with the len of the rectangles anyway. I can't think of a better way to do it without a separate `encode_header` method.

pacmancoder: @CBenoit Yep, I agree sounds perfect to me, I think it was an oversight on my behalf to keep `data` inside `ColorPlanes` in the first place 👍

CBenoit: > This is exactly what I had in mind! As for [`BitmapUpdateData<'a>`](https://github.com/Devolutions/IronRDP/pull/167/files#diff-fa35c6d358b1347196b93a6d496da9869f8c68835d113b41357b1bd3d9614098R56), I think adding a header part is not ideal, as it would have to always be in sync with the len of the rectangles anyway. I can't think of a better way to do it without a separate `encode_header` method.

For BitmapUpdateData, let’s keep your approach for now. It’s not that bad, and if we come up with something better we can refactor at that time.

crates/ironrdp-server/src/builder.rs
```diff
@@ -0,0 +1,128 @@
...
+    security: RdpServerSecurity,
+    handler: H,
+    display: D,
+}
+
+pub struct RdpServerBuilder<State, H, D> {
+    state: State,
+    _handler: PhantomData<H>,
+    _display: PhantomData<D>,
+}
```
CBenoit, line 32: Nice, I really like this kind of type safe builders (like the one in rustls)

crates/ironrdp-server/src/acceptor/mod.rs
```diff
@@ -0,0 +1,99 @@
+use std::io;
```
CBenoit, line 1: Just wondering: could this be in its own separate `ironrdp-acceptor` crate?

mihneabuz: Sure, I could see this as well

mihneabuz: What about the state machine driver? I noticed for the `ironrdp-connector`, that logic is extracted into `ironrdp_async::connector`.

For my case, the driver is more coupled to `tokio` (which is fine since the server will also always run on `tokio`), but that makes it a little harder to extract into `ironrdp-async` since that is runtime independent.
I will leave that logic in the acceptor crate for now, but I could refactor it a bit if you think is should belong to `ironrdp-async`.

crates/ironrdp-server/src/encoder/mod.rs
mihneabuz: For this buffer, a solution I can see is starting with a small size and doing a manual doubling of the buffer when it is not enough.
Something like this:

```rust
let len = loop {
    match self.bitmap.encode(&bitmap, self.buffer.as_mut_slice()) {
        Err(e) => match e.kind() {
            ironrdp_pdu::PduErrorKind::NotEnoughBytes { .. } => {
                self.buffer.resize(self.buffer.len() * 2, 0);
            }

            _ => return None;
        },
        Ok(len) => break len,
    }
};
```

What do you think?

crates/ironrdp-server/src/acceptor/mod.rs
CBenoit: No problem, let’s keep this in ironrdp-acceptor for now. I don’t think everything should go inside ironrdp-async, I’ll think it over more later

crates/ironrdp-server/src/server.rs
mihneabuz: For this buffer, I changed it to start with 4096 bytes and get resized accordingly when it needs to hold a larger amount. Also, the fragments of the display updates are sent one at a time, which should greatly reduce the required size of this buffer.

crates/ironrdp-server/src/encoder/bitmap.rs
mihneabuz: After looking into it some more, removing the need for this buffer would require some changes to the `BitmapData` (TS_BITMAP_DATA), splitting it into a header and data part similar to `BitmapStream`.

I would like to leave it like this for now and improve it in a later pr, if you are ok with it :)

crates/ironrdp-server/src/encoder/mod.rs
CBenoit: It’s probably good enough for now. I’m thinking in the future we could use the [`WriteBuf`](https://github.com/Devolutions/IronRDP/blob/2eedfe7b8d987d25ec3996738c69d1d0bb3d416d/crates/ironrdp-pdu/src/write_buf.rs) (it’s not yet merged on master)

crates/ironrdp-server/src/encoder/bitmap.rs
CBenoit: Sounds good to me!

CBenoit (commented):
Before we merge, do you think you could fuzz the RDP6 bitmap stream encoder implementation? For that:
- you’ll need a [test case generator](https://github.com/Devolutions/IronRDP/blob/bcc4e35ecbc74eef619d183e2262473abe99af5a/crates/ironrdp-fuzzing/src/generators/mod.rs#L12) (or maybe not, if you just need random bytes),
- along [an oracle](https://github.com/Devolutions/IronRDP/blob/bcc4e35ecbc74eef619d183e2262473abe99af5a/crates/ironrdp-fuzzing/src/oracles/mod.rs#L87) to evaluate the generated test cases, and
- a [fuzz target](https://github.com/Devolutions/IronRDP/blob/bcc4e35ecbc74eef619d183e2262473abe99af5a/fuzz/fuzz_targets/bitmap_stream.rs), gluing the test case generator and the oracle together

This architecture is useful because it’s easy to have regression tests [like this](https://github.com/Devolutions/IronRDP/blob/bcc4e35ecbc74eef619d183e2262473abe99af5a/crates/ironrdp-testsuite-core/tests/pdu/regression.rs).

CBenoit:
@mihneabuz is there anything else you wish to modify before I review?

mihneabuz:
@CBenoit no, I'm happy with it right now :)

CBenoit:
Perfect, I’ll take some time to review soon then :+1:

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,249 @@
...
+    where
+        S: FramedWrite + FramedRead,
+    {
+        let mut buffer = vec![0u8; 4096];
+        let mut encoder = UpdateEncoder::new();
+
+        debug!("starting client loop");
+
+        'main: loop {
+            select! {
```
elmarco, line 113: `select!` seems wrong here, as it will terminate one of the branch: "Polls multiple futures and streams simultaneously, executing the branch for the future that finishes first". So you may have unexpected errors when looping and continue reading the input stream for ex, or unfinished display updates. It should spawn tasks instead.

crates/ironrdp-server/src/handler.rs
```diff
@@ -0,0 +1,140 @@
+use ironrdp_pdu::input::{
+    fast_path, mouse::PointerFlags, mouse_x::PointerXFlags, scan_code, unicode, MousePdu, MouseXPdu,
+};
+
+/// Keyboard Event
+///
+/// Describes a keyboard event received from the client
+///
```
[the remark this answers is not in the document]
mihneabuz: Personally, I think enums are a more natural way to represent these than the bitflags found in the pdus. And this way we can also combine both FastPath and SlowPath events into a single type.
What do you think @CBenoit?

crates/ironrdp-server/src/server.rs
mihneabuz: A similar loop is used in the `ironrdp-client`: https://github.com/mihneabuz/IronRDP/blob/master/crates/ironrdp-client/src/rdp.rs#L135C19-L135C19

From what I understand, dropping either of the futures should have no effect if they are not yet completed.
Polling `read_pdu` will read into an internal buffer, and only return the frame when enough bytes are read. Thus, it should pick up from where it left off on the next iteration of the loop.
The display should also act as a channel/receiver, so dropping it early should not lose any items. And once an update is received, it will block the loop until it is sent completely.

Spawning tasks might be better for performance here, so we don't block the input events while we send display updates. Or just improve the display update encoding. I was hoping to do this in a follow-up pr :)

crates/ironrdp-server/src/handler.rs
```diff
@@ -0,0 +1,140 @@
...
+        if flags.contains(scan_code::KeyboardFlags::RELEASE) {
+            KeyboardEvent::Released(key)
+        } else {
+            KeyboardEvent::Pressed(key)
+        }
+    }
+}
+
+impl From<(u16, unicode::KeyboardFlags)> for KeyboardEvent {
+    fn from((key, flags): (u16, unicode::KeyboardFlags)) -> Self {
```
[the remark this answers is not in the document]
mihneabuz: Right, I will split the scan code and unicode events and also add any other events I missed.

crates/ironrdp-pdu/src/basic_output/fast_path.rs
```diff
@@ -20,7 +20,7 @@ use crate::{decode_cursor, per, PduDecode, PduEncode, PduResult};
 pub struct FastPathHeader {
     pub flags: EncryptionFlags,
     pub data_length: usize,
-    forced_long_length: bool,
+    pub forced_long_length: bool,
```
CBenoit, line 23: I would rather keep this private, as an implementation detail. I think it exist for the sole purpose of keeping the round-tripping property even when a "weird" packet is decoded, but it’s not intended to be changed otherwise. Can you confirm? @pacmancoder 
Instead, add a `new` constructor.

crates/ironrdp-server/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+categories.workspace = true
+
+[lib]
+doctest = false
+test = false
+
+[dependencies]
+anyhow = "1.0"
+bytes = "1"
+tokio = { version = "1", features = ["full"] }
```
CBenoit, line 21: Since `ironrdp-server` is used as a library, it’s better to not enable all tokio features and only enable the ones required by the crate.

crates/ironrdp-graphics/src/rdp6/rle.rs
```diff
@@ -152,13 +162,221 @@ pub fn decompress_8bpp_plane(
...
+/// Performs compression of 8bpp color plane pixel stream into a buffer.
+/// Pixel iterator must have at least width * height items.
+/// Destination slice must have enough space for the compressed data.
+///
+/// Returns number of bytes written to the dst buffer.
+pub fn compress_8bpp_plane(
+    src: impl Iterator<Item = u8>,
+    dst: &mut WriteCursor<'_>,
+    width: impl Into<usize>,
+    height: impl Into<usize>,
```
CBenoit, line 370: What about taking `usize` without generics here?

```suggestion
    width: usize,
    height: usize,
```

I’m not sure we want to hide this kind of conversion.

crates/ironrdp-acceptor/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+{
+    let mut buf = Vec::new();
+    let mut framed = TokioFramed::new(stream);
+
+    loop {
+        if let Some(security) = acceptor.reached_security_upgrade() {
+            let result = if security.is_empty() {
+                BeginResult::Continue(framed)
+            } else {
+                BeginResult::ShouldUpgrade(framed.into_inner().0)
```
CBenoit, line 64: For sanity, `.into_inner_no_leftover()` should be used instead. This function [asserts that there is no leftover bytes](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-async/src/framed.rs#L66) before proceeding further. Also, I’m not convinced that this logic should lives here. I would suggest an [API similar to the one we have for the connector](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-client/src/rdp.rs#L97-L109), so it’s easier to support an alternative to tokio in the future.

crates/ironrdp-server/src/handler.rs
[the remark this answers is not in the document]
CBenoit: I’m not sure we’ll ever have to deal with Slow-Path packets, but I do agree. We have [something similar for client side in the `ironrdp-input` crate](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-input/src/lib.rs#L129-L136) because it gives [code easier to read and write](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-client/src/gui.rs#L79-L88) (imo), while [dealing with the raw RDP PDUs is not super fun](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-input/src/lib.rs#L262-L292). What is missing in your opinion? @elmarco

crates/ironrdp-server/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+mod display;
+mod encoder;
+mod handler;
+mod server;
+
+pub use display::*;
+pub use handler::*;
+pub use server::*;
+
+pub use ironrdp_acceptor as acceptor;
```
CBenoit, line 15: Should we re-export `ironrdp_acceptor` from `ironrdp_server` or from `ironrdp`? I feel `ironrdp` should re-export it.

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,259 @@
...
+                FastPathInputEvent::MouseEvent(mouse) => {
+                    self.handler.mouse(mouse.into()).await;
+                }
+
+                FastPathInputEvent::MouseEventEx(mouse) => {
+                    self.handler.mouse(mouse.into()).await;
+                }
+
+                FastPathInputEvent::QoeEvent(quality) => {
+                    eprintln!("received QoE: {}", quality);
```
CBenoit, line 189: `warn!` or `error!`

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,259 @@
...
+            }
+
+            mcs::McsMessage::DisconnectProviderUltimatum(disconnect) => {
+                if disconnect.reason == mcs::DisconnectReason::UserRequested {
+                    return Ok(true);
+                }
+            }
+
+            unexpected => {
+                eprintln!("unexpected mcs message {:?}", ironrdp_pdu::name(&unexpected));
```
CBenoit, line 225: `warn!` or `error!`

CBenoit (commented):
Finally got through it! Let me know what you think.

crates/ironrdp-server/src/handler.rs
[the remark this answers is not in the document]
elmarco: I am afraid the server implementation may have issues with some simpler abstraction if we lack details like flags EXTENDED, or atomic DOWN|RELEASE. But there is a clear benefit of a common type for Slow/Fast events.

crates/ironrdp-graphics/src/rdp6/rle.rs
CBenoit: Oh, I didn’t notice. Not a big deal, but if you could change `decompress_8bpp_plane` the same way, I would appreciate it.

crates/ironrdp-server/src/server.rs
elmarco: `read_pdu` might be ok, but the display update write loop doesn't look ok to me.

CBenoit: Are you referring to the `self.display.get_update()` below that is used as a select branch event? Indeed, depending on the implementation, this trait method may not be cancellation safe. Documentation of `get_update` should specify that the implementer must ensure the returned future is cancellation safe. In case you are referring to the `framed.write_all(&buffer[..len]).await` in the branch itself, this is not a problem: cancellation safety is required only for the futures used as "events" of the `select!`. The branch itself is never cancelled once it is entered. The only concern here is that we will stop reading PDUs until we are done writing the update, I’m okay with waiting for a follow up PR (though this may be addressed immediately as well).

crates/ironrdp-server/src/handler.rs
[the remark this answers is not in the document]
mihneabuz: Right, I will add the EXTENDED flag as a bool similar to [Scancode](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-input/src/lib.rs#L64) from `ironrdp-input`

crates/ironrdp-server/src/display.rs
```diff
@@ -0,0 +1,65 @@
...
+///
+///     async fn get_update(&mut self) -> Option<DisplayUpdate> {
+///         self.receiver.recv().await
+///     }
+/// }
+/// ```
+#[async_trait::async_trait]
+pub trait RdpServerDisplay {
+    async fn size(&mut self) -> DesktopSize;
+    async fn get_update(&mut self) -> Option<DisplayUpdate>;
```
CBenoit, line 64: As mentioned [in another comment](https://github.com/Devolutions/IronRDP/pull/167#discussion_r1297343359), it would be preferable to document cancellation safety. Here is how you could phrase it:
```suggestion
    /// # Cancel safety
    ///
    /// This method is cancellation safe. If you use it as the event in a
    /// `tokio::select!` statement and some other branch
    /// completes first, then it is guaranteed that no data is lost.
    async fn get_update(&mut self) -> Option<DisplayUpdate>;
```

From https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#cancel-safety

crates/ironrdp-acceptor/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub async fn upgrade<S>(security: &RdpServerSecurity, stream: S) -> Result<TlsStream<S>, io::Error>
+where
+    S: Unpin + AsyncRead + AsyncWrite,
+{
+    match security {
+        RdpServerSecurity::None => unreachable!(),
+        RdpServerSecurity::SSL(tls_acceptor) => tls_acceptor.accept(stream).await,
+    }
+}
```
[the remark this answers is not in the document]
mihneabuz: Sure, I should have done this when I extracted the acceptor logic. I will make the `ironrdp-acceptor` crate run-time agnostic and move this logic into the server.

crates/ironrdp-server/src/display.rs
```diff
@@ -0,0 +1,65 @@
...
+///     }
+///
+///     async fn get_update(&mut self) -> Option<DisplayUpdate> {
+///         self.receiver.recv().await
+///     }
+/// }
+/// ```
+#[async_trait::async_trait]
+pub trait RdpServerDisplay {
+    async fn size(&mut self) -> DesktopSize;
```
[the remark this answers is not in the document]
mihneabuz: Currently, there is no way for the client or server to communicate resolution changes. The server will always request this resolution.

Right now, I don't really know how resolution changes work in the protocol or how to implement them here. (I know there is a [virtual channel extension](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/d2954508-f487-48bc-8731-39743e0854a9) for it)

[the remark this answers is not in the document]
CBenoit: I was referring to the initial handshake. During the initial handshake, the client can ask a specific resolution.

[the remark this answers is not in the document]
mihneabuz: Even if the client asks for a specific resolution, it is not passed in any way to the downstream code.

crates/ironrdp-server/src/handler.rs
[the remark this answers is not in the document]
CBenoit: > I am afraid the server implementation may have issues with some simpler abstraction if we lack details like flags EXTENDED, or atomic DOWN|RELEASE. 

Indeed. Probably we can get started with the current approach and revisit later if we stumble upon any problem because of this. Maybe at some point in the future we'll want to unify with `ironrdp-input` and make sure the abstraction is still reflecting the original input accurately if anything is missing.

crates/ironrdp-server/src/server.rs
```diff
@@ -0,0 +1,292 @@
...
+#[derive(Clone)]
+pub struct RdpServerOptions {
+    pub addr: SocketAddr,
+    pub security: RdpServerSecurity,
+}
+
+#[derive(Clone)]
+pub enum RdpServerSecurity {
+    None,
+    TLS(TlsAcceptor),
```
CBenoit, line 35: Nit: the convention in Rust is to count acronyms and contractions of compound words as one word so this should be `Tls` rather than `TLS`

```suggestion
    Tls(TlsAcceptor),
```

crates/ironrdp-server/src/display.rs
```diff
@@ -0,0 +1,75 @@
...
+pub trait RdpServerDisplay {
+    /// This method should return the current size of the display.
+    /// Currently, there is no way for the client to negotiate resolution,
+    /// so the size returned by this method will be enforced.
+    async fn size(&mut self) -> DesktopSize;
+
+    /// # Cancel safety
+    ///
+    /// This method MUST be cancellation safe because it is used in a `tokio::select!` statement.
+    /// If some other branch completes first, it MUST guaranteed that no data is lost.
```
CBenoit, line 73: Nit:
```suggestion
    /// If some other branch completes first, it MUST be guaranteed that no data is lost.
```

crates/ironrdp-server/src/display.rs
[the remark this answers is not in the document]
CBenoit: Sounds good. This is not something we absolutely need to support for the MVP.

CBenoit (commented):
Thank you, this is looking quite good.

> In the future, I hope to also:
>
>    improve display updates encoder (remove that buffer, refactor main loop to not block input events)
>    figure out a way to handle resolution changes/requests.
>
> But that will be in follow-up pr's as this one is already pretty big, and I need a break from reading RDP specifications :)

Haha, sounds good! I’m expecting PRs with a more manageable size next :laughing: 

I’ll wait for the input of @pacmancoder and @elmarco and then it’s good to merge as soon as the remaining conversations are resolved.

Nice work!

crates/ironrdp-server/src/display.rs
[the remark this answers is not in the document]
CBenoit: Sounds good. This is not something we absolutely need to support for the MVP.

crates/ironrdp-pdu/src/basic_output/fast_path.rs
pacmancoder: Agreed, it is not intended to be public

crates/ironrdp-server/src/builder.rs
pacmancoder: I agree that compile-time checks are always preferable to runtime checks. I may suggest to use [`typed_builder`](https://docs.rs/typed-builder/latest/typed_builder/derive.TypedBuilder.html) crate for such purposes, It has great error reporting, minimal dependencies, and is easy to use (in my opinion). This will greatly decrease boilerplate code for the builder types.

 I used it on my previous job, this crate proved to be very useful! @CBenoit what do you think about it? Making builder type is as simple as

mihneabuz: Personally, I still prefer my builder because it works like a state machine. You can easily define multiple setters that take you from one state to the next, and it forces you to always do the steps in the same order.
Like @CBenoit noticed, it was inspired by the [one in rustls](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html).

But I looked into `typed_builder` (thanks for the suggestion @pacmancoder) and it seems like it could replace this entire file with a couple annotations. So if you guys prefer that, I am ok with it.

It would be something like this
```rust
    let mut server = RdpServer::builder()
        .with_options(RdpServerOptions {
            addr: ([127, 0, 0, 1], 3389).into(),
            security: RdpServerSecurity::Tls(tls_acceptor),
        })
        .with_display(Box::new(display_handler))
        .with_handler(Box::new(input_handler))
        .build();
```

Compared to the current one
```rust
     let mut server = RdpServer::builder()
         .with_addr(([127, 0, 0, 1], 3389))
         .with_tls(tls_acceptor)
         .with_input_handler(input_handler)
         .with_display_handler(display_handler)
         .build();
```

mihneabuz: I just realized that, after changing `RdpServer` to use dynamic dispatch, I can also use it to simplify the builder.
But my point still stands, if you would rather use `typed_builder` I'm ok with it.

CBenoit: After further discussion with @pacmancoder on slack, we agreed on merging the current builder and see if `typed-builder` would be used somewhere else before committing to using the crate.
Main reasons are:
- `typed-builder` depends on the `syn` crate, and ideally we want to push this dependency as far as possible from the foundational crates so it doesn’t become too much of a compilation bottleneck (also, this crate often ends up duplicated as it can be seen in the `Cargo.lock` on current master). We’re not yet in this ideal world, but that’s the goal. Note however that I’m open to using some proc-macro crates in `ironrdp-server` since this crate will probably end up either [in the community or extra tier](https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#extra-tier), which is "far enough" from the foundational crates per IronRDP standard.
- The handcrafted API is indeed looking a bit better.

So there is no need to change anything immediately.

CBenoit:
@elmarco Sorry to bother you again. Do you have any remaining concern?

elmarco:
> @elmarco Sorry to bother you again. Do you have any remaining concern?

Nothing that should block this from being merged. We can address various improvements later.

CBenoit (approved):
LGTM!
Thank you for this major contribution @mihneabuz 
Also, thank you @elmarco @pacmancoder for helping with reviewing, this is really appreciated!

PR #168, opened 2023-08-02
[DRAFT] Fast path pointer processing logic implementation

pacmancoder:
This PR is still in draft, a few minor changes are left to be implemented:
- CLI flag to disable pointer messages processing
- Finalize code refactoring
- Unit tests for cursor decoding

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -16,6 +16,7 @@ use crate::{
 #[derive(Clone, Copy, Debug)]
 pub struct CredsspTsRequestHint;
 
+pub const DEFAULT_POINTER_CACHE_SIZE: u16 = 32;
 pub const CREDSSP_TS_REQUEST_HINT: CredsspTsRequestHint = CredsspTsRequestHint;
 
 impl PduHint for CredsspTsRequestHint {
@@ -977,8 +978,8 @@ fn create_client_confirm_active(
             keyboard_ime_filename: config.ime_file_name.clone(),
         }),
         CapabilitySet::Pointer(Pointer {
-            color_pointer_cache_size: 0,
-            pointer_cache_size: 0,
+            color_pointer_cache_size: DEFAULT_POINTER_CACHE_SIZE,
+            pointer_cache_size: DEFAULT_POINTER_CACHE_SIZE,
         }),
         CapabilitySet::Brush(Brush {
             support_level: SupportLevel::Default,
@@ -1007,7 +1008,7 @@ fn create_client_confirm_active(
             flags: SoundFlags::empty(),
         }),
         CapabilitySet::LargePointer(LargePointer {
-            flags: LargePointerSupportFlags::UP_TO_96X96_PIXELS,
+            flags: LargePointerSupportFlags::UP_TO_384X384_PIXELS,
         }),
         CapabilitySet::SurfaceCommands(SurfaceCommands {
             flags: CmdFlags::SET_SURFACE_BITS | CmdFlags::STREAM_SURFACE_BITS | CmdFlags::FRAME_MARKER,
--- a/crates/ironrdp-graphics/src/pointer.rs
+++ b/crates/ironrdp-graphics/src/pointer.rs
@@ -1,0 +1,379 @@
+use ironrdp_pdu::{
+    cursor::ReadCursor,
+    pointer::{ColorPointerAttribute, LargePointerAttribute, PointerAttribute},
+};
+use thiserror::Error;
+
+/// References:
+/// Drawing pointers: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/pointer-drawing
+/// Drawing color pointers: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-color-pointers
+/// Drawing monochrome pointers https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-monochrome-pointers
+///
+///
+/// andMask is always 1bpp.
+/// andMask == 0 -> Copy pixel from xorMask
+/// andMask == 1, xorMask == 0 -> Transparent pixel
+/// andMask == 1, xorMask == 1 -> Pixel is inverted
+///
+/// Input: andMask, xorMask, lastFrame -> Output: pixelColor, invertionMask
+
+#[derive(Debug, Error)]
+pub enum PointerError {
+    #[error("Invalid pointer xorMask size. Expected: {expected}, actual: {actual}")]
+    InvalidXorMaskSize { expected: usize, actual: usize },
+    #[error("Invalid pointer andMask size. Expected: {expected}, actual: {actual}")]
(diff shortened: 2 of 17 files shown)


crates/ironrdp-connector/src/connection.rs
```diff
@@ -977,8 +978,8 @@ fn create_client_confirm_active(
             keyboard_ime_filename: config.ime_file_name.clone(),
         }),
         CapabilitySet::Pointer(Pointer {
-            color_pointer_cache_size: 0,
-            pointer_cache_size: 0,
+            color_pointer_cache_size: DEFAULT_POINTER_CACHE_SIZE,
+            pointer_cache_size: DEFAULT_POINTER_CACHE_SIZE,
```
CBenoit, line 982: Nit: maybe add a comment above this explaining how this is really important (as you said on slack)

crates/ironrdp-connector/src/connection.rs
```diff
@@ -1007,7 +1008,7 @@ fn create_client_confirm_active(
             flags: SoundFlags::empty(),
         }),
         CapabilitySet::LargePointer(LargePointer {
-            flags: LargePointerSupportFlags::UP_TO_96X96_PIXELS,
+            flags: LargePointerSupportFlags::UP_TO_384X384_PIXELS,
```
CBenoit, line 1011: Nit: same here

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -0,0 +1,379 @@
...
+/// Drawing color pointers: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-color-pointers
+/// Drawing monochrome pointers https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-monochrome-pointers
+///
+///
+/// andMask is always 1bpp.
+/// andMask == 0 -> Copy pixel from xorMask
+/// andMask == 1, xorMask == 0 -> Transparent pixel
+/// andMask == 1, xorMask == 1 -> Pixel is inverted
+///
+/// Input: andMask, xorMask, lastFrame -> Output: pixelColor, invertionMask
```
CBenoit, line 18: This is considered a docstring for the `PointerError` enum below. You can either promote it to a module docstring by putting it at the top and using the `//!` prefix instead of `///`, or if you don’t want this to be module-level documentation (i.e.: showing in the rustdoc), use the simple comment prefix `//` as done in https://github.com/Devolutions/IronRDP/blob/2cb0b476b5a6c31b19915de9f1167bd5c54f5eb8/crates/ironrdp-pdu/src/mcs.rs#L10

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -0,0 +1,379 @@
...
+pub enum PointerError {
+    #[error("Invalid pointer xorMask size. Expected: {expected}, actual: {actual}")]
+    InvalidXorMaskSize { expected: usize, actual: usize },
+    #[error("Invalid pointer andMask size. Expected: {expected}, actual: {actual}")]
+    InvalidAndMaskSize { expected: usize, actual: usize },
+    #[error("Not supported pointer bpp: {bpp}")]
+    NotSupportedBpp { bpp: usize },
+    #[error(transparent)]
+    Pdu(#[from] ironrdp_pdu::PduError),
+}
```
CBenoit, line 30: Nit: note that at some point I want to remove the dependency on `thiserror`  from all the core-tier crates. The idea is to not depend on heavy proc macros for these so they can be compiled faster and earlier during the build (prior to `syn`). You don’t need to refactor now though.

CBenoit, line 30: Nit: note that at some point I want to remove the dependency on `thiserror`  from all the core-tier crates. The idea is to not depend on heavy proc macros for these so they can be compiled faster and earlier during the build (prior to `syn`). You don’t need to refactor now though.

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -0,0 +1,379 @@
...
+    let g = (((((color >> 5) & 0x3f) * 259) + 33) >> 6) as u8;
+    let b = ((((color & 0x1f) * 527) + 23) >> 6) as u8;
+
+    [r, g, b, 0xff]
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
CBenoit, line 321: Nit: do you think you could move this to the `ironrdp-testsuite-core` crate? Eventually [we want everything in there](https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#test-at-the-boundaries-test-features-not-code). Is there an okay way to trigger the same behavior by using the public API itself? Maybe we should add a feature flag to expose some inner API to the testsuite in some cases…

CBenoit, line 321: Nit: do you think you could move this to the `ironrdp-testsuite-core` crate? Eventually [we want everything in there](https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#test-at-the-boundaries-test-features-not-code). Is there an okay way to trigger the same behavior by using the public API itself? Maybe we should add a feature flag to expose some inner API to the testsuite in some cases…

crates/ironrdp-session/src/fast_path.rs
```diff
@@ -3,30 +3,57 @@ use ironrdp_graphics::rle::RlePixelFormat;
...
+    use_system_pointer: bool,
+    mouse_pos_update: Option<(usize, usize)>,
 }
 
 impl Processor {
+    pub fn update_mouse_pos(&mut self, x: usize, y: usize) {
+        self.mouse_pos_update = Some((x, y));
+    }
+
     // Returns true if image buffer was updated, false otherwise
```
CBenoit, line 42: It appears this comment is now really outdated :laughing:

CBenoit (commented):
Good job!

crates/ironrdp-graphics/src/pointer.rs
pacmancoder: Yeah, I added that comment temporarily, I need to change it and make it more informative either way

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -24,30 +27,85 @@ impl ActiveStage {
...
+        self.fast_path_processor.update_mouse_pos(x, y);
+    }
+
+    /// Encodes outgoing input events and modifies image if necessary (e.g for client-side pointer
+    /// rendering).
+    pub fn process_fastpath_input(
+        &mut self,
+        image: &mut DecodedImage,
+        events: &[FastPathInputEvent],
+    ) -> SessionResult<Vec<ActiveStageOutput>> {
```
CBenoit, line 51: Note: `ironrdp-session` should not be responsible for any kind of "rendering", but it’s a problem for another day.

CBenoit (approved):
LGTM! :tada:

Issue #107, opened 2023-05-01
Initial Clipboard support (textual data)


CBenoit:
Initial implementation of the Clipboard Virtual Channel Extension (MS-RDPECLIP):
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeclip/fb9b7e0b-6db4-41c2-b83c-f889c1ee7688

To simplify initial implementation, let’s start with textual data only.

Jira Issue: [ARC-125](https://devolutions.atlassian.net/browse/ARC-125)

[ARC-125]: https://devolutions.atlassian.net/browse/ARC-125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #130, opened 2023-05-11
Fix UTF-16 string length computations


CBenoit:
Some code in IronRDP is incorrectly computing the size of UTF-16 strings.

Example:
```rust
// This is not the right way to compute the number of bytes for unicode strings.
// This is a time bomb: it will returns the correct value some times, but not always.
fn utf16_len(utf8_str: &str) -> usize {
    utf8_str.len() * 2
}
```

Both UTF-8 and UTF-16 are using a variable-length encoding and code points may be encoded using multiple code units. The thing is, UTF-16 uses one or two 16-bit code units and UTF-8 uses between one and four 8-bit code units. It’s really not always the case that a code point in UTF-16 is twice as big as the same code point in UTF-8.

This kind of erroneous code is present at multiple places. One such instance is [`ironrdp_pdu::rdp::client_info::string_len`](https://github.com/Devolutions/IronRDP/blob/8857c3c25ea40bd2a2f4d850f67b16b2306393c9/crates/ironrdp-pdu/src/rdp/client_info.rs#L572).

Instead:
```rust
// Something like that must be used:
utf8_str.encode_utf16().count() * 2 // add 2 if we need to account for a null terminator (0x0000)
```

Refer to `ironrdp_pdu::pcb` module for a correct implementation.

PR #170, opened 2023-08-11
Clipboard PDU support 

pacmancoder:
Add support for clipboard virtual channel PDUs encode/decode logic [MS-RDPECLIP].

Although only textual representation is needed in the first iteration(#107), this PR implements all PDUs defined in [MS-RDPECLIP] specification.

The feature was tested against PDU blobs presented in section 4 (examples) of [MS-RDPECLIP]


crates/ironrdp-pdu/src/utils.rs
```diff
@@ -26,6 +27,98 @@ pub(crate) enum CharacterSet {
...
+            // Find null or read all if null is not found
+            cursor
+                .remaining()
+                .chunks_exact(2)
+                .position(|chunk| chunk[0] == 0 && chunk[1] == 0)
+                .map(|codepoints| codepoints + 1) // Read null codepoint
+                .unwrap_or(cursor.len() / 2)
+        } else {
+            // UTF16 uses 2 bytes per code point, so we need to read an even number of bytes
+            cursor.len() / 2
```
CBenoit, line 46: I know this is something done in the current codebase, that dates back to when IronRDP was initially written, but that’s not true. I opened an issue on this: https://github.com/Devolutions/IronRDP/issues/130

pacmancoder: No, it's a different calculation -- not the size of the UTF16 string, but the closest valid buffer size which could be used for a decoding attempt.

I think I just used the wrong naming, which gave this confusion -- I think I wanted to write `code units` instead of `code points`. Code units in UTF16 are always 2 bytes in size.

CBenoit: Oh I see. Fair enough!

crates/ironrdp-pdu/src/clipboard/capabilities.rs
```diff
@@ -0,0 +1,249 @@
...
+impl PduEncode for Capabilities {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
+        if self.capabilities.len() > u16::MAX as usize {
+            return Err(invalid_message_err!(
+                "cCapabilitiesSets",
+                "Too much capability sets specified",
+            ));
+        }
+
+        let header = PartialHeader::new(self.inner_size() as u32);
```
CBenoit, line 30: I suggest either using `u32::from` or the `cast_length!` macro if the conversion is fallible. Same for all other `as` conversions because those are lossy conversions while we need lossless conversion (if the value does not fit in the target type, we have a problem). Since this is not exactly what I would call "elements", you could also add a new `cast_int!` macro like this one:
```rust
#[macro_export]
macro_rules! cast_int {
    ($len:expr, $name:expr, $field:expr) => {{
        $len.try_into().map_err(|_| $crate::Error::InvalidMessage {
            name: $name,
            field: $field,
            reason: "out of range integral type conversion",
        })
    }};
}
```

CBenoit (commented):
This is looking pretty good.

Do you think you could move the clipboard PDUs into its own crate? I’m thinking about a crate named after the name of the channel: `ironrdp-cliprdr`.
This crate will also provides the implementation of the static channel itself.

What do you think?

crates/ironrdp-pdu/src/clipboard/capabilities.rs
pacmancoder: Oh, good point, indeed this is an oversight, I'll make `PartialHeader::new` return `Result` and perform usize->u32 cast with validation inside, Macro would be great addition too!

pacmancoder:
> Do you think you could move the clipboard PDUs into its own crate? I’m thinking about a crate named after the name of the channel: ironrdp-cliprdr.
> This crate will also provides the implementation of the static channel itself in the future.

Sounds good to me, will do that 👍

crates/ironrdp-pdu/src/clipboard/capabilities.rs
CBenoit: Sounds good, but I do think it’s also okay for `PartialHeader::new` to stay the way it currently is:
```rust
let inner_size: u32 = cast_int!(self.inner_size());
let header = PartialHeader::new(inner_size);
```

pacmancoder:
@CBenoit I made the refactoring, PR is ready for review again 😃

CBenoit (commented):
Excellent work, minor comments

crates/ironrdp-cliprdr/src/pdu/format_data/file_list.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    fn size(&self) -> usize {
+        Self::FIXED_PART_SIZE + FileDescriptor::SIZE * self.files.len()
+    }
+}
+
+impl<'de> PduDecode<'de> for PackedFileList {
+    fn decode(src: &mut ReadCursor<'de>) -> PduResult<Self> {
+        ensure_fixed_part_size!(in: src);
+        let file_count = src.read_u32() as usize;
```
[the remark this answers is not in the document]
CBenoit: Demonstration: https://godbolt.org/z/dv7b69xf8

pacmancoder:
@CBenoit fixed last batch of comments 😄

CBenoit (approved):
LGTM! Thank you!

PR #171, opened 2023-08-11
Adds basic plumbing for RDPDR svc

ibeckermayer:
This also includes plumbing for RDPSND, which is required for the server to send us anything on RDPDR


crates/ironrdp-connector/src/connection.rs
```diff
@@ -50,6 +50,8 @@ pub struct ConnectionResult {
     pub user_channel_id: u16,
     pub static_channels: StaticChannels,
     pub desktop_size: DesktopSize,
+    /// The virtual channel capability set sent back to us by the server.
+    pub server_vchan_cap: Option<VirtualChannel>,
```
ibeckermayer, line 54: I added this in anticipation of needing to use this information in the active stage. In particular, we will need the maximum chunk size, which starts to become relevant with the potentially very large messages we need to send for directory sharing.

> VCChunkSize (4 bytes): A 32-bit unsigned integer. When sent from server to client, this field contains the maximum allowed size of a virtual channel chunk. When sent from client to server, the value in this field is ignored by the server; the server determines the maximum virtual channel chunk size. This value MUST be greater than or equal to CHANNEL_CHUNK_LENGTH and less than or equal to 16256.

PR #178, opened 2023-08-18
static virtual channel + rdpdr scaffolding

ibeckermayer:


crates/ironrdp-rdpsnd/src/lib.rs
```diff
@@ -0,0 +1,54 @@
+use ironrdp_pdu::{gcc::ChannelName, write_buf::WriteBuf, PduResult};
+use ironrdp_svc::{AsAny, CompressionCondition, StaticVirtualChannel};
+use std::any::Any;
+use tracing::warn;
+
+/// We currently don't implement any of rdpsnd, however it's required
+/// for rdpdr to work: https://tinyurl.com/2fvrtfjd
```
CBenoit, line 7: Even if the original link is very loong, it’s probably better to still not shorten it:
- tinyurl.com is sometimes blocked for security reasons
- They say it will never expire, but you never know

crates/ironrdp-pdu/src/utils.rs
```diff
@@ -115,31 +115,38 @@ pub fn write_string_to_cursor(
...
         CharacterSet::Unicode => {
             let mut buffer = to_utf16_bytes(value);
             if write_null_terminator {
                 buffer.push(0);
                 buffer.push(0);
             }
-
-            ensure_size!(ctx: "Encode sting (UTF-16)", in: cursor, size: buffer.len());
-            cursor.write_slice(&buffer);
+            (buffer, "Encode sting (UTF-16)")
```
CBenoit, line 125:
```suggestion
            (buffer, "Encode string (UTF-16)")
```

crates/ironrdp-pdu/src/utils.rs
```diff
@@ -115,31 +115,38 @@ pub fn write_string_to_cursor(
...
         }
         CharacterSet::Ansi => {
             let mut buffer = value.as_bytes().to_vec();
             if write_null_terminator {
                 buffer.push(0);
             }
-
-            ensure_size!(ctx: "Encode sting (UTF-8)", in: cursor, size: buffer.len());
-            cursor.write_slice(&buffer);
+            (buffer, "Encode sting (UTF-16)")
```
CBenoit, line 132:
```suggestion
            (buffer, "Encode string (UTF-8)")
```
Also, "Ansi" encoding is different from UTF-8, but IIRC UTF-8 is actually used by freerdp in those cases

crates/ironrdp-pdu/src/write_buf.rs
```diff
@@ -40,6 +40,10 @@ impl WriteBuf {
         self.inner
     }
 
+    pub fn slice(&self, from: usize, to: usize) -> &[u8] {
+        &self.inner[from..to]
+    }
+
```
CBenoit, line 46: You could implement the trait `Index<Range<usize>>` instead if you want

crates/ironrdp-dvc/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-#![cfg_attr(not(feature = "std"), no_std)]
```
CBenoit, line 1: Intentional?

crates/ironrdp-rdpdr/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+name = "ironrdp-rdpdr"
+version = "0.1.0"
+description = "RDPDR channel implementation."
+readme = "README.md"
+edition.workspace = true
+license.workspace = true
+
+homepage.workspace = true
+repository.workspace = true
+# TODO: add individuals and/or Teleport as authors?
```
CBenoit, line 11: At your preference

CBenoit (commented):
Looking nice so far

crates/ironrdp-dvc/src/lib.rs
ibeckermayer: Yes, although I figured out I can do this instead: 3d9462d9abadedadf81e4fcf91b6918ecd21a759

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -74,23 +74,17 @@ impl Processor {
...
-                            .process(data_ctx.initiator_id, channel_id, user_data, &mut static_channel_pdus)
-                            .map_err(crate::SessionError::pdu)?;
-
-                        // For each response PDU, chunkify it it and add
-                        // appropriate static channel headers.
-                        let chunks = chunkify(&mut static_channel_pdus, CHANNEL_CHUNK_LEGNTH)
-                            .map_err(crate::SessionError::pdu)?; // TODO: CHANNEL_CHUNK_LEGNTH is only the default, we should grab this from what the server advertises if possible
+                        let response_pdus = static_channel.process(payload).map_err(crate::SessionError::pdu)?;
+
+                        // For each response PDU, chunkify it it and add appropriate static channel headers.
```
CBenoit, line 86:
```suggestion
                        // For each response PDU, chunkify it and add appropriate static channel headers.
```

crates/ironrdp-pdu/src/lib.rs
```diff
@@ -148,11 +148,21 @@ pub fn encode<T: PduEncode>(pdu: &T, dst: &mut [u8]) -> PduResult<usize> {
     Ok(cursor.pos())
 }
 
+pub fn encode_box(pdu: Box<dyn PduEncode>, dst: &mut [u8]) -> PduResult<usize> {
+    let mut cursor = WriteCursor::new(dst);
+    encode_cursor_box(pdu, &mut cursor)?;
+    Ok(cursor.pos())
+}
```
CBenoit, line 155: If I understand correctly, we don’t need all the `*_box` functions. The following code should compile fine:
```rust
fn chunkify_one(pdu: Box<dyn PduEncode>) -> PduResult<Vec<WriteBuf>> {
    let mut encoded_pdu = WriteBuf::new();
    encode_buf(&pdu, &mut encoded_pdu)?; // <-- `&buf`
    // If it doesn’t, maybe `as_ref` will return what we want
    encode_buf(pdu.as_ref(), &mut encoded_pdu)?; // <-- `buf.as_ref()`
    // …
}
```

Cargo.toml
```diff
@@ -16,7 +16,7 @@ edition = "2021"
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/Devolutions/IronRDP"
 repository = "https://github.com/Devolutions/IronRDP"
-authors = ["Devolutions Inc. <infos@devolutions.net>"]
+authors = ["Devolutions Inc. <infos@devolutions.net>" "Teleport <goteleport.com>"]
```
CBenoit, line 19: Values are separated by comma:
```suggestion
authors = [
  "Devolutions Inc. <infos@devolutions.net>",
  "Teleport <goteleport.com>",
]
```

crates/ironrdp-rdpsnd/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+use ironrdp_pdu::{gcc::ChannelName, PduEncode, PduResult};
+use ironrdp_svc::{AsAny, CompressionCondition, StaticVirtualChannel};
+use std::any::Any;
+
+/// We currently don't implement any of rdpsnd, however it's required
+/// for rdpdr to work: [\[MS-RDPEFS\] Appendix A<1>](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/fd28bfd9-dae2-4a78-abe1-b4efa208b7aa#Appendix_A_1)
```
CBenoit, line 6: Nit: markdown tip, you can [use the reference-style syntax](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#link-all-the-things).

```suggestion
/// for rdpdr to work: [\[MS-RDPEFS\] Appendix A<1>]
///
/// [\[MS-RDPEFS\] Appendix A<1>]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/fd28bfd9-dae2-4a78-abe1-b4efa208b7aa#Appendix_A_1)
```

crates/ironrdp-svc/src/lib.rs
```diff
@@ -36,22 +37,17 @@ pub enum CompressionCondition {
...
-    fn process(
-        &mut self,
-        initiator_id: StaticChannelId,
-        channel_id: StaticChannelId,
-        payload: &[u8],
-        output: &mut WriteBuf,
-    ) -> PduResult<()>;
+    /// Processes a payload received on the virtual channel.
+    /// Returns a list of PDUs to be sent back to the client.
+    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<Box<dyn PduEncode>>>;
```
CBenoit, line 50: I was initially a bit skeptical about returning a `Vec` of trait objects, but it appears to have a few nice properties:
- the trait is more straightforward to implement downstream, and
- it actually opens up the door to further optimizations down the road: we can actually perform inline encoding of the packet into the final output buffer given that `PduEncode::size` returns a value smaller that the maximum chunk size for instance.
Let’s roll with that and see where we end up.

crates/ironrdp-svc/src/lib.rs
```diff
@@ -62,6 +58,79 @@ pub trait StaticVirtualChannel: AsAny + fmt::Debug + Send + Sync {
 
 assert_obj_safe!(StaticVirtualChannel);
 
+/// Takes a vector of PDUs and breaks them into chunks prefixed with a [`ChannelPDUHeader`].
+/// Each chunk is at most `max_chunk_len` bytes long (not including the ChannelPduHeader).
+pub fn chunkify(pdus: Vec<Box<dyn PduEncode>>, max_chunk_len: usize) -> PduResult<Vec<WriteBuf>> {
```
CBenoit, line 63: Just thinking out loud: I could see this function elsewhere as it’s not going to be used by implementers of static channels, but it’s not something we have to think about right now.

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -72,19 +74,42 @@ impl Processor {
...
-
-                        static_channel
-                            .process(data_ctx.initiator_id, channel_id, user_data, &mut buf)
-                            .map_err(crate::SessionError::pdu)?;
-
-                        Ok(buf.into_inner())
+                        let channel_header = ironrdp_pdu::rdp::vc::ChannelPduHeader::from_buffer(&mut payload)?;
+                        debug_assert_eq!(payload.len(), channel_header.length as usize);
+
+                        // Fill the bufs with encoded response PDUs
```
CBenoit, line 83: This comment appears to be outdated with the last change.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+pub enum VersionAndIdPDUKind {
+    /// [2.2.2.2 Server Announce Request (DR_CORE_SERVER_ANNOUNCE_REQ)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/046047aa-62d8-49f9-bf16-7fe41880aaf4)
+    ServerAnnounceRequest,
+    /// [2.2.2.3 Client Announce Reply (DR_CORE_CLIENT_ANNOUNCE_RSP)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/d6fe6d1b-c145-4a6f-99aa-4fe3cdcea398)
+    ClientAnnounceReply,
+}
+
+#[derive(Debug)]
+/// VersionAndIdPDU is a fixed size structure representing multiple PDUs.
+/// The kind field is used to determine the actual PDU type.
```
CBenoit, line 111: Docstring is typically above the annotations such as `#[derive]`.
Also, [the summary is separated from the rest by a blank line](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components).
```suggestion
/// VersionAndIdPDU is a fixed size structure representing multiple PDUs.
///
/// The kind field is used to determine the actual PDU type.
#[derive(Debug)]
```

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -72,19 +74,42 @@ impl Processor {
...
+                            .collect::<Vec<mcs::SendDataRequest>>();
+
+                        // SendDataRequest is [`McsPdu`], which is [`x224Pdu`], which is [`PduEncode`]. [`PduEncode`] for [`x224Pdu`]
+                        // also takes care of adding the Tpkt header, so therefore we can just call `encode_buf` on each of these and
+                        // we will create a buffer of fully encoded PDUs ready to send to the server.
+                        //
+                        // For example, if we had 2 chunks, our fully_encoded_responses buffer would look like:
+                        //
+                        // [ | tpkt | x224 | mcs::SendDataRequest | chunk 1 | tpkt | x224 | mcs::SendDataRequest | chunk 2 | ]
+                        //   |<------------------- PDU 1 ------------------>|<------------------- PDU 2 ------------------>|
```
CBenoit, line 106: Nicely documented code. Thank you.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    fn header(&self) -> PduResult<SharedHeader> {
+        match self.kind {
+            VersionAndIdPDUKind::ClientAnnounceReply => Ok(SharedHeader {
+                component: Component::RDPDR_CTYP_CORE,
+                packet_id: PacketId::PAKID_CORE_CLIENTID_CONFIRM,
+            }),
+            VersionAndIdPDUKind::ServerAnnounceRequest => Err(unexpected_message_type_err!("ServerAnnounceRequest", 0)),
```
CBenoit, line 135: I’m a bit curious about this. I understand we can’t encode a `VersionAndIdPDU` when the kind is "Server Announce Request", is there a reason for that? In general this kind of logic should not live in the PDU itself.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    fn encode(&self, dst: &mut WriteCursor) -> PduResult<()> {
+        dst.write_u16(self.component as u16);
+        dst.write_u16(self.packet_id as u16);
+        Ok(())
+    }
+
+    fn size(&self) -> usize {
+        4 // 2 * u16  = 2 * 2 bytes = 4 bytes
```
pacmancoder, line 35: nit: this could be changed into self-documented code:
```
std::mem::size_of::<u16> * 2
```

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+/// The kind field is used to determine the actual PDU type.
+pub struct VersionAndIdPDU {
+    version_major: u16,
+    version_minor: u16,
+    pub client_id: u32,
+    kind: VersionAndIdPDUKind,
+}
+
+impl VersionAndIdPDU {
+    pub fn new(version_major: u16, version_minor: u16, client_id: u32, kind: VersionAndIdPDUKind) -> Self {
```
pacmancoder, line 120: nit: I think for such simple data types all fields could be made `public`, and `new` could be removed

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn size(&self) -> usize {
+        4 // 2 * u16  = 2 * 2 bytes = 4 bytes
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+#[repr(u16)]
+#[allow(non_camel_case_types)]
+pub enum Component {
+    RDPDR_CTYP_CORE = 0x4472,
```
pacmancoder, line 43: In `ironrdp` We usually transform enum variant to camel case for PDUs, And only for `bitflags!` do we preserve original case (see `ironrdp_cliprdr::pdu::FormatListResponse`  for example

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+            _ => Err(invalid_message_err!("try_from", "Component", "invalid value")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+#[repr(u16)]
+#[allow(non_camel_case_types)]
+pub enum PacketId {
+    PAKID_CORE_SERVER_ANNOUNCE = 0x496E,
```
pacmancoder, line 63: Same comment about camel case as above

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: True! In the case of `bitflags!` generated structs, the reason is that those are actually `const` items.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+#[derive(Debug)]
+pub struct SharedHeader {
+    pub component: Component,
+    pub packet_id: PacketId,
+}
+
+impl SharedHeader {
+    pub fn decode(src: &mut ReadCursor) -> PduResult<Self> {
+        Ok(Self {
+            component: src.read_u16().try_into()?,
```
pacmancoder, line 23: `ensure_size` is missing for decode and encode, this should be done to prevent panics on insufficient input/output buffer

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn size(&self) -> usize {
+        // header bytes + (2 * u16) + u32 = header bytes + (2 * 2 bytes) + 4 bytes = header bytes + 8 bytes
+        self.header().unwrap().size() + 8
+    }
+}
+
+/// 2.2.2.4 Client Name Request (DR_CORE_CLIENT_NAME_REQ)
+/// https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/902497f1-3b1c-4aee-95f8-1668f9b7b7d2
+#[derive(Debug)]
+pub struct ClientNameRequest {
```
pacmancoder, line 174: nit: Such structures could be transformed to be represented in mo rust-y way 😄 
```
pub enum ClientNameRequest {
    Ascii(String),
    Unicode(String),
}

impl ClientNameRequest {
    // - ClientNameRequestUnicodeFlag could be removed, flags could be directly transformed to/from enum variants
    // - `new` could be removed too
}
```

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn size(&self) -> usize {
+        // header bytes + u32 * 3 + computer_name_len = header bytes + 4 bytes * 3 + computer_name_len = header bytes + 12 + computer_name_len
+        self.header().size() + 12 + encoded_str_len(&self.computer_name, self.unicode_flag.into(), true)
+    }
+}
+
+#[repr(u32)]
+#[derive(Debug, Clone, Copy)]
+enum ClientNameRequestUnicodeFlag {
+    #[allow(dead_code)]
```
pacmancoder, line 217: Generally, dead code should be avoided, and PDU parsing/building code should provide the full access to all PDU variants if possible

There are a few options to avoid dead code:
- Add 2 separate public constructors for `ClientNameRequest`: `new_unicode` and `new_ascii` instead of single one
- Remove `new` in `ClientNameRequest` and make its fields public, ClientNameRequestUnicodeFlag also should be made public
- Refactor `ClientNameRequest` as described above

crates/ironrdp-svc/src/lib.rs
pacmancoder: Yep, I agree too!

crates/ironrdp-pdu/src/write_buf.rs
ibeckermayer: fb16114b545a7a09232af3b2c93211460c6b7ad2

Cargo.toml
ibeckermayer: 15ce485f09700dab2cf9928768767fb54e007b82

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: It's not that we actually don't know how to do build and encode that PDU, it was just representative of the fact that we're currently just building the client side code and so don't currently need to be able to do so.

However I suppose there's no harm in just going ahead and implementing this now for later use by IronRDP server: 

6d7af41bb314c4a0ae13db3a7b15303665c1307d

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: My preference is for overriding the standard Rust linter rules here because it makes it easier for me to cross reference from the codebase to the documentation. 

For example, if I Google "[RDPDR_CTYP_CORE](https://www.google.com/search?q=RDPDR_CTYP_CORE)" or "[RDPDR_CTYP_CORE = 0x4472](https://www.google.com/search?q=RDPDR_CTYP_CORE+%3D+0x4472)", I get the exactly relevant piece of documentation at the top, vs if I try "[RdpdrCtypCore](https://www.google.com/search?q=rdp+RdpdrCtypCore)" or "[RdpdrCtypCore = 0x4472](https://www.google.com/search?q=RdpdrCtypCore+%3D+0x4472)" or "[rdp [MS-RDPEFS] RdpdrCtypCore](https://www.google.com/search?q=rdp+%5BMS-RDPEFS%5D+RdpdrCtypCore)" etc., the documentation doesn't show up.

It's also helpful in the other direction, i.e. if you're looking at the documentation and say "I wonder where this is implemented?", you can just grep for `RDPDR_CTYP_CORE`. Of course it's not that difficult to translate into camel case, but that's only because I know what to look for -- if I'm an outside developer interested in exploring the codebase anew, I won't have that knowledge. 

So I would propose that we treat such constants like this here and moving forward (and ideally go back and change the CamelCase ones elsewhere, longer term). But if you prefer consistency with the rest of the codebase over this concern I'm not going to fight back on that.

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: 34ec2d46c46c87c4dfb60537deff0f00b52536af

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: `new_unicode`/`new_ascii` doesn't get rid of dead code, because then it just complains that I'm not using `new_ascii`. I was however able to get rid of the dead_code like 925c42031bc2c72d356ed14ff7c0fb7c78ed8097

crates/ironrdp-pdu/src/lib.rs
ibeckermayer: For `encode_buf(&pdu, &mut encoded_pdu)?;`, I get the error:

```
error[E0277]: the trait bound `std::boxed::Box<dyn ironrdp_pdu::PduEncode>: ironrdp_pdu::mcs::McsPdu<'_>` is not satisfied
   --> crates/ironrdp-svc/src/lib.rs:80:16
    |
80  |     encode_buf(&pdu, &mut encoded_pdu)?;
    |     ---------- ^^^^ the trait `ironrdp_pdu::mcs::McsPdu<'_>` is not implemented for `std::boxed::Box<dyn ironrdp_pdu::PduEncode>`
    |     |
    |     required by a bound introduced by this call
    |
    = help: the following other types implement trait `ironrdp_pdu::mcs::McsPdu<'de>`:
              ironrdp_pdu::mcs::AttachUserConfirm
              ironrdp_pdu::mcs::AttachUserRequest
              ironrdp_pdu::mcs::ChannelJoinConfirm
              ironrdp_pdu::mcs::ChannelJoinRequest
              ironrdp_pdu::mcs::DisconnectProviderUltimatum
              ironrdp_pdu::mcs::ErectDomainPdu
              ironrdp_pdu::mcs::McsMessage<'de>
              ironrdp_pdu::mcs::SendDataIndication<'de>
              ironrdp_pdu::mcs::SendDataRequest<'de>
    = note: required for `std::boxed::Box<dyn ironrdp_pdu::PduEncode>` to implement `ironrdp_pdu::x224::X224Pdu<'_>`
    = note: required for `std::boxed::Box<dyn ironrdp_pdu::PduEncode>` to implement `ironrdp_pdu::PduEncode`
note: required by a bound in `ironrdp_pdu::encode_buf`
   --> /Users/ibeckermayer/Devolutions/IronRDP/crates/ironrdp-pdu/src/lib.rs:168:22
    |
168 | pub fn encode_buf<T: PduEncode>(pdu: &T, buf: &mut WriteBuf) -> PduResult<usize> {
    |                      ^^^^^^^^^ required by this bound in `encode_buf`
```

It's not exactly clear to me why it's complaining about `McsPdu` not being implemented when `PduEncode` is what's required by `encode_buf`, though I think it may be related to the chain of

```rust
impl<'de, T> X224Pdu<'de> for T where T: McsPdu<'de>;
impl<'de, T> PduEncode for T where T: X224Pdu<'de>;
```

---

For `encode_buf(pdu.as_ref(), &mut encoded_pdu)?;`, I get

```
error[E0277]: the size for values of type `dyn ironrdp_pdu::PduEncode` cannot be known at compilation time
   --> crates/ironrdp-svc/src/lib.rs:80:16
    |
80  |     encode_buf(pdu.as_ref(), &mut encoded_pdu)?;
    |     ---------- ^^^^^^^^^^^^ doesn't have a size known at compile-time
    |     |
    |     required by a bound introduced by this call
    |
    = help: the trait `std::marker::Sized` is not implemented for `dyn ironrdp_pdu::PduEncode`
note: required by a bound in `ironrdp_pdu::encode_buf`
   --> /Users/ibeckermayer/Devolutions/IronRDP/crates/ironrdp-pdu/src/lib.rs:168:19
    |
168 | pub fn encode_buf<T: PduEncode>(pdu: &T, buf: &mut WriteBuf) -> PduResult<usize> {
    |                   ^ required by this bound in `encode_buf`
```

but I can't make `PduEncode: Sized` because that breaks its object safety.

ibeckermayer: I think the `McsPdu` error message might just be a weird consequence of how the compiler checks whether `PduEncode` is implemented. E.g. it checks whether `PduEncode` is implemented, then if not it sees `impl<'de, T> PduEncode for T where T: X224Pdu<'de>;` so checks whether `X224Pdu` is implemented, then since that's not the case it knows `impl<'de, T> X224Pdu<'de> for T where T: McsPdu<'de>;` and so checks if `McsPdu`. And since all of those fail, it fails, but it only reports the last one it checked. If that's the case then that's confusing and should probably be fixed by the rust team.

Anyways, I managed to get rid of the `*_box()` functions in 868d7ec962a5de7962f7c08e04e310db04f4bf86

crates/ironrdp-rdpsnd/src/lib.rs
ibeckermayer: a3babe8342e4c869c858d8fe64f6b2416e8f8c59

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: https://github.com/Devolutions/IronRDP/pull/178#discussion_r1300682038

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: 2c8308b48da6d8c7e79c41b6e5f28611c97a1faf

Makes sense, but was tedious and error prone. It makes me wonder how difficult might it be to change the signature in `PduEncode` from `fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()>;` to `fn encode(&self, dst: &mut WriteBuf) -> PduResult<()>;` and thus get rid of this (cc @CBenoit)

crates/ironrdp-pdu/src/lib.rs
CBenoit: Ah okay! Now that I see the error, I understand. Actually, there is a trait bound, `Sized`, that is implicit for all type parameters. I believe the `encode_*` function family should opt out of this [using `?Sized`](https://doc.rust-lang.org/std/marker/trait.Sized.html).
Your approach works by taking a `Box` , which by itself is a `Sized` type, pointing to a dynamically sized type, a trait object.

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: Thank you!

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: To be honest, I don’t feel too strongly about this either. I agree it’s actually useful for cross referencing with documentation. At some places I actually added comments with the actual name from the documentation for that exact same purpose. Eventually, we’ll want consistency: but I’m not against changing the casing, the motivation is good enough for me. What’s your opinion on this @pacmancoder?

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: For decoding I don’t think we can avoid checking for bounds. As for encoding, this change would not be too difficult, but I don’t really want that. This would not be inline encoding anymore. I know this property is adding a few difficulties like this, but it’s actually useful. Some of my colleagues want C# for IronRDP, and the current design make it trivial to read from and write into buffers allocated on C# side via FFI calls. `WriteBuf` requires ownership of the buffer.

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: I might be missing the big picture, but remove the said dead code is also an avenue if it will never be used inside of this crate, and if it’s not intended to be part of the public API.

crates/ironrdp-rdpdr/src/pdu/efs.rs
pacmancoder: I am more inclined to keep casing consistent with other parts of IronRDP. Let's add doc comments with original constant names and rename enum variants to follow standard Rust conventions

CBenoit: > I am more inclined to keep casing consistent with other parts of IronRDP.

Just noting that we would be going back and change the CamelCase ones of other similar enums for consistency. Are you suggesting we make this change for all enums at once, or that we stick to the Rust conventions forever for all the enums?

pacmancoder: > we stick to the Rust conventions forever for all the enums?

Yeah, I think sticking to rust conventions would be better in my opinion

CBenoit: Got you!

ibeckermayer: 8e609e20be22f1243b406585d01172ff0065ba58

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: 8e609e20be22f1243b406585d01172ff0065ba58

crates/ironrdp-pdu/src/lib.rs
ibeckermayer: I see, good find: e31467f45515f17120c1508e9f091a35fe28a534

ibeckermayer: I see! Good find. 

e31467f45515f17120c1508e9f091a35fe28a534

CBenoit (approved):
Great work, thank you!

pacmancoder (approved):
LGTM!

PR #182, opened 2023-08-23
CLIPRDR(clipboard) processing - initialization sequence

pacmancoder:
Adds initialization sequence of CLIPRDR static virtual channel

+ Changes `ironrdp-svc` API to accommodate the need for adding SVC PDU flags on a per-message basis
+ Reduces boilerplate code for `AsAny`

Part of #107

### DIFF
--- a/crates/ironrdp-cliprdr/src/lib.rs
+++ b/crates/ironrdp-cliprdr/src/lib.rs
@@ -4,3 +4,147 @@
 //! - Clipboard SVC processing (TODO)
 
 pub mod pdu;
+
+use ironrdp_pdu::{decode, gcc::ChannelName, PduResult};
+use ironrdp_svc::{impl_as_any, ChannelFlags, CompressionCondition, StaticVirtualChannel, SvcMessage};
+use pdu::{
+    Capabilities, ClientTemporaryDirectory, ClipboardFormat, ClipboardFormatId, ClipboardGeneralCapabilityFlags,
+    ClipboardPdu, ClipboardProtocolVersion, FormatListResponse,
+};
+
+use crate::pdu::FormatList;
+use thiserror::Error;
+use tracing::{error, info};
+
+#[derive(Debug, Error)]
+enum ClipboardError {
+    #[error("Received clipboard PDU is not implemented")]
+    UnimplementedPdu { pdu: &'static str },
+
+    #[error("Sent format list was rejected")]
+    FormatListRejected,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum CliprdrState {
+    Initialization,
+    Ready,
+    Failed,
+}
+
+/// CLIPRDR static virtual channel client endpoint implementation
+#[derive(Debug)]
+pub struct Cliprdr {
+    capabilities: Capabilities,
+    temporary_directory: String,
+    state: CliprdrState,
+}
+
+impl_as_any!(Cliprdr);
+
+impl Cliprdr {
+    const CHANNEL_NAME: ChannelName = ChannelName::from_static(b"cliprdr\0");
+
+    fn build_local_format_list(&self) -> PduResult<FormatList<'static>> {
+        let formats = vec![
+            ClipboardFormat::new(ClipboardFormatId::CF_TEXT),
+            ClipboardFormat::new(ClipboardFormatId::CF_UNICODETEXT),
+        ];
+
+        FormatList::new_unicode(
+            &formats,
+            self.capabilities
+                .flags()
+                .contains(ClipboardGeneralCapabilityFlags::USE_LONG_FORMAT_NAMES),
+        )
+    }
+
+    fn handle_error_transition(&mut self, err: ClipboardError) -> PduResult<Vec<SvcMessage>> {
+        // Failure of clipboard is not an critical error, but we should properly report it
+        // and transition channel to failed state.
+        self.state = CliprdrState::Failed;
+        error!("CLIPRDR(clipboard) failed: {err}");
+
+        Ok(vec![])
+    }
+
+    fn handle_server_capabilities(&mut self, server_capabilities: Capabilities) -> PduResult<Vec<SvcMessage>> {
+        self.capabilities.downgrade(&server_capabilities);
+
+        // Do not send anything, wait for monitor ready pdu
+        Ok(vec![])
+    }
+
+    fn handle_monitor_ready(&mut self) -> PduResult<Vec<SvcMessage>> {
+        let response = [
+            ClipboardPdu::Capabilities(self.capabilities.clone()),
+            ClipboardPdu::TemporaryDirectory(ClientTemporaryDirectory::new(self.temporary_directory.clone())?),
+            ClipboardPdu::FormatList(self.build_local_format_list()?),
+        ]
+        .into_iter()
+        .map(into_cliprdr_message)
+        .collect();
+
+        Ok(response)
+    }
+
+    fn handle_format_list_response(&mut self, response: FormatListResponse) -> PduResult<Vec<SvcMessage>> {
+        match response {
+            FormatListResponse::Ok => {
+                info!("CLIPRDR(clipboard) virtual channel has been initialized");
+                self.state = CliprdrState::Ready;
+            }
+            FormatListResponse::Fail => {
+                return self.handle_error_transition(ClipboardError::FormatListRejected);
+            }
+        }
+
+        Ok(vec![])
+    }
+}
+
+impl Default for Cliprdr {
+    fn default() -> Self {
+        Self {
+            state: CliprdrState::Initialization,
+            capabilities: Capabilities::new(
+                ClipboardProtocolVersion::V2,
+                ClipboardGeneralCapabilityFlags::USE_LONG_FORMAT_NAMES,
+            ),
+            temporary_directory: ".cliprdr".to_string(),
+        }
+    }
+}
+
+impl StaticVirtualChannel for Cliprdr {
+    fn channel_name(&self) -> ChannelName {
+        Self::CHANNEL_NAME
+    }
+
+    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<SvcMessage>> {
+        let pdu = decode::<ClipboardPdu>(payload)?;
+
+        if self.state == CliprdrState::Failed {
+            error!("Attempted to process clipboard static virtual channel in failed state");
+            return Ok(vec![]);
+        }
+
+        match pdu {
+            ClipboardPdu::Capabilities(caps) => self.handle_server_capabilities(caps),
+            ClipboardPdu::FormatListResponse(response) => self.handle_format_list_response(response),
+            ClipboardPdu::MonitorReady => self.handle_monitor_ready(),
+            _ => self.handle_error_transition(ClipboardError::UnimplementedPdu {
+                pdu: pdu.message_name(),
+            }),
+        }
+    }
+
+    fn compression_condition(&self) -> ironrdp_svc::CompressionCondition {
+        CompressionCondition::WhenRdpDataIsCompressed
+    }
+}
+
+fn into_cliprdr_message(pdu: ClipboardPdu<'static>) -> SvcMessage {
+    // Adding [`CHANNEL_FLAG_SHOW_PROTOCOL`] is a must for clipboard svc messages
+    SvcMessage::from(pdu).with_flags(ChannelFlags::SHOW_PROTOCOL)
+}
--- a/crates/ironrdp-cliprdr/src/pdu/format_list.rs
+++ b/crates/ironrdp-cliprdr/src/pdu/format_list.rs
@@ -6,11 +6,172 @@ use ironrdp_pdu::{cast_int, ensure_size, invalid_message_err, PduDecode, PduEnco
 
 use crate::pdu::{ClipboardPduFlags, PartialHeader};
 
+/// Clipboard format id.
+///
+/// [Standard clipboard formats](https://learn.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats)
+/// defined by Microsoft are available as constants.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ClipboardFormatId(u32);
+
+impl ClipboardFormatId {
+    /// Text format. Each line ends with a carriage return/linefeed (CR-LF) combination.
+    /// A null character signals the end of the data. Use this format for ANSI text.
+    pub const CF_TEXT: Self = Self(1);
+
+    /// A handle to a bitmap (HBITMAP).
+    pub const CF_BITMAP: Self = Self(2);
+
+    /// Handle to a metafile picture format as defined by the METAFILEPICT structure.
+    ///
+    /// When passing a CF_METAFILEPICT handle by means of DDE, the application responsible for
+    /// deleting hMem should also free the metafile referred to by the CF_METAFILEPICT handle.
+    pub const CF_METAFILEPICT: Self = Self(3);
+
+    /// Microsoft Symbolic Link (SYLK) format.
+    pub const CF_SYLK: Self = Self(4);
+
+    /// Software Arts' Data Interchange Format.
+    pub const CF_DIF: Self = Self(5);
+
+    /// Tagged-image file format.
+    pub const CF_TIFF: Self = Self(6);
+
+    /// Text format containing characters in the OEM character set. Each line ends with a carriage
+    /// return/linefeed (CR-LF) combination. A null character signals the end of the data.
+    pub const CF_OEMTEXT: Self = Self(7);
+
+    /// A memory object containing a BITMAPINFO structure followed by the bitmap bits.
+    pub const CF_DIB: Self = Self(8);
+
+    /// Handle to a color palette.
+    ///
+    /// Whenever an application places data in the clipboard that
+    /// depends on or assumes a color palette, it should place the palette on the clipboard as well.
+    /// If the clipboard contains data in the CF_PALETTE (logical color palette) format, the
+    /// application should use the SelectPalette and RealizePalette functions to realize (compare)
+    /// any other data in the clipboard against that logical palette. When displaying clipboard
+    /// data, the clipboard always uses as its current palette any object on the clipboard that is
+    /// in the CF_PALETTE format.
+    ///
+    /// NOTE: When transferred over `CLIPRDR`, [`crate::pdu::format_data::ClipboardPalette`] structure
+    /// is used instead of `HPALETTE`.
+    pub const CF_PALETTE: Self = Self(9);
+
+    /// Data for the pen extensions to the Microsoft Windows for Pen Computing.
+    pub const CF_PENDATA: Self = Self(10);
+
+    /// Represents audio data more complex than can be represented in a CF_WAVE standard wave format.
+    pub const CF_RIFF: Self = Self(11);
+
+    /// Represents audio data in one of the standard wave formats, such as 11 kHz or 22 kHz PCM.
+    pub const CF_WAVE: Self = Self(12);
+
+    /// Unicode text format. Each line ends with a carriage return/linefeed (CR-LF) combination.
+    /// A null character signals the end of the data.
+    pub const CF_UNICODETEXT: Self = Self(13);
+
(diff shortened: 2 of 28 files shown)


crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -4,3 +4,197 @@
...
+
+impl StaticVirtualChannel for Cliprdr {
+    fn channel_name(&self) -> ChannelName {
+        Self::CHANNEL_NAME
+    }
+
+    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<SvcMessage>> {
+        let pdu = decode::<ClipboardPdu>(payload)?;
+
+        let messages = match &self.state {
```
ibeckermayer, line 154: It's unclear to me what the advantage of maintaining this explicit state machine here is. Regardless of state, we are `match`ing on `pdu` and responding based on its type. AFAICT, this logic is only checking that the server gives us PDUs in the expected order -- however I think we can safely assume that the server is working appropriately, and simply respond based on what it's giving us.

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -4,3 +4,197 @@
...
+    }
+}
+
+impl StaticVirtualChannel for Cliprdr {
+    fn channel_name(&self) -> ChannelName {
+        Self::CHANNEL_NAME
+    }
+
+    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<SvcMessage>> {
+        let pdu = decode::<ClipboardPdu>(payload)?;
```
ibeckermayer, line 152: Good idea, I'm going to make a point to refactor rdpdr to do it like this.

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -4,3 +4,197 @@
...
+};
+
+use crate::pdu::FormatList;
+use thiserror::Error;
+use tracing::{error, info};
+
+pub mod clipboard_format {
+    pub const CF_TEXT: u32 = 1;
+    pub const CF_UNICODE: u32 = 13;
+}
```
ibeckermayer, line 22: Some info on where these come from could be helpful in case you want to support more formats in the future. Teleport's code for this is [here](https://github.com/gravitational/teleport/blob/36f3a623827dfe0277fa6561d9f2fd2c852fe935/lib/srv/desktop/rdp/rdpclient/src/cliprdr.rs#L712-L747)

crates/ironrdp-cliprdr/src/pdu/format_list.rs
```diff
@@ -13,6 +13,15 @@ pub struct ClipboardFormat {
     pub name: String,
 }
 
+impl ClipboardFormat {
+    pub fn new_standard(id: u32) -> Self {
```
ibeckermayer, line 17: Consider making a `ClipboardFormatId` enum so that it's clear what this id means.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
```diff
@@ -13,6 +13,15 @@ pub struct ClipboardFormat {
...
+        Self {
+            id,
+            name: String::new(),
```
ibeckermayer, line 20: In what cases is `name` empty? Maybe this type should be `Optional<String>` instead

crates/ironrdp-cliprdr/src/pdu/mod.rs
```diff
@@ -115,6 +115,22 @@ pub enum ClipboardPdu<'a> {
...
+            ClipboardPdu::FormatDataRequest(_) => "CLIPRDR_FORMAT_DATA_REQUEST",
+            ClipboardPdu::FormatDataResponse(_) => "CLIPRDR_FORMAT_DATA_RESPONSE",
+            ClipboardPdu::TemporaryDirectory(_) => "CLIPRDR_TEMP_DIRECTORY",
+            ClipboardPdu::Capabilites(_) => "CLIPRDR_CAPABILITIES",
+            ClipboardPdu::FileContentsRequest(_) => "CLIPRDR_FILECONTENTS_REQUEST",
+            ClipboardPdu::FileContentsResponse(_) => "CLIPRDR_FILECONTENTS_RESPONSE",
+            ClipboardPdu::LockData(_) => "CLIPRDR_LOCK_CLIPDATA",
+            ClipboardPdu::UnlockData(_) => "CLIPRDR_UNLOCK_CLIPDATA",
+        }
+    }
```
ibeckermayer, line 133: Looks like this is currently unused, leftover or do you expect to need it in the future?

pacmancoder: It is already used for error messages when unexpected messages are received, it will be used more in the follow-up PRs

crates/ironrdp-cliprdr/src/pdu/format_list.rs
pacmancoder: I thought about that but rejected the idea because clipboard IDs could be actually arbitrary, OS-specific, enums are not ideal for such things, I think.

Maybe the right solution would be the new type struct `struct ClipboardFromatId(pub u32)`
and then define constants inside its `impl`:

```rust
impl ClipboardFromatId {
    const CF_TEXT: Self = Self(1);
    const CF_UNICODE_TEXT: Self = Self(13);
    // ...
}
```

This should make usage more clear:

```rust
let format_id = ClipboardFromatId::CF_TEXT;

// some custom id
let format_id = ClipboardFromatId(42);
```

I think it will be good compromise between raw `u32` and too constrained `enum`

crates/ironrdp-cliprdr/src/lib.rs
pacmancoder: The advantage of this is that when we keep track of the state via a state machine, it gives us the opportunity to make error reporting more clean (on which exact communication stage we received unexpected/faulty PDU), it also allows to more logically split code into methods/types when code will grow upon adding different formats handling (initialization code is almost independent of an actual data transition logic). Also, it leaves less place for human error (the states will eventually store operation-specific fields which will be only accessible when matched on the correct state).

I think for such a complex thing as a clipboard it makes sense, but I may be wrong. @CBenoit what is your opinion on that?

crates/ironrdp-cliprdr/src/pdu/format_list.rs
pacmancoder: AFIK It is non-empty for os-specific formats, so it could be used for mapping remote and local clipboard IDs between each other.

I am not sure that we need `Option` here. because CLIPRDR specification tells us that when the name is not specified, it should be represented as the single null terminator, which translates to an empty string, Idk if it will make an API cleaner, @CBenoit what is your opinion on that?

crates/ironrdp-cliprdr/src/pdu/format_list.rs
CBenoit: > Maybe the right solution would be the new type struct `struct ClipboardFromatId(pub u32)` and then define constants inside its impl

Yes, this is generally a good approach in network protocols for parsing resilience. :+1: 

There is already some occurrence for this pattern:
- [`Code` in `ironrdp_graphics::rle` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-graphics/src/rle.rs#L386-L446)
- [`FailureCode` in `ironrdp_pdu::nego` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-pdu/src/nego.rs#L61-L85)
- [SecurityType in IronVNC](https://github.com/Devolutions/IronVNC/blob/547b482b2b516d176a80a5bdf09e1d4dae2878bc/crates/pdu/src/handshake.rs#L607-L708) (repository is private)

That’s the way to go when knowledge of _ALL_ the values is not required, not desired or not possible and we need to handle gracefully these situations.

crates/ironrdp-cliprdr/src/pdu/mod.rs
ibeckermayer: When I grep for `message_name`, only this definition comes up.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
CBenoit: It’s generally more idiomatic to distinguish between absence of value, and the empty string in Rust, so here are the questions:
- Do we actually need to distinguish the empty string from the "no string" state?
- How ignoring this distinction could be problematic? What kind of issues the consumer of the API may run into?

In this case, what is the meaning of the empty string in the context of `ClipboardFormat`? I believe it means "non standard" or "OS-specific" format. So my understanding is that the empty value is actually something different than _just_ an empty string. It’s likely that special handling is required in such situations.

Another way to see it: the consumer of the API may not even be aware that the empty string requires any special handling and may miss this detail, or may not know how to initialize the value properly. Could the user end up asking himself "I don’t know what I’m supposed to put inside this string"? It’s likely that `Option` will make it obvious that `None` is a valid avenue.

Not only `Option` would be a valid avenue, sum types in general are a good fit, and another way to make this point crystal clear would be to define another enum such as:
```rust
enum FormatKind {
    Standard,
    NonStandard { name: String },
}
```

I believe that all of this also applies to people reading, reviewing or auditing the code later (including the author himself after a few months :stuck_out_tongue:). The information conveyed by `String`, `Option<String>` or `FormatKind` is obviously not the same and plays a role in readability.

How the value itself is actually decoded or encoded can remain an implementation detail in the body of the encoding and decoding functions, i.e.: if a single null terminator means that "no name is specified" and that is also means the format is "standard", then we can say that `Option::None` or `FormatKind::Standard` can be represented by this single null terminator.

I didn’t review everything yet though, so it’s possible I’m misunderstanding something. I hope it helps you to decide on that!

ibeckermayer: Makes sense to use an empty string if it translates correctly to bytes.

crates/ironrdp-cliprdr/src/lib.rs
CBenoit: Though no state machine at all is perfectly viable and more readable for a small channel, I agree with you @pacmancoder 
If you think the CLIPRDR is going to grow into something way more complex that it currently is, then it makes sense to me. :+1:

ibeckermayer: > it gives us the opportunity to make error reporting more clean (on which exact communication stage we received unexpected/faulty PDU)

I think this point still begs the question of whether tracking unexpected/faulty PDUs is actually a valid concern for the client. IMO a faulty server implementation isn't worth designing for.

> it also allows to more logically split code into methods/types when code will grow upon adding different formats handling (initialization code is almost independent of an actual data transition logic)

I'm not seeing how this is the case. From my perspective it seems that we can break down the internal methods/types used by `process` to whatever degree of resolution we want to in either approach.

> Also, it leaves less place for human error (the states will eventually store operation-specific fields which will be only accessible when matched on the correct state).

I suppose that's true at least theoretically. IMO, presuming my other critiques hold water, then this concern alone isn't large enough to justify all of the additional code needed for explicit state management.

I do notice that there's a `Sequence` trait which we `impl` for various state machine types that do something similar, e.g. 

https://github.com/Devolutions/IronRDP/blob/b9de8ba635ac0974ad792dd4170f3a94907a9e70/crates/ironrdp-connector/src/connection.rs#L262

I can see an argument for keeping consistency there, however in my view the "implicit state machine" version gets the job done just fine and is much easier to read / follow, e.g. 9c7e9daca02400b1a1e4f9c19d7a08d2677abc3d:

```rust
impl StaticVirtualChannel for Cliprdr {
    // ...

    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<SvcMessage>> {
        let pdu = decode::<ClipboardPdu>(payload)?;

        match pdu {
            ClipboardPdu::Capabilites(caps) => self.handle_server_capabilities(caps),
            ClipboardPdu::MonitorReady => self.handle_monitor_ready(),
            ClipboardPdu::FormatListResponse(response) => self.handle_format_list_response(response),
            _ => todo!(),
        }
    }

    // ...
}
```

CBenoit: Fair arguments.

> > Also, it leaves less place for human error (the states will eventually store operation-specific fields which will be only accessible when matched on the correct state).
>
> I suppose that's true at least theoretically. IMO, presuming my other critiques hold water, then this concern alone > isn't large enough to justify all of the additional code needed for explicit state management.

Operation-specific fields only accessible when matching on the correct state is still an interesting property, but it’s likely we can benefit from this without having a different state for absolutely each step in the sequence.

When it’s really needed, this approach could works as well:
```rust
#[derive(Default)]
enum CliprdrState {
    #[default]
    Idle,
    SpecialState { some_value: u64 },
}

impl Cliprdr {
    fn handle_start_special_thing(&mut self, context: SpecialContext) -> PduResult<Vec<SvcMessage>> {
        self.state = CliprdrState::SpecialState {
            special_value: context.value,
        };

        Ok(Vec::new())
    }

    fn handle_continue_special_thing(&mut self, other_context: SpecialOtherContext) -> PduResult<Vec<SvcMessage>> {
        let CliprdrState::SpecialState { special_value } = std::mem::take(&mut self.state) else { // consume the special state
            return Err(…);
        };

        // do something with `special_value` and `other_context`
    }
}

impl StaticVirtualChannel for Cliprdr {
    // ...

    fn process(&mut self, payload: &[u8]) -> PduResult<Vec<SvcMessage>> {
        let pdu = decode::<ClipboardPdu>(payload)?;

        match pdu {
            ClipboardPdu::Capabilites(caps) => self.handle_server_capabilities(caps),
            ClipboardPdu::MonitorReady => self.handle_monitor_ready(),
            ClipboardPdu::FormatListResponse(response) => self.handle_format_list_response(response),
            ClipboardPdu::StartSpecialThing(context) => self.handle_start_special_thing(context),
            ClipboardPdu::ContinueSpecialThing(other_context) => self.handle_continue_special_thing(other_context),
            _ => todo!(),
        }
    }

    // ...
}
```

It all depends on how much there is to keep track of imo.

crates/ironrdp-cliprdr/src/pdu/mod.rs
pacmancoder: oops, you are right, I am using `name()` from `PduEncode` instead of `message_name()` thanks for spotting this! 👍

crates/ironrdp-cliprdr/src/lib.rs
pacmancoder: Ok, let's do this -- I'll remove the state machine, Currently for initialization we don't need it anyway. Later we will see if we need something to improve state management and refactor code appropriately.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
pacmancoder: @CBenoit Well, in the case of `ClipboardFormat` PDU structure there is no option to specify absence of string, only an empty string (null terminator only) is possible, therefore I decided to reflect that in the Rust strucutre, But I agree that is it not idiomatic.

I just don't like the fact that if we go with Option<String> we will have the following situation when comparing encoded and decode:

```rust
let original = FormatList::new_unicode(&[ClipboardFormat { id: 42, name: Some("".to_string()) }], true);
let encoded = encode(&original);

let decoded = decode::<FormatList>(&encoded);

let actual = FormatList::new_unicode(&[ClipboardFormat { id: 42, name: None }], true)
assert_eq!(decoded, actual);

assert_eq!(decoded, original); // Will fail, re-encoding produces different results
```


I think the best idea would be to hide implementation details of `ClibpoardFormat` (make fields private) and automatically store empty strings as None.

```rust
pub struct ClipboardFormat {
    id: u32,
    name: Option<String>
}

impl ClipboardFormat  {
    pub fn new(id: u32) -> Self { ... }
    /// NOTE: Method does nothing if empty string is specified
    pub fn with_name(self, name: &str) -> Self { ... }

    pub fn id(&self) -> u32 { ... }
    pub fn name(&self) -> Option<&str> {...}
}
```

CBenoit: Oh, I see. Sounds good to me!

CBenoit: Note: since this is an implementation detail, you could indeed use `String` internally and simply check for emptiness when `name()` is called.
At your preference.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
```diff
@@ -6,11 +6,164 @@ use ironrdp_pdu::{cast_int, ensure_size, invalid_message_err, PduDecode, PduEnco
...
+    name: Option<ClipboardFormatName>,
+}
+
+impl ClipboardFormat {
+    /// Creates unnamed `ClipboardFormat` with given id.
+    pub const fn new(id: ClipboardFormatId) -> Self {
+        Self { id, name: None }
+    }
+
+    /// Sets clipboard format name. If specified name is empty, then name will be set to `None`.
```
CBenoit, line 145: I imagined something more like
```suggestion
    /// Sets clipboard format name.
    ///
    /// This is typically used for custom/OS-specific formats where a name must be associated to
    /// the `ClipboardFomatId` in order to distinguish between vendors.
```

No need to mention how it’s handled internally. Feel free to change if this does not reflect the actual purpose of format names.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
```diff
@@ -6,11 +6,164 @@ use ironrdp_pdu::{cast_int, ensure_size, invalid_message_err, PduDecode, PduEnco
...
+    /// The data is a handle (HGLOBAL) to the locale identifier (LCID) associated with text in the
+    /// clipboard. When you close the clipboard, if it contains CF_TEXT data but no CF_LOCALE data,
+    /// the system automatically sets the CF_LOCALE format to the current input language. You can
+    /// use the CF_LOCALE format to associate a different locale with the clipboard text. An
+    /// application that pastes text from the clipboard can retrieve this format to determine which
+    /// character set was used to generate the text. Note that the clipboard does not support plain
+    /// text in multiple character sets. To achieve this, use a formatted text data type such as
+    /// RTF instead.The system uses the code page associated with CF_LOCALE to implicitly convert
+    /// from CF_TEXT to CF_UNICODETEXT. Therefore, the correct code page table is used for the
+    /// conversion.
```
CBenoit, line 89: Keep in mind that the first few lines must be a summary, and this summary is separated from the rest by a newline:

```suggestion
    /// The data is a handle (HGLOBAL) to the locale identifier (LCID) associated with text in the
    /// clipboard.
    ///
    /// When you close the clipboard, if it contains CF_TEXT data but no CF_LOCALE data,
    /// the system automatically sets the CF_LOCALE format to the current input language. You can
    /// use the CF_LOCALE format to associate a different locale with the clipboard text. An
    /// application that pastes text from the clipboard can retrieve this format to determine which
    /// character set was used to generate the text. Note that the clipboard does not support plain
    /// text in multiple character sets. To achieve this, use a formatted text data type such as
    /// RTF instead.The system uses the code page associated with CF_LOCALE to implicitly convert
    /// from CF_TEXT to CF_UNICODETEXT. Therefore, the correct code page table is used for the
    /// conversion.
```

You can check how it’s rendered using `cargo doc --open`.

See [rustdoc book](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components) for reference.

I think there are other places that need to be addressed in a similar fashion.

CBenoit (commented):
Looking pretty good to me. Small rust doc nitpicks below.

CBenoit (commented):
This is looking good to me. I’ll wait for ARC-143 branch to be merged on master before approving.

crates/ironrdp-cliprdr/src/pdu/format_list.rs
pacmancoder: done

crates/ironrdp-cliprdr/src/pdu/format_list.rs
pacmancoder: done

pacmancoder:
@CBenoit Sure, I am already doing it 😄 , w'll update PR soon

Issue #104, opened 2023-05-01
Support for user-provided static channels


CBenoit:
Provide an API for custom static channels.

Jira Issue: [ARC-147](https://devolutions.atlassian.net/browse/ARC-147)

[ARC-147]: https://devolutions.atlassian.net/browse/ARC-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #183, opened 2023-08-25
feat: API for custom static virtual channels

CBenoit:
Fixes https://github.com/Devolutions/IronRDP/issues/104

This adds a new API to extend IronRDP client with user-defined static virtual channels.

See the updated `crates/ironrdp-pdu/README.md` file for more on `WriteCursor` vs `WriteBuf`.

@ibeckermayer’s work was reviewed in this other PR: https://github.com/Devolutions/IronRDP/pull/178

Note that for some reason, when the DRDYNVC channel is enabled the client ends up losing the connection randomly during the dynamic channels advertising at the beginning of the active stage.
I’m not sure why, but I don’t see the advertising sequence happening on the current master. That would explain why the bug didn’t manifest before and it’s likely that we already have the bug on master.
This will be fixed in another PR.

Issue: ARC-147


pacmancoder (approved):
Nice! Great work @CBenoit and @ibeckermayer , lets merge it! 🎉

crates/ironrdp-dvc/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+
+            write!(f, "{}", channel.channel_name())?;
+        }
+
+        write!(f, "])")
+    }
+}
+
+impl Drdynvc {
+    pub const NAME: ChannelName = ChannelName::from_static(b"drdynvc\0");
```
ibeckermayer, line 73: nit: sometimes this interferes with some of the `PduError` macros who expect a `NAME: &'static str`, maybe we can change it to `CHANNEL_NAME`.

crates/ironrdp-pdu/src/gcc/network_data.rs
```diff
@@ -12,17 +12,92 @@ const CHANNELS_MAX: usize = 31;
...
+            .zip(inner.iter_mut())
+            .try_for_each(|(src, dst)| {
+                let c = u8::try_from(src).ok()?;
+                c.is_ascii().then(|| *dst = c)
+            })?;
+
+        Some(Self {
+            inner: Cow::Owned(inner),
+        })
+    }
```
ibeckermayer, line 60: This would be more useful if it just returned a `Self` and panicked on any errors. With `Option<Self>` it makes it impossible to use like

```rust
const CHANNEL_NAME: ChannelName = ChannelName::from_utf8("cliprdr");
```

crates/ironrdp-dvc/src/lib.rs
CBenoit: I see, yes we could do that! (I’m not sure [this](https://github.com/Devolutions/IronRDP/blob/cd91756305eea95702510e48367754eba5e2ac64/crates/ironrdp-pdu/src/macros.rs#L21) should be used outside of `PduEncode` / `PduDecode` though)

crates/ironrdp-pdu/src/gcc/network_data.rs
CBenoit: You won’t be able to use it like this because `from_utf8` is not `const`-friendly (and can’t be made `const` because we can’t use iterators in const context as of now). That’s why `from_utf8` is only a convenience function for dynamic contexts where it’s preferable to catch the error instead of panicking, and that’s why we also have `ChannelName::from_static`.

CBenoit: (Though it would be nice to revisit this when `const` is made more powerful in some future Rust version)

crates/ironrdp-pdu/README.md
```diff
@@ -1,3 +1,133 @@
...
+pub fn encode_buf<T: PduEncode + ?Sized>(pdu: &T, buf: &mut WriteBuf) -> PduResult<usize> {
+    let pdu_size = pdu.size();
+    let dst = buf.unfilled_to(pdu_size);
+    let mut cursor = WriteCursor::new(dst);
+    encode_cursor(pdu, &mut cursor)?;
+    let written = cursor.pos();
+    buf.advance(written);
+    Ok(written)
+}
+```
```
ibeckermayer, line 65: I find this explanation confusing, a couple of points that come to mind:

1. You say "One purpose is to facilitate buffer re-use: it’s okay to keep around the same `WriteBuf` and repeatedly write into it.", but there's no mention of calling `WriteBuf::clear()`, or why this is preferable to a `Vec<u8>`, which you can also call `clear()` on. (As I understand it, `WriteBuf::clear()` is preferable because it always keeps some non-trivial piece of initialized memory around, while also making sure that the internal `Vec<u8>` never becomes a gigantic memory hog over the course of the program. Vs using `Vec<u8>`, for which if you call `clear()` when it's 1GB long, will remain 1GB until it's destroyed).
2. "This way, it’s easy and safe to retrieve a
mutable slice of the unfilled (but already initialized) region. This slice can be used to build
a `WriteCursor`. Another purpose of the filled region tracking is to enable multiple items to be
written consecutively in the same buffer." This assumes we want `WriteCursor` around, but it's not clear why that's a constraint.

Generally I might reframe this by starting with our constraints, then a quick overview of the `Vec<u8>` model, and then a comparison with the `WriteBuf` model, explaining how `Vec<u8>` doesn't meet those constraints but `WriteBuf` does. 

All of that said, maybe it's not worth the time clarifying right now.

CBenoit: You are very right, this needs to be fleshed out further. To be honest, it’s still a WIP document, but I wanted to start writing it now since we already had some discussions in PR comments that I wanted to keep around before I forget. I’ll use your comment as reference when I get time for that :+1:

crates/ironrdp-dvc/src/lib.rs
pacmancoder: @CBenoit however `NAME` note that is used for a short form of `PduError` building macros such as `custom_err!` https://github.com/Devolutions/IronRDP/blob/feat/clipboard-text/crates/ironrdp-pdu/src/macros.rs#L120-L122

CBenoit: Maybe this should be de-sugared into `Self::NAME` for all macros so it’s more versatile, but I think there was a problem because of "conflicting" definitions with multiple traits. I have an idea on how to improve that long term.

PR #184, opened 2023-08-28
fix(pdu): don't be strict about platform type

elmarco:
Allow values outside the range of known values. FreeRDP has been using 0xFFFF - N for custom values.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
+++ b/crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
@@ -1,12 +1,10 @@
 #[cfg(test)]
 mod tests;
 
-use std::io;
+use std::{fmt, io};
 
 use bitflags::bitflags;
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
-use num_derive::{FromPrimitive, ToPrimitive};
-use num_traits::{FromPrimitive, ToPrimitive};
 
 use crate::rdp::capability_sets::CapabilitySetsError;
 use crate::PduParsing;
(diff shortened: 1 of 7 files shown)


crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
```diff
@@ -25,9 +24,45 @@ pub enum MajorPlatformType {
...
+            5 => Self::IOs,
+            6 => Self::OsX,
+            7 => Self::Android,
+            8 => Self::ChromeOs,
+            _ => Self::Other(value),
+        }
+    }
+}
+
+impl Into<u16> for MajorPlatformType {
```
CBenoit, line 47: `From` should be implemented over `Into`
```suggestion
impl From<MajorPlatformType> for u16 {
```

[Here is why](https://doc.rust-lang.org/std/convert/trait.Into.html#):
```
/// One should always prefer implementing `From` over [`Into`]
/// because implementing `From` automatically provides one with an implementation of [`Into`]
/// thanks to the blanket implementation in the standard library.
///
/// Only implement [`Into`] when targeting a version prior to Rust 1.41 and converting to a type
/// outside the current crate.
/// `From` was not able to do these types of conversions in earlier versions because of Rust's
/// orphaning rules.
```

I believe CI is failing because of the clippy lint for this.

CBenoit (commented):
Thank you! Indeed, this kind of type should be less "stiff" because knowledge of all the existing values is not required or not possible and we still need to handle gracefully the situation. But to me there is an issue (in the future) with this fix, a forwards-compatibility hazard that is hard to catch: if/when we add a new value to the enum, the `Other` variant for this value will stop being emitted, and downstream code matching on `Other` in order to handle that "other" value will break silently.

```rust
// Library //
pub enum MajorPlatformType {
    // …
    NewPopularOs = 20, // <- NEW
    Other(u16),
}

// Downstream //
match pdu.minor_platform_type {
    Other(20) => {
        // special logic for "new popular os"
    }
    _ => {} // ignore everything else
}
```

I think the right tool here is actually to not use an `enum` at all, but a new type wrapper around an `u16` instead.

There is already some occurrence for this pattern:
- [`Code` in `ironrdp_graphics::rle` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-graphics/src/rle.rs#L386-L446)
- [`FailureCode` in `ironrdp_pdu::nego` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-pdu/src/nego.rs#L61-L85)

What do you think?

crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
elmarco: done

elmarco:
> Thank you! Indeed, this kind of type should be less "stiff" because knowledge of all the existing values is not required or not possible and we still need to handle gracefully the situation. But to me there is an issue (in the future) with this fix, a forwards-compatibility hazard that is hard to catch: if/when we add a new value to the enum, the `Other` variant for this value will stop being emitted, and downstream code matching on `Other` in order to handle that "other" value will break silently.
> 
> ```rust
> // Library //
> pub enum MajorPlatformType {
>     // …
>     NewPopularOs = 20, // <- NEW
>     Other(u16),
> }
> 
> // Downstream //
> match pdu.minor_platform_type {
>     Other(20) => {
>         // special logic for "new popular os"
>     }
>     _ => {} // ignore everything else
> }
> ```
> 
> I think the right tool here is actually to not use an `enum` at all, but a new type wrapper around an `u16` instead.
> 
> There is already some occurrence for this pattern:
> 
>     * [`Code` in `ironrdp_graphics::rle` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-graphics/src/rle.rs#L386-L446)
> 
>     * [`FailureCode` in `ironrdp_pdu::nego` module](https://github.com/Devolutions/IronRDP/blob/55176ff99b8c72076c49ec4a2333ba9324798bc5/crates/ironrdp-pdu/src/nego.rs#L61-L85)
> 
> 
> What do you think?

Well, I am not convinced this is better than an enum here. If you use a newtype wrapping the underlying uint, then you won't notice `NewPopularOs` at all. But with the `Other(o)` case, at least you can get an error from the compiler that there is a new value to handle.

CBenoit:
> Well, I am not convinced this is better than an enum here. If you use a newtype wrapping the underlying uint, then you won't notice `NewPopularOs` at all. But with the `Other(o)` case, at least you can get an error from the compiler that there is a new value to handle.

Fair enough, I agree it’s not perfect by any means.

I think we have to consider this question: do we actually need to handle all the values for this? I could be wrong, but in general I don’t think it’s really required, I expect this to be more of a case-by-case basis ("I need to handle THIS specific case and ignore the rest"), or for logging / metrics purposes; in both cases it’s fine to use the value directly without matching on everything. I believe it’s common practice especially in network protocols where both backward and forward compatibility is a concern. Neither the client nor the server should break when the protocol gets extended. An older client (not handling the new value) should generally be able to function properly with a newer server, and vice-versa, and not handling the new value is not an immediate problem.

My main concern with the current approach is that `Other(20)` and `NewPopularOs` are conceptually the same thing, but are still considered to be different values. For instance, the `PartialEq` implementation derived will returns false for `Other(20) == NewPopularOs`. And, you still have the problem mentioned in my initial comment: when `ironrdp-pdu` is updated, `Other(20)` will stop being emitted, but user code might still be relying on this in non-exhaustive `match`, in which case you end up with the exact same issue. To me it’s worst, because if the user actually wrote special handling code for this case, then something is now guaranteed to be broken silently.

Another avenue would be for `Other` to not hold the original value at all so the user have to wait for `ironrdp-pdu` to be updated before he can do any special handling for `NewPopularOs` at all (and document that `Other` should not be relied upon), but I don’t like this either because one can’t know anymore what is the value from the original payload, and we lose the "round-trip" property because this is destructive parsing and we don’t know how it should be (re)encoded (back). Overall, it’s not a good option.

I think the newtype wrapper is best in this kind of situation because:
- it fixes the problem at hand (parsing resilience),
- there is no weird `PartialEq` logic (derived implementation will return true for `MajorPlatformType(20) == MajorPlatformType::NEW_POPULAR_OS` (correctness), and
- user can handle new values without waiting for `ironrdp-pdu` to be updated without fear of silent breakage, and could even support a custom platform that IronRDP would not support (extendable).

Also, the main downside (no exhaustive matching), can be mitigated by providing a higher-level abstraction, an enum, that could be obtained from the "raw" `MajorPlatformType` value. This pattern is applied in the `ironrdp-rdcleanpath` crate in order to convert from [a lower-level, future-proof, resilient but awkward to use type](https://github.com/Devolutions/IronRDP/blob/d0eed2515794b6bcb896675238ea5becb71e0150/crates/ironrdp-rdcleanpath/src/lib.rs#L49-L95), to [a high-level, easier to use type](https://github.com/Devolutions/IronRDP/blob/d0eed2515794b6bcb896675238ea5becb71e0150/crates/ironrdp-rdcleanpath/src/lib.rs#L261-L277), and [user code](https://github.com/Devolutions/IronRDP/blob/d0eed2515794b6bcb896675238ea5becb71e0150/crates/ironrdp-web/src/session.rs#L659-L674) will stop compiling if action is required, because in this case we should really make sure to handle everything.
If you think that exhaustive matching is really useful for this type, I would prefer this approach even if it requires more boilerplate.

Overall, it’s pretty much the same requirements and needs as HTTP status code, and it’s exactly [how the `http` crate deals with it](https://docs.rs/http/0.2.9/http/status/struct.StatusCode.html).

CBenoit (approved):
Thank you! Sorry for the delay.

PR #186, opened 2023-08-28
feat(CoreCapability): Core Capability PDUs and refactoring ironrdp-rdpdr internals

ibeckermayer:
Adds
- [2.2.2.7 Server Core Capability Request (DR_CORE_CAPABILITY_REQ)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/702789c3-b924-4bc2-9280-3221bc7d6797)
- [2.2.2.8 Client Core Capability Response (DR_CORE_CAPABILITY_RSP)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/f513bf87-cca0-488a-ac5c-18cf18f4a7e1)

This also refactors rdpdr code (inspired by @pacmancoder's design) such that all PDUs are packed into an `RdpdrPdu` enum, and only that enum is `PduEncode`/`PduDecode`.

Note: It looks like I managed to goof up my git history somehow s.t. some of @CBenoit's commits are considered part of this PR. I will be sure to squash-and-merge so that `master`'s commit history is sane.


crates/ironrdp-rdpdr/Cargo.toml
```diff
@@ -1,14 +1,13 @@
 [package]
 name = "ironrdp-rdpdr"
 version = "0.1.0"
-description = "RDPDR static channel for file system implemented as described in MS-RDPEFS"
+description = "RDPDR channel implementation."
```
CBenoit, line 4: The original description was too verbose?

crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -30,42 +27,49 @@ use crate::pdu::efs::{
 pub struct Rdpdr {
     /// TODO: explain what this is
     computer_name: String,
+    /// See [`CapabilityMessage`]
```
CBenoit, line 30: Nit: not a big deal, but it seems to me that this doc string is of little use (not an explanation and we can already jump to the definition of CapabilityMessage without the link it adds)

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -264,3 +163,479 @@ impl From<ClientNameRequestUnicodeFlag> for CharacterSet {
...
+
+/// [2.2.1.2.1 Capability Message (CAPABILITY_SET)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/f1b9dd1d-2c37-4aac-9836-4b0df02369ba)
+#[derive(Debug, Clone, Copy)]
+pub struct CapabilityMessage {
+    header: CapabilityHeader,
+    capability_data: CapabilityData,
+}
+
+impl CapabilityMessage {
+    const NAME: &str = "CapabilityMessage";
```
CBenoit, line 262: Nit: let’s use the upper cased structure name from Microsoft documentation for consistency (this is what we’ll get in errors)
```suggestion
    const NAME: &str = "CAPABILITY_SET";
```
In some cases there is no such name, and in such cases the name of the Rust type is used instead (`ironrdp_pdu::geometry::InclusiveRectangle` for instance)

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -264,3 +163,479 @@ impl From<ClientNameRequestUnicodeFlag> for CharacterSet {
         }
     }
 }
+
+/// [2.2.2.7 Server Core Capability Request (DR_CORE_CAPABILITY_REQ)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/702789c3-b924-4bc2-9280-3221bc7d6797)
+/// [2.2.2.8 Client Core Capability Response (DR_CORE_CAPABILITY_RSP)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/f513bf87-cca0-488a-ac5c-18cf18f4a7e1)
+#[derive(Debug)]
+pub struct CoreCapability {
+    pub num_capabilities: u16,
```
CBenoit, line 171: It’s better to not keep this field because:
- it is redundant with `capabilities.len()`
- it’s easy to end up with `num_capabilities != capabilities.len()`

When encoding, you can use the `cast_length!` macro to handle the usize → u16 conversion properly:
```rust
dst.write_u16(cast_length!("CoreCapability", "numCapabilities", self.capabilities.len()?);
```
(first parameter is not required when `<Self as Pdu>::NAME` is defined)

In the unlikely event of `self.capabilities.len()` value being too big to fit a `u16`, the final error will look like:
```
[CoreCapability] invalid `numCapabilities`: too many elements, caused by: out of range integral type conversion attempted
```

crates/ironrdp-rdpdr/Cargo.toml
ibeckermayer: The RDPDR static channel isn't entirely defined in MS-RDPEFS, there are other MS-* which define significant portions of it's functionality such as MS-RDPESC (smartcards) or MS-RDPEPC (printers), so I decided to just keep it high level.

crates/ironrdp-rdpdr/src/lib.rs
ibeckermayer: 001d937d366b978775b62aa9455105ebd60bfc6c

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: 001d937d366b978775b62aa9455105ebd60bfc6c

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: 554e22a7ed8bc3c665e9ef0045c77ff9021b1450

crates/ironrdp-rdpdr/Cargo.toml
CBenoit: Fair enough

CBenoit (commented):
Looks good to me! You may want to update the description in `ARCHITECTURE.md` as well, but otherwise I’m okay with squashing and merging this. Maybe wait for @pacmancoder if you want.

CBenoit (approved):

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -127,67 +25,81 @@ pub enum VersionAndIdPduKind {
...
+                    "PacketId",
+                    "invalid value"
+                ))
+            }
+        };
+
+        ensure_size!(ctx: kind.name(), in: src, size: Self::FIXED_PART_SIZE);
+
+        Ok(Self {
+            version_major: src.read_u16(),
```
pacmancoder, line 90: [nit] This is definitely a nitpicking by me, you could ignore this comment if you think otherwise, but In my opinion struct field initialization should not contain code, which depends on execution order -- I always use additional `let`s for such cases:

```rust
let version_major = src.read_u16();
let version_minor = src.read_u16();
let client_id = src.read_u32();

Ok(Self {
    version_major,
    version_minor,
    client_id,
    kind,
})
```

But in case of PDUs I guess it is not so important, as it is hard to mix field ordering during consecutive refactoring due to immutable nature of PDUs

pacmancoder (approved):
LGTM, great work!

PR #187, opened 2023-08-29
build(client): update softbuffer dependency

CBenoit:


crates/ironrdp-client/Cargo.toml
```diff
@@ -34,25 +34,25 @@ ironrdp-tokio.workspace = true
...
-winit = "0.28.3"
+softbuffer = "0.3"
+winit = "0.28"
 
 # CLI
 clap = { version = "4.2", features = ["derive", "cargo"] }
-exitcode = "1.1.2"
-inquire = "0.6.0"
+exitcode = "1.1"
+inquire = "0.6"
```
CBenoit, line 43: `cargo upgrade` (`cargo-edit`) is now able to handle versions without always adding the patch number

pacmancoder: 🔥

pacmancoder (approved):
LGTM!

PR #188, opened 2023-08-30
Make it easier to use IronRDP with Tokio

probakowski:
Various types used in connection sequence were not picked up as `Send` or `Sync` which made it impossible to use inside `tokio::spawn` blocks

### DIFF
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -11,14 +11,17 @@ pub trait FramedRead {
     fn read<'a>(
         &'a mut self,
         buf: &'a mut BytesMut,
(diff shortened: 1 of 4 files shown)


CBenoit:
Build is failing for `ironrdp-futures`. Also, I think this approach will be a problem for `ironrdp-web` because [`WebSocket` is `!Send`](https://docs.rs/gloo-net/latest/gloo_net/websocket/futures/struct.WebSocket.html#impl-Send-for-WebSocket), we probably need another approach.

CBenoit:
I found an alternative approach using [GATs](https://github.com/rust-lang/rust/issues/44265).

probakowski:
Closing in favor of https://github.com/Devolutions/IronRDP/pull/189

PR #190, opened 2023-09-05
Remove redundant debug comment

pacmancoder:
@CBenoit I accidentally discovered I made an oopsie and forgot to remove inappropriate `eprintln!` which was left after pointer rendering implementation, sorry 😨 It somehow slipped through the cracks during self-review & PR


CBenoit (approved):
It’s okay, let’s remove this :smile:

PR #191, opened 2023-09-08
feat(web): use softbuffer to draw into the canvas

CBenoit:
cc @ngirot-devolutions 

Instead of copying and sending image buffers to JavaScript, the WASM module now draws into the canvas by itself.
This removes some overhead associated with the previous approach and open the door for further optimizations.

In order to achieve good performance, the newest API of `softbuffer@0.3.0` is used: the ["owned buffer"](https://docs.rs/softbuffer/0.3.0/softbuffer/struct.Buffer.html) that can be written into by us with direct access and `present_with_damage` to apply partial updates.
The presentation itself is currently not yet "no-copy" in the case of the web backend because the current API of `softbuffer` is expecting a pixel buffer in the BGRX format while the underlying canvas can only takes RGBA pixels. There is an [open issue](https://github.com/rust-windowing/softbuffer/issues/98) for this.

There was a bug with the `present_with_damage` implementation for the web backend.
I fixed the issue and opened a PR to upstream the patch: https://github.com/rust-windowing/softbuffer/pull/150
The cargo dependency patch will be removed once the fix is published on crates.io.

Issue: ARC-164


pacmancoder (approved):
Looks nice! 👍

PR #192, opened 2023-09-12
doc: elaborate high-level documentation

CBenoit:
cc @pacmancoder @ibeckermayer 

I took some time to elaborate the high-level documentation of the repository a bit more.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -12,78 +12,226 @@ This document describes the high-level architecture of IronRDP.
 
 This section talks briefly about various important directories and data structures.
 
+Note also which crates are **API Boundaries**.
+Remember, [rules at the boundary are different](https://www.tedinski.com/2018/02/06/system-boundaries.html).
+
 ### Core Tier
 
 Set of foundational libraries for which strict quality standards must be observed.
+Note that all crates in this tier are **API Boundaries**.
 Pay attention to the "**Architecture Invariant**" sections.
-
-- `crates/ironrdp`: meta crate re-exporting important crates.
-- `crates/ironrdp-pdu`: PDU encoding and decoding. (TODO: talk about important types and traits such as PduDecode, PduEncode…)
-- `crates/ironrdp-graphics`: image processing primitives.
-- `crates/ironrdp-svc`: traits to implement RDP static virtual channels.
-- `crates/ironrdp-dvc`: DRDYNVC static channel implementation and traits to implement dynamic virtual channels.
-- `crates/ironrdp-rdpdr`: RDPDR channel implementation.
-- `crates/ironrdp-rdpsnd`: RDPSND static channel for audio output implemented as described in MS-RDPEA.
-- `crates/ironrdp-connector`: state machines to drive an RDP connection sequence.
-- `crates/ironrdp-session`: state machines to drive an RDP session.
-- `crates/ironrdp-input`: utilities to manage and build input packets.
-- `crates/ironrdp-rdcleanpath`: RDCleanPath PDU structure used by IronRDP web client and Devolutions Gateway.
-- `crates/ironrdp-error`: lightweight and `no_std`-compatible generic `Error` and `Report` types.
-  The `Error` type wraps a custom consumer-defined type for domain-specific details (such as `PduErrorKind`).
 
 **Architectural Invariant**: doing I/O is not allowed for these crates.
 
 **Architectural Invariant**: all these crates must be fuzzed.
 
-**Architectural Invariant**: no non-essential dependency is allowed.
-
 **Architectural Invariant**: must be `#[no_std]`-compatible (optionally using the `alloc` crate). Usage of the standard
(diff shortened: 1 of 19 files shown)


ARCHITECTURE.md
```diff
@@ -41,49 +31,201 @@ Pay attention to the "**Architecture Invariant**" sections.
...
+will always try to fix things when it's a minor change.
+
+#### [`crates/ironrdp-acceptor`](./crates/ironrdp-acceptor) (@mihneabuz)
+
+State machines to drive an RDP connection acceptance sequence
+
+#### [`crates/ironrdp-server`](./crates/ironrdp-server) (@mihneabuz)
+
+Extendable skeleton for implementing custom RDP servers.
+
```
CBenoit, line 217: Hi! @mihneabuz
Since you are the one who contributed them, I figured that maybe you could be the maintainer for these.
Is it okay for you?

Note that I’m willing to move these crates to the extra tier at some point in the future, when we are done consolidating what we already have in the core and extra tiers.

mihneabuz: Sure, that sounds great! :)

crates/ironrdp-pdu/README.md
```diff
@@ -19,88 +19,154 @@ TODO: elaborate this section
 
 ## Difference between `WriteBuf` and `WriteCursor`
 
+`WriteCursor` is a wrapper around `&mut [u8]` and its purpose is to:
+
+- Provide convenient methods such as `write_u8`, `write_u16`, `write_u16_be`, etc.
+- Guarantee syscall-free, infaillible write access to a continuous slice of memory.
```
ibeckermayer, line 25:
```suggestion
- Guarantee syscall-free, infallible write access to a continuous slice of memory.
```

ARCHITECTURE.md
CBenoit: Great to hear, thank you!

crates/ironrdp-pdu/README.md
```diff
@@ -130,3 +196,129 @@ When hiding some fields is really required, one of the following approach is sug
...
+    Unknown(u32),
+}
+```
+
+This type can indeed hold additional values in its fallback variant, and the implementation can
+remain compatible with future protocol versions as desired.
+
+However, this solution is not without its issues. There is a hard to catch forward-compatibility
+hazard at the library level: when a new value, such as `ThirdValue = 2`, is added to the enum, the
+`Unknown` variant for this value will no longer be emitted by the library. This can lead to _silent_
```
ibeckermayer, line 245: I don't understand what it means for a value to be "emitted" by the library.

ibeckermayer (approved):

crates/ironrdp-pdu/README.md
CBenoit: I meant something like "the library does not construct and return the value".

To illustrate:

```rust
// library code

#[repr(u32)]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum MyNetworkCode {
    FirstValue = 0,
    SecondValue = 1,
    Unknown(u32),
}

impl MyNetworkCode {
    fn parse_network_code(value: u32) -> MyNetworkCode {
        match value {
            0 => MyNetworkCode::FirstValue,
            1 => MyNetworkCode::SecondValue,
            _ => MyNetworkCode::Unknown(value),
        }
    }
}

// user code

fn handle_network_code(reader: /* … */) {
    let value = reader.read_u32();
    let code = MyNetworkCode::from_u32(value);
    
    if code == MyNetworkCode::Unknown(2) {
        // The library doesn’t know about this value yet, but we need to handle it because X
    }
}
```

Once the library is updated to handle the third value we get:

```rust
// library code

#[repr(u32)]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum MyNetworkCode {
    FirstValue = 0,
    SecondValue = 1,
    ThirdValue = 2, // NEW
    Unknown(u32),
}

impl MyNetworkCode {
    fn parse_network_code(value: u32) -> MyNetworkCode {
        match value {
            0 => MyNetworkCode::FirstValue,
            1 => MyNetworkCode::SecondValue,
            2 => MyNetworkCode::ThirdValue, // NEW
            _ => MyNetworkCode::Unknown(value),
        }
    }
}

// user code (unchanged)

fn handle_network_code(reader: /* … */) {
    let value = reader.read_u32();
    let code = MyNetworkCode::from_u32(value);
    
    if code == MyNetworkCode::Unknown(2) { // <- the library does not construct this value as it used to…
        // …  the special case is not handled anymore; no warning and no error
        // was emitted by the compiler, so it’s very easy to overlook this in practice
    }
}
```

The bottom line is to avoid types which may represent the same thing in two different ways

ibeckermayer: I see, thanks for the example. Maybe it's worth adding a TL;DR to the top of that section (rather than at the end), otherwise all looks good to me.

CBenoit: You are right!

PR #196, opened 2023-09-22
Encapsulates [de]chunkification in an `SvcPreprocessor`

ibeckermayer:


pacmancoder (approved):
Looks good to me!

However, I have a few thoughts on API -- I think that we may hit some roadblocks in the future when using the same hard-wired in-memory dechunkification/chunkification logic for all SVC channels. E.g. when some kind of file transfer will be implemented over SVC, we may want to parse PDU header from the first chunk, and then perform direct I/O operations on each chunk received instead of accumulating them on RAM for a single `process` call, avoiding in-memory dechunkification at all. Same thing with chunkification, It would be great if the channels could have some API in future to set `SvcMessage` kind as chunked to perform manual chunkification logic.

Currently, this is most likely not an issue and could be improved in the future if we are in need of such optimizations.

crates/ironrdp-svc/src/lib.rs
```diff
@@ -126,81 +133,132 @@ pub trait StaticVirtualChannel: AsAny + fmt::Debug + Send + Sync {
...
+
+    /// Takes a single PDU and breaks it into chunks prefixed with a [`ChannelPduHeader`].
+    ///
+    /// Each chunk is at most `max_chunk_len` bytes long (not including the Channel PDU Header).
+    ///
+    /// For example, if the PDU is 4000 bytes long and `max_chunk_len` is 1600, this function will
+    /// return 3 chunks, each 1600 bytes long, and the last chunk will be 800 bytes long.
+    ///
+    /// [[ Channel PDU Header | 1600 bytes of PDU data ] [ Channel PDU Header | 1600 bytes of PDU data ] [ Channel PDU Header | 800 bytes of PDU data ]]
+    fn chunkify_one(&self, message: SvcMessage, max_chunk_len: usize) -> PduResult<Vec<WriteBuf>> {
```
CBenoit, line 199: Looks like there is no benefit for `chunkify` and `chunkify_one` to be methods of `SvcProcessor`, the `&self` parameter is not used

crates/ironrdp-svc/src/lib.rs
```diff
@@ -126,81 +133,132 @@ pub trait StaticVirtualChannel: AsAny + fmt::Debug + Send + Sync {
...
-    for message in messages {
-        results.extend(chunkify_one(message, max_chunk_len)?);
-    }
-    Ok(results)
+/// SvcPreprocessor is used to chunkify/de-chunkify static virtual channel PDUs.
+#[derive(Debug)]
+pub struct SvcPreprocessor {
+    /// Buffer for de-chunkification of clipboard PDUs. Everything bigger than ~1600 bytes is
+    /// usually chunked when transfered over svc.
+    chunked_pdu: Option<Vec<u8>>,
```
CBenoit, line 141: Tell me if I’m wrong: `Option` is used so that it’s easy to `take` the inner vector out and to hand it to the user, but at no point you really need to handle a `None` state (a new vector is always inserted when `None`). This sounds like you could use `Vec<u8>` directly, and `std::mem::take` it instead. This will leave a default vector in place (no allocation is performed for empty vectors, created using `Vec::new`, `Vec::default` or `vec![]`), and remove the need to `unwrap` though I’m confident there was no risk of panicking in this case.

crates/ironrdp-svc/src/lib.rs
```diff
@@ -126,81 +133,132 @@ pub trait StaticVirtualChannel: AsAny + fmt::Debug + Send + Sync {
...
-/// Each chunk is at most `max_chunk_len` bytes long (not including the Channel PDU Header).
-pub fn chunkify(messages: Vec<SvcMessage>, max_chunk_len: usize) -> PduResult<Vec<WriteBuf>> {
-    let mut results = vec![];
-    for message in messages {
-        results.extend(chunkify_one(message, max_chunk_len)?);
-    }
-    Ok(results)
+/// SvcPreprocessor is used to chunkify/de-chunkify static virtual channel PDUs.
+#[derive(Debug)]
+pub struct SvcPreprocessor {
```
CBenoit, line 138: I feel that `ChunkProcessor` or `SvcChunkProcessor` would be a somewhat more obvious name for this helper

CBenoit (commented):
Nice, I was observing the idea of building something similar.

However, I agree with @pacmancoder’s points.

I was thinking about a somewhat more intermediate approach. Tell me what you think.

- Remove `preprocessor` and `preprocessor_mut` trait methods
- Add a new trait method, something like `fn chunked_input_supported(&self) -> bool`. This method advertises whether the channel implementation is able to process fragmented input manually or not
- When `chunked_input_supported` returns `true`, either call `process_chunked` or `process` depending on whether the input is fragmented or not
- When `chunked_input_supported` returns `false`, never call `process_chunked` on this channel, only `process` will be called
- Manage the `SvcProcessor` accumulator for each channel that returned `false` in `ironrdp-session`, stored in something like `HashMap<TypeId, SvcProcessor>` or `HashMap<StaticChannelId, SvcProcessor>`
- Most channels don’t have to do anything else than implementing `process`, when a custom chunk processing logic is required, the channel can return `true` and implement `process_chunked`

crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub struct Rdpdr {
     /// TODO: explain what this is
     computer_name: String,
     capabilities: Vec<CapabilityMessage>,
+    preprocessor: SvcPreprocessor,
```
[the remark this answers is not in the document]
CBenoit: [I also think the whole chunkify and dechunkify could live somewhere else](https://github.com/Devolutions/IronRDP/pull/178#discussion_r1298631881). Though I’m okay with not bothering about that right now.

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -100,25 +99,15 @@ impl Processor {
                 Some(drdynvc_id) if channel_id == drdynvc_id => self.process_dyvc(data_ctx),
                 _ => {
                     if let Some(static_channel) = self.static_channels.get_by_channel_id_mut(channel_id) {
-                        let mut payload = data_ctx.user_data;
+                        let payload = static_channel
+                            .preprocessor_mut()
+                            .dechunkify(data_ctx.user_data)
+                            .map_err(crate::SessionError::pdu)?
+                            .unwrap_or_default();
```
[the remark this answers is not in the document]
ibeckermayer: 74c47ab415221f6fa747e73f3d184bd9b93877ee

crates/ironrdp-svc/src/lib.rs
ibeckermayer: Technically true, I added these as `&self` methods primarily for having the symmetry of `preprocessor.chunkify` vs `preprocessor.dechunkify` (or `chunk_processor.chunkify`/`chunk_processor.dechunkify` after I rename). I prefer this style but will change it if you _truly_ insist :). For now I've compromised: 4dad3b2486539038cee09cc3d6e7614b6fe2834f

crates/ironrdp-svc/src/lib.rs
ibeckermayer: 5409cbf6302508d1d384b0065545f4904479db0c

pacmancoder:
UPD: After discussion in Slack /w @ibeckermayer and @CBenoit : 

We actually don't need any "custom" chunkification/dechunkification logic for SVC, because in the real world there are only SvcMessages with size <= 1MB, such size is a small enough to do chunkification/dechunkification in memory and provide a simplified interface for SvcChannels, so my comment above about custom preprocessing is resolved

crates/ironrdp-svc/src/lib.rs
CBenoit: Fair enough, yes, I generally tend to prefer removing non required dependencies so we don’t end up with a "Banana Gorilla Jungle" problem and my reflex would be to keep `chunkify` as a free-standing function, unrelated to `SvcPreprocessor`, like it was originally. Also, I see `SvcProcessor` as a "chunk accumulator", so I don’t feel bad if it doesn’t deal with the inverse operation.

That’s my take on it, but I understand your desire to keep the symmetry, and in this case it’s reasonable, so you can consider that I’m not _truly_ insisting :stuck_out_tongue:

crates/ironrdp-svc/src/lib.rs
```diff
@@ -126,81 +120,131 @@ pub trait StaticVirtualChannel: AsAny + fmt::Debug + Send + Sync {
...
+        let last = self.process_header(payload)?;
 
-        // Encode the header for this chunk.
-        encode_buf(&header, &mut chunk)?;
-        // Append the piece of the encoded_pdu that belongs in this chunk.
-        chunk.write_slice(&encoded_pdu[chunk_start_index..chunk_end_index]);
-        // Push the chunk onto the results.
-        chunks.push(chunk);
+        // Extend the chunked_pdu buffer with the payload
+        self.chunked_pdu.extend_from_slice(payload); // unwrap ok because we just ensured we have a chunked_pdu buffer above
```
CBenoit, line 158: Leftover comment

```suggestion
        self.chunked_pdu.extend_from_slice(payload);
```

CBenoit (approved):
Thank you!

Issue #115, opened 2023-05-01
User-Initiated message: Shutdown Request


CBenoit:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/27915739-8f77-487e-9927-55008af7fd68

[TODO comment in code](https://github.com/Devolutions/IronRDP/blob/ccdd3825b08c64d7736dd3207c3d7281d0445b30/crates/client/src/rdp.rs#L224)

Jira Issue: [ARC-148](https://devolutions.atlassian.net/browse/ARC-148)

[ARC-148]: https://devolutions.atlassian.net/browse/ARC-148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #198, opened 2023-09-27
Server sample & mstsc related connection fixes

elmarco:
This is WIP branch to add a simple server, so we can further debug issues more easily.

In particular mstsc fails to connect succesfully, and I am having trouble to debug the issue. Notably, not being able to setup wireshark to decode the protocol...

Help welcome!


elmarco:
> @elmarco luckily, I have just the right guide and tooling for you to properly decode mstsc traffic in Wireshark! This should help you diagnose the issue quickly: https://github.com/awakecoding/wireshark-rdp

thanks, unfortunately that didn't help me much :( I am trying to dump ironrdp, on server side (and sfreerdp-server).  `SSLKEYLOGFILE=` doesnt seem to be taken into account there, I'll need to investigate why.  I tried to set the RSA key list myself, but that didnt help either.. I am bit clueless, I would appreciate your help! :)

crates/ironrdp/examples/server.rs
```diff
@@ -0,0 +1,169 @@
...
+
+fn acceptor(cert_path: &str, key_path: &str) -> anyhow::Result<TlsAcceptor> {
+    let cert = certs(&mut BufReader::new(File::open(cert_path)?))?[0].clone();
+    let key = pkcs8_private_keys(&mut BufReader::new(File::open(key_path)?))?[0].clone();
+
+    let server_config = ServerConfig::builder()
+        .with_safe_defaults()
+        .with_no_client_auth()
+        .with_single_cert(vec![rustls::Certificate(cert)], rustls::PrivateKey(key))
+        .expect("bad certificate/key");
```
CBenoit, line 101: > thanks, unfortunately that didn't help me much :( I am trying to dump ironrdp, on server side (and sfreerdp-server). SSLKEYLOGFILE= doesnt seem to be taken into account there, I'll need to investigate why.

@elmarco  You need to enable support for SSLKEYLOGFILE at this place.
That’s why we do for the client: https://github.com/Devolutions/IronRDP/blob/6b660fa298a3d7bc92ea3dc80468a7d4cbc45bc2/crates/ironrdp-tls/src/lib.rs#L22-L23

awakecoding:
> > @elmarco luckily, I have just the right guide and tooling for you to properly decode mstsc traffic in Wireshark! This should help you diagnose the issue quickly: https://github.com/awakecoding/wireshark-rdp
> 
> thanks, unfortunately that didn't help me much :( I am trying to dump ironrdp, on server side (and sfreerdp-server). `SSLKEYLOGFILE=` doesnt seem to be taken into account there, I'll need to investigate why. I tried to set the RSA key list myself, but that didnt help either.. I am bit clueless, I would appreciate your help! :)

is there a particular reason why the technique where you'd try to dump the SChannel pre-master secrets out of the Windows *client* machine running mstsc would not be suitable in this case? It's either that, or SSLKEYLOGFILE in the IronRDP server, but as my colleague Benoit pointed out there's a small code modification to be made there, I don't think it's enabled in the server right now.

CBenoit:
(Converted the PR to an draft explicitly, given the current title)

crates/ironrdp/examples/server.rs
elmarco: thanks, done

elmarco:
> @elmarco luckily, I have just the right guide and tooling for you to properly decode mstsc traffic in Wireshark! This should help you diagnose the issue quickly: https://github.com/awakecoding/wireshark-rdp

I can get the decrypted TLS now. However, it doesn't interpret it as RDP. Any idea? 

image

elmarco:
> @elmarco hum... that's odd, normally the RDP dissector kicks in automatically when TLS gets decrypted. This doesn't look like the Windows version of Wireshark, maybe it's an older release on another OS? In any case, if you look at the last packets exchanged, can you spot anything interesting?

Wireshark 4.0.8 (Git commit 81696bb74857) (wireshark-4.0.8-2.fc38.x86_64)

> 
> It may be easier to discuss some of this in the FreeRDP chat, which we use for more than just FreeRDP these days: https://www.freerdp.com/ (see the matrix.org channel, I hang in there from time to time).

I am there too, ok.

> 
> I cloned your branch, and I'm unsure how to really get started running the sample server. Here's what I did so far:
> 
> `cargo run --example=server --features=server -- --host 127.0.0.1 --port 4489`
> 
> I see there's a parameter for a server certificate which I didn't pass, but I don't see how I'd be supposed to pass server username+password that the server would use to authenticate the client. If I can reproduce the issue on my side, I could probably get the Wireshark part right

There is no user/pass handling atm in ironrdp-server (I suppose it accepts all users): https://github.com/Devolutions/IronRDP/blob/master/crates/ironrdp-acceptor/src/connection.rs#L329

You need to pass cert/keys (I generated mine with something like `openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out MyCertificate.crt -keyout MyKey.key`)

elmarco:
> I hit a protocol error in mstsc, and the IronRDP server shows the following errors:

Yep, same error here. nice! :-)

awakecoding:
I got the traffic decrypted in Wireshark, and see two fishy things:

A malformed packet sent by the server at the beginning, containing the GCC conference create response:
image

Later down the connection sequence, channel id 1003 is weirdly undissected, before causing a protocol error in the client:
image

With FreeRDP, I don't have a traffic capture, but I have a bunch of warnings:

```
 type=PDU_TYPE_DATA[0x00000007], tpktLength=22, remainingLength=16
 recv Synchronize Data PDU (0x1F), length: 8
 [CONNECTION_STATE_FINALIZATION_CLIENT_SYNC] received flag FINALIZE_SC_SYNCHRONIZE_PDU [0x00000001]
 CONNECTION_STATE_FINALIZATION_CLIENT_SYNC --> CONNECTION_STATE_FINALIZATION_CLIENT_COOPERATE
 x=38, y=59, w=1024, h=768
 x=1926, y=27, w=1024, h=768
 type=PDU_TYPE_DATA[0x00000007], tpktLength=26, remainingLength=20
 recv Control Data PDU (0x14), length: 12
 [CONNECTION_STATE_FINALIZATION_CLIENT_COOPERATE] received flag FINALIZE_SC_CONTROL_COOPERATE_PDU [0x00000002]
 CONNECTION_STATE_FINALIZATION_CLIENT_COOPERATE --> CONNECTION_STATE_FINALIZATION_CLIENT_GRANTED_CONTROL
 type=PDU_TYPE_DATA[0x00000007], tpktLength=26, remainingLength=20
 recv Control Data PDU (0x14), length: 12
 [CONNECTION_STATE_FINALIZATION_CLIENT_GRANTED_CONTROL] received flag FINALIZE_SC_CONTROL_GRANTED_PDU [0x00000004]
 CONNECTION_STATE_FINALIZATION_CLIENT_GRANTED_CONTROL --> CONNECTION_STATE_FINALIZATION_CLIENT_FONT_MAP
 type=PDU_TYPE_DATA[0x00000007], tpktLength=26, remainingLength=20
 recv Font Map Data PDU (0x28), length: 12
 [MS-RDPBCGR] 2.2.1.22.1 Font Map PDU Data (TS_FONT_MAP_PDU)::numberEntries != 0 [1]
 [MS-RDPBCGR] 2.2.1.22.1 Font Map PDU Data (TS_FONT_MAP_PDU)::totalNumEntries != 0 [1]
 [MS-RDPBCGR] 2.2.1.22.1 Font Map PDU Data (TS_FONT_MAP_PDU)::mapFlags != 0x0003 (FONTLIST_FIRST | FONTLIST_LAST) [0x0000]
 [MS-RDPBCGR] 2.2.1.22.1 Font Map PDU Data (TS_FONT_MAP_PDU)::entrySize != 4 [0]
 [CONNECTION_STATE_FINALIZATION_CLIENT_FONT_MAP] received flag FINALIZE_SC_FONT_MAP_PDU [0x00000008]
 CONNECTION_STATE_FINALIZATION_CLIENT_FONT_MAP --> CONNECTION_STATE_ACTIVE
```

CBenoit:
Hi @elmarco!
You should rebase on top of master and fix the conflicts when you get a chance

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -262,6 +262,7 @@ pub enum ShareDataPdu {
     FrameAcknowledge(FrameAcknowledgePdu),
     ServerSetErrorInfo(ServerSetErrorInfoPdu),
     Input(InputEventPdu),
+    ShutdownRequest,
```
CBenoit, line 265: Nice! You don’t have to address the client-side immediately, but I had an issue related to that: https://github.com/Devolutions/IronRDP/issues/115

crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
```diff
@@ -111,7 +111,7 @@ impl PduParsing for General {
 
         let protocol_ver = buffer.read_u16::<LittleEndian>()?;
         if protocol_ver != PROTOCOL_VER {
-            return Err(CapabilitySetsError::InvalidProtocolVersion);
+            warn!("Invalid protocol version: {protocol_ver:?}");
         }
```
CBenoit, line 115: You’re right that this should not be a fatal error, but I feel this check should not exist in `ironrdp-pdu` at all in the first place. I would see it somewhere in `ironrdp-connector` / `ironrdp-acceptor`. As much as possible, I want the `ironrdp-pdu` crate to focus more on decoding / encoding PDUs in a lenient way and hold less such logic, leaving it to high layers.

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -221,7 +221,7 @@ impl PduParsing for ConferenceCreateResponse {
         // ConnectData::connectPDU: length (MUST be ignored by the client according to [MS-RDPBCGR])
         per::legacy::write_length(
             &mut stream,
-            gcc_blocks_buffer_length as u16 + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE,
+            gcc_blocks_buffer_length as u16 + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE + 1,
```
CBenoit, line 224: Good catch. I think this fix will break some tests. FYI a lot of tests in ironrdp-pdu are very old and I wouldn’t be surprised if some of them are snapshots generated using IronRDP itself (i.e.: not official samples from the documentation or extracted from an actual RDP session)

crates/ironrdp-server/src/capabilities.rs
```diff
@@ -74,6 +74,8 @@ fn virtual_channel_capabilities() -> capability_sets::VirtualChannel {
 
 fn multifragment_update() -> capability_sets::MultifragmentUpdate {
     capability_sets::MultifragmentUpdate {
-        max_request_size: u32::MAX,
+        // FIXME: use an acceptable value for msctc.
+        // What is the actual server max size?
+        max_request_size: 16_777_215,
```
CBenoit, line 79: cc @awakecoding 

If we can’t figure this out before merging, do you think you can open an issue to track this?

crates/ironrdp-pdu/src/gcc/core_data/client.rs
```diff
@@ -475,5 +475,7 @@ bitflags! {
         const SUPPORT_DYNAMIC_TIME_ZONE = 0x0200;
         const SUPPORT_HEART_BEAT_PDU = 0x0400;
         const SUPPORT_SKIP_CHANNELJOIN = 0x0800;
+        // The source may set any bits
+        const _ = !0;
```
CBenoit, line 479: I think the correct fix here is to use [`from_bits_retain`](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Flags.html#method.from_bits_retain) instead of `from_bits` of `from_bits_truncate`

elmarco: I'll bump the bitflags crate. However, I feel uneasy addressing the whole #217 issue, since they are a lot of bitflags around...

crates/ironrdp-pdu/src/gcc/conference_create.rs
elmarco: I am not sure why WireShark complains about this field though.. I tried to read their code, but it's not obvious :) anyway the field is marked as being ignored in RDP spec, so I guess we can just add +1 to make WS happy.

crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
elmarco: ok, I added a protocol_version field to General, and an "impl Default", should be enough for now.

crates/ironrdp-pdu/src/gcc/core_data/client.rs
CBenoit: By any means, don’t address #217 as part of this PR, I was just mentioning it but it should be addressed in its own PR

elmarco: well, it fixes server connections from freerdp/mstcs clients, otherwise I wouldn't have included it here.

CBenoit: Of course

crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
```diff
@@ -97,22 +97,33 @@ bitflags! {
...
     type Error = CapabilitySetsError;
 
     fn from_buffer(mut buffer: impl io::Read) -> Result<Self, Self::Error> {
         let major_platform_type = MajorPlatformType(buffer.read_u16::<LittleEndian>()?);
         let minor_platform_type = MinorPlatformType(buffer.read_u16::<LittleEndian>()?);
 
-        let protocol_ver = buffer.read_u16::<LittleEndian>()?;
-        if protocol_ver != PROTOCOL_VER {
-            return Err(CapabilitySetsError::InvalidProtocolVersion);
-        }
```
CBenoit, line 115: To be a bit more conservative, do you think you could add a warning when `protocol_ver` is different from 0x0200 at this place?

https://github.com/Devolutions/IronRDP/blob/78caad219a9c7f701c016b8dc45a6f94c46626d0/crates/ironrdp-connector/src/connection.rs#L706

We may want a similar check in the acceptor as well, so we when a version we didn’t originally expect is received

CBenoit:
FYI, the coverage workflow is now fixed for forks on master: https://github.com/Devolutions/IronRDP/commit/a4fee87c49b1021a673191e69b259b9f9ee8cfe6
CI should be green next time you rebase

crates/ironrdp-pdu/src/rdp/capability_sets/general.rs
elmarco: @CBenoit there is no issue adding "tracing" as dependency of "ironrdp-pdu"? ok.

elmarco: oh ok!

CBenoit:
Great to see progress on this! I noticed there are many 'misc' changes. Would it make sense to merge those as a separate PR? We could benefit from these immediately and reduce the risk of conflicts. What are your thoughts on this?

elmarco:
It turns out that mstsc is happy with TS_SURFCMD_SET_SURF_BITS commands.. I will try to clean up this branch now, as I experimented with various other stuff.

crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -11,8 +11,8 @@ use super::channel_connection::ChannelConnectionSequence;
 use super::finalization::FinalizationSequence;
 use crate::util::{self, wrap_share_data};
 
-const IO_CHANNEL_ID: u16 = 1003;
-const USER_CHANNEL_ID: u16 = 1002;
+pub const IO_CHANNEL_ID: u16 = 1003;
+pub const USER_CHANNEL_ID: u16 = 1002;
```
CBenoit, line 18: Suggestion: for exposing these constants `ironrdp-svc` may be a better place.

Rationale: 1) these are static channel IDs and 2) in the future it would be preferable for `ironvnc-acceptor` to not depend on `ironvnc-connector`.

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+    allow_display_updates: AllowDisplayUpdatesType,
+    desktop_rect: Option<InclusiveRectangle>,
+}
+
+impl PduParsing for SuppressOutputPdu {
+    type Error = io::Error;
+
+    fn from_buffer(mut stream: impl io::Read) -> Result<Self, Self::Error> {
+        let allow_display_updates = AllowDisplayUpdatesType::from_u8(stream.read_u8()?)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidData, "Invalid display update type"))?;
```
CBenoit, line 27: Nitpick: [here](https://github.com/Devolutions/IronRDP/blob/d69ff6e6e7ebf76864f4d812a33677b8d0d70eac/STYLE.md#formatting-of-error-messages) is the error message style we are committed to

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+            allow_display_updates,
+            desktop_rect,
+        })
+    }
+
+    fn to_buffer(&self, mut stream: impl io::Write) -> Result<(), Self::Error> {
+        stream.write_u8(self.allow_display_updates.to_u8().unwrap())?;
+        stream.write_u8(0)?; // padding
+        stream.write_u8(0)?; // padding
+        stream.write_u8(0)?; // padding
```
CBenoit, line 46: Nitpick: we have the `write_padding!` macro specifically for this

Rationale: it communicates the intent by its name and always ensure padding is zeroed

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl PduParsing for SuppressOutputPdu {
+    type Error = io::Error;
+
+    fn from_buffer(mut stream: impl io::Read) -> Result<Self, Self::Error> {
+        let allow_display_updates = AllowDisplayUpdatesType::from_u8(stream.read_u8()?)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidData, "Invalid display update type"))?;
+        let _padding = stream.read_u8()?; // padding
+        let _padding = stream.read_u8()?; // padding
+        let _padding = stream.read_u8()?; // padding
```
CBenoit, line 30: Nitpick: likewise, we have `read_padding!`

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Debug, PartialEq, Eq, FromPrimitive, ToPrimitive, Clone)]
+pub enum AllowDisplayUpdatesType {
+    SuppressDisplayUpdates = 0x00,
+    AllowDisplayUpdates = 0x01,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SuppressOutputPdu {
+    allow_display_updates: AllowDisplayUpdatesType,
+    desktop_rect: Option<InclusiveRectangle>,
```
CBenoit, line 19: Thought: `allow_display_updates` seems redundant with `desktop_rect`.
> desktopRect (8 bytes): An Inclusive Rectangle (section 2.2.11.1) which contains the coordinates of the desktop rectangle if the allowDisplayUpdates field is set to ALLOW_DISPLAY_UPDATES (1). If the allowDisplayUpdates field is set to SUPPRESS_DISPLAY_UPDATES (0), this field MUST NOT be included in the PDU.

Suggestion: only keep `desktop_rect`, and infer `allow_display_updates` depending on whether it is `None` or `Some`.

crates/ironrdp-server/src/encoder/bitmap.rs
```diff
@@ -20,20 +20,20 @@ impl BitmapEncoder {
     }
 
     pub(crate) fn encode(&mut self, bitmap: &BitmapUpdate, output: &mut [u8]) -> Result<usize, PduError> {
-        let row_len = bitmap.width * u32::from(bitmap.format.bytes_per_pixel());
-        let chunk_height = u32::from(u16::MAX) / row_len;
+        let row_len = bitmap.width.get() as usize * bitmap.format.bytes_per_pixel() as usize;
+        let chunk_height = u16::MAX as usize / row_len;
```
CBenoit, line 24: Suggestion: `as` conversions should be avoided when possible. In this case it’s possible to use `usize::from` to convert from `u16` in a guaranteed to be lossless way.

Rationale: `as` conversions will perform many kinds of conversions, including silently lossy conversions and dangerous coercions.

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -28,8 +32,47 @@ impl UpdateEncoder {
...
+        //             _ => {
+        //                 debug!("bitmap encode error: {:?}", e);
+        //                 return None;
+        //             }
+        //         },
+        //         Ok(len) => break len,
+        //     }
+        // };
+
+        // Some(UpdateFragmenter::new(UpdateCode::Bitmap, &self.buffer[..len]))
```
CBenoit, line 52: Note: leftover which can be removed before merging.

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -67,7 +135,7 @@ impl<'a> UpdateFragmenter<'a> {
     pub(crate) fn next(&mut self, dst: &mut [u8]) -> Option<usize> {
         let (consumed, written) = self.encode_next(dst)?;
         self.data = &self.data[consumed..];
-        self.index.checked_add(1)?;
+        self.index = self.index.checked_add(1)?;
```
CBenoit, line 138: Praise: good catch!

crates/ironrdp-svc/src/lib.rs
```diff
@@ -145,6 +150,86 @@ impl StaticVirtualChannel {
...
         self.chunk_processor.dechunkify(payload)
     }
+
+    /// Processes a vector of [`SvcMessage`] in preparation for sending them on the `channel_id` channel.
+    ///
+    /// This includes chunkifying the messages, adding MCS, x224, and tpkt headers, and encoding them into a buffer.
+    /// The messages returned here are ready to be sent to the server.
+    ///
+    /// The caller is responsible for ensuring that the `channel_id` corresponds to the correct channel.
+    pub fn client_process_svc_messages(
```
CBenoit, line 160: Thought: I feel this should either be a free-standing function, or be renamed to `client_process_messages`, or even both.

Rationale: "svc" stands for "Static Virtual Channel", which is the name of the struct. The crate itself is also named `ironrdp-svc`. Overall it’s a bit redundant.

This comment can be applied to `server_process_svc_messages` as well.

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+        stream.write_u8(0)?; // padding
+        if let Some(rect) = &self.desktop_rect {
+            rect.to_buffer(&mut stream)?;
+        }
+
+        Ok(())
+    }
+
+    fn buffer_length(&self) -> usize {
+        4 + self.desktop_rect.as_ref().map_or(0, |r| r.buffer_length())
```
CBenoit, line 55: Note (style): FYI, we are following [this convention](https://github.com/Devolutions/IronRDP/blob/d69ff6e6e7ebf76864f4d812a33677b8d0d70eac/STYLE.md#formatting-for-sizes--lengths-eg-in-encodesize-and-fixed_part_size-definitions)

In this case, it could be something like this:
```rust
1 // allowDisplayUpdates
+ 3 // pad3Octets
+ self.desktop_rect.as_ref().map_or(0, |r| r.buffer_length()) // desktopRect
```

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[repr(u8)]
+#[derive(Debug, PartialEq, Eq, FromPrimitive, ToPrimitive, Clone)]
+pub enum AllowDisplayUpdatesType {
+    SuppressDisplayUpdates = 0x00,
+    AllowDisplayUpdates = 0x01,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SuppressOutputPdu {
```
CBenoit, line 17: Note: FYI, we are now trying to make cross-referencing with spec easier when implementing new PDUs
https://github.com/Devolutions/IronRDP/blob/d69ff6e6e7ebf76864f4d812a33677b8d0d70eac/STYLE.md#doc-comments-should-link-to-reference-documents

crates/ironrdp-acceptor/src/connection.rs
CBenoit: Fair enough

CBenoit: Then I think this should not be exposed as a public item at all. Instead, add fields in the `AcceptorResult` and use that instead in server (leaving the acceptor decide which ID to use and hiding the details from the server crate, even though in practice it will never change).

crates/ironrdp-pdu/src/rdp/suppress_output.rs
elmarco: ah.. it doesn't work with streams..

CBenoit: Oh right, this code is implementing the legacy trait (I assume because the rest of the module is too). You can forget about my comment

crates/ironrdp-pdu/src/rdp/finalization_messages.rs
```diff
@@ -89,6 +89,18 @@ pub struct FontPdu {
     pub entry_size: u16,
 }
 
+impl Default for FontPdu {
+    fn default() -> Self {
+        // 2.2.1.22.1 Font Map PDU Data (TS_FONT_MAP_PDU)
```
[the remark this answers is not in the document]
elmarco: done

crates/ironrdp-pdu/src/rdp/suppress_output.rs
elmarco: ok

crates/ironrdp-svc/src/lib.rs
elmarco: dropped from this  MR

crates/ironrdp-pdu/src/rdp/suppress_output.rs
elmarco: ok

crates/ironrdp-server/src/encoder/mod.rs
elmarco: fixed

crates/ironrdp-server/src/server.rs
```diff
@@ -160,15 +176,32 @@ impl RdpServer {
     where
         S: FramedWrite + FramedRead,
     {
-        let mut buffer = vec![0u8; 4096];
-        let mut encoder = UpdateEncoder::new();
+        debug!("Starting client loop");
+        self.io_channel_id = Some(result.io_channel_id);
```
CBenoit, line 180: Issue (non-blocking): we never handle more than one client at a time, so there is no actual problem, but I feel mixed about this approach of updating a global state (as far as the server is concerned) for something which is only relevant for the current client. This could lead to confusion, especially since this state isn't reset after client disconnection, leaving a redundant "dummy" value.

Thought: that being said, we know it’s technically always the same value, so the "dummy" value will always be the "good" value.

Suggestion: two avenues:
- Remove the `io_channel_id` from `RdpServer`: it’s possible to just keep the value in the function of the body and to give it to `handle_x224`
- Go the "constant" route completely. I directed otherwise in a previous comment, but on second thought it was not bad all things considered. The only request I have is to keep things consistent, i.e., use the constant in the connector itself too (previously the approach was mixed).

crates/ironrdp-server/src/server.rs
```diff
@@ -274,26 +307,39 @@ impl RdpServer {
...
-                        }
-                    },
-
-                    unexpected => {
-                        warn!(?unexpected, "Unexpected share control");
-                    }
+                debug!(?data, "McsMessage::SendDataRequest");
+                if Some(data.channel_id) == self.io_channel_id && self.handle_io_channel_data(data).await? {
+                    return Ok(true);
                 }
```
CBenoit, line 343: Suggestion: a more idiomatic approach would be to call `handle_io_channel_data` inside the if body and avoid nesting control flow inside the condition expression itself.
```suggestion
                if Some(data.channel_id) == self.io_channel_id {
                    return self.handle_io_channel_data(data).await;
                }
```

crates/ironrdp/examples/server.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    // This adds support for the SSLKEYLOGFILE env variable (https://wiki.wireshark.org/TLS#using-the-pre-master-secret)
+    server_config.key_log = Arc::new(rustls::KeyLogFile::new());
+
+    Ok(TlsAcceptor::from(Arc::new(server_config)))
+}
+
+#[derive(Debug)]
+struct InnerHandler {}
```
CBenoit, line 116: Issue (non-blocking): this appears to be dead code. Strangely enough, the lint is not fired.

CBenoit (approved):
LGTM! No blocker, I’ll merge and help with the follow up in a subsequent PR.
Thank you a lot for moving forward this!!

elmarco:
> LGTM! No blocker, I’ll merge and help with the follow up in a subsequent PR. Thank you a lot for moving forward this!!

thanks a lot,

btw are you on #ironrdp on matrix?

crates/ironrdp-pdu/src/rdp/refresh_rectangle.rs
```diff
@@ -0,0 +1,54 @@
...
+        let _padding = stream.read_u8()?; // padding
+        let areas_to_refresh = (0..number_of_areas)
+            .map(|_| InclusiveRectangle::from_buffer(&mut stream))
+            .collect::<Result<Vec<_>, Self::Error>>()?;
+
+        Ok(Self { areas_to_refresh })
+    }
+
+    fn to_buffer(&self, mut stream: impl io::Write) -> Result<(), Self::Error> {
+        stream.write_u8(self.areas_to_refresh.len() as u8)?;
```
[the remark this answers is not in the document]
elmarco: I was wondering. I think there are a couple of other places that could also use it then.

crates/ironrdp-server/src/capabilities.rs
```diff
@@ -74,6 +75,8 @@ fn virtual_channel_capabilities() -> capability_sets::VirtualChannel {
 
 fn multifragment_update() -> capability_sets::MultifragmentUpdate {
     capability_sets::MultifragmentUpdate {
-        max_request_size: u32::MAX,
+        // FIXME: use an acceptable value for msctc.
```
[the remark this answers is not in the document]
elmarco: right

CBenoit:
> btw are you on #ironrdp on matrix?

Just joined the room!

PR #201, opened 2023-09-29
rdpdr next steps

ibeckermayer:
Adds `DeviceIoRequest`, `DeviceControlResponse`, the beginnings of `RdpdrBackend`.

Corresponds with https://github.com/gravitational/teleport/pull/32814


crates/ironrdp-pdu/src/lib.rs
```diff
@@ -128,7 +128,7 @@ pub trait Pdu {
 /// The resulting binary payload is a fully encoded PDU that may be sent to the peer.
 ///
 /// This trait is object-safe and may be used in a dynamic context.
-pub trait PduEncode {
+pub trait PduEncode: Send + Sync + fmt::Debug {
```
ibeckermayer, line 131: These traits were needed for integration with Teleport's current system, they didn't break anything on IronRDP's side afaik so I figured they were acceptable.

pacmancoder: Hmm, this change seems harmless for PDUs, but maybe @CBenoit will have a different opinion on that?

CBenoit: It’s more flexible to add the additional trait bounds at the place that really needs them and this is reducing a bit the number of types on which we may implement the trait, but I think this is quite reasonable for all PDUs to be `Send` + `Sync`. To be honest I’m not sure we actually want to enforce all PDUs to be `Debug`, but we derive this trait most of the time as well, so it’s probably not too bad either.
One concern I have is that one major way to reduce the size of our WASM artifacts would be to turn off the `Debug` derives in production builds, and by requiring the `Debug` trait this way we are probably making this harder for us in the future.

Mind if I ask exactly how this solves the issue with Teleport’s current system?

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -0,0 +1,515 @@
...
+            _ => {
+                error!("Unsupported ScardIoctlCode: 0x{:08x}", value);
+                Err(invalid_message_err!("try_from", "ScardIoctlCode", "unsupported value"))
+            }
+        }
+    }
+}
+
+/// Allow [`ScardIoctlCode`] to be used as an [`IoctlCode`].
+impl IoctlCode for ScardIoctlCode {}
```
CBenoit, line 178: Nit/Question: should this be named this way?
```suggestion
impl IoCtlCode for ScardIoCtlCode {}
```

crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -108,11 +153,19 @@ impl StaticVirtualChannelProcessor for Rdpdr {
...
             }
+            RdpdrPdu::ServerDeviceAnnounceResponse(pdu) => self.handle_server_device_announce_response(pdu),
+            RdpdrPdu::DeviceIoRequest(pdu) => self.handle_device_io_request(pdu, &mut payload),
             RdpdrPdu::Unimplemented => {
                 warn!("received unimplemented packet: {:?}", pdu);
                 Ok(Vec::new())
             }
-            _ => Err(other_err!("rdpdr", "internal error")),
+            // TODO: This can eventually become a `_ => {}` block, but being explicit for now
+            // to make sure we don't miss handling new RdpdrPdu variants here during active development.
```
CBenoit, line 163: :+1:

crates/ironrdp-pdu/src/lib.rs
ibeckermayer: > Mind if I ask exactly how this solves the issue you have with Teleport’s current system?

`Send` + `Sync` are required for Teleport for a couple of reasons:

1. As written, so that we can use `RdpdrPdu` in a `Send` + `Sync` manner.
2. So that `SvcProcessorMessages`, which contain a `Vec<SvcMessage>` which contains a `Box<dyn PduEncode>`, can be used in a `Send` + `Sync` manner.

`Debug` was added because:

`Box<dyn PduEncode>` is part of `DeviceControlResponse`:

https://github.com/Devolutions/IronRDP/blob/1a98d69d9a7652e59d55c136b18233dd7da7ad88/crates/ironrdp-rdpdr/src/pdu/efs.rs#L1154-L1160

and `DeviceControlResponse` is part of `RdpdrPdu`:

https://github.com/Devolutions/IronRDP/blob/1a98d69d9a7652e59d55c136b18233dd7da7ad88/crates/ironrdp-rdpdr/src/pdu/mod.rs#L15-L26

and `RdpdrPdu` has a custom `Debug` implementation which I've needed to verify that these intermediate implementation steps are working correctly

https://github.com/Devolutions/IronRDP/blob/1a98d69d9a7652e59d55c136b18233dd7da7ad88/crates/ironrdp-rdpdr/src/pdu/mod.rs#L144-L173

Currently the only thing that goes into that `Box<dyn Encode>` is the `LongReturn`, however soon there will be a variety of other structures. However I can make another trait which has these constraints rather than using PduEncode so that I can rid us of the Debug:

0a9fbf06d3de1f01dcd202fffc0cf9f94606b12f

CBenoit: > `Send` + `Sync` are required for Teleport for a couple of reasons:
> 
>     1. As written, so that we can use `RdpdrPdu` in a `Send` + `Sync` manner.
> 
>     2. So that `SvcProcessorMessages`, which contain a `Vec<SvcMessage>` which contains a `Box<dyn PduEncode>`, can be used in a `Send` + `Sync` manner.

Okay I see! I was able to spot that in your code on Teleport side and I now understand better.

Note that in that case, a less restrictive approach is to add this bound at [the place where this is required directly](https://github.com/Devolutions/IronRDP/blob/d8c6efa3d180110812e2f6b28c9313981b3c020b/crates/ironrdp-svc/src/lib.rs#L60). Also, and I might be wrong, but it looks like only the `Send` bound is required:
```rust
pdu: Box<dyn PduEncode + Send>,
```

That being said, a PDU structure that is not `Sync` (non thread-safe interior mutability) or even non `Send` (non thread-safe ref counting) would be pretty weird in my opinion, so I’m not really against leaving it as-is either.

I’ll leave you decide on that!

> `Debug` was added because […]

Fair enough, yes I agree with your approach of using another trait internally in that case

crates/ironrdp-pdu/src/x224.rs
```diff
@@ -28,7 +28,7 @@ where
 
 impl<'de, T> PduEncode for T
 where
-    T: X224Pdu<'de>,
+    T: X224Pdu<'de> + Send + Sync + std::fmt::Debug,
```
CBenoit, line 31: Leftover
```suggestion
    T: X224Pdu<'de> + Send + Sync,
```

crates/ironrdp-pdu/src/x224.rs
```diff
@@ -86,6 +86,7 @@ where
     }
 }
 
+#[derive(Debug)]
```
CBenoit, line 89: Leftover

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1,14 +1,17 @@
 //! PDUs for [\[MS-RDPEFS\]: Remote Desktop Protocol: File System Virtual Channel Extension]
 //!
 //! [\[MS-RDPEFS\]: Remote Desktop Protocol: File System Virtual Channel Extension]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/34d9de58-b2b5-40b6-b970-f82d4603bdb5
+use std::convert::TryFrom;
```
CBenoit, line 4: I’m pretty sure this trait is in the prelude nowadays

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -0,0 +1,528 @@
...
+    //!
+    //! [\[MS-RPCE\]: Remote Procedure Call Protocol Extensions]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/290c38b1-92fe-4229-91e6-4fc376610c15
+
+    use ironrdp_pdu::{cursor::WriteCursor, ensure_size, invalid_message_err, PduError, PduResult};
+
+    /// Trait for types that can be encoded into an [MS-RPCE] message.
+    ///
+    /// Implementers should typically avoid implementing this trait directly
+    /// and instead implement [`HeaderlessEncode`].
+    pub trait Encode: Send + Sync + std::fmt::Debug {
```
CBenoit, line 377: Note that you may declare `Encode` here as a super trait of `PduEncode` which requires `std::fmt::Debug`
```suggestion
    pub trait Encode: PduEncode + std::fmt::Debug {
```

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -0,0 +1,528 @@
...
+        fn name(&self) -> &'static str;
+
+        /// Computes the size in bytes for this RPCE PDU.
+        fn size(&self) -> usize;
+    }
+
+    /// Trait for types that can be encoded into an [MS-RPCE] message.
+    ///
+    /// Implementers should typically implement this trait instead of [`Encode`].
+    pub trait HeaderlessEncode: Send + Sync + std::fmt::Debug {
```
CBenoit, line 391: Same at this place

crates/ironrdp-pdu/src/lib.rs
ibeckermayer: > Note that in that case, a less restrictive approach is to add this bound at [the place where it is required directly](https://github.com/Devolutions/IronRDP/blob/d8c6efa3d180110812e2f6b28c9313981b3c020b/crates/ironrdp-svc/src/lib.rs#L60).

It makes sense generally to be as conservative as possible with constraints in order to keep the trait maximally flexible. I agree that a `!Send` or `!Sync` `PduEncode` seems odd, but will avoid arbitrarily restricting us/others on principle.

Initially I opted for constraining `PduEncode` because `pdu: Box<dyn PduEncode + Send + Sync + std::fmt::Debug>` isn't allowed:

> > only auto traits can be used as additional traits in a trait object
consider creating a new trait with all of these as supertraits and using that trait here instead: `trait NewTrait: PduEncode + Debug {}`
auto-traits like `Send` and `Sync` are traits that have special properties; for more information on them, visit <https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits>

> Also, and I might be wrong, it looks like only the `Send` bound is required

That's correct, I was confused. Again I'll opt for maximal conservatism.

1b31b3625ffc1993a8a533a9f2a951bd1dff5ec3

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -882,3 +891,336 @@ impl std::convert::TryFrom<u32> for DeviceType {
...
+
+    pub fn decode(header: DeviceIoRequest, payload: &mut ReadCursor<'_>) -> PduResult<Self> {
+        ensure_size!(ctx: "DeviceControlRequest", in: payload, size: Self::headerless_size());
+        let output_buffer_length = payload.read_u32();
+        let input_buffer_length = payload.read_u32();
+        let io_control_code = T::try_from(payload.read_u32())
+            // TODO: precise error reporting is lost here, figure out how to fix that
+            .map_err(|_| {
+                error!("Failed to parse IoctlCode");
+                invalid_message_err!("DeviceControlRequest", "IoctlCode", "invalid IoctlCode")
```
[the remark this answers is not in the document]
ibeckermayer: > Maybe it’s best to merge PR at hand as-is, and then we can merge https://github.com/Devolutions/IronRDP/pull/202

Let's go with that.

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: Having removed the constraints on `PduEncode`, I went with this: a892f4eca2267db91787c6ede8c589cdae873b19

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: Having removed the constraints on `PduEncode`, I went with this: a892f4eca2267db91787c6ede8c589cdae873b19

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: Sure, I was just copying what we had in Teleport but that works for me: 5376e0b5aa1421b73bcc899b7df69adfc79db51f

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -882,3 +891,336 @@ impl std::convert::TryFrom<u32> for DeviceType {
...
+    const FIXED_PART_SIZE: usize = size_of::<u32>() * 2; // DeviceId, ResultCode
+
+    pub fn name(&self) -> &'static str {
+        Self::NAME
+    }
+
+    pub fn encode(&self, dst: &mut WriteCursor) -> PduResult<()> {
+        ensure_size!(in: dst, size: self.size());
+        dst.write_u32(self.device_id);
+        dst.write_u32(self.result_code as u32);
```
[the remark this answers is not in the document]
ibeckermayer: fed52fdf6b54b3a260ad5f3872d3fab17695d7cd

I opted for e.g.

```rust
impl From<NtStatus> for u32 {
    fn from(status: NtStatus) -> Self {
        status as u32
    }
}
```

The only as conversations I wasn't able to get rid of were from `usize` to `u32`. The use of `usize` is primarily a function of `PduEncode::size -> usize`, we might consider changing that in the future if IronRDP really needs to support the odd 16-bit architecture.

pacmancoder (approved):

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -0,0 +1,539 @@
...
+    _unused: u32,
+}
+
+impl ScardAccessStartedEventCall {
+    pub fn decode(payload: &mut ReadCursor<'_>) -> PduResult<Self> {
+        Ok(Self {
+            _unused: payload.read_u32(),
+        })
+    }
+}
```
CBenoit, line 193: Nit: should this be ignored and discarded like we do for padding?
```suggestion
#[derive(Debug)]
pub struct ScardAccessStartedEventCall;

impl ScardAccessStartedEventCall {
    pub fn decode(payload: &mut ReadCursor<'_>) -> PduResult<Self> {
        ironrdp_pdu::read_padding!(payload, 4);
        Ok(Self)
    }
}
```
(equivalent to `ironrdp_pdu::padding::read(payload, 4)`)

CBenoit (commented):
Minor comments

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -0,0 +1,539 @@
...
+    pub trait Encode: Send + std::fmt::Debug {
+        /// Encodes this RPCE PDU in-place using the provided `WriteCursor`.
+        fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()>;
+
+        /// Returns the name associated with this RPCE PDU.
+        fn name(&self) -> &'static str;
+
+        /// Computes the size in bytes for this RPCE PDU.
+        fn size(&self) -> usize;
+    }
```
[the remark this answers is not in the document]
ibeckermayer: My fault, I assumed that comment was no longer relevant with some of my changes but I was wrong.

One reason we might want a standalone trait in this crate is that it allows us to do this

https://github.com/Devolutions/IronRDP/blob/ff6b959bd8290fdd4ea307d4514eaf7595f8ecab/crates/ironrdp-rdpdr/src/pdu/esc.rs#L405-L431

and share common rcpe encoding logic between all rpce pdus.

However I figured out that I can do it like this and get rid of some of the redundancy with `PduEncode`

fcf47979984398fe38ae53be280b6ff78989fe08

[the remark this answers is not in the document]
CBenoit: > One reason we might want a standalone trait in this crate is that it allows us to do this

You’re right

I also like your other approach

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: 8571c44c501f6efdf1406530cb279008e421ef86

CBenoit (approved):
Looking good to me!

PR #203, opened 2023-10-04
refactor: enable elided_lifetimes_in_paths lint

CBenoit:
Hiding the lifetimes in "paths" is making the code less obvious.

That’s not a problem for types that I know about intimately such as `ReadCursor` or `WriteCursor`, but I actually found myself surprised more than once when reading code I didn’t authored, discovering  later there was in fact a hidden lifetime parameter.
I expect this problem to be worse for someone not familiar with our codebase.

I understand this lint is "allow" by default because in some cases it leads to writing unergonomic ugly code when a type has many generic lifetimes parameters:

```rust
TyCtxt<'_, '_, '_>
```

However we _never_ work with more than one generic lifetime parameter in IronRDP codebase, so it seems to me that the tradeoff towards clarity is worth it, in our case.

### DIFF
--- a/crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
+++ b/crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
@@ -10,7 +10,9 @@ pub use self::palette::*;
 use std::borrow::Cow;
 
 use ironrdp_pdu::cursor::{ReadCursor, WriteCursor};
+use ironrdp_pdu::impl_pdu_borrowing;
 use ironrdp_pdu::utils::{read_string_from_cursor, to_utf16_bytes, CharacterSet};
+use ironrdp_pdu::IntoOwnedPdu;
 use ironrdp_pdu::{cast_int, ensure_fixed_part_size, ensure_size, PduDecode, PduEncode, PduResult};
 
 use super::ClipboardFormatId;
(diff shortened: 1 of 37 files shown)


crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
```diff
@@ -23,6 +25,19 @@ pub struct FormatDataResponse<'a> {
...
+impl IntoOwnedPdu for FormatDataResponse<'_> {
+    type Owned = OwnedFormatDataResponse;
+
+    fn into_owned_pdu(self) -> Self::Owned {
+        OwnedFormatDataResponse {
+            is_error: self.is_error,
+            data: Cow::Owned(self.data.into_owned()),
+        }
+    }
+}
```
CBenoit, line 39: cc @pacmancoder 
This needs to be better documented, but that’s how we handle PDUs that may either borrows or owns data in other parts of IronRDP

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -63,15 +63,15 @@ impl VersionAndIdPdu {
...
+    pub fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
         ensure_size!(ctx: self.name(), in: dst, size: Self::FIXED_PART_SIZE);
         dst.write_u16(self.version_major);
         dst.write_u16(self.version_minor);
         dst.write_u32(self.client_id);
         Ok(())
     }
 
-    pub fn decode(header: SharedHeader, payload: &mut ReadCursor) -> PduResult<Self> {
+    pub fn decode(header: SharedHeader, src: &mut ReadCursor<'_>) -> PduResult<Self> {
```
CBenoit, line 74: cc @ibeckermayer 
I remembered [you were favorable to using `src` for consistency](https://github.com/Devolutions/IronRDP/pull/199#discussion_r1342265487) with the rest of the codebase so I figured it was fine if I search/replaced the remaining occurrences. This code was probably already in-flight when you commented back.

ibeckermayer (approved):

PR #208, opened 2023-10-06
`GetStatusChangeCall` and `GetStatusChangeReturn`

ibeckermayer:
Corresponds with https://github.com/gravitational/teleport/pull/33056

### DIFF
--- a/crates/ironrdp-rdpdr/src/pdu/esc.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/esc.rs
@@ -660,7 +660,6 @@ impl rpce::HeaderlessDecode for GetStatusChangeCall {
             let state = ReaderState::decode_ptr(src, &mut index)?;
             states.push(state);
         }
-
         for state in states.iter_mut() {
             state.decode_value(src)?;
         }
@@ -689,9 +688,10 @@ pub struct ReaderStateCommonCall {
 
 impl ReaderStateCommonCall {
     const NAME: &'static str = "ReaderState_Common_Call";
+    const FIXED_PART_SIZE: usize = size_of::<u32>() * 3 /* dwCurrentState, dwEventState, cbAtr */ + 36 /* rgbAtr */;
 
     fn decode(src: &mut ReadCursor<'_>) -> PduResult<Self> {
-        ensure_size!(in: src, size: Self::size());
+        ensure_size!(in: src, size: Self::FIXED_PART_SIZE);
         let current_state = CardStateFlags::from_bits_truncate(src.read_u32());
         let event_state = CardStateFlags::from_bits_truncate(src.read_u32());
         let atr_length = src.read_u32();
@@ -705,9 +705,16 @@ impl ReaderStateCommonCall {
         })
     }
 
-    fn size() -> usize {
-        size_of::<u32>() * 3 // dwCurrentState, dwEventState, cbAtr
-        + 36 // rgbAtr
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
+        dst.write_u32(self.current_state.bits());
+        dst.write_u32(self.event_state.bits());
+        dst.write_u32(self.atr_length);
+        dst.write_slice(&self.atr);
+        Ok(())
+    }
+
+    fn size(&self) -> usize {
+        Self::FIXED_PART_SIZE
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -758,12 +923,12 @@ pub mod rpce {
     ///
     /// Implementers should typically avoid implementing this trait directly
     /// and instead implement [`HeaderlessEncode`], and wrap it in a [`Pdu`].
-    pub trait Encode: PduEncode + Send + std::fmt::Debug {}
+    pub trait Encode: PduEncode + Send + Sync + std::fmt::Debug {}
```
CBenoit, line 926: This looks good to me, but now I’m curious about what is actually requiring the `Sync` bound on your side. Is it because you are sharing something like `Arc<DeviceControlResponse>` in various threads?

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -616,6 +618,168 @@ impl rpce::HeaderlessEncode for ListReadersReturn {
...
+        let timeout = src.read_u32();
+        let states_ptr_length = src.read_u32();
+
+        let states_ptr = ndr::decode_ptr(src, &mut index)?;
+
+        context.decode_value(src)?;
+
+        ensure_size!(in: src, size: size_of::<u32>() * 2);
+        let states_length = src.read_u32();
+
```
CBenoit, line 657: Copy-paste error?
```suggestion
        ensure_size!(in: src, size: size_of::<u32>());
        let states_length = src.read_u32();

```

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -616,6 +618,168 @@ impl rpce::HeaderlessEncode for ListReadersReturn {
...
+        })
+    }
+}
+
+impl rpce::HeaderlessEncode for GetStatusChangeReturn {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
+        dst.write_u32(self.return_code.into());
+        let mut index = 0;
+        ndr::encode_ptr(Some(self.reader_states.len() as u32), &mut index, dst)?;
+        dst.write_u32(self.reader_states.len() as u32);
```
CBenoit, line 764: Use `cast_length!` macro

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: It's because we need to hold a `DeviceControlResponse` (holding `Box<dyn rpce::Encode>`) in our `RdpdrBackend` implementation, and `RdpdrBackend` is

```rust
RdpdrBackend: fmt::Debug + Send + Sync
```

That happens [here](https://github.com/gravitational/teleport/blob/06453019ed42ac94a04cac4efbb03cae287b04a3/lib/srv/desktop/rdp/rdpclient/src/rdpdr.rs#L200-L203) in the Teleport code, with the comment explaining it at a high level.

CBenoit: Thank you, I see.

I note we have the same "problem" with `CliprdrBackend`.
I understand we need the `Sync` bound on the backend trait because it is required by `StaticVirtualChannelProcessor`, because in turn it is also required by `ironrdp_connector::Sequence`.
But it’s probably not very useful to require this at all: out of curiosity, I tried removing the bound and everything still works, at least in IronRDP repository, and to be fair I don’t think it’s sensible to share the connector or the client itself across multiple threads (like in wrapping in into a `Mutex` and driving it across many threads), so it would make sense if `Sync` was not required everywhere like it is today.

The reason I’m considering this carefully is (I’m sure you’re used to it) WASM. As you know, in WASM, when running in the browser, we can’t block the main thread at all, [**including `Mutex`**](https://github.com/WebAssembly/threads/issues/106#issue-368968008).

Here is also a relevant thread on the Rust Users forum I came across not long ago:
https://users.rust-lang.org/t/wasm-winit-wgpu-weird-error-about-waiting-on-the-main-thread-despite-not-waiting-on-the-main-thread/96804/1

This means that, in the Web Client, if at some point we really need interior mutability for some reason in order to implement a backend, we can’t cheat our way using `Mutex`. `Cell` or `RefCell` would be the way to go. Incidentally, it’s also something that could be used in a multi-threaded application as long as the `Cell` itself is not shared, which is actually not useful as often as I initially thought, and when sharing a `Cell` among several tasks is actually required it’s actually [possible even with tokio using a `LocalSet`](https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html).

Anyway, I’m not blocking the PR for this because this should be addressed in its own PR, and it’s not necessarily urgent. Pinging @pacmancoder for information as he’s currently implementing the cliprdr backend for the Web Client.

CBenoit (approved):
Looking good! CI is reporting some lints, but I’m approving and enabling auto-merge now so it’s automatically merged when addressed.

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: Makes sense to me at a high level, this works for Teleport as well. I added this to this PR in 11b2e7e2f7f19f52a3d4b3e105853ce97efe25b2.

ibeckermayer: Oh, I realize you enabled auto-merge so I reverted that, I will put it up in a separate PR so you and @pacmancoder have a chance to confirm.

ibeckermayer: https://github.com/Devolutions/IronRDP/pull/209

pacmancoder: Yes, I completely agree that Sync should be removed

PR #210, opened 2023-10-09
`ConnectCall` and `ConnectReturn`

ibeckermayer:
This also refactors rpce and ndr into their own files, as well as modifying each. The diff is a little confusing/overwhelming, but reviewing commit-by-commit should make it easier to track the changes.

Corresponds to https://github.com/gravitational/teleport/pull/33136


crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -931,12 +931,12 @@ pub mod rpce {
...
     /// Implementers should typically avoid implementing this trait directly
     /// and instead implement [`HeaderlessEncode`], and wrap it in a [`Pdu`].
-    pub trait Encode: PduEncode + Send + std::fmt::Debug {}
+    pub trait Encode: PduEncode + Send + Sync + std::fmt::Debug {}
 
     /// Trait for types that can be encoded into an [MS-RPCE] message.
     ///
     /// Implementers should typically implement this trait instead of [`Encode`].
-    pub trait HeaderlessEncode: Send + std::fmt::Debug {
+    pub trait HeaderlessEncode: Send + Sync + std::fmt::Debug {
```
CBenoit: You’re used to it now, but here is the customary question (mostly for my own understanding): why do we need this bound on `Encode` and `HeaderlessEncode`?

Is `Sync` required for `DeviceControlResponse` (the struct holding a `Box<dyn rpce::Encode>`)?

If this is absolutely required, I suggest adding this line below `DeviceControlResponse`:
```rust
// Small comment explaining the rationale
ironrdp_pdu::assert_impl!(Config: Send, Sync);
```

ibeckermayer: I'm not quite sure what happened, but it looks like you reviewed something from before I rebased. This is fixed now.

crates/ironrdp-rdpdr/src/pdu/esc/ndr.rs
```diff
@@ -0,0 +1,169 @@
...
+//! - in the code below, "*_ptr" is the pointer value and "*_value" the actual data
+//! - note that some fields (like arrays) will have a length prefix before the pointer and also
+//!   before the actual data at the end of the message
+//!
+//! To deal with this, fixed-size structs only have encode/decode methods, while variable-size ones
+//! have encode_ptr/decode_ptr and encode_value/decode_value methods. Messages are parsed linearly,
+//! so decode_ptr/decode_value are called at different stages (same for encoding).
+//!
+//! Most of the above was reverse-engineered from FreeRDP:
+//! https://github.com/FreeRDP/FreeRDP/blob/master/channels/smartcard/client/smartcard_pack.c
```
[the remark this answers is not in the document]
ibeckermayer: 464f154ea0558133b556af541ee35f342335abd3

crates/ironrdp-rdpdr/src/pdu/esc.rs
CBenoit: Oh, all good then!

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -787,483 +886,202 @@ impl rpce::HeaderlessEncode for GetStatusChangeReturn {
...
+/// [2.2.1.2 REDIR_SCARDHANDLE]
+///
+/// [2.2.1.2 REDIR_SCARDHANDLE]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpesc/b6276356-7c5f-4d3e-be92-a6c85e58d008
+#[derive(Debug)]
+pub struct ScardHandle {
+    context: ScardContext,
+    length: u32,
+    // Shortcut: we always create 4-byte handle values.
+    // The spec allows this field to have variable length.
+    value: u32,
```
CBenoit, line 975: Not a big deal, but since we have an hardcoded length of 4, you may consider to not hold the `length` field at all and use a constant instead.

```rust
pub struct ScardHandle {
    context: ScardContext,
    // Shortcut: we always create 4-byte handle values.
    // The spec allows this field to have variable length.
    value: u32,
}

impl ScardHandle {
    const NAME: &'static str = "REDIR_SCARDHANDLE";
    const VALUE_LENGTH: u32 = 4; // <-

    /* … */
}

impl ndr::Decode for ScardHandle {
    fn decode_ptr(src: &mut ReadCursor<'_>, index: &mut u32) -> PduResult<Self>
    where
        Self: Sized,
    {
        let context = ScardContext::decode_ptr(src, index)?;
        ensure_size!(ctx: "Handle::decode_ptr", in: src, size: size_of::<u32>());
        let length = src.read_u32();
        if length != Self::VALUE_LENGTH { // <-
            todo!("return an error, reporting unsupported value length");
        }
        let _ptr = ndr::decode_ptr(src, index)?;
        Ok(Self {
            context,
            value: 0,
        })
    }

    fn decode_value(&mut self, src: &mut ReadCursor<'_>) -> PduResult<()> {
        self.context.decode_value(src)?;
        ensure_size!(in: src, size: size_of::<u32>());
        let length = src.read_u32();
        if length != Self::VALUE_LENGTH { // <-
            Err(invalid_message_err!(
                "decode_value",
                "mismatched length in Handle reference and value"
            ))
        } else {
            ensure_size!(in: src, size: size_of::<u32>());
            self.value = src.read_u32();
            Ok(())
        }
    }
}

impl ndr::Encode for ScardHandle {
    /* … */

    fn encode_value(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
        ensure_size!(in: dst, size: self.size_value());
        self.context.encode_value(dst)?;
        dst.write_u32(Self::VALUE_LENGTH); // <-
        dst.write_u32(self.value);
        Ok(())
    }

    /* … */
}
```

I think it’s easier to reason about: since we fix the value to `u32`, we also fix the length field to `4`.

(Of course, if supporting variable-length values is a thing we’ll need in the future, you may want to keep the current code.)

Note: I’ve seen something similar a several places, you may want to reflect this everywhere.

ibeckermayer: IMO its preferable to keep it as is: currently, if the server happens to send `ScardContext` or `ScardHandle` of a different length, we can handle that, whereas with the proposed changes the program would fail. Granted it's unlikely that a server sends us back anything other than exactly what we sent it, but I don't see good reason to make changes that make this more fragile.

CBenoit: I agree with the rationale, but I’m not sure to understand how the current implement is actually more robust: is it actually handling payloads bigger or smaller than 4 bytes? For example, at [this place](https://github.com/Devolutions/IronRDP/blob/464f154ea0558133b556af541ee35f342335abd3/crates/ironrdp-rdpdr/src/pdu/esc.rs#L108-L117):
```rust
fn decode_value(&mut self, src: &mut ReadCursor<'_>) -> PduResult<()> {
    ensure_size!(in: src, size: size_of::<u32>()*2);
    let length = src.read_u32();
    if length != self.length {
        error!(?length, ?self.length, "Mismatched length in ScardContext reference and value");
        Err(invalid_message_err!(
            "decode_value",
            "mismatched length in ScardContext reference and value"
        ))
    } else {
        self.value = src.read_u32();
        Ok(())
    }
}
```
I understand that if length is not 4, then `src.read_u32()` is basically reading either not enough or too much, which is just a bug instead of an explicit error.

ibeckermayer: Ah true, mistake on my part.

ibeckermayer: d978bf9cb19b6dc606aef23dd528f50e993e767d

CBenoit: Thank you!

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -5,7 +5,7 @@
 pub mod ndr;
 pub mod rpce;
 use super::efs::IoCtlCode;
-use crate::pdu::esc::ndr::{Decode, Encode};
+use crate::pdu::esc::ndr::{Decode as NdrDecode, Encode as NdrEncode};
```
CBenoit, line 8: Nit: not a blocker, but it might be sensible to simply name this trait `NdrDecode` instead of using `Decode as NdrDecode`

CBenoit (approved):
LGTM!

PR #215, opened 2023-10-11
`TransmitCall` and `TransmitReturn`

ibeckermayer:
Corresponds with https://github.com/gravitational/teleport/pull/33282


crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -943,7 +945,8 @@ impl ndr::Decode for ConnectCommon {
         let context = ScardContext::decode_ptr(src, index)?;
         ensure_size!(in: src, size: size_of::<u32>() * 2);
         let share_mode = src.read_u32();
-        let preferred_protocols = CardProtocol::from_bits_truncate(src.read_u32());
+        let preferred_protocols = CardProtocol::from_bits(src.read_u32())
+            .ok_or_else(|| invalid_message_err!("decode_ptr", "ConnectCommon", "invalid CardProtocol"))?;
```
CBenoit, line 949: Out of curiosity: why `from_bits_truncate` wasn’t doing the job? I wonder if [`from_bits_retain`](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Flags.html#method.from_bits_retain) wouldn’t be a better alternative (keeping the original payload as-is)

ibeckermayer: It was doing the job, I just noticed that `from_bits` was more precise so decided to switch to using that. I tried it here as well

https://github.com/Devolutions/IronRDP/blob/fbbfe31ab45eff0acff4171d06215a3a271d9afd/crates/ironrdp-rdpdr/src/pdu/esc.rs#L810-L811

but it turns out that we actually do sometimes get junk bits in those fields 🤷‍♂️.

CBenoit: Re-thinking about this, I really think we should use [`from_bits_retain`](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Flags.html#method.from_bits_retain) everywhere now. My understanding is that it wasn’t used up until now because this method didn’t exist yet in older versions of the `bitflags` crate.

As much as possible I want parsing to be lossless and to uphold the "round- trip" property (`m = encode(decode(m))`)

ibeckermayer: I don't quite understand how `from_bits_retain` works. What if there are undefined bits?

CBenoit: Those bits are ignored, but not unset (unlike `from_bits_truncate`).

Actually, let me correct myself: we should either use `from_bits` or `from_bits_retain`, but never `from_bits_truncate`, and I would tend to reach for `from_bits_retain` over `from_bits`

Rationale:
- We want the round-tripping property to hold, and for that parsing must be non destructive (lossless), but `from_bits_truncate` is destructive (undefined bits are discarded)
- We generally want lenient parsing, ignoring unknown values as long as we don’t need them, but `from_bits` is not lenient

However, it’s okay to use `from_bits` if more strictness is required at some place, but I would like such places to be documented with a comment explaining why we have to refuse unknown flags

ibeckermayer: a8fb511de93916a47dfed80d273d6a9536b80129

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -1122,3 +1124,186 @@ impl rpce::HeaderlessDecode for HCardAndDispositionCall {
...
+        let recv_buffer_is_null = src.read_u32() == 1;
+        let recv_length = src.read_u32();
+
+        handle.decode_value(src)?;
+        send_pci.decode_value(src)?;
+
+        ensure_size!(in: src, size: size_of::<u32>());
+        let send_length = src.read_u32();
+        ensure_size!(in: src, size: send_length as usize);
+        let send_buffer = src.read_slice(send_length as usize).to_vec();
```
CBenoit, line 1169: Nit: `cast_length` should be used instead of `as usize`:
```suggestion
        let send_length: usize = cast_length!("TransmitCall", "send_buffer", src.read_u32());
        ensure_size!(in: src, size: send_length);
        let send_buffer = src.read_slice(send_length).to_vec();
```

crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -1122,3 +1124,186 @@ impl rpce::HeaderlessDecode for HCardAndDispositionCall {
...
+        Ok(Self {
+            protocol,
+            extra_bytes_length,
+            extra_bytes,
+        })
+    }
+
+    fn decode_value(&mut self, src: &mut ReadCursor<'_>) -> PduResult<()> {
+        ensure_size!(in: src, size: self.extra_bytes_length as usize);
+        self.extra_bytes = src.read_slice(self.extra_bytes_length as usize).to_vec();
```
CBenoit, line 1224: Nit: same as above

CBenoit (approved):
Looking good!

crates/ironrdp-rdpdr/src/pdu/esc.rs
ibeckermayer: 18b05ca972ed19e165fff7ab6066c433c905b600

CBenoit (approved):
Thank you!

crates/ironrdp-rdpdr/src/pdu/esc.rs
pacmancoder: Huh, yes, `from_bits_retain` is much better than truncation via `from_bits_truncate`, which may prevent a hard-to-catch bug for us in the future! We may want to make refactoring on the whole codebase, I created an issue to do refactoring in future https://github.com/Devolutions/IronRDP/issues/217

PR #216, opened 2023-10-11
`StatusCall` and `StatusReturn`

ibeckermayer:
Corresponds with https://github.com/gravitational/teleport/pull/33284


crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -1309,3 +1311,138 @@ impl rpce::HeaderlessEncode for TransmitReturn {
...
+    fn name(&self) -> &'static str {
+        Self::NAME
+    }
+
+    fn size(&self) -> usize {
+        size_of::<u32>() * 5 // dst.write_u32(self.return_code.into()); dst.write_u32(self.state.into()); dst.write_u32(self.protocol.bits()); dst.write_slice(&self.atr); dst.write_u32(self.atr_length);
+        + ndr::ptr_size(true) // ndr::encode_ptr(Some(reader_names_length), &mut index, dst)?;
+        + self.atr.len() // dst.write_slice(&self.atr);
+        + encoded_multistring_len(&self.reader_names, self.encoding) // write_multistring_to_cursor(dst, &self.reader_names, self.encoding)?;
+    }
```
ibeckermayer, line 1420: @CBenoit I know we've gone in circles on this, I think in these particular cases for PDUs which require this kind of tricky ndr encoding, this kind of notation makes the most sense. (I use this to help me write the function in the first place).

MS-RDPESC is something of an exception in this respect, the sizing isn't as straightforward to follow from the docs (e.g. https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpesc/987c1358-ad6b-4c8e-88e1-06210c28a66f) as compared to the rest of MS-RDPDR (e.g. https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/a00fbce4-95bb-4e15-8182-be2b5ef9076a).

IMO I should use this style in these instances, and then switch to what we discussed with @pacmancoder once I get back to MS-RDPDR proper.

CBenoit: Fair point, you’re right, let’s use this style for MS-RDPESC

CBenoit (approved):
LGTM

CBenoit:
Just need to rebase on master, I’m enabling auto-merge

PR #218, opened 2023-10-13
ContextCall

ibeckermayer:


crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -109,6 +109,7 @@ impl Rdpdr {
             let call = ScardCall::decode(req.io_control_code, payload)?;
 
             debug!(?req);
+            debug!(?req.io_control_code);
             debug!(?call);
```
CBenoit, line 113: Nit: to limit pressure on the logger and reduce output size, I think this should be in a single log record
```suggestion
            debug!(?req, ?req.io_control_code, ?call);
```
And I would expect `req.io_control_code` to be included into the Debug representation of `req`. 

That being said, maybe this is generating an output that is easier to exploit for you so I don’t insist on this

CBenoit (approved):
Looks good to me. I left a nit comment that you may address or not. Tell me if I should merge the PR as-is or wait for you to change something.

crates/ironrdp-rdpdr/src/lib.rs
ibeckermayer: Just marginally easier for me to read this way: 93b8c24f4e74d590505e600c87925fb2199c64bd

ibeckermayer:
Ready to merge!

PR #219, opened 2023-10-16
Makes `CliprdrBackend` `AsAny`

ibeckermayer:
Teleport could use this feature, afaik it's harmless and simple to add (we're using `impl_as_any!`).

Corresponds with https://github.com/gravitational/teleport/pull/33506

### DIFF
--- a/crates/ironrdp-cliprdr/src/lib.rs
+++ b/crates/ironrdp-cliprdr/src/lib.rs
@@ -47,7 +47,7 @@ enum CliprdrState {
 /// CLIPRDR static virtual channel client endpoint implementation
 #[derive(Debug)]
 pub struct Cliprdr {
-    pub backend: Box<dyn CliprdrBackend>,
+    backend: Box<dyn CliprdrBackend>,
     capabilities: Capabilities,
     state: CliprdrState,
 }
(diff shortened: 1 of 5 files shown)


crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -47,7 +47,7 @@ enum CliprdrState {
 /// CLIPRDR static virtual channel client endpoint implementation
 #[derive(Debug)]
 pub struct Cliprdr {
-    backend: Box<dyn CliprdrBackend>,
+    pub backend: Box<dyn CliprdrBackend>,
```
pacmancoder, line 50: Let's make getter for this instead

CBenoit: What about exposing it with methods like these?

```rust
fn downcast_backend<T: Any>(&self) -> Option<&T> {
    self.backend.as_any().downcast::<T>()
}

fn downcast_backend_mut<T: Any>(&mut self) -> Option<&mut T> {
    self.backend.as_any_mut().downcast_mut::<T>()
}
```

ibeckermayer: @pacmancoder sounds good.

@CBenoit I tried that initially but it breaks object safety (due to generics in trait methods) which is required for having `dyn CliprdrBackend`

ibeckermayer: @CBenoit ah I see, adding those to `Cliprdr` itself 👍.

CBenoit (approved):
LGTM!

CBenoit:
You need to implement `AsAny` for `WinCliprdrBackend` as well

PR #221, opened 2023-10-16
ci: make sure the coverage workflow is disabled for forks

CBenoit:
The check to disable the coverage job is not working properly.


CBenoit:
Opened from the wrong repository.

PR #224, opened 2023-10-17
Adds `EndTransaction`, `Disconnect`, `Cancel`, and `IsValidContext`

ibeckermayer:
Adds handling for multiple ioctl codes which don't require new PDUs.


CBenoit (approved):
LGTM!

PR #227, opened 2023-10-18
`WriteCacheCall`

ibeckermayer:


CBenoit (approved):
LGTM!

Remark: I didn’t realize until now, but it would be better to implement both decoding and encoding for all messages. Two reasons:

- We currently don’t do it as much as we could, but I want to add more property tests in the future in order to validate stuff like round-tripping property, ensure correct computation for size and so on, and this would help with that

- This will be useful for implementing server side RDPDR

However, I understand you may not have enough time for that and/or you may not have both sides in your RDPDR original implementation in the first place. Just mentioning this, so that if you find some bandwidth for that, it would help us for the long term.

PR #230, opened 2023-10-19
allows additional `client_info_flags` to be defined in `Config`

ibeckermayer:
Teleport needs this to pipe in `ClientInfoFlags::PASSWORD_IS_SC_PIN | ClientInfoFlags::AUTOLOGON`


CBenoit (commented):
I’m not very comfortable with this API as the user can now enable RDP features we don’t actually support. 
Maybe this should be exposed differently? For instance, if IronRDP actually supports `INFO_AUTOLOGON`, a boolean like `enable_autologon` would be enough.

As for `INFO_PASSWORD_IS_SC_PIN`, I feel this should be made more explicit by replacing the `password` field with something different.

I don’t know if the username field is required for the smartcard as well, but we could imagine something like this:
```rust
enum Authentication {
    UsernamePassword { username: String, password: String },
    SmartCard { pin: String },
}
```

CBenoit:
Another question: does it makes sense to use `INFO_PASSWORD_IS_SC_PIN` when the RDPDR channel is not enabled?

(I’m trying to understand the problem space better)

ibeckermayer:
@CBenoit see e1220e65fa3e0be33e2e33fab1ab150283317920

> For instance, if IronRDP actually supports INFO_AUTOLOGON, a boolean like enable_autologon would be enough.

IronRDP does support `INFO_AUTOLOGON`, at least to an extent: when user `SecurityProtocol::SSL` this causes the system to autologon, when using `HYBRID` it appears to have no effect. I didn't test other configurations.

> Another question: does it make sense to use `INFO_PASSWORD_IS_SC_PIN` when the RDPDR channel is not enabled?

AFAIK that flag is there for the case where a user wants to autologon with a smartcard without needing to enter the pin during the authentication sequence -- in short, no.

crates/ironrdp-async/src/connector.rs
```diff
@@ -53,13 +53,17 @@ where
...
+    }
 
     while connector.is_credssp_step() {
         single_connect_step(framed, &mut connector, &mut buf).await?;
     }
 
-    debug!("Remaining of connection sequence");
+    if connector.is_credssp_step() {
+        debug!("CredSSP complete, continue with remainder of connection sequence");
+    }
```
CBenoit, line 66: This is surprising. Is it a leftover?

crates/ironrdp-pdu/src/rdp/server_license/server_license_request/tests.rs
```diff
@@ -272,9 +272,9 @@ fn from_buffer_correctly_parses_server_license_request() {
 
 #[test]
 fn from_buffer_correctly_parses_server_license_request_no_certificate() {
-    let server_certificate_header_buffer = vec![
+    let server_certificate_header_buffer = [
         0x03, 0x00, // blob type
-        0x00, 0x00, // blob len
```
CBenoit, line 277: It’s probably fine to leave the comment here

CBenoit:
> @CBenoit see [e1220e6](https://github.com/Devolutions/IronRDP/commit/e1220e65fa3e0be33e2e33fab1ab150283317920)

Thank you! Looking good to me.

> > For instance, if IronRDP actually supports INFO_AUTOLOGON, a boolean like enable_autologon would be enough.
> 
> IronRDP does support `INFO_AUTOLOGON`, at least to an extent: when user `SecurityProtocol::SSL` this causes the system to autologon, when using `HYBRID` it appears to have no effect.

Oh okay. So when using a smart card, the flow is actually to use `SecurityProtocol::SSL`, skipping CredSSP? I see.

Question 1: looks like `INFO_AUTOLOGON` + `SecurityProtocol::SSL` + `Credentials::SmartCard` are all coming together. It’s likely okay to infer `INFO_AUTOLOGON` based on the other two. What do you think?

Question 2: What happens when setting `autologon` to `false` with `SecurityProtocol::SSL` and any one of `Credentials::UsernamePassword` or `Credentials::SmartCard`?

> > Another question: does it make sense to use `INFO_PASSWORD_IS_SC_PIN` when the RDPDR channel is not enabled?
> 
> AFAIK that flag is there for the case where a user wants to autologon with a smartcard without needing to enter the pin during the authentication sequence -- in short, no.

Sounds reasonable.

crates/ironrdp-async/src/connector.rs
ibeckermayer: Ah yeah, derp. That's what I meant.

ibeckermayer: How's this: 2595917e64712d4c4f4b497b903ce998d196fc1e

crates/ironrdp-pdu/src/rdp/server_license/server_license_request/tests.rs
ibeckermayer: Added back a108bf64038dca4d3f9e455d38508d3b19c6f322

crates/ironrdp-async/src/connector.rs
CBenoit: Good for me!

ibeckermayer:
> Oh okay. So when using a smart card, the flow is actually to use `SecurityProtocol::SSL`, skipping CredSSP? I see.

Yes

> Question 2: What happens when setting `autologon` to `false` with `SecurityProtocol::SSL` and any one of `Credentials::UsernamePassword` or `Credentials::SmartCard`?

With `Credentials::UsernamePassword` you end up at the login screen, with the username pre-entered for you:

image

With `Credentials::SmartCard` I just ended up here

image

I'm not sure how it chose Administrator, maybe it's because that was the last account I logged in as. It doesn't appear to have read that from the smartcard (based on Teleport logs).

The first case is at least somewhat coherent, although it doesn't quite make sense that we require a password and then the user needs to enter it again. The latter doesn't make much sense since you can only log in with a password, even though you selected smart card. 

> Question 1: looks like `INFO_AUTOLOGON` + `SecurityProtocol::SSL` + `Credentials::SmartCard` are all coming together. It’s likely okay to infer `INFO_AUTOLOGON` based on the other two. What do you think?

Based on the above, as long as IronRDP always requires username + password, it makes sense that we just always set `INFO_AUTOLOGON`: 62b29fa6744f6eb4a5672095b01e34d482767388

crates/ironrdp-client/src/config.rs
```diff
@@ -168,7 +169,7 @@ struct Args {
     password: Option<String>,
 
     /// Specify the security protocols to use
-    #[clap(long, value_enum, value_parser, default_value_t = SecurityProtocol::Hybrid)]
+    #[clap(long, value_enum, value_parser, default_value_t = SecurityProtocol::Ssl)]
```
CBenoit, line 172: The only request I have before we actually merge, is to leave this at `Hybrid` by default. I believe the intention is to use CredSSP by default.

CBenoit (commented):
Thank you for the investigation!!

Sounds to me like we actually want to expose an `autologon` boolean in the connector config.
In our web client we probably want `autologon = false` + `Hybrid` + `UsernamePassword` (until we add support for smartcards anyway).

I’m sorry to ask you to re-add it now that you removed it. I’m okay with re-adding it myself later too.

> However if a design goal for IronRDP is to support all possible modes then we either need more complex logic or to just leave it up to the user.

That’s a goal (though I can see us not supporting RC4 ever…).
My intention here is to find the right balance for the API we expose, but I don’t really want to disable a valid use case.

crates/ironrdp-client/src/config.rs
ibeckermayer: Ah yeah good catch, that was just for testing I didn't mean to push it.

CBenoit (approved):
Looks good to me! Thank you for all this!

PR #234, opened 2023-10-24
Feat/drive announce

ibeckermayer:
Adds the beginnings of drive redirection.

### DIFF
--- a/crates/ironrdp-pdu/src/utils.rs
+++ b/crates/ironrdp-pdu/src/utils.rs
@@ -23,6 +23,14 @@ pub fn combine_u64(lo: u32, hi: u32) -> u64 {
     u64::from_le_bytes(position_bytes)
 }
 
+pub fn to_utf8_bytes(value: &str, write_null_terminator: bool) -> Vec<u8> {
+    let mut bytes = value.as_bytes().to_vec();
+    if write_null_terminator {
+        bytes.push(0u8);
+    }
+    bytes
+}
+
 pub fn to_utf16_bytes(value: &str) -> Vec<u8> {
     value
         .encode_utf16()
--- a/crates/ironrdp-pdu/src/utils.rs
+++ b/crates/ironrdp-pdu/src/utils.rs
@@ -23,14 +23,6 @@ pub fn combine_u64(lo: u32, hi: u32) -> u64 {
     u64::from_le_bytes(position_bytes)
 }
 
-pub fn to_utf8_bytes(value: &str, write_null_terminator: bool) -> Vec<u8> {
-    let mut bytes = value.as_bytes().to_vec();
-    if write_null_terminator {
-        bytes.push(0u8);
-    }
-    bytes
-}
-
 pub fn to_utf16_bytes(value: &str) -> Vec<u8> {
     value
         .encode_utf16()
--- a/crates/ironrdp-rdpdr/src/backend/mod.rs
+++ b/crates/ironrdp-rdpdr/src/backend/mod.rs
@@ -9,7 +9,7 @@ use ironrdp_pdu::PduResult;
 use ironrdp_svc::AsAny;
 
 /// OS-specific device redirection backend inteface.
-pub trait RdpdrBackend: AsAny + fmt::Debug + Send + 'static {
+pub trait RdpdrBackend: AsAny + fmt::Debug + Send {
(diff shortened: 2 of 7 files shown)


crates/ironrdp-pdu/src/utils.rs
```diff
@@ -23,6 +23,14 @@ pub fn combine_u64(lo: u32, hi: u32) -> u64 {
     u64::from_le_bytes(position_bytes)
 }
 
+pub fn to_utf8_bytes(value: &str, write_null_terminator: bool) -> Vec<u8> {
+    let mut bytes = value.as_bytes().to_vec();
+    if write_null_terminator {
+        bytes.push(0u8);
+    }
+    bytes
+}
```
CBenoit, line 32: In my opinion, this should be more something like this, in order to opportunistically reuse the underlying buffer without extra allocation.
```suggestion
pub fn into_utf8_bytes(value: String, write_null_terminator: bool) -> Vec<u8> {
    let mut bytes = value.into_bytes();
    if write_null_terminator {
        bytes.push(0u8);
    }
    bytes
}
```
Though I’m not even sure how useful this helper will actually be in the future, and inlining the logic at the only callsite sounds reasonable to me as well in this case. When `write_null_terminator = false` I don’t think I would even bother using this function because `my_string.into_bytes()` / `my_string.as_bytes()` is already exactly that.

crates/ironrdp-rdpdr/src/backend/mod.rs
```diff
@@ -6,9 +6,10 @@ use crate::pdu::{
 };
 use core::fmt;
 use ironrdp_pdu::PduResult;
+use ironrdp_svc::AsAny;
 
 /// OS-specific device redirection backend inteface.
-pub trait RdpdrBackend: fmt::Debug + Send {
+pub trait RdpdrBackend: AsAny + fmt::Debug + Send + 'static {
```
CBenoit, line 12: Nit: it’s technically okay since we currently only use this trait for boxed trait-objects, but I would tend to add the `'static` bound at the place where it’s necessary in order to not over-constrain the definition. For instance:
```rust
    pub fn into_box<T>(backend: T) -> Box<dyn RdpdrBackend>
    where
        T:  RdpdrBackend + 'static, // <- here
    {
        Box::new(backend)
    }

```
Or is there is another reason for it that I’m missing?

crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -56,11 +59,42 @@ impl Rdpdr {
 
     #[must_use]
     pub fn with_smartcard(mut self, device_id: u32) -> Self {
-        self.device_list.add_smartcard(device_id);
         self.capabilities.add_smartcard();
+        self.device_list.add_smartcard(device_id);
```
CBenoit, line 63: Question: if the order is especially important, do you think you could add a comment explaining why? (seems easy to break otherwise)

ibeckermayer: It's not actually important, I just swapped it because I used the other ordering in `with_drives`.

crates/ironrdp-pdu/src/utils.rs
ibeckermayer: b7a30f42cf0652e968b06e7033df01b8c1b6901a

ibeckermayer: b7a30f42cf0652e968b06e7033df01b8c1b6901a

crates/ironrdp-rdpdr/src/backend/mod.rs
ibeckermayer: It's a requirement of for `T` of `downcast_ref::<T>`, I've moved it there in f7c14ea1ea77713c7b502a73559294c371bd2619

CBenoit: Oh right. My bad. [`Any`](https://doc.rust-lang.org/std/any/trait.Any.html) requires `'static` in the first place, so we might as well add this bound on `AsAny`. I pushed a commit for that: https://github.com/Devolutions/IronRDP/pull/234/commits/6286dd70038c65ffbc47ca5199930c5ef31318f5

CBenoit (approved):
LGTM!

PR #235, opened 2023-10-28
`DeviceCreateRequest` and `DeviceCreateResponse`

ibeckermayer:


crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1012,7 +1022,7 @@ impl From<NtStatus> for u32 {
 /// [2.2.1.4 Device I/O Request (DR_DEVICE_IOREQUEST)]
 ///
 /// [2.2.1.4 Device I/O Request (DR_DEVICE_IOREQUEST)]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/a087ffa8-d0d5-4874-ac7b-0494f63e2d5d
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
```
CBenoit, line 1025: Is derivating the `Copy` trait a deliberate decision?
Since this type is a POD, not _too_ big (20 bytes), it’s not unreasonable at all though I don’t think we did that on the others (more of an API-wise consideration than a performance-wise one; fwiw, maybe we should implement `Copy` on more types at some point).

ibeckermayer: This was added so that I could pass a `DeviceIoRequest` without taking ownership here

https://github.com/gravitational/teleport/blob/bcb7125512155f9036401181a02efa380e90e385/lib/srv/desktop/rdp/rdpclient/src/rdpdr/filesystem.rs#L399-L408

however I didn't give any thought to the implications for the API. Having thought about it more, I think we can be conservative with respect to deriving `Copy` on full PDUs (I'll remove it here), and revisit it again later if I find myself bumping into it commonly over in Teleport.

CBenoit: Sounds good, keep me informed!

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -971,15 +971,25 @@ impl ServerDeviceAnnounceResponse {
 #[derive(Debug, PartialEq, Copy, Clone)]
 #[repr(u32)]
 pub enum NtStatus {
```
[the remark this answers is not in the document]
ibeckermayer: Makes sense, changed here: ef6ab44bb1b4347b439fe3aa9aa12ec0287c58c9

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -968,57 +970,60 @@ impl ServerDeviceAnnounceResponse {
...
+            0xC000_0035 => Ok(NtStatus::OBJECT_NAME_COLLISION),
+            0xC000_0022 => Ok(NtStatus::ACCESS_DENIED),
+            0xC000_0103 => Ok(NtStatus::NOT_A_DIRECTORY),
+            0xC000_000F => Ok(NtStatus::NO_SUCH_FILE),
+            0xC000_00BB => Ok(NtStatus::NOT_SUPPORTED),
+            0xC000_0101 => Ok(NtStatus::DIRECTORY_NOT_EMPTY),
+            _ => Err(custom_err!("NtStatus", UnsupportedNtStatus(value))),
         }
     }
 }
```
CBenoit, line 1007: Replace this with `From<u32>`:
```rust
impl From<u32> for NtStatus {
    fn from(value: u32) -> Self {
        Self(value)
    }
}
```

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -968,57 +970,60 @@ impl ServerDeviceAnnounceResponse {
...
+    pub const SUCCESS: Self = Self(0x0000_0000);
+    pub const UNSUCCESSFUL: Self = Self(0xC000_0001);
+    pub const NOT_IMPLEMENTED: Self = Self(0xC000_0002);
+    pub const NO_MORE_FILES: Self = Self(0x8000_0006);
+    pub const OBJECT_NAME_COLLISION: Self = Self(0xC000_0035);
+    pub const ACCESS_DENIED: Self = Self(0xC000_0022);
+    pub const NOT_A_DIRECTORY: Self = Self(0xC000_0103);
+    pub const NO_SUCH_FILE: Self = Self(0xC000_000F);
+    pub const NOT_SUPPORTED: Self = Self(0xC000_00BB);
+    pub const DIRECTORY_NOT_EMPTY: Self = Self(0xC000_0101);
```
CBenoit, line 986: I guess we can still keep the full name as docstrings for cross-referencing?
```suggestion
    /// STATUS_SUCCESS
    pub const SUCCESS: Self = Self(0x0000_0000);
    /// STATUS_UNSUCCESSFUL
    pub const UNSUCCESSFUL: Self = Self(0xC000_0001);
    /// STATUS_NOT_IMPLEMENTED
    pub const NOT_IMPLEMENTED: Self = Self(0xC000_0002);
    /// STATUS_NO_MORE_FILES
    pub const NO_MORE_FILES: Self = Self(0x8000_0006);
    /// STATUS_OBJECT_NAME_COLLISION
    pub const OBJECT_NAME_COLLISION: Self = Self(0xC000_0035);
    /// STATUS_ACCESS_DENIED
    pub const ACCESS_DENIED: Self = Self(0xC000_0022);
    /// STATUS_NOT_A_DIRECTORY
    pub const NOT_A_DIRECTORY: Self = Self(0xC000_0103);
    /// STATUS_NO_SUCH_FILE
    pub const NO_SUCH_FILE: Self = Self(0xC000_000F);
    /// STATUS_NOT_SUPPORTED
    pub const NOT_SUPPORTED: Self = Self(0xC000_00BB);
    /// STATUS_DIRECTORY_NOT_EMPTY
    pub const DIRECTORY_NOT_EMPTY: Self = Self(0xC000_0101);
```

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -968,57 +970,60 @@ impl ServerDeviceAnnounceResponse {
...
+pub struct UnsupportedNtStatus(u32);
+
+impl std::fmt::Display for UnsupportedNtStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:#010X}", self.0)
+    }
+}
+
+impl std::error::Error for UnsupportedNtStatus {}
+
```
CBenoit, line 1026: Do we actually need to restrict the values we put into `NtStatus`?

ibeckermayer: Ah yes, now I understand. No, we have no need for restricting these.

998151bcbd39e274296e0f7799f4e5896b661fa2

CBenoit (approved):
LGTM! Ready to merge once CI is green

Issue #217, opened 2023-10-12
Preserve unknown bitmask bits values in PDUs (eliminate use of 'from_bits_truncate')


pacmancoder:
To prevent losing information after PDU encode/decode roundtrip, we should always keep bitmask values unchanged, but currently, we use `from_bits_truncate` excessively in the codebase. As @CBenoit explained in [the comment](https://github.com/Devolutions/IronRDP/pull/215#discussion_r1355706172), we need to avoid usage of `from_bits_truncate`:

> we should either use from_bits or from_bits_retain, but never from_bits_truncate, and I would tend to reach for from_bits_retain over from_bits
> 
> Rationale:
> 
> - We want the round-tripping property to hold, and for this property to hold, parsing must be non destructive (lossless), but from_bits_truncate is destructive (undefined bits are discarded)
> - We generally want lenient parsing, ignoring unknown values as long as we don’t need them and/or as long as ignoring them is causing no harm, but from_bits is not lenient
> 
> However, it’s okay to use from_bits if strictness is required at some place, but I would like such places to be documented with a comment explaining why we have to refuse unknown flags

CBenoit:
Following up on this, looks like a new syntax was added to the macro in `bitflags` v2.4: https://github.com/bitflags/bitflags/pull/371

```rust
bitflags! {
    pub struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;

        // The source may set any bits
        const _ = !0; // <- this guy
    }
}
```

There is documentation crate-level documentation for this: https://docs.rs/bitflags/2.4.0/bitflags/#externally-defined-flags

This is basically make `from_bits_truncate` behaves exactly like `from_bits_retain` because all bits are then considered as known. `from_bits` also never fails.

So I suggest this:

- Use **both** `from_bits_retain` and `const _ = !0` when lenient parsing is required
    - `const _ = !0` ensure we don’t accidentally have non lenient or destructive parsing
    - `from_bits_retain` make it clear that preserving all the bits is intentional at the call site
- Use `from_bits` WITHOUT `const _ = !0` when strictness is required

PR #237, opened 2023-10-29
docs: add "on bit flags" section

CBenoit:
Related to https://github.com/Devolutions/IronRDP/issues/217

[Rendered](https://github.com/Devolutions/IronRDP/tree/docs/on-bitflags/crates/ironrdp-pdu#on-bit-flags)


crates/ironrdp-pdu/README.md
```diff
@@ -400,3 +407,65 @@ MUST always be rejected. In such cases, an `enum` is deemed appropriate.
...
+then considered to be known and defined. In this case `from_bits` also never fails and provides no
+real benefit over `from_bits_retain`.
+
+Here is the bottom line:
+
+- Use **both** `from_bits_retain` and `const _ = !0` when resilient parsing is required.
+    - `const _ = !0` ensures we don’t accidentally have non resilient or destructive parsing.
+    - `from_bits_retain` makes it clear at the call site that preserving all the bits is intentional.
+- Use `from_bits` WITHOUT `const _ = !0` when strictness is required (almost never in IronRDP), and
+    document why with an in-source comment.
```
ibeckermayer, line 466: This might work better as a tl;dr near the top of the section.

crates/ironrdp-pdu/README.md
```diff
@@ -400,3 +407,65 @@ MUST always be rejected. In such cases, an `enum` is deemed appropriate.
...
+There is crate-level documentation for this: <https://docs.rs/bitflags/2.4.0/bitflags/#externally-defined-flags>
+
+This is making `from_bits_truncate` behaves exactly like `from_bits_retain` because all bits are
+then considered to be known and defined. In this case `from_bits` also never fails and provides no
+real benefit over `from_bits_retain`.
+
+Here is the bottom line:
+
+- Use **both** `from_bits_retain` and `const _ = !0` when resilient parsing is required.
+    - `const _ = !0` ensures we don’t accidentally have non resilient or destructive parsing.
```
ibeckermayer, line 463: IMO this case is less compelling than the reason given by the crate itself (https://docs.rs/bitflags/2.4.0/bitflags/#externally-defined-flags)

CBenoit: What do you mean? Are you referring to this?

> It helps compatibility when the external source may start setting additional bits at any time. 

This is what I mean by avoiding non resilient parsing. Maybe I should change the wording?

ibeckermayer: I meant

> Generated methods like `all` and truncating operators like `!` only consider bits in defined flags. Adding an unnamed flag makes those methods consider additional bits, without generating additional constants for them.

is a more compelling reason to also add `const _ = !0` rather than protecting from accidentally having destructive parsing (e.g. if we use `from_bits` accidentally). Though arguably that all falls under the umbrella of your wording, so feel free to just ignore.

pacmancoder (approved):

CBenoit:
@ibeckermayer I think everything is properly addressed so I’ll merge the current version

PR #238, opened 2023-10-29
Maintenance and follow-up

CBenoit:
Small teaks and improvements I had in my git working tree.

Each teak has its own commit.

### DIFF
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -66,9 +66,13 @@ where
     S: StreamWrapper,
 {
     pub fn new(stream: S::InnerStream) -> Self {
+        Self::new_with_leftover(stream, BytesMut::new())
+    }
+
+    pub fn new_with_leftover(stream: S::InnerStream, leftover: BytesMut) -> Self {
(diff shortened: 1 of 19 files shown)


crates/ironrdp-async/src/framed.rs
```diff
@@ -66,9 +66,13 @@ where
     S: StreamWrapper,
 {
     pub fn new(stream: S::InnerStream) -> Self {
+        Self::from_leftover(stream, BytesMut::new())
+    }
+
+    pub fn from_leftover(stream: S::InnerStream, leftover: BytesMut) -> Self {
```
pacmancoder, line 72: Maybe `with_leftower` would be a better name for a constructor? `from_` sounds like `leftover` should be the only argument, but we also have stream as a first argument. What do you think?

CBenoit: You’re right. I’ll rename to `new_with_leftover` (`with_leftover` alone would suggest some sort of builder pattern)

CBenoit:
> Disabled automerge and approved for faster merge

Good call, thank you!

PR #245, opened 2023-10-31
Adds `ServerDriveQueryInformationRequest`

ibeckermayer:
Also changes `FilesystemRequest` to be named `ServerDriveIoRequest`.

### DIFF
--- a/crates/ironrdp-rdpdr/src/pdu/efs.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/efs.rs
@@ -1705,9 +1705,9 @@ impl ClientDriveQueryInformationResponse {
     }
 }
 
-/// [2.4 File Information Classes [MS-FSCC]]
+/// [2.4 File Information Classes \[MS-FSCC\]]
 ///
-/// [2.4 File Information Classes [MS-FSCC]]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/4718fc40-e539-4014-8e33-b675af74e3e1
+/// [2.4 File Information Classes \[MS-FSCC\]]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/4718fc40-e539-4014-8e33-b675af74e3e1
 #[derive(Debug, Clone)]
 pub enum FileInformationClass {
     Basic(FileBasicInformation),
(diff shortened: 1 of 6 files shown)


crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1589,3 +1597,250 @@ bitflags! {
         const FILE_OVERWRITTEN = 0x00000003;
     }
 }
+
+/// [2.2.3.3.8 Server Drive Query Information Request (DR_DRIVE_QUERY_INFORMATION_REQ)]
+///
+/// Note that Length, Padding, and QueryBuffer fields are all ignored in keeping with the [analogous code in FreeRDP].
+///
+/// [2.2.3.3.8 Server Drive Query Information Request (DR_DRIVE_QUERY_INFORMATION_REQ)]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/e43dcd68-2980-40a9-9238-344b6cf94946
```
CBenoit, line 1605: Nit: you can ignore if you want, but I was thinking about only putting a link on the section number instead:
```suggestion
/// [2.2.3.3.8] Server Drive Query Information Request (DR_DRIVE_QUERY_INFORMATION_REQ)
///
/// Note that Length, Padding, and QueryBuffer fields are all ignored in keeping with the [analogous code in FreeRDP].
///
/// [2.2.3.3.8]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/e43dcd68-2980-40a9-9238-344b6cf94946
```
This requires less duplication and results in shorter lines. What do you think?

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1589,3 +1597,250 @@ bitflags! {
...
+/// [2.4.7 FileBasicInformation [MS-FSCC]]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/16023025-8a78-492f-8b96-c873b042ac50
+#[derive(Debug, Clone)]
+pub struct FileBasicInformation {
+    pub creation_time: i64,
+    pub last_access_time: i64,
+    pub last_write_time: i64,
+    pub change_time: i64,
+    pub file_attributes: FileAttributes,
+    // NOTE: The `reserved` field in the spec MUST not be serialized and sent over RDP, or it will break the server implementation.
+    // FreeRDP does the same: https://github.com/FreeRDP/FreeRDP/blob/1adb263813ca2e76a893ef729a04db8f94b5d757/channels/drive/client/drive_file.c#L508
```
CBenoit, line 1747: Thank you for documenting this :pray:

CBenoit (approved):
LGTM. Leaving some time if you want to address the comment about links, but otherwise good to merge.

crates/ironrdp-rdpdr/src/pdu/efs.rs
ibeckermayer: I'm fine with that style in general, but I'd be bothered by the inconsistency with the rest of [this part of] the codebase if I were to fix just this one. Seems more trouble than it's worth to fix them all at this point.

CBenoit: Fear not! [`sd`](https://github.com/chmln/sd) is here to save the day!
I’ll run that in another PR though, you’re right.

CBenoit: https://github.com/Devolutions/IronRDP/pull/247

PR #247, opened 2023-11-01
docs: uniformize links to specification

CBenoit:
@ibeckermayer you’re right, there were quite a bunch of these!

I ran `sd` commands to change most places, and modified using multi-cursors the last few places it didn’t covered (mainly the places which were not consistent in the first place).

```shell
$ fd --extension=rs --exec sd "/// \[([\d\.]+)( .+)\]:" "/// [\$1]:"
$ fd --extension=rs --exec sd "/// \[([\d\.]+)( .+)\]" "/// [\$1]\$2"
```


ibeckermayer (approved):
💥 good moves

PR #259, opened 2023-11-02
feat: clipboard support for web client

### LINKED ISSUE #107: Initial Clipboard support (textual data)

pacmancoder:
This PR adds clipboard support for IronRDP-web

**Features:**
- Most complete support could be achieved on chromium-based browsers
    - Formats: `text/plain`, `text/html`, `text/png`
    - Clipboard monitoring is implemented via monitoring loop
- Firefox has reduced clipboard support due to browser restrictions:
    - Copy: `text/plain`
    - Paste: `text/plain`, `text/html`, `text/png`
    - Periodic clipboard monitoring is absent, copy/paste actions executed on demand, may introduce input freezes, the there is no known to me way to overcome this (synchronous onPaste event API is used; clipboardWrite action could only be performed as a response to user input)

- Internally, the following `CLIPRDR` formats are handled and converted to/from their MIME counterparts:
    - `CF_UNICODETEXT` <=> `text/plain`
    - `Registered(CF_HTML<HTML Format>)`, `Registered(text/html)`  <=> `text/html`
    - `CF_DIB`, `CF_DIBV5`, `Registered(PNG)`, `Registered(image/png)` <=> `image/png`

Closes #107

Issue: ARC-125


web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -182,6 +219,12 @@ export class WasmBridgeService {
         this.canvas = canvas;
     }
 
+    /// Triggered by the browser when local clipboard is updated. Clipbard backend should
```
CBenoit, line 222:
```suggestion
    /// Triggered by the browser when local clipboard is updated. Clipboard backend should
```

pacmancoder:
FYI: I was clippy lints regarding arithmetic side-effects, I'll fix that, other than that PR is ready for the review

crates/ironrdp-cliprdr-format/src/html.rs
```diff
@@ -0,0 +1,141 @@
...
+    replace_placeholder(start_fragment_placeholder_pos, &start_fragment_pos_value);
+    replace_placeholder(end_fragment_placeholder_pos, &end_fragment_pos_value);
+
+    buffer
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+}
```
CBenoit, line 141: Leftover

crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -118,3 +118,17 @@ pub fn rdp6_decode_bitmap_stream_to_rgb24(input: &BitmapInput<'_>) {
         input.height as usize,
     );
 }
+
+pub fn cliprdr_format(input: &[u8]) {
```
CBenoit, line 122: Did you intend to add a new fuzz target? (Much welcomed.) You may have forgotten to commit it.

web-client/iron-remote-gui/src/iron-remote-gui.svelte
```diff
@@ -26,16 +27,357 @@
 
     let viewerStyle;
     let wrapperStyle;
-    
+
     let wasmService = new WasmBridgeService();
     let publicAPI = new PublicAPI(wasmService);
 
+    // Firefox's clipboard API is very limited, and doesn't reading from the clipboard
```
CBenoit, line 34: Typo:
```suggestion
    // Firefox's clipboard API is very limited, and doesn't support reading from the clipboard
```

crates/ironrdp-web/src/clipboard/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+    RemoteClipboardChanged(Vec<ClipboardFormat>),
+    RemoteDataResponse(FormatDataResponse<'static>),
+
+    FormatListReceived,
+    ForceClipboardUpdate,
+}
+
+/// Clipboard backend implementation for web. This object should be created once per session and
+/// kept alive until session is terminated.
+pub struct WasmCipboard {
```
CBenoit, line 104:
```suggestion
pub struct WasmClipboard {
```

CBenoit (commented):
Truly excellent work!
I appreciate you created a re-usable crate `ironrdp-cliprdr-format` :+1:

crates/ironrdp-cliprdr-format/src/bitmap.rs
```diff
@@ -0,0 +1,663 @@
...
+        if size != Self::FIXED_PART_SIZE {
+            return Err(invalid_message_err!("biSize", "invalid V1 bitmap info header size"));
+        }
+
+        Ok(header)
+    }
+}
+
+/// Header used in `CF_DIBV5` formats, defined as `BITMAPV5HEADER` in
+/// [wingdi](https://learn.microsoft.com/ru-ru/windows/win32/api/wingdi/ns-wingdi-bitmapv5header)
```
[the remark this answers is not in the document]
pacmancoder: > Nit: looks like you linked to Russian documentation

Dammit. My Google Search is very actively tries to show russian pages for the search output (and I hate that, google please fix, for god's sake, I am not from russia). I accidentally opened the first link and pasted it 😢

crates/ironrdp-cliprdr-format/src/bitmap.rs
```diff
@@ -0,0 +1,663 @@
...
+fn transform_bitmap(
+    header: &BitmapInfoHeader,
+    input: &[u8],
+    preserve_alpha: bool,
+) -> Result<PngEncoderInput, BitmapError> {
+    // If height is positive, DIB is bottom-up, but target PNG format is top-down.
+    let flip = header.height > 0;
+
+    let width: usize = cast_int!(BitmapInfoHeader::NAME, "biWidth", header.width).expect("width is positive");
+    let height: usize = cast_int!(BitmapInfoHeader::NAME, "biHeight", header.height.abs()).expect("height is positive");
```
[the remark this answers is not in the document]
pacmancoder: Yep, it makes sense at least for height, negative height as per wingdi, requires to flip the vertical coordinates for the image (scanlines top-down instead of bottom-up). Negative width does not have a special meaning as per wingdi, but i think we could keep it i32 just for consistency

crates/ironrdp-fuzzing/src/oracles/mod.rs
pacmancoder: Huh.... no wonder it passed CI the first time... 🤣 Thanks for spotting this!

crates/ironrdp-web/src/clipboard/mod.rs
pacmancoder: Huh... Interesting why `typos` did not catch that even though I ran it multiple times... Maybe I need to update it...

CBenoit: Oh, I thought it would indeed!
Looks like they don’t have this typo in their database: https://github.com/epage/typos/blob/194cd85dd6eecb097568b803d374ab21dccace1b/crates/typos-dict/assets/words.csv#L10034
I commented their issue for the November batch: https://github.com/crate-ci/typos/issues/865#issuecomment-1792626518
Should be properly detected next time :slightly_smiling_face:

crates/ironrdp-cliprdr-format/src/bitmap.rs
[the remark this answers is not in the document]
CBenoit: Fair enough! :+1:

crates/ironrdp-cliprdr-format/src/bitmap.rs
```diff
@@ -166,11 +178,16 @@ impl BitmapInfoHeader {
     fn encode_with_size(&self, dst: &mut WriteCursor<'_>, size: u32) -> PduResult<()> {
         ensure_fixed_part_size!(in: dst);
 
+        check_invariant(self.width.abs() <= u16::MAX.into())
+            .ok_or_else(|| invalid_message_err!("biWidth", "width is too big"))?;
+        check_invariant(self.height.abs() <= u16::MAX.into())
+            .ok_or_else(|| invalid_message_err!("biHeight", "height is too big"))?;
```
CBenoit, line 184: Invariants should be checked when constructing the struct (when decoding or calling new), so that remaining of the code may rely on this (encoding, casting…)

crates/ironrdp-cliprdr-format/src/html.rs
```diff
@@ -0,0 +1,144 @@
...
+            // Extract fragment from the original buffer.
+            if start > end || end > input.len() {
+                return Err(HtmlError::InvalidFormat);
+            }
+
+            break String::from_utf8(input[start..end].to_vec())?;
+        }
+
+        // INVARIANT: end_pos < headers_cursor.len() - 1
+        #[allow(clippy::arithmetic_side_effects)]
```
CBenoit, line 71: The `INVARIANT:` comment should come where `end_pos` is declared so the reader knows what is being enforced for this value before reading the logic itself.

```rust
// INVARIANT: end_pos < headers_cursor.len() - 1
let end_pos = /* … */;
```

Example in rust-analyzer code:
https://github.com/rust-lang/rust-analyzer/blob/c1c9e10f72ffd2e829d20ff1439ff49c2e121731/crates/hir-ty/src/lower.rs#L1118-L1127

When it’s not obvious why some piece of code is indeed ensuring the invariant holds, a comment may be added:
https://github.com/rust-lang/rust-analyzer/blob/c1c9e10f72ffd2e829d20ff1439ff49c2e121731/crates/hir-ty/src/lower.rs#L1143-L1144

Then, the invariant can be used to explain why it’s okay to do something non obvious such as disabling a lint:
```rust
#[allow(clippy::arithmetic_side_effects)] // given the invariant on `end_pos`, this code can’t cause integer overflow
{
    /* … */
}
```

CBenoit (commented):
In general, `INVARIANT:` comments should come where the variable is declared, either in the type definition, or above the `let` for local variables. The rationale is that invariants should be uphold at all times, and it’s a useful to keep them in mind when analyzing the flow of the code.

```rust
// INVARIANT: end_pos < headers_cursor.len() - 1
let end_pos = /* … */;
```

Typically, when a local invariant is defined using `INVARIANT:`, the reader should not have to backtrack in order to verify it, but read forward instead. 

Example in rust-analyzer code:
https://github.com/rust-lang/rust-analyzer/blob/c1c9e10f72ffd2e829d20ff1439ff49c2e121731/crates/hir-ty/src/lower.rs#L1118-L1127

When it’s not obvious why some piece of code is indeed ensuring the invariant holds, a comment may be added:
https://github.com/rust-lang/rust-analyzer/blob/c1c9e10f72ffd2e829d20ff1439ff49c2e121731/crates/hir-ty/src/lower.rs#L1143-L1144

And when explaining an assumption for doing something non obvious such as disabling a lint locally, the invariant may be referenced to without prefixing the comment with `INVARIANT:` (because it is already defined elsewhere)

```rust
// Given the invariant on `end_pos`, this code can’t cause integer overflow
#[allow(clippy::arithmetic_side_effects)]
{
    /* … */
}
```

crates/ironrdp-cliprdr-format/src/html.rs
```diff
@@ -0,0 +1,144 @@
...
+    buffer.extend_from_slice(b"<!--EndFragment-->\r\n</body>\r\n</html>");
+
+    let end_html_pos = buffer.len();
+
+    let start_html_pos_value = format!("{:0>10}", start_html_pos);
+    let end_html_pos_value = format!("{:0>10}", end_html_pos);
+    let start_fragment_pos_value = format!("{:0>10}", start_fragment_pos);
+    let end_fragment_pos_value = format!("{:0>10}", end_fragment_pos);
+
+    // INVARIANT: buffer.len() > placeholder_pos + POS_PLACEHOLDER.len()
```
CBenoit, line 132: It’s generally better to use `<` and `<=` over `>` and `>=`
```
placeholder_pos + POS_PLACEHOLDER.len() < buffer.len()
```

CBenoit (commented):
You’ll have to rebase on master to fix conflicts on iron-remote-gui.
It’s probably not much.

FYI, formatting and lints checks for TypeScript were [added recently](https://github.com/Devolutions/IronRDP/pull/263).
Here are the commands you should run when checking the code:
```shell
$ npm run check
$ npm run format
$ npm run lint
```

CBenoit (approved):
I’ll leave you address the remaining comments and and merge yourself

PR #260, opened 2023-11-02
feat(credssp): support for Kerberos

irvingoujAtDevolution:
Bump sspi-rs version to the lastest master, restructure the network accessing model when using credssp.

Draft Until the [formatter/linter change]( https://github.com/Devolutions/IronRDP/pull/25)  is merged, since there are changes overlap.

Reviews welcomed, will fix then rebase to master after the linter change is merged


crates/ironrdp-client/src/network_cllient.rs
CBenoit, line 1: Typo in the name of the file

crates/ironrdp-client/src/rdp.rs
```diff
@@ -105,9 +104,6 @@ async fn connect(
 
     let mut connector = connector::ClientConnector::new(config.connector.clone())
         .with_server_addr(server_addr)
-        .with_server_name(&config.destination)
-        .with_credssp_network_client(RequestClientFactory)
-        // .with_static_channel(ironrdp::dvc::Drdynvc::new()) // FIXME: drdynvc is not working
```
CBenoit, line 110: Do not remove the `FIXME`

crates/ironrdp-client/src/network_cllient.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(buf)
+    }
+
+    async fn send_udp(&self, url: &Url, data: &[u8]) -> ConnectorResult<Vec<u8>> {
+        let port =
+            portpicker::pick_unused_port().ok_or_else(|| general_err!("No ports free for sending udp request"))?;
+
+        let udp_socket = UdpSocket::bind((IpAddr::V4(Ipv4Addr::LOCALHOST), port))
+            .await
+            .map_err(|e| custom_err!("Cannot bind udp socket", e))?;
```
CBenoit, line 76: Remove the dependency on `portpicker`, and ask for a port assigned by the operating system:
```suggestion
        let udp_socket = UdpSocket::bind((IpAddr::V4(Ipv4Addr::LOCALHOST), 0))
            .await
            .map_err(|e| custom_err!("Cannot bind udp socket", e))?;
```

crates/ironrdp-web/Cargo.toml
```diff
@@ -65,4 +65,4 @@ smallvec = "1.11"
 x509-cert = { version = "0.2", default-features = false, features = ["std"] }
 tap = "1"
 semver = "1"
-url = "2"
+reqwest = "0.11.22"
```
CBenoit, line 68: Stick to `gloo_net` instead of `reqwest`. `reqwest` is much bigger than `gloo_net` (which don’t do much more than wrapping the browser API), and we don’t need all that. (WASM files should be as small as possible.)

crates/ironrdp-connector/src/lib.rs
```diff
@@ -97,6 +100,38 @@ pub struct Config {
     pub no_server_pointer: bool,
     /// If true, the INFO_AUTOLOGON flag is set in the [`ironrdp_pdu::rdp::ClientInfoPdu`].
     pub autologon: bool,
+    // for kerberos authentication
+    pub sspi_config: Option<SspiConfig>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct SspiConfig {
+    pub kdc_url: Option<url::Url>,
```
CBenoit, line 109: Is it a KDC URL, or a KDC **Proxy** URL?

crates/ironrdp-client/Cargo.toml
```diff
@@ -56,6 +56,10 @@ anyhow = "1"
 smallvec = "1.11"
 tap = "1"
 semver = "1"
+byteorder = "1.5.0"
```
CBenoit, line 59: Unused dependency

crates/ironrdp-web/src/session.rs
```diff
@@ -165,6 +169,12 @@ impl SessionBuilder {
         self.clone()
     }
 
+    /// Optional
+    pub fn kdc_url(&self, kdc_url: String) -> SessionBuilder {
```
CBenoit, line 173: Is it a KDC URL, or a KDC **Proxy** URL?
My understanding is that either we have a KDC address (`tcp://domain:88`, `udp://domain:88`) or a KDC Proxy URL (`https://gateway.devolutions.net/jet/KdcProxy?token=<…>`).
That is, when sending the messages via HTTP, it’s always using the KDC Proxy endpoint, with the payload wrapped in a [`KdcProxyMessage`](https://docs.rs/picky-krb/latest/picky_krb/messages/struct.KdcProxyMessage.html).
Unless I’m wrong, the naming should reflect this:
```suggestion
    pub fn kdc_proxy_url(&self, kdc_proxy_url: String) -> SessionBuilder {
```

irvingoujAtDevolution: agreed, kdc_proxy_url makes more sense

crates/ironrdp-client/Cargo.toml
irvingoujAtDevolution: updated

crates/ironrdp-web/Cargo.toml
irvingoujAtDevolution: updated

crates/ironrdp-client/src/rdp.rs
irvingoujAtDevolution: updated

Cargo.toml
```diff
@@ -51,7 +51,7 @@ ironrdp = { version = "0.5", path = "crates/ironrdp" }
 expect-test = "1"
 proptest = "1.3"
 rstest = "0.18"
-sspi = "0.10"
+sspi = {git = "https://github.com/Devolutions/sspi-rs.git"}
```
CBenoit, line 54: Note: I’ll publish a new version for the sspi crate before we merge this PR

crates/ironrdp-connector/src/connection.rs
```diff
@@ -53,6 +56,173 @@ pub struct ConnectionResult {
...
+    next_request: Option<credssp::TsRequest>,
+    state: CredSSPState,
+    selected_protocol: nego::SecurityProtocol,
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum CredSSPState {
+    CredsspInitial,
+    CredsspReplyNeeded,
+    CredsspEarlyUserAuthResult,
```
irvingoujAtDevolution, line 75: Need Help to Test this state

CBenoit: I’ll try to run the code on my side and see if I can hit it

web-client/iron-svelte-client/src/lib/login/login.svelte
```diff
@@ -6,14 +6,14 @@
     import { toast } from '$lib/messages/message-store';
     import { showLogin } from '$lib/login/login-store';
     import type { DesktopSize } from '../../models/desktop-size';
-
```
CBenoit, line 9: You should leave the newline before the `import`s

crates/ironrdp-web/src/session.rs
```diff
@@ -50,6 +50,8 @@ struct SessionBuilderInner {
     hide_pointer_callback_context: Option<JsValue>,
     show_pointer_callback: Option<js_sys::Function>,
     show_pointer_callback_context: Option<JsValue>,
+
+    kdc_proxy_url: Option<String>,
```
CBenoit, line 54: No need to separate this from the rest, can go after `pcb`

crates/ironrdp-web/src/session.rs
```diff
@@ -460,6 +481,8 @@ fn build_config(
     domain: Option<String>,
     client_name: String,
     desktop_size: DesktopSize,
+    kdc_url: Option<String>,
```
CBenoit, line 484: Should kdc_proxy_url everywhere

crates/ironrdp-web/src/session.rs
```diff
@@ -522,17 +546,26 @@ async fn connect(
...
-    // .with_static_channel(ironrdp::dvc::Drdynvc::new()); // FIXME: drdynvc is not working
+    let mut connector = connector::ClientConnector::new(config);
 
-    let upgraded = connect_rdcleanpath(&mut framed, &mut connector, destination, proxy_auth_token, pcb).await?;
+    let (upgraded, server_public_key) =
+        connect_rdcleanpath(&mut framed, &mut connector, destination.clone(), proxy_auth_token, pcb).await?;
 
-    let connection_result = ironrdp_futures::connect_finalize(upgraded, &mut framed, connector).await?;
+    info!("kdc url = {:?}", &kdc_url);
+    let networkc_client = kdc_url.map(WasmNetworkClient::new);
```
CBenoit, line 559: Typo

crates/ironrdp-connector/src/lib.rs
CBenoit, line 109: kdc_proxy_url

crates/ironrdp-connector/src/lib.rs
```diff
@@ -97,6 +100,38 @@ pub struct Config {
...
+    pub hostname: Option<String>,
+}
+
+impl SspiConfig {
+    pub fn new(kdc_proxy_url: Option<impl Into<String>>, hostname: Option<impl Into<String>>) -> ConnectorResult<Self> {
```
CBenoit, line 114: Overall, it’s not worth to have stuff like `Option<impl Into<String>>`
```suggestion
    pub fn new(kdc_proxy_url: Option<String>, hostname: Option<String>) -> ConnectorResult<Self> {
```

crates/ironrdp-connector/src/lib.rs
```diff
@@ -97,6 +100,38 @@ pub struct Config {
...
+        let kdc_url: Option<url::Url> = match kdc_proxy_url {
+            Some(inner) => {
+                let url_str: String = inner.into();
+                let url = url::Url::parse(&url_str).map_err(|_| general_err!("kdc url cannot be parsed"))?;
+                Some(url)
+            }
+            None => None,
+        };
+        let hostname: Option<String> = hostname.map(|inner| inner.into());
+        Ok(Self { kdc_url, hostname })
```
CBenoit, line 124: Can be greatly simplified (along above comment):
```suggestion
        let kdc_url = kdc_proxy_url.and_then(|url| url::Url::parse(&url).ok());
        Ok(Self { kdc_url, hostname })
```

CBenoit, line 124: Nit: FYI, this is another way to write such code:
```suggestion
        let kdc_url = kdc_proxy_url.map(|url| url::Url::parse(&url)).transpose().map_err(|e| custom_err!("invalid KDC URL", e)?;
        Ok(Self { kdc_url, hostname })
```

crates/ironrdp-connector/src/connection.rs
```diff
@@ -53,6 +56,173 @@ pub struct ConnectionResult {
     pub no_server_pointer: bool,
 }
 
+// FIXME: move that to a new module at ironrdp_connector::credssp
```
CBenoit, line 59: You should do this

web-client/iron-remote-gui/src/services/PublicAPI.ts
```diff
@@ -23,6 +23,7 @@ export class PublicAPI {
         authToken: string,
         desktopSize?: DesktopSize,
         preConnectionBlob?: string,
+        kdc_proxy_url?: string,
```
CBenoit, line 26: You should update the README.md for iron-remote-gui

crates/ironrdp-client/src/network_cllient.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/session.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/session.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/session.rs
irvingoujAtDevolution: updated

crates/ironrdp-connector/src/lib.rs
irvingoujAtDevolution: updated

crates/ironrdp-connector/src/connection.rs
irvingoujAtDevolution: updated

web-client/iron-remote-gui/src/services/PublicAPI.ts
irvingoujAtDevolution: updated

web-client/iron-svelte-client/src/lib/login/login.svelte
irvingoujAtDevolution: updated

web-client/iron-remote-gui/README.md
```diff
@@ -52,12 +52,31 @@ You can add some parameters for default initialization on the component `<iron-r
 
 ## `UserInteractionService` methods
 
-> `connect(username: string, password: string, host: string, authtoken: string): Observable<NewSessionInfo>`
+```typesccript
+ private connect(
```
CBenoit, line 56: I think you should copy from `UserInteraction.ts` instead:
```ts
    connect(
        username: string,
        password: string,
        destination: string,
        proxyAddress: string,
        serverDomain: string,
        authToken: string,
        desktopSize?: DesktopSize,
        preConnectionBlob?: string,
        kdc_proxy_url?: string,
    ): Observable<NewSessionInfo>;
```

(no `private`)

web-client/iron-remote-gui/README.md
```diff
@@ -52,12 +52,31 @@ You can add some parameters for default initialization on the component `<iron-r
 
 ## `UserInteractionService` methods
 
-> `connect(username: string, password: string, host: string, authtoken: string): Observable<NewSessionInfo>`
+```typesccript
```
CBenoit, line 55: Typo
You can use `ts` instead

web-client/iron-remote-gui/README.md
```diff
@@ -52,12 +52,31 @@ You can add some parameters for default initialization on the component `<iron-r
...
+        authToken: string,
+        desktopSize?: DesktopSize,
+        preConnectionBlob?: string,
+        kdc_proxy_url?: string,
+    ): Observable<NewSessionInfo> 
+```
 >
 > `username` and `password` are the credentials to use on the remote host.
+
 > `host` refers to the Devolutions Gateway hostname and port.
```
CBenoit, line 71: This is out of date (named `destination` above)

web-client/iron-remote-gui/README.md
irvingoujAtDevolution: updated

web-client/iron-remote-gui/README.md
irvingoujAtDevolution: updated

web-client/iron-remote-gui/README.md
irvingoujAtDevolution: updated

CBenoit (commented):
Did you test NTLM?

I’m getting this error when I do:
```
Connection error: [read frame by hint] custom error, caused by: received fatal alert: InternalError
```

The "fatal alert: InternalError" is coming from sspi-rs

crates/ironrdp-async/src/connector.rs
```diff
@@ -68,6 +84,87 @@ where
...
+                state = generator.resume(Ok(response));
+            }
+            GeneratorState::Completed(client_state) => {
+                break Ok(client_state.map_err(|e| custom_err!("cannot resolve generator state", e))?)
+            }
+        }
+    }
+}
+
+#[instrument(level = "trace", skip(network_client, framed, buf, server_name, server_public_key))]
```
CBenoit, line 105: It’s okay to just `skip_all` in this case

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+        match self.state {
+            CredSSPState::CredsspInitial => None,
+            CredSSPState::CredsspReplyNeeded => Some(&CREDSSP_TS_REQUEST_HINT),
+            CredSSPState::CredsspEarlyUserAuthResult => Some(&CREDSSP_EARLY_USER_AUTH_RESULT_HINT),
+            CredSSPState::Finishied => panic!(),
+        }
+    }
+
+    pub fn new(
+        connector: &ClientConnector,
```
CBenoit, line 66: The connector should not be passed to the CredSspSequence, I think it communicates the wrong intention: a ClientConnector is not required to build the CredSspSequence per se. The precondition (i.e. : the connector is in the CredSSP state), should be checked beforehand, and it’s indeed checked in the `perform_credssp_step` function.

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+        server_name: impl Into<ServerName>,
```
CBenoit, line 67: Take a `ServerName` directly, to avoid unnecessary monomorphization.
```suggestion
        server_name: ServerName,
```
See: https://github.com/Devolutions/IronRDP/blob/23777d6b8b9da20f575b861cdc36430a9e1c975a/ARCHITECTURE.md#core-tier

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    pub fn is_done(&self) -> bool {
+        self.state == CredSSPState::Finishied
+    }
+
+    pub fn wants_request_from_server(&self) -> bool {
+        self.next_request.is_none()
+    }
+
+    // FIXME: support for EarlyUserAuthResult
```
CBenoit, line 130: Leftover?

crates/ironrdp-async/src/connector.rs
```diff
@@ -78,7 +175,7 @@ where
 {
     buf.clear();
 
-    let written = if let Some(next_pdu_hint) = connector.next_pdu_hint() {
+    let written: ironrdp_connector::Written = if let Some(next_pdu_hint) = connector.next_pdu_hint() {
```
CBenoit, line 178: Needless type annotation

crates/ironrdp-async/src/connector.rs
```diff
@@ -68,6 +84,87 @@ where
...
+}
+
+#[instrument(level = "trace", skip(network_client, framed, buf, server_name, server_public_key))]
+async fn perform_credssp_step<S>(
+    framed: &mut Framed<S>,
+    connector: &mut ClientConnector,
+    buf: &mut WriteBuf,
+    server_name: impl Into<ServerName>,
+    server_public_key: Vec<u8>,
+    mut network_client: Option<impl AsyncNetworkClient>,
```
CBenoit, line 112: FYI, it’s possible to use a trait object behind an exclusive reference (`&mut`).
```suggestion
    mut network_client: Option<&mut dyn AsyncNetworkClient>,
```
This also cause the code to not be monomorphized, which is likely an acceptable trade off in this case

crates/ironrdp-connector/src/connection.rs
```diff
@@ -253,13 +178,15 @@ impl ClientConnector {
...
-        )
+    pub fn should_perform_credssp(&self) -> bool {
+        matches!(self.state, ClientConnectorState::CredSsp { .. })
+    }
+
+    pub fn mark_credssp_as_done(&mut self) -> Written {
+        assert!(self.should_perform_credssp());
+        let res = self.step(&[], &mut WriteBuf::new()).expect("transition to next state");
+        debug_assert!(!self.should_perform_credssp());
+        res
```
CBenoit, line 189: This function should not return a `Written`, because we don’t expect anything from the connector in addition to the state transition. However, as a sanity check, you can assert for `Written::Nothing` if you want.
```suggestion
    pub fn mark_credssp_as_done(&mut self) {
        assert!(self.should_perform_credssp());
        let res = self.step(&[], &mut WriteBuf::new()).expect("transition to next state");
        assert_eq!(res, Written::Nothing);
        debug_assert!(!self.should_perform_credssp());
```

crates/ironrdp-async/src/connector.rs
```diff
@@ -68,6 +84,87 @@ where
...
+) -> ConnectorResult<ironrdp_connector::Written>
```
CBenoit, line 113: For the same reason.
```suggestion
) -> ConnectorResult<()>
```

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+    state: CredSSPState,
+    selected_protocol: nego::SecurityProtocol,
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum CredSSPState {
+    CredsspInitial,
+    CredsspReplyNeeded,
+    CredsspEarlyUserAuthResult,
+    Finishied,
```
CBenoit, line 52:
```suggestion
    Finished,
```

crates/ironrdp-web/src/network_client.rs
```diff
@@ -1,73 +1,80 @@
...
+        let result = JsFuture::from(reader.read())
+            .await
+            .map_err(|_e| general_err!("error read stream"))?;
+
+        // Cast the result into an object and check if the stream is done
+        let result_obj = result.dyn_into::<js_sys::Object>().unwrap();
+        let done = js_sys::Reflect::get(&result_obj, &"done".into())
+            .map_err(|_| general_err!("error read stream"))?
+            .as_bool()
+            .ok_or_else(|| general_err!("error resolve reader promise proerty: done"))?;
```
CBenoit, line 66:
```suggestion
            .ok_or_else(|| general_err!("error resolve reader promise property: done"))?;

```

crates/ironrdp-client/src/network_client.rs
```diff
@@ -0,0 +1,116 @@
...
+            .await
+            .map_err(|e| custom_err!("Error sending udp request", e))?;
+
+        // 48 000 bytes: default maximum token len in Windows
+        let mut buf = vec![0; 0xbb80];
+
+        let n = udp_socket
+            .recv(&mut buf)
+            .await
+            .map_err(|e| custom_err!("Error receving udp request", e))?;
```
CBenoit, line 87:
```suggestion
            .map_err(|e| custom_err!("Error receiving UDP request", e))?;
```

web-client/iron-remote-gui/README.md
```diff
@@ -52,12 +52,30 @@ You can add some parameters for default initialization on the component `<iron-r
...
+```
+
 > `username` and `password` are the credentials to use on the remote host.
-> `host` refers to the Devolutions Gateway hostname and port.
+
+> `destination` refers to the Devolutions Gateway hostname and port.
+
 > `authtoken` is the authentication token to send to the Devolutions Gateway.
 
+> `serverDomain` is the Microsoft Doman name(if the computer has one)
```
CBenoit, line 75:
```suggestion
> `serverDomain` is the Windows domain name (if the target computer has one)
```

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+impl CredSspSequence {
+    pub fn next_pdu_hint(&self) -> Option<&dyn PduHint> {
+        match self.state {
+            CredSSPState::CredsspInitial => None,
+            CredSSPState::CredsspReplyNeeded => Some(&CREDSSP_TS_REQUEST_HINT),
+            CredSSPState::CredsspEarlyUserAuthResult => Some(&CREDSSP_EARLY_USER_AUTH_RESULT_HINT),
+            CredSSPState::Finishied => panic!(),
```
CBenoit, line 61: As done in `ClientConnector`
```suggestion
            CredSSPState::Finishied => None,
```

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+                };
+                Ok(())
+            }
+            _ => Err(general_err!("CredSsp Sequence is Finished")),
+        }
+    }
+
+    pub fn process(&mut self) -> CredSspProcessGenerator<'_> {
+        let request = self.next_request.take().expect("next request");
+        info!("Ts request = {:?}", &request);
```
CBenoit, line 158: With tracing, you can log values in a structured manner. Also this should not be `info` level. `debug` or `trace` is appropriate.
```suggestion
        let ts_request = self.next_request.take().expect("next request");
        debug!(?ts_request);
```

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Debug)]
+pub struct CredSspSequence {
+    client: CredSspClient,
+    next_request: Option<credssp::TsRequest>,
+    state: CredSSPState,
+    selected_protocol: nego::SecurityProtocol,
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum CredSSPState {
```
CBenoit, line 48: [Rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) is
```suggestion
pub(crate) enum CredsspState {
```
> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    pub fn wants_request_from_server(&self) -> bool {
+        self.next_request.is_none()
+    }
+
+    // FIXME: support for EarlyUserAuthResult
+    pub fn read_request_from_server(&mut self, input: &[u8]) -> ConnectorResult<()> {
+        match self.state {
+            CredSSPState::CredsspInitial | CredSSPState::CredsspReplyNeeded => {
+                info!("read request from server: {:?}", input);
```
CBenoit, line 134: Should not be INFO level

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        let service_principal_name = format!("TERMSRV/{}", &server_name);
+
+        let credssp_config: Box<dyn ProtocolConfig>;
+        if let Some(ref krb_config) = config.sspi_config {
+            credssp_config = Box::new(Into::<KerberosConfig>::into(krb_config.clone()));
+        } else {
+            credssp_config = Box::<sspi::ntlm::NtlmConfig>::default();
+        }
+        info!("using config : {:?}", &credssp_config);
```
CBenoit, line 94: Should not be INFO level

crates/ironrdp-connector/src/connection.rs
```diff
@@ -65,15 +38,7 @@ pub enum ClientConnectorState {
...
-    CredsspInitial {
-        selected_protocol: nego::SecurityProtocol,
-    },
-    /// At this point, user code may use `sspi::credssp::TsRequest::read_length` to find the correct frame length.
-    CredsspReplyNeeded {
-        selected_protocol: nego::SecurityProtocol,
-        credssp_client: Box<credssp::CredSspClient>,
-    },
-    CredsspEarlyUserAuthResult {
+    CredSsp {
```
CBenoit, line 41: Similarly to above, the convention would be:
```suggestion
    Credssp {
```

crates/ironrdp-connector/src/lib.rs
```diff
@@ -97,6 +99,36 @@ pub struct Config {
     pub no_server_pointer: bool,
     /// If true, the INFO_AUTOLOGON flag is set in the [`ironrdp_pdu::rdp::ClientInfoPdu`].
     pub autologon: bool,
+    // for kerberos authentication
+    pub sspi_config: Option<SspiConfig>,
```
CBenoit, line 103: This should not be part of the "connector" configuration. Instead, you can pass it to the `CredsspSquence::new` function. Also, this should probably be named `KerberosConfig` since it is exclusively used to build an sspi-rs `KerberosConfig`.
```rust
impl CredSspSequence {
    /* snip */

    pub fn new(
        connector: &ClientConnector,
        kerberos_config: Option<KerberosConfig>,
        server_name: ServerName,
        server_public_key: Vec<u8>,
    ) -> ConnectorResult<Self> {
        /* snip */
    }
}
```

crates/ironrdp-client/Cargo.toml
```diff
@@ -56,8 +56,10 @@ anyhow = "1"
 smallvec = "1.11"
 tap = "1"
 semver = "1"
+reqwest = "0.11.22"
+url = "2.4.1"
 
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.48", features = [
-    "Win32_Foundation",
+    "Win32_Foundation"
```
CBenoit, line 64: FYI, trailing commas are the convention in Rust (it makes it easier to insert new values later without modifying both lines)

crates/ironrdp-client/src/network_client.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::{future::Future, pin::Pin};
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::{TcpStream, UdpSocket};
+
+use ironrdp::connector::{custom_err, general_err, ConnectorResult};
+use reqwest::Client;
+use url::Url;
+
+use ironrdp_tokio::AsyncNetworkClient;
+pub(crate) struct AsyncTokioNetworkClient {
```
CBenoit, line 12: This is probably a more accurate name:
```suggestion
pub(crate) struct ReqwestNetworkClient {
```

crates/ironrdp-web/src/network_client.rs
```diff
@@ -1,73 +1,80 @@
...
+pub(crate) struct WasmNetworkClient {
+    kdc_proxy_url: String,
+}
+impl AsyncNetworkClient for WasmNetworkClient {
+    fn send<'a>(
+        &'a mut self,
+        network_request: &'a NetworkRequest,
+    ) -> Pin<Box<dyn Future<Output = ConnectorResult<Vec<u8>>> + 'a>> {
+        Box::pin(async move {
+            info!("network requwest = {:?}", &network_request);
```
CBenoit, line 22: Should not be INFO level

crates/ironrdp-web/src/network_client.rs
```diff
@@ -1,73 +1,80 @@
...
+use ironrdp::connector::sspi::network_client::NetworkProtocol;
+use ironrdp::connector::{general_err, reason_err, ConnectorResult};
+use ironrdp_futures::AsyncNetworkClient;
+use wasm_bindgen::JsCast;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::ReadableStream;
+
+#[derive(Debug)]
+pub(crate) struct WasmNetworkClient {
+    kdc_proxy_url: String,
```
CBenoit, line 14: Isn’t that supposed to be retrieved from the `NetworkRequest`? (`request.url`)

irvingoujAtDevolution: updated

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
+use ironrdp_pdu::write_buf::WriteBuf;
+use ironrdp_pdu::{nego, PduHint};
+use sspi::credssp::{self, ClientState, CredSspClient};
+use sspi::generator::{Generator, NetworkRequest};
+use sspi::negotiate::ProtocolConfig;
+use sspi::KerberosConfig as SSPIKerberosConfig;
```
CBenoit, line 6: You can use `sspi::KerberosConfig` instead of renaming it

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+
+const CREDSSP_EARLY_USER_AUTH_RESULT_HINT: CredsspEarlyUserAuthResultHint = CredsspEarlyUserAuthResultHint;
+
+impl PduHint for CredsspEarlyUserAuthResultHint {
+    fn find_size(&self, _: &[u8]) -> ironrdp_pdu::PduResult<Option<usize>> {
+        Ok(Some(sspi::credssp::EARLY_USER_AUTH_RESULT_PDU_SIZE))
+    }
+}
+
+pub type CredSspProcessGenerator<'a> = Generator<'a, NetworkRequest, sspi::Result<Vec<u8>>, sspi::Result<ClientState>>;
```
CBenoit, line 38: Naming convention
```suggestion
pub type CredsspProcessGenerator<'a> = Generator<'a, NetworkRequest, sspi::Result<Vec<u8>>, sspi::Result<ClientState>>;
```

crates/ironrdp-async/src/connector.rs
```diff
@@ -68,17 +87,99 @@ where
...
+            None
+        };
+
+        if credssp_sequence.wants_request_from_server() {
+            credssp_sequence.read_request_from_server(&input.unwrap_or_else(|| [].to_vec()))?;
+        }
+        let client_state = {
+            let mut generator = credssp_sequence.process();
+            if let Some(ref network_client_ref) = network_client {
+                info!("resolving network");
```
CBenoit, line 149: Should not be INFO level

crates/ironrdp-web/src/session.rs
```diff
@@ -522,17 +534,38 @@ async fn connect(
...
+        connector,
+        Some(KerberosConfig {
+            kdc_proxy_url: kdc_proxy_url
+                .map(|url| url::Url::parse(&url))
+                .transpose()
+                .context("invalid KDC URL")?,
+            hostname: Some(destination),
+        }),
+    )
+    .await?;
```
CBenoit, line 568: Why not pass the network client unconditionally?
```suggestion
    let mut network_client = WasmNetworkClient::new();

    let connection_result = ironrdp_futures::connect_finalize(
        upgraded,
        &mut framed,
        (&destination).into(),
        server_public_key,
        Some(&mut network_client),
        connector,
        Some(KerberosConfig {
            kdc_proxy_url: kdc_proxy_url
                .map(|url| url::Url::parse(&url))
                .transpose()
                .context("invalid KDC URL")?,
            hostname: Some(destination),
        }),
    )
    .await?;
```

crates/ironrdp-web/src/session.rs
```diff
@@ -522,17 +534,38 @@ async fn connect(
...
+        (&destination).into(),
+        server_public_key,
+        network_client,
+        connector,
+        Some(KerberosConfig {
+            kdc_proxy_url: kdc_proxy_url
+                .map(|url| url::Url::parse(&url))
+                .transpose()
+                .context("invalid KDC URL")?,
+            hostname: Some(destination),
```
CBenoit, line 565: IIRC, this is not supposed to be the "destination" computer name, but the client computer name. There is no way to easily retrieve the computer name from the browser, and if this is a workaround, you should explicitly mark it as such:
```suggestion
            // HACK: It’s supposed to be the computer name of the client, but since it’s not easy to retrieve the information, we set the destination hostname
            // because it happens to work.
            hostname: Some(destination),
```

crates/ironrdp-connector/src/credssp_sequence.rs
```diff
@@ -0,0 +1,214 @@
...
+        let server_name = server_name.into_inner();
+
+        let service_principal_name = format!("TERMSRV/{}", &server_name);
+
+        let credssp_config: Box<dyn ProtocolConfig>;
+        if let Some(ref krb_config) = kerberos_config {
+            credssp_config = Box::new(Into::<SSPIKerberosConfig>::into(krb_config.clone()));
+        } else {
+            credssp_config = Box::<sspi::ntlm::NtlmConfig>::default();
+        }
```
CBenoit, line 94: We should use the "negotiate" security provider: https://github.com/Devolutions/sspi-rs/blob/7779af59d32af845f25778897dc5ff8e5febde2f/src/negotiate.rs#L37-L42
https://learn.microsoft.com/en-us/windows/win32/secauthn/microsoft-negotiate

This provider will automatically use Kerberos and fallback to NTLM when necessary.

crates/ironrdp-async/src/connector.rs
```diff
@@ -68,17 +87,99 @@ where
     Ok(result)
 }
 
+async fn resolve_generator(
+    generator: &mut CredSspProcessGenerator<'_>,
+    mut network_client: Box<dyn AsyncNetworkClient>,
```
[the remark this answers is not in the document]
irvingoujAtDevolution: This is my initial thoughts as well, but since we passed the network down the stack as Option<&mut AsyncNetworkClient> and we use it in a loop, we cannot move out of options.

crates/ironrdp-connector/src/credssp_sequence.rs
irvingoujAtDevolution: yes, that is indeed what I use. it takes a protocol config as parameter, which if it is Kerbero config, it will user Kerberos, then if it fails, it will fall back to NTLM, if Kerberos config not persented at all, it starts with NTLM. 
``` Rust
 let credssp_config: Box<dyn ProtocolConfig>;
        if let Some(ref krb_config) = kerberos_config {
            credssp_config = Box::new(Into::<sspi::KerberosConfig>::into(krb_config.clone()));
        } else {
            credssp_config = Box::<sspi::ntlm::NtlmConfig>::default();
        }
        
        info!("using config : {:?}", &credssp_config);

        let client = credssp::CredSspClient::new(
            server_public_key,
            credentials.into(),
            credssp::CredSspMode::WithCredentials,
            credssp::ClientMode::Negotiate(sspi::NegotiateConfig { //<-- here
                protocol_config: credssp_config,
                package_list: None,
                hostname: server_name,
            }),
            service_principal_name,
        )
```

crates/ironrdp-client/src/network_client.rs
irvingoujAtDevolution: updated

crates/ironrdp-connector/src/credssp_sequence.rs
irvingoujAtDevolution: updated

crates/ironrdp-connector/src/credssp_sequence.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/network_client.rs
irvingoujAtDevolution: updated

crates/ironrdp-async/src/connector.rs
irvingoujAtDevolution: updated

crates/ironrdp-async/src/connector.rs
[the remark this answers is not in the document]
irvingoujAtDevolution: if this is not desired way of doing it, we can try consume the network client at top level. Let me know which way you perfer

crates/ironrdp-connector/src/credssp_sequence.rs
CBenoit: Oh, okay! Sounds good

irvingoujAtDevolution:
Messed Up, hold on for any review, fixing merge/rebase conflict with Vlad's change

crates/ironrdp-async/src/connector.rs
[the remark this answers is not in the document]
CBenoit: Oh yeah, the error is not straightforward. Actually, stuff like `Option<&mut T>` are not super ergonomic in some cases. However, you can definitely retrieve the `&mut T` inside.
```rust
async fn perform_credssp_step<S>(
    /* … */
    mut network_client: Option<&mut dyn AsyncNetworkClient>, // <- make the binding mutable
    /* … */
) -> ConnectorResult<()>
where
    S: FramedRead + FramedWrite,
{
    /* … */
            if let Some(network_client) = network_client.as_deref_mut() { // <- use `.as_deref_mut()`
                trace!("resolving network");
                resolve_generator(&mut generator, network_client).await?
            } else {
                generator
                    .resolve_to_result()
                    .map_err(|e| custom_err!(" cannot resolve generator without a network client", e))?
            }
    /* … */
```

This way, you can borrow "exclusively" the `Option<&mut T>`, ensuring no one else have access to the inner `&mut T` at the same time, without moving `&mut T` out of it (you should think of `&mut` as "exclusive reference" in Rust).

Also you can basically forget about the `ref` keyword. It’s not useful since https://rust-lang.github.io/rfcs/2005-match-ergonomics.html (you may see it in old codebases, but that’s pretty much about it)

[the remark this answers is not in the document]
irvingoujAtDevolution: updated

irvingoujAtDevolution:
Fixed NTLM not authenticate problem, see https://github.com/Devolutions/sspi-rs/pull/187

CBenoit (approved):
Let’s merge this now, I’ll send a follow up PR for the remaining suggestions.

PR #264, opened 2023-11-06
Update prettier end of line configuration

irvingoujAtDevolution:
Updated Prettier configurations to avoid ending character conflict.
Reference:
https://prettier.io/docs/en/options 
end of line options
https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-eslint-delete-cr-prettier-prettier

### DIFF
--- a/web-client/.prettierrc.yaml
+++ b/web-client/.prettierrc.yaml
@@ -10,6 +10,7 @@ singleQuote: true
 semi: true
 trailingComma: all
 printWidth: 120
+endOfLine: auto
 
 plugins:
   - prettier-plugin-svelte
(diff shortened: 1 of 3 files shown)


CBenoit (commented):
I want line endings to be unified to LF so I don’t want to modify the prettierrc config this way
However, I understand eslint is giving you errors because your editor is inserting CRLF? In that case, I think it’s fine to modify the eslint configuration instead in order to relax this, as mentioned in the stackoverflow answer
Alternatively, make sure your editor is auto-formatting when saving

irvingoujAtDevolution:
> I want line endings to be unified to LF so I don’t want to modify the prettierrc config this way However, I understand eslint is giving you errors because your editor is inserting CRLF? In that case, I think it’s fine to modify the eslint configuration instead in order to relax this, as mentioned in the stackoverflow answer Alternatively, make sure your editor is auto-formatting when saving

I have pushed a new commit, changes eslint config isntead of prettier

CBenoit (approved):
Thank you!

PR #284, opened 2023-11-15
Ignore *.css in Eslint

irvingoujAtDevolution:
Ignore css files in Eslint process for now. We don't have a lot of css file at this moment and it blocks other PR from merging. 

Eslint will give this error is ./web/iron-svelte-client
```
C:\Users\jou\code\IronRDP\web-client\iron-svelte-client\src\lib\login\login.css
  0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/src\lib\login\login.css` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
The extension for the file (`.css`) is non-standard. It should be added to your existing `parserOptions.extraFileExtensions`

✖ 1 problem (1 error, 0 warnings)
```

### DIFF
--- a/web-client/iron-svelte-client/.eslintrc.yaml
+++ b/web-client/iron-svelte-client/.eslintrc.yaml
@@ -31,6 +31,7 @@ parserOptions:
 
 ignorePatterns:
   - '*.cjs'
+  - '*.css'
 
 overrides:
   - files: '*.svelte'


CBenoit:
@irvingoujAtDevolution I don’t see the problem in CI for #260

irvingoujAtDevolution:
> @irvingoujAtDevolution I don’t see the problem in CI for #260

That is wired, I thought that is the problem. I would get this if I run locally. Do you have it?
image

CBenoit:
> > @irvingoujAtDevolution I don’t see the problem in CI for #260
> 
> That is wired, I thought that is the problem. I would get this if I run locally. Do you have it? image

Strange, I don’t have this:
image

CBenoit (approved):
Unsure about what is causing this difference in behavior, but as far as I understand, we don’t really lint *.css files anyway (at least, we’re currently not using something like `eslint-plugin-css` as of today). Let’s add this ignore pattern :+1:

PR #285, opened 2023-11-15
chore(web-client): bump prettier to 3.1.0

CBenoit:
A bug in 3.0.x versions was causing the --plugin setting to be required in order to let prettier know about svelte. This is now fixed.


irvingoujAtDevolution (approved):
LGTM

PR #286, opened 2023-11-16
feat(cliprdr): BI_BITFIELDS compression is supported

CBenoit:
As it turns out we actually support the BI_BITFIELDS compression (which is not actually for compressed bitmaps).

Quote from [MS-WMF] section 2.1.1.7:

> The bitmap is not compressed, and the color table consists of three
> DWORD (defined in [MS-DTYP] section 2.2.9) color masks that specify the
> red, green, and blue components, respectively, of each pixel. This is
> valid when used with 16 and 32-bits per pixel bitmaps.

[dibv5-bitfields-compression.webm](https://github.com/Devolutions/IronRDP/assets/3809077/440695e7-88ce-4d64-96e1-670ec8b5409f)

All I did was to add `BitmapCompression::BITFIELDS` to the `SUPPORTED_COMPRESSION` array, and things just worked out of the box.
@pacmancoder maybe you understand why it works better than me?

### DIFF
--- a/crates/ironrdp-cliprdr-format/src/bitmap.rs
+++ b/crates/ironrdp-cliprdr-format/src/bitmap.rs
@@ -417,13 +417,6 @@ fn validate_v1_header(header: &BitmapInfoHeader) -> Result<(), BitmapError> {
         return Err(BitmapError::Unsupported("unsupported bit count"));
     }
 
-    // We support only uncompressed DIB bitmaps as it is the most common case for clipboard-copied
-    // bitmaps.
-    const SUPPORTED_COMPRESSION: &[BitmapCompression] = &[BitmapCompression::RGB, BitmapCompression::BITFIELDS];
-    if !SUPPORTED_COMPRESSION.contains(&header.compression) {
-        return Err(BitmapError::Unsupported("unsupported compression"));
-    }
-
     // This is only relevant for bitmaps with bpp < 24, which are not supported.
     if header.clr_used != 0 {
         return Err(BitmapError::Unsupported("color table is not supported"));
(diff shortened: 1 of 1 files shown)


pacmancoder (changes requested):
@CBenoit I believe It works in this example because the target remote machine uses `CF_DIBV5` and the default color order/masks for its output.

Correct `BITFIELDS` support would require the support of variable-size header for `CF_DIB` and interpreting `*_mask` fields correctly for `CF_DIBV5` -- those color masks should select the order of color channels in the image.

If we just enable `BITFIELDS` flag, `CF_DIB`s will break, because we will not read additional color mask data following the fixed header  (which will result in a corrupted bitmap). For `CF_DIBV5`, it will not break the world, but if color masks will be non-standard, then color channels could be read in the wrong order because we will assume the following color channel order/masks:

```
    // Written as little-endian, interpreted as big-endian -- standard order is BGRA
        red_mask: 0x00FF0000,
        green_mask: 0x0000FF00,
        blue_mask: 0x000000FF,
        alpha_mask: 0xFF000000,
```

I haven't encountered `BITFIELDS` in 24/32 bit mode when copying images from paint/gimp/paint.net/browsers, therefore, assumed it should be REALLY rarely used... But if the Windows snip tool uses that, then it should be supported 😅 . 

To merge this support for minimal `BITFIELDS` support with default masks, we need to add additional checks for DIBV5:
```rust
let is_color_order_bgra = header.red_mask == 0x00FF0000
    || header.green_mask == 0x0000FF00
    || header.blue_mask == 0x000000FF
    || header.alpha_mask == 0xFF000000;

if (header.compression == BitmapCompression::BITFIELDS && !is_color_order_bgra) {
    return Err(BitmapError::Unsupported("non-standard color masks for `BITFIELDS` is not supported"));
}
```

For `CF_DIB` we need to reject `BITFIELDS` until variable size header is supported

CBenoit:
@pacmancoder thank you! I adjusted a bit according to your suggestion: [Code review](https://github.com/Devolutions/IronRDP/pull/286/commits/f10526b22d359948cd7b94f817cd6556f9d54ddd)

crates/ironrdp-cliprdr-format/src/bitmap.rs
```diff
@@ -435,6 +428,29 @@ fn validate_v1_header(header: &BitmapInfoHeader) -> Result<(), BitmapError> {
...
+    }
+
+    if header.header_v1.compression == BitmapCompression::BITFIELDS {
+        // Currently, we only support the standard order, BGRA, for the bitfields compression.
+        // Note: we are making sure to interpret the masks as big-endian using `u32::to_be`.
+        // (`to_be` is a no-op on big endian architectures.)
+        let is_bgra = header.red_mask.to_be() == 0x00FF0000
+            || header.green_mask.to_be() == 0x0000FF00
+            || header.blue_mask.to_be() == 0x000000FF
+            || header.alpha_mask.to_be() == 0xFF000000;
```
CBenoit, line 445: Although we may never hit this problem in practice, I ensured proper interpretation across all native endianness using `to_be`

pacmancoder (approved):
LGTM 🚀

crates/ironrdp-cliprdr-format/src/bitmap.rs
pacmancoder: `to_be` will actually break the code here, constants on the `rhs` (e.g. `0x00FF0000`) are implicit `le` on most of the systems, but could be `be` in some bizarre cases. However, the left-hand side is always represented as `be` because of `to_be`. 

AFIK, numeric values in our PDUs have values with endianness native to the current machine, therefore comparison without `to_le`/`to_be` is fine

CBenoit: I meant to use `to_le` in this case. This is fixed in https://github.com/Devolutions/IronRDP/pull/286/commits/76379ab2600e7a65edca5c188150de9957f5a9c0

`to_le` is a no-op when little-endian is the native endianness, so it’s basically free on most systems. However, on a big endian architecture, the code is wrong `to_le`, because the hexadecimal literal will not be converted to big endian automatically when compiling for such target. (Albeit, I don’t expect to see such target in practice.)

CBenoit: Update: my assumption was wrong.

```rust
fn main() {
    let network_packet = [0xFF, 0x00];
    let int = u16::from_le_bytes(network_packet);
    println!("{:02X?}", int.to_ne_bytes());
    assert_eq!(int, 0x00FF);
}
```

image

As you said!

pacmancoder (approved):

PR #290, opened 2023-11-20
feat(web): support for hardware cursor

### LINKED ISSUE #250: Support for hardware cursor in the browser web client
The cursor image should be set by updating the CSS.

It’s possible to inline the raw bitmap by encoding it in base64:
```
url(data:image/bmp;base64,Qk06AAAAAAAAADYAAAAoAAAAAQAAAAEAAAABABgAAAAAAAQAAADEDgAAxA4AAAAAAAAAAAAAAgD+AA==)
```

Note that:
- It’s okay to not handle animated cursor in an efficient way because it’s a pretty rare scenario. If it works out of the box or if there is an easy solution to increase the performance, it’s nice, but no need to put too much effort into supporting this.
- Some cursor icons such as the I-beam have "invert color" pixels. I’m not sure it’s possible to set invert pixels using CSS. Maybe there is something smart to do with [`mix-blend-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode), but it’s fine to consider these pixels as black if it’s too much trouble.

pacmancoder:
Adds support for native cursors for `IronRDP-web`.

- Uses CSS `cursor` property to set the custom cursor (which we encode as Base64 data URL
- Software rendering is still in place and could be enabled via the config flag
- "Inverted colors" are implemented for native cursors as a "checkerboard pattern" (As in FreeRDP)
- Native cursors have size **limitation** - _Any remote cursor bigger than 32x32 will be scaled down._ This is due to browser limitations described [here](https://chromestatus.com/feature/5825971391299584) and [here](https://bugs.chromium.org/p/chromium/issues/detail?id=880863)

Closes #250

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ rstest = "0.18"
 sspi = "0.11"
 tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
+png = "0.17"
 
 [profile.dev]
 opt-level = 1
--- a/crates/ironrdp-graphics/src/pointer.rs
+++ b/crates/ironrdp-graphics/src/pointer.rs
@@ -50,33 +50,30 @@ pub struct DecodedPointer {
 /// Pointer bitmap rendering target. Defines properties and format of the decoded bitmap.
 #[derive(Clone, Copy, Debug)]
 pub enum PointerBitmapTarget {
-    /// Software rendering target will produce RGBA bitmaps with premultiplied alpha
-    /// and inverted pixels represented as `[0xFF, 0xFF, 0xFF, 0x00]` color.
+    /// Software rendering target will produce RGBA bitmaps with premultiplied alpha.
+    ///
+    /// Colors with alpha channel set to 0x00 are always invisible no matter their color
+    /// component. We could take advantage of that, and use a special color to represent
+    /// inverted pixels. [0xFF, 0xFF, 0xFF, 0x00] is used for such purpose in software
+    /// rendering mode.
     Software,
     /// Accelerated rendering target will produce RGBA bitmaps with non-premultiplied alpha.
-    /// Inverted pixels will be rendered checkboard pattern.
+    /// Inverted pixels will be rendered following the check pattern.
     Accelerated,
 }
 
 impl PointerBitmapTarget {
-    fn premultiply_alpha(self) -> bool {
+    fn should_premultiply_alpha(self) -> bool {
         match self {
             Self::Software => true,
             Self::Accelerated => false,
         }
     }
 
-    fn inverted_pixels_as_checkboard(self) -> bool {
+    fn should_invert_pixels_using_check_pattern(self) -> bool {
         match self {
             Self::Software => false,
             Self::Accelerated => true,
-        }
-    }
-
-    fn inverted_pixels_color(self) -> [u8; 4] {
-        match self {
-            Self::Software => [0xFF, 0xFF, 0xFF, 0x00],
-            Self::Accelerated => [0x00, 0x00, 0x00, 0x00],
         }
     }
 }
@@ -197,6 +194,19 @@ impl DecodedPointer {
             let mut color_reader = ColorStrideReader::new(data.xor_bpp, xor_stride);
             let mut bitmask_reader = BitmaskStrideReader::new(and_stride);
 
+            let compute_inverted_pixel = if target.should_invert_pixels_using_check_pattern() {
+                |row_idx: u16, col_idx: u16| -> [u8; 4] {
+                    // Checkered pattern is used to represent inverted pixels.
+                    if (row_idx + col_idx) % 2 == 0 {
+                        [0xff, 0xff, 0xff, 0xff]
+                    } else {
+                        [0x00, 0x00, 0x00, 0xff]
+                    }
+                }
+            } else {
+                |_, _| [0xFF, 0xFF, 0xFF, 0x00]
+            };
+
             for col_idx in 0..data.width {
                 let and_bit = bitmask_reader.next_bit(&mut and_stride_cursor);
                 let color = color_reader.next_pixel(&mut xor_stride_cursor);
@@ -207,18 +217,8 @@ impl DecodedPointer {
                     bitmap_data.extend_from_slice(&[0, 0, 0, 0]);
                 } else if and_bit == 1 && color == [0xff, 0xff, 0xff, 0xff] {
                     // Inverted pixel.
-
-                    if target.inverted_pixels_as_checkboard() {
-                        // Checkboard pattern is used to represent inverted pixels.
-                        if (row_idx + col_idx) % 2 == 0 {
-                            bitmap_data.extend_from_slice(&[0xff, 0xff, 0xff, 0xff]);
-                        } else {
-                            bitmap_data.extend_from_slice(&[0x00, 0x00, 0x00, 0xff]);
-                        }
-                    } else {
-                        bitmap_data.extend_from_slice(&target.inverted_pixels_color());
-                    }
-                } else if target.premultiply_alpha() {
+                    bitmap_data.extend_from_slice(&compute_inverted_pixel(row_idx, col_idx));
+                } else if target.should_premultiply_alpha() {
                     // Calculate premultiplied alpha via integer arithmetic
                     let with_premultiplied_alpha = [
                         ((color[0] as u16 * color[0] as u16) >> 8) as u8,
--- a/crates/ironrdp-graphics/src/pointer.rs
+++ b/crates/ironrdp-graphics/src/pointer.rs
@@ -1,12 +1,12 @@
 //! This module implements logic to decode pointer PDUs into RGBA bitmaps ready for rendering.
 //!
-//! #### References:
+//! # References:
 //! - Drawing pointers: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/pointer-drawing
 //! - Drawing color pointers: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-color-pointers
 //! -Drawing monochrome pointers https://learn.microsoft.com/en-us/windows-hardware/drivers/display/drawing-monochrome-pointers
 //!
 //!
-//! #### Notes on xor/and masks encoding:
+//! # Notes on xor/and masks encoding:
 //! RDP's pointer representation is a bit weird. It uses two masks to represent a pointer -
 //! andMask and xorMask. Xor mask is used as a base color for a pointer pixel, and andMask
 //! mask is used co control pixel's full transparency (`src_color.a = 0`), full opacity
--- a/crates/ironrdp-graphics/src/pointer.rs
+++ b/crates/ironrdp-graphics/src/pointer.rs
@@ -32,7 +32,7 @@ pub enum PointerError {
     #[error("invalid pointer andMask size. Expected: {expected}, actual: {actual}")]
     InvalidAndMaskSize { expected: usize, actual: usize },
     #[error("not supported pointer bpp: {bpp}")]
-    NotSupportedBpp { bpp: usize },
+    NotSupportedBpp { bpp: u16 },
     #[error(transparent)]
     Pdu(#[from] ironrdp_pdu::PduError),
 }
@@ -40,11 +40,45 @@ pub enum PointerError {
 /// Represents RDP pointer in decoded form (color channels stored as RGBA pre-multiplied values)
 #[derive(Debug)]
 pub struct DecodedPointer {
-    pub width: usize,
-    pub height: usize,
-    pub hot_spot_x: usize,
-    pub hot_spot_y: usize,
+    pub width: u16,
+    pub height: u16,
+    pub hotspot_x: u16,
+    pub hotspot_y: u16,
     pub bitmap_data: Vec<u8>,
+}
+
+/// Pointer bitmap rendering target. Defines properties and format of the decoded bitmap.
+#[derive(Clone, Copy, Debug)]
+pub enum PointerBitmapTarget {
+    /// Software rendering target will produce RGBA bitmaps with premultiplied alpha
+    /// and inverted pixels represented as `[0xFF, 0xFF, 0xFF, 0x00]` color.
+    Software,
+    /// Accelerated rendering target will produce RGBA bitmaps with non-premultiplied alpha.
+    /// Inverted pixels will be rendered checkboard pattern.
+    Accelerated,
+}
+
+impl PointerBitmapTarget {
+    fn premultiply_alpha(self) -> bool {
+        match self {
+            Self::Software => true,
+            Self::Accelerated => false,
+        }
+    }
+
+    fn inverted_pixels_as_checkboard(self) -> bool {
+        match self {
+            Self::Software => false,
+            Self::Accelerated => true,
+        }
+    }
+
+    fn inverted_pixels_color(self) -> [u8; 4] {
+        match self {
+            Self::Software => [0xFF, 0xFF, 0xFF, 0x00],
+            Self::Accelerated => [0x00, 0x00, 0x00, 0x00],
+        }
+    }
 }
 
 impl DecodedPointer {
@@ -53,49 +87,67 @@ impl DecodedPointer {
             width: 0,
             height: 0,
             bitmap_data: Vec::new(),
-            hot_spot_x: 0,
-            hot_spot_y: 0,
+            hotspot_x: 0,
+            hotspot_y: 0,
         }
     }
 
-    pub fn decode_pointer_attribute(src: &PointerAttribute<'_>) -> Result<Self, PointerError> {
-        Self::decode_pointer(PointerData {
-            width: src.color_pointer.width as usize,
-            height: src.color_pointer.height as usize,
-            xor_bpp: src.xor_bpp as usize,
-            xor_mask: src.color_pointer.xor_mask,
-            and_mask: src.color_pointer.and_mask,
-            hot_spot_x: src.color_pointer.hot_spot.x as usize,
-            hot_spot_y: src.color_pointer.hot_spot.y as usize,
-        })
+    pub fn decode_pointer_attribute(
+        src: &PointerAttribute<'_>,
+        target: PointerBitmapTarget,
+    ) -> Result<Self, PointerError> {
+        Self::decode_pointer(
+            PointerData {
+                width: src.color_pointer.width,
+                height: src.color_pointer.height,
+                xor_bpp: src.xor_bpp,
+                xor_mask: src.color_pointer.xor_mask,
+                and_mask: src.color_pointer.and_mask,
+                hot_spot_x: src.color_pointer.hot_spot.x,
+                hot_spot_y: src.color_pointer.hot_spot.y,
+            },
+            target,
+        )
     }
 
-    pub fn decode_color_pointer_attribute(src: &ColorPointerAttribute<'_>) -> Result<Self, PointerError> {
-        Self::decode_pointer(PointerData {
-            width: src.width as usize,
-            height: src.height as usize,
-            xor_bpp: 24,
-            xor_mask: src.xor_mask,
-            and_mask: src.and_mask,
-            hot_spot_x: src.hot_spot.x as usize,
-            hot_spot_y: src.hot_spot.y as usize,
-        })
+    pub fn decode_color_pointer_attribute(
+        src: &ColorPointerAttribute<'_>,
+        target: PointerBitmapTarget,
+    ) -> Result<Self, PointerError> {
+        Self::decode_pointer(
+            PointerData {
+                width: src.width,
+                height: src.height,
+                xor_bpp: 24,
+                xor_mask: src.xor_mask,
+                and_mask: src.and_mask,
+                hot_spot_x: src.hot_spot.x,
+                hot_spot_y: src.hot_spot.y,
+            },
+            target,
+        )
     }
 
-    pub fn decode_large_pointer_attribute(src: &LargePointerAttribute<'_>) -> Result<Self, PointerError> {
-        Self::decode_pointer(PointerData {
-            width: src.width as usize,
-            height: src.height as usize,
-            xor_bpp: src.xor_bpp as usize,
-            xor_mask: src.xor_mask,
-            and_mask: src.and_mask,
-            hot_spot_x: src.hot_spot.x as usize,
-            hot_spot_y: src.hot_spot.y as usize,
-        })
+    pub fn decode_large_pointer_attribute(
+        src: &LargePointerAttribute<'_>,
+        target: PointerBitmapTarget,
+    ) -> Result<Self, PointerError> {
+        Self::decode_pointer(
+            PointerData {
+                width: src.width,
+                height: src.height,
+                xor_bpp: src.xor_bpp,
+                xor_mask: src.xor_mask,
+                and_mask: src.and_mask,
+                hot_spot_x: src.hot_spot.x,
+                hot_spot_y: src.hot_spot.y,
+            },
+            target,
+        )
     }
 
-    fn decode_pointer(data: PointerData<'_>) -> Result<Self, PointerError> {
-        const SUPPORTED_COLOR_BPP: [usize; 4] = [1, 16, 24, 32];
+    fn decode_pointer(data: PointerData<'_>, target: PointerBitmapTarget) -> Result<Self, PointerError> {
+        const SUPPORTED_COLOR_BPP: [u16; 4] = [1, 16, 24, 32];
 
         if data.width == 0 || data.height == 0 {
             return Ok(Self::new_invisible());
@@ -109,19 +161,19 @@ impl DecodedPointer {
 
         let flip_vertical = data.xor_bpp != 1;
 
-        let and_stride = Stride::from_bits(data.width);
-        let xor_stride = Stride::from_bits(data.width * data.xor_bpp);
+        let and_stride = Stride::from_bits(data.width.into());
+        let xor_stride = Stride::from_bits(usize::from(data.width) * usize::from(data.xor_bpp));
 
-        if data.xor_mask.len() != xor_stride.length * data.height {
+        if data.xor_mask.len() != xor_stride.length * usize::from(data.height) {
             return Err(PointerError::InvalidXorMaskSize {
-                expected: xor_stride.length * data.height,
+                expected: xor_stride.length * usize::from(data.height),
                 actual: data.xor_mask.len(),
             });
         }
 
-        if data.and_mask.len() != and_stride.length * data.height {
+        if data.and_mask.len() != and_stride.length * usize::from(data.height) {
             return Err(PointerError::InvalidAndMaskSize {
-                expected: and_stride.length * data.height,
+                expected: and_stride.length * usize::from(data.height),
                 actual: data.and_mask.len(),
             });
         }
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -55,6 +55,7 @@ rstest = "0.18"
 sspi = "0.11"
 tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
+png = "0.17"
```
CBenoit, line 58: Thought: besides the local crates, I would prefer we only put "ubiquitous" dependencies in the workspace manifest. Following this logic, we would probably only keep `tracing` and `thiserror`. Since all tests should go into the testsuite crate, it’s likely `proptest`, `rstest` and `expect-test` don’t belong here either.

pacmancoder:
@kristahouse FYI: This PR adds native cursors support (previously, it was done via software rendering).

crates/ironrdp-web/Cargo.toml
```diff
@@ -40,6 +40,9 @@ tracing-web = "0.1"
 
 # Rendering
 softbuffer = { version = "0.3", default-features = false }
+png.workspace = true
+resize = "0.8"
```
CBenoit, line 44: Question: looks like the `rayon` feature is enabled by default. Does it work in WASM? Should we disable the default features?

crates/ironrdp-web/Cargo.toml
```diff
@@ -67,3 +70,4 @@ x509-cert = { version = "0.2", default-features = false, features = ["std"] }
 tap = "1"
 semver = "1"
 url = "2.4"
+base64 = "0.21"
```
CBenoit, line 73: Suggestion: [base64ct](https://lib.rs/crates/base64ct) is a smaller alternative, probably a good idea for a WASM module.

crates/ironrdp-web/Cargo.toml
```diff
@@ -40,6 +40,9 @@ tracing-web = "0.1"
 
 # Rendering
 softbuffer = { version = "0.3", default-features = false }
+png.workspace = true
+resize = "0.8"
+rgb = "0.8"
```
CBenoit, line 45: Praise: I would like us to use this crate more :+1:

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -282,12 +280,28 @@ export class WasmBridgeService {
...
+        if (style.startsWith('data')) {
+            if (hotspot_x == undefined || hotspot_y == undefined) {
+                console.error('Hot spot is not defined for a custom cursor.');
+                return;
+            }
 
-    private showPointerCallback() {
-        this.canvas!.style.cursor = 'default';
+            // IMPORTANT: We need to make proxy `Image` object to actually load the image and make
+            // it usable for CSS property. Without this projy object, URL will be rejected.
```
CBenoit, line 296: Typo:
```suggestion
            // it usable for CSS property. Without this proxy object, URL will be rejected.
```

crates/ironrdp-web/Cargo.toml
pacmancoder: You are right, It will be useless on web... I'll disable `rayon` feature

crates/ironrdp-web/Cargo.toml
pacmancoder: Well, `base64` API is better suited for our purpose, since we need to append base64 to an existing string (`base64::Engine::to_string` does just that), while `base64ct` either will always allocate a new string or could write into raw `&mut [u8]`). 

I think `base64` will work better here, but if you have a string opinion on that, I could switch the library

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,120 @@ impl Session {
...
+                        };
+
+                        let (png_width, png_height, hot_spot_x, hot_spot_y, rgba_buffer) = if let Some(scale) = scale {
+                            // Per invariants: Following conversions are losslesss.
+                            let scaled_width = cast_f64_to_u16_lossy(f64::from(pointer.width) * scale);
+                            let scaled_height = cast_f64_to_u16_lossy(f64::from(pointer.height) * scale);
+                            let hotspot_x = cast_f64_to_u16_lossy(f64::from(pointer.hotspot_x) * scale);
+                            let hotspot_y = cast_f64_to_u16_lossy(f64::from(pointer.hotspot_y) * scale);
+
+                            // INVARIANT: scaled_width * scaled_height * 4 <= 32 * 32 * 4
```
CBenoit, line 531: Note (style): here, we are [not defining an invariant, but using one](https://github.com/Devolutions/IronRDP/blob/08d165d71f70d4b819bcd689f339e1a4ce06ece5/STYLE.md#explain-non-obvious-assumptions-by-referencing-the-invariants).

```suggestion
                            // Per invariants: scaled_width * scaled_height * 4 <= 32 * 32 * 4 < usize::MAX
```

It’s also useful specify the `< usize::MAX` part here, because that’s the reason why there is no overflow (i.e.: 32 * 32 * 4 is smaller than `usize::MAX` on any architecture, which means we are not overflowing).

Cargo.toml
pacmancoder: Well, on the other hand, globally defined dependency versions allow very easy version upgrades and reduce the possibility of duplicated libraries (via human error)

crates/ironrdp-web/src/session.rs
pacmancoder: My idea during writing of this was that `32 * 32 * 4` part is enough to explicitly state `< usize::MAX`

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,120 @@ impl Session {
...
+                        // - 128x128 and bigger cursors are not supported in browsers.
+                        // - 32x32 and bigger cursors will default to the system cursor if their
+                        //   sprite does not fit in the browser's viewport, introducing anabrupt
+                        //   cursor style changes when the cursor is moved to the edge of the
+                        //   browser window.
+                        //
+                        // Therefore, we need to scale the cursor sprite down to 32x32 if it is
+                        // bigger than that.
+                        const MAX_CURSOR_SIZE: u16 = 32;
+                        // INVARIANT: 0 < scale <= 1.0
```
CBenoit, line 515: Note: those are additional invariants that are useful when reasoning about the following code:
```rust
// INVARIANT: pointer.width * scale = MAX_CURSOR_SIZE
// INVARIANT: pointer.height * scale = MAX_CURSOR_SIZE
```

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,120 @@ impl Session {
...
+                        let scale = if pointer.width >= pointer.height && pointer.width > MAX_CURSOR_SIZE {
+                            Some(f64::from(MAX_CURSOR_SIZE) / f64::from(pointer.width))
+                        } else if pointer.height > MAX_CURSOR_SIZE {
+                            Some(f64::from(MAX_CURSOR_SIZE) / f64::from(pointer.height))
+                        } else {
+                            None
+                        };
+
+                        let (png_width, png_height, hot_spot_x, hot_spot_y, rgba_buffer) = if let Some(scale) = scale {
+                            // Per invariants: Following conversions are losslesss.
```
CBenoit, line 525: Question: are they really lossless? For instance, when the cursor is not a square, either `width * scale` or `height * scale` may result in something that is not an integer. Did you mean that the cast will never "saturate"?

pacmancoder: Ah, yes, the working is a bit off. Sure, that is about saturation

crates/ironrdp-web/src/session.rs
CBenoit: You are right, indirectly we have `< usize::MAX`. However, ` < usize::MAX` is also the important part (the reason why no overflow can happen), so it’s useful to mention it, so the justification is "complete".

Cargo.toml
CBenoit: This is true, but we use don't upgrade manually anymore (dependabot does that for us)

crates/ironrdp-web/src/session.rs
CBenoit: Sounds good! In that case, the naming for `cast_f64_to_u16_lossy` should be changed too

crates/ironrdp-web/Cargo.toml
CBenoit: Fair enough. I agree `base64` API is superior. Also, this is just a suggestion.

That being said, let me elaborate.
It’s possible to allocate a buffer of the right size in one go. For instance:
```rust
use base64ct::{Base64, Encoding as _};

const PREFIX: &str = "data:image/png;base64,";

let png_encoded_len = Base64::encoded_len(png_buffer);
let mut buffer = vec![0; PREFIX.len() + png_encoded_len];
buffer[..PREFIX.len].copy_from_slice(PREFIX);
Base64::encode(bytes, &mut buffer[PREFIX.len()..])?;

// buffer is holding the final result
```

I agree the code is less simple. However, the size of the resulting WASM module _is_ important too.
It would be a good idea to measure the impact of `base64` vs the impact of `base64ct` before making the move, making sure we indeed see a smaller artifact.
I think it’s okay to look into this separately and to open a follow-up PR.

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -32,70 +32,122 @@ pub enum PointerError {
...
+    /// Software rendering target will produce RGBA bitmaps with premultiplied alpha
+    /// and inverted pixels represented as `[0xFF, 0xFF, 0xFF, 0x00]` color.
+    Software,
+    /// Accelerated rendering target will produce RGBA bitmaps with non-premultiplied alpha.
+    /// Inverted pixels will be rendered checkboard pattern.
+    Accelerated,
+}
+
+impl PointerBitmapTarget {
+    fn premultiply_alpha(self) -> bool {
```
CBenoit, line 62: Nitpick: the name `should_premultiply_alpha` would make it obvious that no operation is performed by this method.
Thought: I’m not sure how useful these very specific function are over something more generic like `is_software` and `is_accelerated`.

crates/ironrdp-session/src/fast_path.rs
```diff
@@ -36,6 +37,7 @@ pub struct Processor {
     use_system_pointer: bool,
     mouse_pos_update: Option<(u16, u16)>,
     no_server_pointer: bool,
+    pointer_software_rendering: bool,
```
CBenoit, line 40: Note: I would be interested in some doc comment explaining what is `use_system_pointer` as opposed to `pointer_software_rendering`.

CBenoit, line 40: Nitpick: "blending" sounds more accurate to me (over "rendering")

crates/ironrdp-session/src/image.rs
```diff
@@ -287,23 +287,23 @@ impl DecodedImage {
             _ => return,
         };
 
-        let left_virtual = x as i16 - pointer.hot_spot_x as i16;
-        let top_virtual = y as i16 - pointer.hot_spot_y as i16;
+        let left_virtual = x as i16 - pointer.hotspot_x as i16;
+        let top_virtual = y as i16 - pointer.hotspot_y as i16;
```
CBenoit, line 291: Note: all the `as i16` here are conversions with possible "wrapping" (e.g.: `hotspot_y` > `i16::MAX`). I don’t remember what is around this code, but I suppose some operation _must_ be performed using unsigned integers? Would it make sense to use `i32::from` instead?

crates/ironrdp-session/src/fast_path.rs
```diff
@@ -25,6 +25,7 @@ pub enum UpdateKind {
     PointerDefault,
     PointerHidden,
     PointerPosition { x: u16, y: u16 },
+    NativePointerUpdate(Rc<DecodedPointer>),
```
CBenoit, line 28: Suggestion: I feel "Native" prefix is a bit misleading. My understanding is that this is really a "PointerIcon" (we decoded the icon). Even if this is indeed a "native" cursor, it’s surprising we have "PointerHidden" (is the "Native" part really relevant?). Note also that the "Update" suffix is redundant (the enum is called "UpdateKind").

crates/ironrdp-web/src/session.rs
```diff
@@ -159,27 +158,27 @@ impl SessionBuilder {
...
-        self.0.borrow_mut().hide_pointer_callback = Some(callback);
-        self.clone()
-    }
-
-    /// Required
-    pub fn hide_pointer_callback_context(&self, context: JsValue) -> SessionBuilder {
-        self.0.borrow_mut().hide_pointer_callback_context = Some(context);
+    /// Required.
+    ///
+    /// #### Callback signature:
```
CBenoit, line 163: Nitpick: I think a single `#` is enough for sections in doc comments

crates/ironrdp-web/src/session.rs
```diff
@@ -159,27 +158,27 @@ impl SessionBuilder {
...
+    /// ```typescript
+    /// function callback(cursor_kind: string, hotspot_x: number, hotspot_y: number): void
```
CBenoit, line 165: Suggestion: for `cursor_kind`, use an enum instead of a string.
Rationale: self documented and type checked (less error prone).

For that, add a new variant "url", and a new parameter "cursor_data" which is something like: `string | undefined`.

crates/ironrdp-web/src/session.rs
```diff
@@ -159,27 +158,27 @@ impl SessionBuilder {
...
+    /// ```
+    ///
+    /// #### Cursor kinds:
+    /// - `default` (default system cursor); second and third arguments are `UNDEFINED`
+    /// - `none` (hide cursor); second and third arguments are `UNDEFINED`
```
CBenoit, line 170: Question: should this variant be named "hide" instead?
```suggestion
    /// - `hide` (hide cursor); second and third arguments are `UNDEFINED`
```

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,122 @@ impl Session {
...
                     }
                     ActiveStageOutput::PointerPosition { .. } => {
-                        // Not applicable for web
+                        // Not applicable for web.
+                    }
+                    ActiveStageOutput::NativePointerUpdate(pointer) => {
+                        // Maximum allowed cursor size for browsers is 32x32, because bigger sizes
+                        // will cause the following issues:
+                        // - 128x128 and bigger cursors are not supported in browsers.
+                        // - 32x32 and bigger cursors will default to the system cursor if their
```
CBenoit, line 507: Nitpick: "cursors bigger than 128x128" and "cursors bigger than 32x32" would be a more correct wording, because 128x128 and 32x32 are actually _excluded_ (32x32 cursors are supported).

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,122 @@ impl Session {
...
+                        //   sprite does not fit in the browser's viewport, introducing anabrupt
```
CBenoit, line 508: Typo:
```suggestion
                        //   sprite does not fit in the browser's viewport, introducing an abrupt
```

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,122 @@ impl Session {
...
+                        //   cursor style changes when the cursor is moved to the edge of the
```
CBenoit, line 509: Typo:
```suggestion
                        //   cursor style change when the cursor is moved to the edge of the
```

crates/ironrdp-web/src/session.rs
```diff
@@ -492,18 +477,122 @@ impl Session {
...
+                            let mut encoder =
+                                png::Encoder::new(&mut png_buffer, u32::from(png_width), u32::from(png_height));
+
+                            encoder.set_color(png::ColorType::Rgba);
+                            encoder.set_depth(png::BitDepth::Eight);
+                            let mut writer = encoder.write_header().context("PNG encoder header write failed")?;
+                            writer
+                                .write_image_data(rgba_buffer.as_ref())
+                                .context("failed to encode pointer PNG")?;
+                        }
```
CBenoit, line 581: Question: could we use `image/bmp` instead, and avoid the cost of converting to a PNG?

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -282,12 +280,28 @@ export class WasmBridgeService {
...
 
-    private showPointerCallback() {
-        this.canvas!.style.cursor = 'default';
+            // IMPORTANT: We need to make proxy `Image` object to actually load the image and make
+            // it usable for CSS property. Without this proxy object, URL will be rejected.
+            const image = new Image();
+            image.src = style;
+
+            const style_x = Math.round(hotspot_x);
+            const style_y = Math.round(hotspot_y);
```
CBenoit, line 301: Suggestion: rename to `rounded_hotspot_x` and `rounded_hotspot_y` for clarity

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -282,12 +280,28 @@ export class WasmBridgeService {
...
+        if (style.startsWith('data')) {
+            if (hotspot_x == undefined || hotspot_y == undefined) {
+                console.error('Hot spot is not defined for a custom cursor.');
+                return;
+            }
 
-    private showPointerCallback() {
-        this.canvas!.style.cursor = 'default';
+            // IMPORTANT: We need to make proxy `Image` object to actually load the image and make
+            // it usable for CSS property. Without this proxy object, URL will be rejected.
```
CBenoit, line 296: Praise: thank you for documenting this :pray:

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -156,12 +208,17 @@ impl DecodedPointer {
...
+                    if target.inverted_pixels_as_checkboard() {
+                        // Checkboard pattern is used to represent inverted pixels.
+                        if (row_idx + col_idx) % 2 == 0 {
+                            bitmap_data.extend_from_slice(&[0xff, 0xff, 0xff, 0xff]);
+                        } else {
+                            bitmap_data.extend_from_slice(&[0x00, 0x00, 0x00, 0xff]);
+                        }
+                    } else {
+                        bitmap_data.extend_from_slice(&target.inverted_pixels_color());
+                    }
```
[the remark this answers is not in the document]
pacmancoder: We can't push this outside the loop, because this condition is only relevant for "invisible" pixels, which are enumerated amd found inside this loop. Maybe I missunderstood the suggestion?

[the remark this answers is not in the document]
CBenoit: To clarify: I’m not suggesting to remove all conditions from the loop, only `if target.inverted_pixels_as_checkboard` one. It must be possible to push it out of the loop, because `target` is an immutable binding, and the result of `inverted_pixels_as_checkboard` will never change, even on the next iteration. Here is one possible avenue for that:
```diff
@@ -180,6 +180,19 @@ impl DecodedPointer {

         let mut bitmap_data = Vec::new();

+        let compute_inverted_pixel: fn(u16, u16) -> [u8; 4] = if target.inverted_pixels_as_checkboard() {
+            |row_idx, col_idx| {
+                // Checkered pattern is used to represent inverted pixels.
+                if (row_idx + col_idx) % 2 == 0 {
+                    [0xff, 0xff, 0xff, 0xff]
+                } else {
+                    [0x00, 0x00, 0x00, 0xff]
+                }
+            }
+        } else {
+            |_, _| [0xFF, 0xFF, 0xFF, 0x00]
+        };
+
         for row_idx in 0..data.height {
             // For non-monochrome cursors we read strides from bottom to top
             let (mut xor_stride_cursor, mut and_stride_cursor) = if flip_vertical {
@@ -207,17 +220,7 @@ impl DecodedPointer {
                     bitmap_data.extend_from_slice(&[0, 0, 0, 0]);
                 } else if and_bit == 1 && color == [0xff, 0xff, 0xff, 0xff] {
                     // Inverted pixel.
-
-                    if target.inverted_pixels_as_checkboard() {
-                        // Checkboard pattern is used to represent inverted pixels.
-                        if (row_idx + col_idx) % 2 == 0 {
-                            bitmap_data.extend_from_slice(&[0xff, 0xff, 0xff, 0xff]);
-                        } else {
-                            bitmap_data.extend_from_slice(&[0x00, 0x00, 0x00, 0xff]);
-                        }
-                    } else {
-                        bitmap_data.extend_from_slice(&target.inverted_pixels_color());
-                    }
+                    bitmap_data.extend_from_slice(&compute_inverted_pixel(row_idx, col_idx));
                 } else if target.premultiply_alpha() {
                     // Calculate premultiplied alpha via integer arithmetic
                     let with_premultiplied_alpha = [
```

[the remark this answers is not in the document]
pacmancoder: Ah, I see the idea now, thanks for the clarification

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -156,12 +208,17 @@ impl DecodedPointer {
                 } else if and_bit == 1 && color == [0xff, 0xff, 0xff, 0xff] {
                     // Inverted pixel.
 
-                    // Colors with alpha channel set to 0x00 are always invisible no matter their color
-                    // component. We could take advantage of that, and use a special color to represent
-                    // inverted pixels. For example, we could use [0xFF, 0xFF, 0xFF, 0x00] for such
-                    // purpose.
```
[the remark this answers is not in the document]
pacmancoder: I think you are right, I'll bring it back as a doc comment for `PointerBitmapTarget`

crates/ironrdp-session/src/fast_path.rs
pacmancoder: I'll add doc comment in `ProcessorBuilder`

pacmancoder: I think naming is about right -- It not only controls how to draw the cursor to the framebuffer but also changes pointer messages generation a bit. I would rather keep more generic `rendering` wording

crates/ironrdp-session/src/image.rs
pacmancoder: Yeah, we need it here to check if the cursor sprite is partially outside the framebuffer, to cut its sides appropriately

crates/ironrdp-session/src/fast_path.rs
pacmancoder: Yeah, let's rename it to `PointerBitmap`.

`PointerHidden` is still needed if the remote wants to hide the pointer completely (e.g. it is the case with 3D-accelerated application which draws the cursor via its game/graphics engine

crates/ironrdp-web/src/session.rs
pacmancoder: I don't like it, its huge 😆 
image

crates/ironrdp-web/src/session.rs
pacmancoder: Great idea

crates/ironrdp-web/src/session.rs
pacmancoder: Yeah, we could use it, but:
- `bmp` crate API very sketchy, it preallocates the filled buffer first and then requires a call to set each pixel on specific coordinates separately. 
- Base64-encoded representation will be bigger in size

I would rather stick with PNG, I don't think we will win a lot of performance by switching to BMP. The only thing which I really agree should be added is to set PNG compression type explicitly to `png::Compression::Fast`

crates/ironrdp-web/src/session.rs
```diff
@@ -823,3 +913,9 @@ where
         Ok((upgraded, server_public_key))
     }
 }
+
+#[allow(clippy::cast_sign_loss)]
+#[allow(clippy::cast_possible_truncation)]
+fn cast_f64_to_u16_unchecked(value: f64) -> u16 {
```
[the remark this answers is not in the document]
pacmancoder: agreed

crates/ironrdp-session/src/image.rs
CBenoit: Sounds good! Fair enough.

crates/ironrdp-session/src/fast_path.rs
CBenoit: Oh, sorry. I didn’t mean to remove `PointerHidden` of course. I wanted to point out the difference between the two (all the others does not have the "Native" prefix). I think I forgot to write the rest of the sentence :sweat_smile:

pacmancoder:
@CBenoit Just updated the review, please re-review 😄

crates/ironrdp-web/src/session.rs
CBenoit: Ah, you mean for the header? That’s fair. You could probably just use the definition we have in `ironrdp-cliprdr-format` crate, but  I understand your point too. Sounds good to me.

crates/ironrdp-web/src/session.rs
```diff
@@ -353,6 +344,16 @@ pub(crate) enum RdpInputEvent {
...
 
+enum CursorStyle {
+    Default,
+    Hidden,
+    Url {
+        data: String,
+        hotspot_x: u16,
+        hotspot_y: u16,
+    },
+}
```
CBenoit, line 355: Suggestion: it’s also possible to export an enum for TypeScript / JavaScript (see IronRdpErrorKind). It would be useful to have a stronger type there too. It’s not possible to use algebraic data types like in Rust, but at least we get something similar to `match` for the kind.

CBenoit (commented):
Looking great!

kristahouse (approved):
TS looks good!

CBenoit (approved):
I’m approving too. Feel free to merge yourself once remaining comments are addressed

web-client/iron-svelte-client/src/lib/login/login.svelte
```diff
@@ -7,13 +7,12 @@
...
     import type { DesktopSize } from '../../models/desktop-size';
 
-    let username = 'Administrator';
-    let password = 'DevoLabs123!';
-    let gatewayAddress = 'ws://localhost:7171/jet/rdp';
-    let hostname = '10.10.0.3:3389';
-    let domain = '';
-    let authtoken =
-        'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImN0eSI6IkFTU09DSUFUSU9OIn0.eyJkc3RfaHN0IjoiMTkyLjE2OC41Ni4xMDE6MzM4OSIsImV4cCI6MTY5MzQyMzY1NSwiamV0X2FpZCI6IjMwNzZjZGIwLWYxNTctNDJlNy1iOWMzLThhMTdlNDFkYjYwNyIsImpldF9hcCI6InJkcCIsImpldF9jbSI6ImZ3ZCIsImp0aSI6IjAwYjY4OTY2LWJiYjAtNDU0NS05ZDZiLWRjNmFmMjAzNjY5MiIsIm5iZiI6MTY5MzQyMjc1NX0.SYQv4HtWQbdHMHgoCLYejCfO3TtsMAyjjILB6-Nir3mBznKiSad3POeLf02n05JFc5QhCeSGxspAaoNU7-znQFhHr0Tt0MnZJ1YMQt4UoR3PR2fTuUqv8M5TKdm4lKwCIjh73tTD001glTkXHaxuCQBTFCUSzfZhXDIqq5-CQueKtCrgJfYepJLmlvgH-ujGcxfXoGJGmeUy3Fmaijiy0uaC98j9GNCfnAd6JENmSAOkxfroMFhq601PSEizRbPzq2exDakfJ0EkaANz15udBX1a7NP-RyANHWQb8hp0rj6hyuyg1-vfUKYusw5qNUjAGXaWOjHC5bLgnqfE2V8Xnw';
+    let username = "vmbox";
```
pacmancoder, line 10: Oops, this should not be committed, I'll revert

PR #303, opened 2023-11-27
Fix broken Package 0.11.0, bundle Rxjs back into umd.js 

irvingoujAtDevolution:
Fixes the broken package 0.11.0, as rxjs is excluded as umd js, but our project all includes rxjs as module. Wrap rxjs back into the umd js defination, so it operates. This is temporary solution, as umd js causes larger bundles size and optimization slow down. in fact, we cannot use the exported enum value, as our TS is in module js format, but our actual js is not module, hence, we can only export types not any values.

but anyway, it works for now.


CBenoit (approved):
Thank you!

PR #305, opened 2023-11-27
chore(release): also update public package.json

CBenoit:


irvingoujAtDevolution:
Future improvement:
remove public/package.json, points iron-remote-gui/package.json direct on to /dist

PR #311, opened 2023-11-28
fix(rfx): Fix invalid rectangle intersection intersection logic (RFX visual artifacts fix)

### LINKED ISSUE #249: Visual artifacts when using RDP6 Bitmap Codec
I noticed there are visual artifacts when the RDP6 Bitmap Codec is used with some versions of Windows (Server 2022).
I suspect there may be an off-by-1 error somewhere in the decoder.

### Target Windows specifications:

```
Edition	Windows Server 2022 Standard
Version	21H2
Installed on	‎2023-‎08-‎10
OS build	20348.1906
```

### With RDP6 Bitmap Codec (faulty)

[faulty.webm](https://github.com/Devolutions/IronRDP/assets/3809077/b5b5f295-f84c-4739-8c02-4b34e2c2f737)

### With 16 bits RLE Bitmap Codec (healthy)

[healthy.webm](https://github.com/Devolutions/IronRDP/assets/3809077/7f373689-f9d4-41f4-ab02-0d90595435b6)

pacmancoder:
- Fixed bug and unit tests
- Validated that RFX and other codecs are still working as expected

Closes #249

### DIFF
--- a/crates/ironrdp-graphics/src/rectangle_processing.rs
+++ b/crates/ironrdp-graphics/src/rectangle_processing.rs
@@ -71,7 +71,7 @@ impl Region {
                 let rectangles = self
                     .rectangles
                     .iter()
-                    .take_while(|r| r.top < rectangle.bottom)
+                    .take_while(|r| r.top <= rectangle.bottom)
                     .filter_map(|r| r.intersect(rectangle))
                     .collect::<Vec<_>>();
                 let extents = InclusiveRectangle::union_all(rectangles.as_slice());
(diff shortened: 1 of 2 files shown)


crates/ironrdp-graphics/src/rectangle_processing.rs
```diff
@@ -1481,10 +1481,10 @@ mod tests {
             rectangles: Vec::new(),
         };
```
pacmancoder, line 1482: Fixed 2 failing tests, it looks like previous tests were written before renaming `Rectangle` to `InclusiveRectangle` and therefore both intersection and tests logic/data were incorrect

CBenoit (approved):
Thank you! :bug: :fire:

PR #312, opened 2023-11-28
`SetInformation` and `UnsupportedMinorFunctionRequest`

ibeckermayer:
`SetInformation` is straightforward.

`UnsupportedMinorFunctionRequest` because I realized that the server is liable to send us back an occasional `NotifyChangeDirectory` or `None` which we want to handle. This allows us to handle these cases gracefully (but without writing all the decode/encode logic), and kicks the handling up the stack to the `RdpdrBackend` implementer (Teleport in our case). Teleport simply mimics what [FreeRDP currently does with these](https://github.com/FreeRDP/FreeRDP/blob/511444a65e7aa2f537c5e531fa68157a50c1bd4d/channels/drive/client/drive_main.c#L660-L666).

### DIFF
--- a/crates/ironrdp-pdu/src/cursor.rs
+++ b/crates/ironrdp-pdu/src/cursor.rs
@@ -46,12 +46,6 @@ impl<'a> ReadCursor<'a> {
     }
 
     #[inline]
-    pub fn discard(&mut self) {
-        let remaining_len = self.len();
-        self.advance(remaining_len);
-    }
-
-    #[inline]
     #[track_caller]
     pub fn read_array<const N: usize>(&mut self) -> [u8; N] {
         let bytes = &self.inner[self.pos..self.pos + N];
--- a/crates/ironrdp-pdu/src/cursor.rs
+++ b/crates/ironrdp-pdu/src/cursor.rs
@@ -168,6 +168,28 @@ impl<'a> ReadCursor<'a> {
     }
 
     #[inline]
+    pub fn read_i64(&mut self) -> i64 {
+        i64::from_le_bytes(self.read_array::<8>())
+    }
+
+    #[inline]
+    pub fn read_i64_be(&mut self) -> i64 {
+        i64::from_be_bytes(self.read_array::<8>())
+    }
+
+    #[inline]
+    pub fn try_read_i64(&mut self, ctx: &'static str) -> PduResult<i64> {
+        ensure_size!(ctx: ctx, in: self, size: 8);
+        Ok(i64::from_le_bytes(self.read_array::<8>()))
+    }
+
+    #[inline]
+    pub fn try_read_i64_be(&mut self, ctx: &'static str) -> PduResult<i64> {
+        ensure_size!(ctx: ctx, in: self, size: 8);
+        Ok(i64::from_be_bytes(self.read_array::<8>()))
+    }
+
+    #[inline]
     #[track_caller]
     pub fn peek<const N: usize>(&mut self) -> [u8; N] {
         self.inner[self.pos..self.pos + N].try_into().expect("N-elements array")
--- a/crates/ironrdp-pdu/src/cursor.rs
+++ b/crates/ironrdp-pdu/src/cursor.rs
@@ -46,6 +46,12 @@ impl<'a> ReadCursor<'a> {
     }
 
     #[inline]
+    pub fn discard(&mut self) {
+        let remaining_len = self.len();
+        self.advance(remaining_len);
+    }
+
+    #[inline]
     #[track_caller]
     pub fn read_array<const N: usize>(&mut self) -> [u8; N] {
         let bytes = &self.inner[self.pos..self.pos + N];
--- a/crates/ironrdp-rdpdr/src/pdu/efs.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/efs.rs
@@ -1088,20 +1088,7 @@ impl DeviceIoRequest {
         let file_id = src.read_u32();
         let completion_id = src.read_u32();
         let major_function = MajorFunction::try_from(src.read_u32())?;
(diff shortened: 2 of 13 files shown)


crates/ironrdp-pdu/src/cursor.rs
```diff
@@ -45,6 +45,12 @@ impl<'a> ReadCursor<'a> {
         self.pos
     }
 
+    #[inline]
+    pub fn discard(&mut self) {
+        let remaining_len = self.len();
+        self.advance(remaining_len);
+    }
```
CBenoit, line 52: Issue: I feel mixed about this. My understanding is that this method is making two assumptions:

1) the cursor inner slice only ever contains a single PDU, and
2) it’s fine to skip the rest of the PDU no matter from where this method is called.

For 1), we typically never include more than one PDU per slice, so it should be generally fine, but following code will stop working as soon as `Pdu1` is relying on this method.
```
let some_slice = &[/* pdu1 payload */, /* pdu2 payload */];
let mut cursor = ReadCursor(some_slice);
let pdu1: Pdu1 = decode(&mut cursor)?;
let pdu2: Pdu2 = decode(&mut cursor)?;
```
It’s not something I would expect to see outside of tests, so it’s probably not too bad.

For similar reasons 2) is fine as long as there is no other item coming after. For instance decoding a list of item where one of the item could discard the rest of the payload is a problem.

I.e: the more layers, the more something could go wrong at unexpected places.

Overall, we should be very careful when using this method and I suggest we recommend against liberally using it. At the very least, a doc comment should explain all that.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1390,10 +1390,15 @@ pub enum ServerDriveIoRequest {
     ServerDriveQueryInformationRequest(ServerDriveQueryInformationRequest),
     DeviceCloseRequest(DeviceCloseRequest),
     ServerDriveQueryDirectoryRequest(ServerDriveQueryDirectoryRequest),
+    /// A special type signifying that the minor function is not supported.
+    /// This is not a real PDU type, rather it is used to pass control of
+    /// what to do with the unsupported minor function to the caller.
+    UnsupportedMinorFunctionRequest(DeviceIoRequest),
```
CBenoit, line 1396: Issue (non-blocking): it seems to suffer from the ["fallback variant problem"](https://github.com/Devolutions/IronRDP/blob/b554c6a6d1ecc7f7070b5fcebc3d78fd2bdbfbbc/crates/ironrdp-pdu/README.md#enumeration-like-types-should-allow-resilient-parsing), and is also not really part of `DR_DRIVE_CORE_DEVICE_IOREQUEST` so it’s not really ideal. Fixing this is probably not very straightforward however, I’ll not block the PR for that.

CBenoit:
> `SetInformation` is straightforward.

:+1: 

> `UnsupportedMinorFunctionRequest` because I realized that the server is liable to send us back an occasional `NotifyChangeDirectory` or `None` which we want to handle. This allows us to handle these cases gracefully (but without writing all the decode/encode logic), and kicks the handling up the stack to the `RdpdrBackend` implementer (Teleport in our case). Teleport simply mimics what [FreeRDP currently does with these](https://github.com/FreeRDP/FreeRDP/blob/511444a65e7aa2f537c5e531fa68157a50c1bd4d/channels/drive/client/drive_main.c#L660-L666).

I expressed my concern about the new variant in the comment above, but I would be okay with this if this is really simpler for your use case.
I just want to make sure of one thing. FreeRDP is calling `Discard` on the `IRP` handle. The purpose of this is just to free the memory.
So it _seems_ to me that it would in fact map to us skipping this line:
```rust
self.backend.handle_drive_io_request(req)?;
```
But I don’t know how it’s actually handled in your backend. Maybe this callback is really required for other kind of clean up.

ibeckermayer:
@CBenoit it wasn't too much effort to actually just handle the decoding per usual, see now with 5b0a29ea45f38236bf6d1ebfd5990a0dbb1cc814

CBenoit (approved):
Nice. This looks good to me. Approving and enabling auto-merge.
For some reason CI is failing with an error at the fuzzing step, but I don’t see how your code could be the reason for that. :thinking:

CBenoit:
On, in fact there is errors related to your changes.
image
Try to run `cargo xtask ci` and see if you get the errors on your side too.

ibeckermayer:
@CBenoit it's passing for me.

CBenoit:
> it's passing for me locally

I tried clear the cache and run again, unfortunately the problem is not gone.
image

Also, I have the same output when I run locally:
```shell
$ cargo clean
$ cargo xtask ci
```
image

Strange :shrug:

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1615,6 +1628,8 @@ bitflags! {
         const FILE_ATTRIBUTE_PINNED = 0x00080000;
         const FILE_ATTRIBUTE_UNPINNED = 0x00100000;
         const FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS = 0x00400000;
+
+        const _ = !0;
```
CBenoit, line 1632: It just occurred to me this is the culprit. The `Cargo.lock` in the `fuzz/` folder is not updated to `bitflags v2.4`. You need to specify this in the `Cargo.toml`:

```rust
bitflags = "2.4"
```

ibeckermayer: We were using the same `bitflags = "2"` throughout, so I added `bitflags = "2.4"` to the workspace root Cargo.toml and updated all of the `Cargo.lock`s

1e5d8e11656cd7c2ff0eb7baba694a0ba4372e5f

PR #315, opened 2023-12-01
ci: check for typos using typos-cli

pacmancoder:
I frequently fail to remember to run `typos-cli` manually, so I added it in the CI.

Installed `typos-cli` binary should be cached, so overall CI speed should not change much (on GitHub runners).


pacmancoder:
Aha, after merge one more false-positive should be fixed :)

pacmancoder:
> Before we merge: a new "typos" job should be added to the CI:

Silly me, I only tested it with `xtask ci` 🤦‍♂️

pacmancoder:
@CBenoit Done, ready for re-review

.github/workflows/ci.yml
```diff
@@ -35,10 +35,23 @@ jobs:
       - name: Check formatting
         run: cargo xtask check fmt -v
 
+  typos:
+    name: Check typos
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v3
+
```
CBenoit, line 44: Suggestion: let’s cache the binaries, as done in other jobs:
```yml
      - name: Binary cache
        uses: actions/cache@v3
        with:
          path: ./.cargo/local_root/bin
          key: ${{ runner.os }}-bin-${{ github.job }}-${{ hashFiles('xtask/src/bin_version.rs') }}
```

CBenoit (approved):
Approving so you can merge after the last few details are addressed :slightly_smiling_face: 

Note: I renamed the PR title to match the conventional commit style that we already mostly follow correctly. Make sure to keep it when squashing.

.github/workflows/ci.yml
pacmancoder: I am all over the place today 🤦‍♂️

xtask/src/main.rs
```diff
@@ -59,8 +60,16 @@ fn main() -> anyhow::Result<()> {
...
             }
         }
+        Action::CheckTypos => {
+            check::typos(&sh)?;
+        }
+        Action::CheckInstall => {
+            check::install(&sh)?;
+        }
         Action::Ci => {
+            check::install(&sh)?;
```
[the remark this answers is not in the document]
pacmancoder: I think we need to keep this consistent with the code below -- we also call `web::install(&sh)?;` here to run `web::check(&sh)?;` later, so lets keep install for `check` too

crates/ironrdp-graphics/src/zgfx/mod.rs
```diff
@@ -578,7 +578,7 @@ mod tests {
             0x20, // "The quick brown "
             0x0E, 0x00, 0x00, 0x00, // second segment is the next 14 bytes:
             0x04, // type 4, not PACKET_COMPRESSED
-            0x66, 0x6F, 0x78, 0x20, 0x6A, 0x75, 0x6D, 0x70, 0x73, 0x20, 0x6F, 0x76, 0x65, // "fox jumps ove"
+            0x66, 0x6F, 0x78, 0x20, 0x6A, 0x75, 0x6D, 0x70, 0x73, 0x20, 0x6F, 0x76, 0x65, // "fox jumps over"
```
CBenoit, line 581: Actually, it seems this is not a typo in the comment, the payload on the left appears to be indeed holding "fox jumps ove" without "r" :joy:

PR #317, opened 2023-12-04
refactor: function! macro for error context

ibeckermayer:
@zmb3 suggested we add this for Teleport to avoid copy-paste errors wherever we're using these macros, however it occurred to me that this might best fit as a more general solution throughout IronRDP.

I tested this by applying the following patch
<details>

```patch
diff --git a/crates/ironrdp-cliprdr/src/pdu/capabilities.rs b/crates/ironrdp-cliprdr/src/pdu/capabilities.rs
index 508b4e5..34381af 100644
--- a/crates/ironrdp-cliprdr/src/pdu/capabilities.rs
+++ b/crates/ironrdp-cliprdr/src/pdu/capabilities.rs
@@ -166,10 +166,10 @@ impl<'de> PduDecode<'de> for CapabilitySet {
         let _length = src.read_u16();
 
         match caps_type {
-            Self::CAPSTYPE_GENERAL => {
-                let general = GeneralCapabilitySet::decode(src)?;
-                Ok(Self::General(general))
-            }
+            // Self::CAPSTYPE_GENERAL => {
+            //     let general = GeneralCapabilitySet::decode(src)?;
+            //     Ok(Self::General(general))
+            // }
             _ => Err(invalid_message_err!(
                 "capabilitySetType",
                 "invalid clipboard capability set type"
```
</details>

and running a session (via teleport), which results in a log message like

```log
SessionError(Error { context: "invalid payload", kind: Pdu(Error { context: "<ironrdp_cliprdr::pdu::capabilities::CapabilitySet as ironrdp_pdu::PduDecode>::decode", kind: InvalidMessage { field: "capabilitySetType", reason: "invalid clipboard capability set type" }, source: None }), source: None })
```

the key piece being `context: "<ironrdp_cliprdr::pdu::capabilities::CapabilitySet as ironrdp_pdu::PduDecode>::decode"`

### DIFF
--- a/crates/ironrdp-cliprdr/src/pdu/capabilities.rs
+++ b/crates/ironrdp-cliprdr/src/pdu/capabilities.rs
@@ -233,8 +233,6 @@ impl ClipboardProtocolVersion {
     const VERSION_VALUE_V1: u32 = 0x0000_0001;
     const VERSION_VALUE_V2: u32 = 0x0000_0002;
 
-    const NAME: &str = "CLIPRDR_CAPS_VERSION";
-
     #[must_use]
     pub fn downgrade(self, other: Self) -> Self {
         if self != other {
--- a/crates/ironrdp-pdu/src/basic_output/bitmap/rdp6.rs
+++ b/crates/ironrdp-pdu/src/basic_output/bitmap/rdp6.rs
@@ -282,7 +282,7 @@ mod tests {
             &[],
             expect![[r#"
                 Error {
-                    context: "Rdp6BitmapStream",
+                    context: "<ironrdp_pdu::basic_output::bitmap::rdp6::BitmapStream as ironrdp_pdu::PduDecode>::decode",
                     kind: NotEnoughBytes {
                         received: 0,
                         expected: 1,
@@ -297,7 +297,7 @@ mod tests {
             &[0x20],
             expect![[r#"
                 Error {
-                    context: "Rdp6BitmapStream",
+                    context: "<ironrdp_pdu::basic_output::bitmap::rdp6::BitmapStream as ironrdp_pdu::PduDecode>::decode",
                     kind: InvalidMessage {
                         field: "padding",
                         reason: "missing padding byte from zero-sized non-RLE bitmap data",
--- a/crates/ironrdp-pdu/src/macros.rs
+++ b/crates/ironrdp-pdu/src/macros.rs
@@ -2,6 +2,20 @@
 //!
 //! Some are exported and available to external crates
 
+/// Automatically returns the full pathname to a
+/// function. Taken from https://stackoverflow.com/a/40234666.
+#[macro_export]
+macro_rules! function {
+    () => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        name.strip_suffix("::f").unwrap()
+    }};
+}
+
 /// Creates a `PduError` with `NotEnoughBytes` kind
 ///
 /// Shorthand for
@@ -10,7 +24,7 @@
 /// ```
 /// and
 /// ```rust
-/// <ironrdp_pdu::PduError as ironrdp_pdu::PduErrorExt>::not_enough_bytes(Self::NAME, received, expected)
+/// <ironrdp_pdu::PduError as ironrdp_pdu::PduErrorExt>::not_enough_bytes($crate::function!(), received, expected)
 /// ```
 #[macro_export]
 macro_rules! not_enough_bytes_err {
@@ -18,7 +32,7 @@ macro_rules! not_enough_bytes_err {
         <$crate::PduError as $crate::PduErrorExt>::not_enough_bytes($context, $received, $expected)
     }};
     ( $received:expr , $expected:expr $(,)? ) => {{
-        not_enough_bytes_err!(Self::NAME, $received, $expected)
+        not_enough_bytes_err!($crate::function!(), $received, $expected)
     }};
 }
 
@@ -30,7 +44,7 @@ macro_rules! not_enough_bytes_err {
 /// ```
 /// and
 /// ```rust
-/// <ironrdp_pdu::PduError as ironrdp_pdu::PduErrorExt>::invalid_message(Self::NAME, field, reason)
+/// <ironrdp_pdu::PduError as ironrdp_pdu::PduErrorExt>::invalid_message($crate::function!(), field, reason)
 /// ```
 #[macro_export]
 macro_rules! invalid_message_err {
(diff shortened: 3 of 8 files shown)


CBenoit:
I explored this approach in the past and decided to not use it at the time. More specifically, I used a similar trick to find the name of the type instead of the name of the function. A minor inconvenience is that the actual representation is more verbose, not guaranteed to be stable, and [some tests may break](https://github.com/Devolutions/IronRDP/blob/f5f5e9c087162b39dabb6af39e4cb1252f26beff/crates/ironrdp-testsuite-core/tests/pcb.rs#L130-L140) from time to time because of that. IMO, this is not a big deal because it’s easy to regenerate these (we are using `expect_test`), and we typically don’t match on the `context` field of the error for control flow (that would not be a great idea anyway).

An alternative I was thinking about is:
```rust
/// A type with a static name.
pub trait StaticName {
    /// Name associated to this type.
    const NAME: &'static str;
}

/// A type from which a name can be retrieved.
pub trait GetName {
    /// Returns the name associated to this value.
    fn get_name(&self) -> &'static str;
}

impl<T: StaticName> GetName for T {
    fn get_name(&self) -> &'static str {
        T::NAME
    }
}

assert_obj_safe!(GetName);

/// Gets the name of this value.
pub fn get_name<T: GetName>(value: &T) -> &'static str {
    value.get_name()
}

macro_rules! impl_static_name {
    ($ty:ident) => {{
        impl StaticName for $ty {
            const NAME: &'static str = stringify!($ty);
        }
    }};
}

struct Example;

impl_static_name!(Example);
```

And this could be part of [another macro](https://github.com/Devolutions/IronRDP/blob/f5f5e9c087162b39dabb6af39e4cb1252f26beff/crates/ironrdp-pdu/src/macros.rs#L218-L251) that we should typically apply on most decodable / encodable types, so we would get it for "free" as well.

I’m not yet settled on either one as of today. Thoughts?

ibeckermayer:
@CBenoit it may be worth clarifying precisely what we want to get from this `context` field (and these errors more generally). From the example that I gave, we appear to arguably have competing semantics currently:

```
SessionError(Error { context: "invalid payload", kind: Pdu(Error { context: "<ironrdp_cliprdr::pdu::capabilities::CapabilitySet as ironrdp_pdu::PduDecode>::decode", kind: InvalidMessage { field: "capabilitySetType", reason: "invalid clipboard capability set type" }, source: None }), source: None })
```

In the `SessionError`, the `context` is something like a reason/general error message, whereas in the `Pdu` error it's more like the name of the pdu where the error occurred (before these changes), and now it's roughly name-of-pdu + function. Granted one could argue that for different error types `context` can reasonably have different meanings, but I wonder if doing so is making things unnecessarily complicated.

Personally when I hit any error, I'm looking for 
1. A way to efficiently, accurately identify where in the code the error came from
2. Information about the cause of the error
3. A message to display to the user in the UI (in cases where that's relevant) which doesn't expose irrelevant or security-sensitive information.

From those axioms I would imagine a "base class" that contains that information, and any "subclasses" (like `PduError`/`PduErrorKind`) being ways to efficiently communicate `2`, and/or as a means for programmatically handling errors further up the call stack based on different semantics. (This is just a first pass at a "theory of error types", having read no other thinking on the topic in depth, so take it with that grain of salt. I'm also not suggesting we redo the entire error system at this point, just trying to clarify some of the thought process for myself).

With the use of this `function!()` macro I'm essentially attempting to define these `context` fields as a means to achieve `1`. I've been presuming that was the original intent, though feel free to correct me if I'm wrong.

A key advantage I see to `function!()` macro as compared to your alternative is that it gets the job of `1` done without needing to know much about the structure of the underlying error system. This can be useful as a caller of the library, because I can just start using these error types and macros without needing to implement any other macros for my classes. For example in Teleport we're using `PduResult` and returning a lot of `custom_err!` (e.g. [here](https://github.com/gravitational/teleport/blob/8b6a3595779dbb51c65d490c79556d1b13fdcade/lib/srv/desktop/rdp/rdpclient/src/rdpdr/filesystem.rs#L145-L151)). Arguably that's an abuse of `PduResult` semantics, but point is that it makes it easier to get `1` consistently wherever we are in the codebase.

That's a long rambling take, I could say more but I want to hear your take on the proper meaning of `context` before rambling any further.

CBenoit:
> it may be worth clarifying precisely what we want to get from this `context` field (and these errors more generally).

Sure, I think clarification is needed too!

> From the example that I gave, we appear to arguably have competing semantics currently:
>
> […]
>
> In the `SessionError`, the `context` is something like a reason/general error message, whereas in the `Pdu` error it's more like the name of the pdu where the error occurred (before these changes), and now it's roughly name-of-pdu + function. Granted one could argue that for different error types `context` can reasonably have different meanings, but I wonder if doing so is making things unnecessarily complicated.

I would argue that the "invalid payload" in `SessionError` is _not_ a good context string, because…

> Personally when I hit any error, I'm looking for
> 
>     1. A way to efficiently, accurately identify where in the code the error came from
> 
>     2. Information about the cause of the error
> 
>     3. A message to display to the user in the UI (in cases where that's relevant) which doesn't expose irrelevant or security-sensitive information.

This is in line with my vision.

> With the use of this `function!()` macro I'm essentially attempting to define these `context` fields as a means to achieve `1`. I've been presuming that was the original intent, though feel free to correct me if I'm wrong.

This is the original intent. For PDUs, the context field is supposed to hold the name of the PDU being decoded or encoded, so it’s easy to find which part of the code is raising the error.

> A key advantage I see to `function!()` macro as compared to your alternative is that it gets the job of `1` done without needing to know much about the structure of the underlying error system. This can be useful as a caller of the library, because I can just start using these error types and macros without needing to implement any other macros for my classes. For example in Teleport we're using `PduResult` and returning a lot of `custom_err!` (e.g. [here](https://github.com/gravitational/teleport/blob/8b6a3595779dbb51c65d490c79556d1b13fdcade/lib/srv/desktop/rdp/rdpclient/src/rdpdr/filesystem.rs#L145-L151)). Arguably that's an abuse of `PduResult` semantics, but point is that it makes it easier to get `1` consistently wherever we are in the codebase.

However, `PduErrorKind` was not intended to be used pervasively this way. The intention was to define an `*ErrorKind` enum for the various parts of the application, which can be wrapped into a `ironrdp_error::Error<*ErrorKind>`. For channels, it could have been `ChannelErrorKind`, with `ChannelError` defined as `ironrdp_error::Error<ChannelErrorKind>`.
The side effect of this error kind being used pervasively is that we are now using the error macros at places where they’re not intended to be used to begin with, and we end up having no `Self::NAME` available for the context.

I’ve been wanting to revisit all that, but didn’t had the chance yet. I have a [WIP branch](https://github.com/Devolutions/IronRDP/tree/ironrdp-common) with some experiments though.

All that being said, I think I would like `custom_err!` macro to work anywhere in the codebase too, and the `function!` macro is likely one of the easiest way to achieve that. I think you convinced me to move forward with this approach. I may revisit things, but I don’t think it would create any hindrance. :thinking:

CBenoit (approved):
Approving and enabling auto-merge

PR #322, opened 2023-12-05
build: bump rust toolchain to 1.74.0

CBenoit:


CBenoit:
FYI @pacmancoder @ibeckermayer
After this PR, for `cargo xtask ci` to continue to work, you may need to run bootstrap the dev environment again:
```shell
$ cargo xtask clean
$ cargo xtask bootstrap
```
This will ensure all the Rust toolchains are up to date (wasm, nightly, …). (As a bonus, this will clean your target/ folder which may amount to 10~20 Go or so if you don’t trim it using something like [cargo-sweep](https://crates.io/crates/cargo-sweep))

xtask/src/check.rs
```diff
@@ -7,7 +7,7 @@ const EXTRA_LINTS: &[&str] = &[
     "invalid_reference_casting",
     "pointer_structural_match",
     "clippy::undocumented_unsafe_blocks",
-    "clippy::multiple_unsafe_ops_per_block",
+    // "clippy::multiple_unsafe_ops_per_block", // FIXME: https://github.com/rust-lang/rust-clippy/issues/11312 (fixed for 1.75.0)
```
CBenoit, line 10: Not a big deal to disable this until 1.75.0 as we are not authoring many unsafe code anyway

thenextman (approved):

PR #328, opened 2023-12-07
fix(connector): better security protocol selection

### LINKED ISSUE #314: Cannot connect to XRDP
I have some Ubuntu Linux servers running XRDP (0.9.12 and 0.9.17)

The connection always fails with this error message:
```
Connection error: [TLS upgrade] custom error, caused by: received corrupt message of type InvalidContentType
```

It works fine with a Windows 11 ARM VM.

CBenoit:
Replace the bitflag-based connector config API with a boolean-based one:

- `enable_tls`: set the PROTOCOL_SSL flag
- `enable_credssp`: set the PROTOCOL_HYBRID and PROTOCOL_HYBRID_EX flags

NLA can be enforced on client-side with `enable_tls = false`  + `enable_credssp = true`.

The `--security_protocol` argument was removed from the native client CLI, and instead it’s possible to disable specific protocols with `--no-tls` and `--no-credssp`. By default, both protocols are enabled for maximum compatibility with most RDP servers. We may change the defaults in the future.

For additional background on this change, see [this comment](https://github.com/Devolutions/IronRDP/issues/314#issuecomment-1843925374).

cc @awakecoding @thenextman 

FYI @ibeckermayer 
In your case, you’ll have to set `enable_tls` to `true` and `enable_credssp` to `false` (if you want to use the autologon flag).

### DIFF
--- a/crates/ironrdp-async/src/connector.rs
+++ b/crates/ironrdp-async/src/connector.rs
@@ -120,12 +120,36 @@ where
 {
     assert!(connector.should_perform_credssp());
 
-    let mut credssp_sequence = CredsspSequence::new(connector, server_name, server_public_key, kerberos_config)?;
+    let (mut sequence, mut ts_request) =
+        CredsspSequence::init(connector, server_name, server_public_key, kerberos_config)?;
 
-    while !credssp_sequence.is_done() {
+    loop {
+        let client_state = {
+            let mut generator = sequence.process_ts_request(ts_request);
+
+            if let Some(network_client_ref) = network_client.as_deref_mut() {
+                trace!("resolving network");
+                resolve_generator(&mut generator, network_client_ref).await?
+            } else {
+                generator
+                    .resolve_to_result()
+                    .map_err(|e| custom_err!("resolve without network client", e))?
+            }
+        }; // drop generator
+
         buf.clear();
+        let written = sequence.handle_process_result(client_state, buf)?;
 
-        if let Some(next_pdu_hint) = credssp_sequence.next_pdu_hint() {
+        if let Some(response_len) = written.size() {
+            let response = &buf[..response_len];
+            trace!(response_len, "Send response");
+            framed
+                .write_all(response)
+                .await
+                .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
+        }
+
+        if let Some(next_pdu_hint) = sequence.next_pdu_hint() {
             debug!(
                 connector.state = connector.state.name(),
                 hint = ?next_pdu_hint,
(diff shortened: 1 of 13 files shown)


crates/ironrdp-client/src/config.rs
```diff
@@ -222,10 +200,25 @@ struct Args {
...
+    /// Disabling this in order to enforce usage of CredSSP (NLA) is recommended.
+    #[clap(long)]
+    no_winlogon: bool,
+
+    /// Disable TLS + Network Level Authentication (NLA) using CredSSP
+    ///
+    /// NLA is used to authenticates RDP clients and servers before sending credentials over the network.
+    /// It’s not recommended to disable this.
+    #[clap(long)]
+    no_credssp: bool,
```
CBenoit, line 221: The updated CLI

crates/ironrdp-connector/src/credssp.rs
```diff
@@ -180,31 +182,19 @@ impl CredsspSequence {
         }
     }
 
-    pub fn process(&mut self) -> CredsspProcessGenerator<'_> {
-        let request = self.next_request.take().expect("next request"); // FIXME: error handling
```
CBenoit, line 184: The extended CredSSP sequence was broken by the refactoring done in this PR: https://github.com/Devolutions/IronRDP/pull/260
And it wasn’t enabled by default anymore since https://github.com/Devolutions/IronRDP/pull/131
For this reason, we failed at catching this bug during the refactoring.
This PR is also fixing the CredSSP sequence because we enable it by default again (mstsc does this as well).

thenextman:
I don't care for the `WinLogon` vernacular; although I don't have any better suggestion. However I see that this is all well-documented which should remove any ambiguity. Good job!

CBenoit:
> I don't care for the `WinLogon` vernacular; although I don't have any better suggestion.

To be fair, I don’t think it’s ideal either, but it sounded less misleading than "TLS" and `enable_tls = false` which is suggesting that the TLS transport will be disabled. (Equally true for variations such as `enable_tls_sec`, `enable_tls_security`, etc.)
That being said, the "TLS" terminology appears to be used otherwise, so it may not be a great idea to diverge.
I’m not too sure. Anyway, it’s only an API thing. In the frontend, we’ll probably follow whatever RDM is using.

crates/ironrdp-async/src/connector.rs
```diff
@@ -165,6 +148,28 @@ where
                 .await
                 .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
         }
+
+        if let Some(next_pdu_hint) = sequence.next_pdu_hint() {
```
pacmancoder, line 152: Nit: Maybe we could reduce nesting a bit for a long code block here:
```rust
let next_pdu_hint = if let Some(next_pdu_hint) = sequence.next_pdu_hint() {
    next_pdu_hint
} else {
    break;
};

debug!(
    connector.state = connector.state.name(),
...
```

same suggestion for `crates/ironrdp-blocking/src/connector.rs`

crates/ironrdp-connector/src/lib.rs
```diff
@@ -78,13 +78,61 @@ impl Credentials {
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Config {
+    /// The initial desktop size to request
     pub desktop_size: DesktopSize,
-    pub security_protocol: nego::SecurityProtocol,
+    /// TLS + Graphical login (legacy)
```
pacmancoder, line 83: This documentation is top-notch! Thanks! 👍

pacmancoder (approved):
Brilliant job! I really appreciate the amount of added documentation👍 I only have a minor comment about formatting (I disabled auto-merge to not merge PR too early by accident)

crates/ironrdp-async/src/connector.rs
CBenoit: Good catch, I adjusted the code accordingly

CBenoit:
> Brilliant job! I really appreciate the amount of added documentation👍 I only have a minor comment about formatting (I disabled auto-merge to not merge PR too early by accident)

Thank you! I adjusted the code accordingly

PR #329, opened 2023-12-08
fix(connector): better errors for MCS channel confirm

CBenoit:
For context: https://github.com/Devolutions/IronRDP/issues/314#issuecomment-1844871275

### DIFF
--- a/crates/ironrdp-connector/src/channel_connection.rs
+++ b/crates/ironrdp-connector/src/channel_connection.rs
@@ -173,11 +173,29 @@ impl Sequence for ChannelConnectionSequence {
 
                 debug!(message = ?channel_join_confirm, "Received");
 
-                if channel_join_confirm.initiator_id != user_channel_id
-                    || channel_join_confirm.channel_id != channel_join_confirm.requested_channel_id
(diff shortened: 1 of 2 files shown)


crates/ironrdp-connector/src/channel_connection.rs
```diff
@@ -173,11 +173,29 @@ impl Sequence for ChannelConnectionSequence {
...
+                }
+
+                if channel_id != channel_join_confirm.channel_id {
+                    return Err(reason_err!(
+                        "ChannelJoinConfirm",
+                        "unexpected channel_id in MCS Channel Join Confirm: received {}, got {}",
+                        channel_id,
+                        channel_join_confirm.channel_id,
+                    ));
                 }
```
CBenoit, line 199: More lenient handling and error reporting of this sanity check:
- Warn when the initiator ID does not match the user channel ID
- Fail when the requested channel ID doesn’t match the channel ID set in the Channel Join Request sent previously
- Fail when the joined channel ID does not match the requested channel ID. The server should follow what it advertised in the GCC Server Network Data.

pacmancoder (approved):
👍

PR #331, opened 2023-12-08
feat(session): handle surface with codec ID 0 (raw bitmap)

elmarco:
This fixes the compability with the server example.


CBenoit (approved):
LGTM!

PR #332, opened 2023-12-08
fix(server): advertize FASTPATH_OUTPUT

elmarco:
This helps FreeRDP, despite not being required by mstcs: https://github.com/FreeRDP/FreeRDP/issues/9612


CBenoit (approved):
Looks reasonable. Thank you for investigating this!

PR #333, opened 2023-12-08
server: add static channel and basic clipboard 

elmarco:
Add basic building blocks for static channel handling and testing with a dummy clipboard implementation.

### DIFF
--- a/crates/ironrdp-client/src/main.rs
+++ b/crates/ironrdp-client/src/main.rs
@@ -4,7 +4,7 @@
 extern crate tracing;
 
 use anyhow::Context as _;
-use ironrdp_client::config::Config;
+use ironrdp_client::config::{ClipboardType, Config};
 use ironrdp_client::gui::GuiContext;
 use ironrdp_client::rdp::{RdpClient, RdpInputEvent};
 use tokio::runtime;
@@ -31,30 +31,37 @@ fn main() -> anyhow::Result<()> {
 
     let (input_event_sender, input_event_receiver) = RdpInputEvent::create_channel();
 
-    #[cfg(not(windows))]
-    let cliprdr_factory = None;
+    let (_cliprdr, cliprdr_factory) = match config.clipboard_type {
+        ClipboardType::Stub => {
+            use ironrdp_cliprdr_native::StubClipboard;
 
-    #[cfg(windows)]
-    let (_win_clipboard, cliprdr_factory) = {
-        use ironrdp_client::clipboard::ClientClipboardMessageProxy;
-        use ironrdp_cliprdr_native::WinClipboard;
-        use windows::Win32::Foundation::HWND;
-        use winit::platform::windows::WindowExtWindows;
+            let cliprdr = StubClipboard::new();
+            let factory = cliprdr.backend_factory();
+            (Some(cliprdr), Some(factory))
+        }
+        #[cfg(windows)]
+        ClipboardType::Windows => {
+            use ironrdp_client::clipboard::ClientClipboardMessageProxy;
+            use ironrdp_cliprdr_native::WinClipboard;
+            use windows::Win32::Foundation::HWND;
+            use winit::platform::windows::WindowExtWindows;
 
-        // SAFETY: provided window handle from `winit` is valid and is guaranteed to be alive
-        // while the gui window is still open.
-        let win_clipboard = unsafe {
-            WinClipboard::new(
-                HWND(gui.window().hwnd()),
-                ClientClipboardMessageProxy::new(input_event_sender.clone()),
-            )?
-        };
+            // SAFETY: provided window handle from `winit` is valid and is guaranteed to be alive
+            // while the gui window is still open.
+            let win_clipboard = unsafe {
+                WinClipboard::new(
+                    HWND(gui.window().hwnd()),
+                    ClientClipboardMessageProxy::new(input_event_sender.clone()),
+                )?
+            };
 
-        let factory = Some(win_clipboard.backend_factory());
+            let factory = Some(win_clipboard.backend_factory());
 
-        // NOTE: we need to keep `win_clipboard` alive, otherwise it will be dropped before IronRDP
-        // starts and clipboard functionality will not be available.
-        (win_clipboard, factory)
+            // NOTE: we need to keep `win_clipboard` alive, otherwise it will be dropped before IronRDP
+            // starts and clipboard functionality will not be available.
+            (Some(win_clipboard), Some(factory))
+        }
+        _ => (None, None),
     };
 
     let client = RdpClient {
--- a/crates/ironrdp-client/src/main.rs
+++ b/crates/ironrdp-client/src/main.rs
@@ -31,13 +31,18 @@ fn main() -> anyhow::Result<()> {
 
     let (input_event_sender, input_event_receiver) = RdpInputEvent::create_channel();
 
-    let (_cliprdr, cliprdr_factory) = match config.clipboard_type {
+    // NOTE: we need to keep `win_clipboard` alive, otherwise it will be dropped before IronRDP
+    // starts and clipboard functionality will not be available.
+    #[cfg(windows)]
+    let mut win_clipboard = None;
+
+    let cliprdr_factory = match config.clipboard_type {
         ClipboardType::Stub => {
             use ironrdp_cliprdr_native::StubClipboard;
 
             let cliprdr = StubClipboard::new();
             let factory = cliprdr.backend_factory();
-            (Some(cliprdr), Some(factory))
+            Some(factory)
         }
         #[cfg(windows)]
         ClipboardType::Windows => {
@@ -48,20 +53,18 @@ fn main() -> anyhow::Result<()> {
 
             // SAFETY: provided window handle from `winit` is valid and is guaranteed to be alive
             // while the gui window is still open.
-            let win_clipboard = unsafe {
+            let cliprdr = unsafe {
                 WinClipboard::new(
                     HWND(gui.window().hwnd()),
                     ClientClipboardMessageProxy::new(input_event_sender.clone()),
                 )?
             };
 
-            let factory = Some(win_clipboard.backend_factory());
-
-            // NOTE: we need to keep `win_clipboard` alive, otherwise it will be dropped before IronRDP
-            // starts and clipboard functionality will not be available.
-            (Some(win_clipboard), Some(factory))
+            let factory = cliprdr.backend_factory();
+            win_clipboard = Some(cliprdr);
+            Some(factory)
         }
-        _ => (None, None),
+        _ => None,
     };
 
     let client = RdpClient {
@@ -75,6 +78,9 @@ fn main() -> anyhow::Result<()> {
     std::thread::spawn(move || {
         rt.block_on(client.run());
     });
+
+    #[cfg(windows)]
+    drop(win_clipboard);
 
     debug!("Run GUI");
     gui.run(input_event_sender);
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -8,6 +8,7 @@ use ironrdp_pdu::input::InputEventPdu;
 use ironrdp_pdu::mcs::SendDataRequest;
 use ironrdp_pdu::rdp::capability_sets::{CapabilitySet, CmdFlags, GeneralExtraFlags};
 use ironrdp_pdu::{self, mcs, nego, rdp, Action, PduParsing};
+use ironrdp_svc::{server_encode_svc_messages, StaticChannelSet};
 use ironrdp_tokio::{Framed, FramedRead, FramedWrite, TokioFramed};
 use tokio::net::{TcpListener, TcpStream};
 use tokio_rustls::TlsAcceptor;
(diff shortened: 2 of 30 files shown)


crates/ironrdp-svc/src/lib.rs
```diff
@@ -11,17 +11,24 @@ use alloc::boxed::Box;
...
 use pdu::cursor::WriteCursor;
 use pdu::gcc::ChannelDef;
 use pdu::rdp::vc::ChannelControlFlags;
 use pdu::{custom_err, encode_buf, PduEncode, PduParsing};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ChannelSide {
+    Server,
+    Client,
+}
```
CBenoit, line 30: note: I’m okay with keeping this as a small utility, but it seems to me that this enum does not really need to be part of public API of `ironrdp-svc` and does not add much value.

rationale: `ironrdp-svc`  does not provide any interesting logic relying on this type. Consumers of `ironrdp-svc` don’t need to use it for anything. I could imagine many channels using two different types, one per side, instead of unifying both sides on a single type. I could also imagine a boolean being used. (The built-in type actually provides many interesting utility methods which could make it a superior alternative in some cases.)

suggestion: At least in this case, my preference goes for this type of API which does not require the library user to know about the type `ChannelSide` at all:
```rust
/// CLIPRDR static virtual channel client endpoint implementation
#[derive(Debug)]
pub struct Cliprdr {
    /* … */
    is_server: bool, // or: side: ChannelSide,
}

impl Cliprdr {
  pub fn client_new(backend: Box<dyn CliprdrBackend>) -> Self {
    Self {
      /* … */,
      is_server: false, // or: side: ChannelSide::Client,
    }
  }

  pub fn server_new(backend: Box<dyn CliprdrBackend>) -> Self {
    Self {
      /* … */,
      is_server: true, // or: side: ChannelSide::Server,
    }
  }
}
```

rationale: It’s typically not useful to flip the "server" flag in an isolated manner. Acting as a server (or as a client) requires many differences in the code itself, and I can’t imagine a good case for something like this:
```rust
let side = if act_as_a_server {
    ChannelSide::Server
} else {
    ChannelSide::Client
};

Cliprdr::new(Box::new(clipboard_backend), side);
```

CBenoit (commented):
Thank you for moving forward this! It’s not a simple task.

cc @pacmancoder 
I would like to hear your thoughts.

crates/ironrdp-server/src/server.rs
```diff
@@ -166,16 +185,31 @@ impl RdpServer {
...
+            self.handle_input_backlog(
+                &mut framed,
+                result.io_channel_id,
+                result.user_channel_id,
+                result.input_events,
+            )
+            .await?;
+        }
+
+        self.static_channels = std::mem::take(&mut result.static_channels);
```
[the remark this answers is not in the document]
elmarco: ok

crates/ironrdp-client/src/main.rs
```diff
@@ -32,7 +32,13 @@ fn main() -> anyhow::Result<()> {
...
 
     #[cfg(not(windows))]
-    let cliprdr_factory = None;
+    let (_cliprdr, cliprdr_factory) = {
+        use ironrdp_cliprdr_native::StubClipboard;
+
+        let cliprdr = StubClipboard::new();
+        let factory = Some(cliprdr.backend_factory());
+        (cliprdr, factory)
+    };
```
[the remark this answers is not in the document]
elmarco: ok, I added a `--clipboard-type` option, which will allow to choose a backend or disable the support.

crates/ironrdp-svc/src/lib.rs
```diff
@@ -162,10 +233,17 @@ pub trait StaticVirtualChannelProcessor: AsAny + fmt::Debug + Send {
         CompressionCondition::Never
     }
 
+    /// Start a channel, after the connection is established and the channel is joined.
+    ///
+    /// Returns a list of PDUs to be sent back.
+    fn start(&mut self) -> PduResult<Vec<SvcMessage>> {
+        Ok(vec![])
+    }
```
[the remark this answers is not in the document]
elmarco: > issue: As far as I understand, this method should only be called by the server. It would be good to document which function is intended to be called by the client and which is intended to be called by the server.

For all kind of channels, I am not sure client side doesn't send data first, or doesn't have to wait for the server initial message. Furthermore, with dynamic channels, you probably want to know when the channel is actually active. And with custom channels, there are even more unknown.

crates/ironrdp-svc/src/lib.rs
elmarco: I agree with your feeling, but ironrdp-svc is also used to implement channels themself, and channel implementation usually need to know which side they act as. I am fine with using different constructuors if you prefer, but new(Client) or new_client() is kind of the same to me at this point.

elmarco: And channel that are asymmetricals don't have to use the ChannelSide in their public API. For ex, a GFX channel implementation could expose GFXClient & GFXServer types.

pacmancoder: Also, I agree with @CBenoit that adding `enum` to the constructor is not optimal. Also, this could result in a situation, where API allows to create, for example, server-side svc, but accidentally use client-side `ChannelProcessor`. Therefore I propose the following solution:

Move client/server decision to `ChannelProcessor` by introducing `is_server()` or `peer_kind()` method to this trait. That way, the svc channel could add correct headers based on the value returned by `ChannelProcessor`, while the `ChannelProcessor` implementors themselves are always will be aware of its peer kind (client/server).

I also suggest splitting `Cliprdr` to `CliprdrClient` and `CliprdrServer`, without introduction of PeerKind constructor argument - we could always re-use the same communication sequences by moving it to a shared module, and svc logic will be aware of the peer kind by returning `is_server() -> true` in `CliprdrServer` and `is_server() -> false`

pacmancoder (commented):
Thanks for the PR! I added my thoughts about SVC/Cliprdr API, @elmarco what are your thoughts on the suggested approach?

crates/ironrdp-svc/src/lib.rs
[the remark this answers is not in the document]
CBenoit: > For all kind of channels, I am not sure client side doesn't send data first

Right, that’s true.

> Furthermore, with dynamic channels, you probably want to know when the channel is actually active.

This API is for static channels only. We’ll expose a different API for dynamic channels, because they are very different.

I would still suggest adding `SvcClientProcessor` and `SvcServerProcessor` traits, as marker traits then. This way we can encode at the type level which type may be used in which side and build other APIs on top of that (connector only takes `SvcClientProcessor`s and acceptor only takes `SvcServerProcessor`s).

crates/ironrdp-svc/src/lib.rs
CBenoit: > I also suggest splitting Cliprdr to CliprdrClient and CliprdrServer, without introduction of PeerKind constructor argument - we could always re-use the same communication sequences by moving it to a shared module

I also prefer this approach. Less room for misuses.

> Move client/server decision to `ChannelProcessor` by introducing `is_server()` or `peer_kind()` method to this trait. That way, the svc channel could add correct headers based on the value returned by `ChannelProcessor`, while the `ChannelProcessor` implementors themselves are always will be aware of its peer kind (client/server).

I’m not sure to follow this part, I may be missing something.
To warrant a new method in a trait, the new method must serve a purpose in generic code.
Is the "peer kind" information _generically_ useful somewhere?
Do you think of a specific place specifically?
My understanding is that the only component which needs to know, is the static channel itself, and it already knows whether it is "client" or "server" depending on whether it is a `CliprdrClient` or `CliprdrServer` (or, depending on the approach taken, whether `client_new` was called instead of `server_new`, but I still prefer different types).

pacmancoder: @CBenoit  the rationale about adding new trait method is that `StaticVirtualChannel` is also need to know how to wrap svc messages inside MDUs PDUs (see logic [here](https://github.com/Devolutions/IronRDP/pull/333/files/c7ec00480270b70ab7b7c86e33bdbb13e1e543c4#diff-419140c5aaf75e62c1ccce87c9c3f5848a8180914fa4b9608e8a4eaba502a5acR180-R196)). So we can't leave `StaticVirtualChannel` without knowledge about underlying `ChannelProcessor` kind 😢

CBenoit: > @CBenoit the rationale about adding new trait method is that `StaticVirtualChannel` is also need to know how to wrap svc messages inside MDUs PDUs (see logic [here](https://github.com/Devolutions/IronRDP/pull/333/files/c7ec00480270b70ab7b7c86e33bdbb13e1e543c4#diff-419140c5aaf75e62c1ccce87c9c3f5848a8180914fa4b9608e8a4eaba502a5acR180-R196)). So we can't leave `StaticVirtualChannel` without knowledge about underlying `ChannelProcessor` kind 😢

Oh, I see. Then, actually we don’t need to ask the processor for this information. We inherently know the side, because client code will call [client_encode_svc_messages](https://github.com/Devolutions/IronRDP/pull/333/files/c7ec00480270b70ab7b7c86e33bdbb13e1e543c4#diff-419140c5aaf75e62c1ccce87c9c3f5848a8180914fa4b9608e8a4eaba502a5acR207), and server code will call [server_encode_svc_messages](https://github.com/Devolutions/IronRDP/pull/333/files/c7ec00480270b70ab7b7c86e33bdbb13e1e543c4#diff-419140c5aaf75e62c1ccce87c9c3f5848a8180914fa4b9608e8a4eaba502a5acR217). This function is simply refactoring the logic, but could be two separate functions, without ifs. (Actually, that would be my preference, but it’s more of a nitpick than anything else.)

crates/ironrdp-svc/src/lib.rs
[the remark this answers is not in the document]
elmarco: done

crates/ironrdp-svc/src/lib.rs
elmarco: I have limited the role with a trait specific to cliprdr, assuming the nearly symmetrical implementation of the channel is going to be an exception:
```
pub type CliprdrClient = Cliprdr<Client>;
pub type CliprdrServer = Cliprdr<Server>;

impl SvcClientProcessor for CliprdrClient {}
impl SvcServerProcessor for CliprdrServer {}
```

If you have better ideas, this could later be changed too.

elmarco:
@pacmancoder @CBenoit happy new year! anything missing here? thanks

CBenoit:
> happy new year! anything missing here? thanks

Happy new year!
I’ll give another review by Monday and let you know!

crates/ironrdp-client/src/main.rs
```diff
@@ -69,6 +79,9 @@ fn main() -> anyhow::Result<()> {
         rt.block_on(client.run());
     });
 
+    #[cfg(windows)]
+    std::mem::forget(win_clipboard);
```
CBenoit, line 83: question: Do we really need to forget the `win_clipboard`? I think it already lives until `gui.run` terminates, because it is in the scope of the main function.

crates/ironrdp-server/src/server.rs
```diff
@@ -174,8 +193,23 @@ impl RdpServer {
...
+            )
+            .await?;
+        }
+
+        self.static_channels = result.static_channels;
+        for (_type_id, channel, channel_id) in self.static_channels.iter_mut() {
+            let Some(channel_id) = channel_id else {
+                continue;
+            };
+            let svcs_response = channel.start()?;
```
CBenoit, line 210: nitpick:
```suggestion
            let svc_responses = channel.start()?;
```

CBenoit (approved):
Looking good to me! I left two additional comments, but I don’t think it’s blocking so I’ll merge the PR now and this can be addressed in follow-up PRs. Thank you!

CBenoit:
Sorry, I tried to fix the conflict using a merge commit from GitHub, but I think you need to update the lock file again using cargo directly.

crates/ironrdp-client/src/main.rs
elmarco: right, it shouldn't be necessary now

crates/ironrdp-server/src/server.rs
elmarco: ok

crates/ironrdp-client/src/main.rs
elmarco: actually it's needed, as nothing else holds it otherwise and the variable isn't read.

CBenoit: Oh, it's just for the warning?
You can prefix the variable name with _ to tell the compiler that it's intentionally not used and the warning will go away

elmarco: "error: collection is never read"(https://github.com/Devolutions/IronRDP/actions/runs/7519920189/job/20469012292). /me getting bored :)

CBenoit: I see! Sounds like a clippy false-positive. I don’t think `forget`ting the variable is the proper way to silent it, but it will not create any problem either in this case.

CBenoit:
Thank you!

PR #348, opened 2024-01-10
Drop PduParsing

elmarco:
The process to move away from to the legacy traits is actually quite boring and long.

I should probably focus on the pain points I encoutered when the two are mixed instead of trying to move the whole code base away.

Anyway, let me know if this is a good approach, starting from leaves types and gradually up.


crates/ironrdp-pdu/src/rdp.rs
```diff
@@ -28,34 +29,70 @@ pub struct ClientInfoPdu {
...
+        let len = cursor.pos();
+        stream.write_all(&buf[..len]).unwrap();
         Ok(())
     }
 
     fn buffer_length(&self) -> usize {
-        self.security_header.buffer_length() + self.client_info.buffer_length()
+        self.size()
     }
 }
```
CBenoit, line 97: This could probably be a blanket implementation for all types implementing `PduEncode` and `PduDecode` :+1: 
We would need to remove the `unwrap`s however, but we can probably just set `type Error = PduError`?

CBenoit (commented):
Thank you very much for helping with this effort! I really appreciate that.

> Anyway, let me know if this is a good approach, starting from leaves types and gradually up.

I think this approach works well. There is a small performance cost for bridging both APIs together, but the benefit of allowing incremental migration is significant. (And the performance hit will go away once the higher types are ported as well.)

crates/ironrdp-pdu/src/rdp.rs
CBenoit: I think it’s probably best to take the big `MAX_SIZE` approach, because even if it’s not great, we can keep the "observable" behavior unchanged, which is very important at least for bisecting regressions as you mention. And if it’s only happening during the handshake sequence, it’s not too bad either.

Also, we could create a branch in which we accumulate all the incremental changes until we have one "big PR" changing everything, but properly reviewed via intermediate smaller PRs :thinking: 
Would you be open to that?

elmarco: Ok, that sounds good. Give me a few days to rework my changes and publish. Thanks!

CBenoit:
Small heads-up: we are currently in a release rush, and I prefer merging afterwards. I’ll let you know! Thank you very much for the work!

CBenoit:
What are the commits with `(break)` and `(fix)` in the message? Are we able to build the project on each commit? (If I rebase merge, git bisect will include all the commits as candidates)

elmarco:
> What are the commits with `(break)` and `(fix)` in the message? Are we able to build the project on each commit? (If I rebase merge, git bisect will include all the commits as candidates)

Yes, each commit compiles. However, the intermediary impl_pdu_parsing_max() is pretty bad.. if it is passed too much data, it will consume it all, and it will break following code that expects data. I haven't found a solution to this... std::io::Read is too limited.

> 
> If you can rebase on top of master, I’ll handle the rest! 👍

done

elmarco:
awesome, thanks a lot!

PR #350, opened 2024-01-15
Add DVC server, and ainput channel to server

elmarco:


CBenoit (commented):
Thank you for this work!

To make sure we are on the same page, turning the methods of the SvcProcessor trait into async is a very controversial change that will be hard to merge.
We already had quite lengthy conversations about that with @pacmancoder and @ibeckermayer , and we concluded it was better for these to _not_ be async.

There are many considerations to keep in mind.
One of the biggest problem I remember is that processing a given channel was causing everything else to hang.

I can see this code is also suffering from this:

```rust
active_stage.process(&mut image, action, &payload).await?
```

This `await` operator is invoked inside the main loop, nested into a `tokio::select!`.
The problem is that it may take an arbitrary amount of time to return, and this is entirely dependent on the channel implementation.
While we await, the loop is stopped and we stop processing new events from both the client and the server.

Fixing this requires taking other controversial decisions with even deeper consequences.

In short, I’m afraid we can’t take this route, even if it could be convenient.

The approach we decided to take is to spawn a platform-dependent "backend" on the side, and communicate using explicit message passing. The main example for that is the clipboard channel. The code for the processor itself is still portable which is nice because we can reuse it for all the platforms, including WASM in the browser, where the execution model is quite different.

elmarco:
> To make sure we are on the same page, turning the methods of the SvcProcessor trait into async is a very controversial change that will be hard to merge.

Hm... then we may have to change the server traits to not be async either, and use event system. This is less convenient and perhaps less idiomatic. Tracking associated request/reply may be troublesome too..

CBenoit:
> > To make sure we are on the same page, turning the methods of the SvcProcessor trait into async is a very controversial change that will be hard to merge.
> 
> Hm... then we may have to change the server traits to not be async either, and use event system. This is less convenient and perhaps less idiomatic. Tracking associated request/reply may be troublesome too..

If you can find a nice alternative using async, I would sure be happy to see it. However, I don’t think it’s evident that we’ll get a more idiomatic and ergonomic API just by adding async / await keywords, at least not without much deeper refactoring. Also, achieving both "idiomatic" and "unidiomatic" API is possible with both "blocking" and async Rust. All I want to convey is that it’s not a trivial change and it will take probably more efforts than expected to make it happen. (And we actually already spent a good amount of time into trying this approach in the past.)

elmarco:
@CBenoit ok, so the current version should be ok, as I dropped async stuff

crates/ironrdp-dvc/src/server.rs
```diff
@@ -0,0 +1,255 @@
...
+    let mut res = Vec::new();
+    for msg in messages {
+        let total_size = msg.size();
+
+        let msg = encode_vec(msg.as_ref())?;
+        let mut off = 0;
+
+        while off < total_size {
+            let rem = total_size.checked_sub(off).unwrap();
+            let size = core::cmp::min(rem, 1590);
```
pacmancoder, line 234: nit: move `1590` to named constant

crates/ironrdp-pdu/src/rdp/vc/dvc/ainput.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub time: u64,
+    pub flags: MouseEventFlags,
+    pub x: i32,
+    pub y: i32,
+}
+
+impl MousePdu {
+    const NAME: &'static str = "AInputMousePdu";
+
+    const FIXED_PART_SIZE: usize = core::mem::size_of::<u64>() * 2 + core::mem::size_of::<i32>() * 2;
```
pacmancoder, line 163: (style) Our new code style states the following about size fields:

>Use an inline comment for each field of the structure.

```rust
// GOOD
const FIXED_PART_SIZE: usize = 1 /* Version */ + 1 /* Endianness */ + 2 /* CommonHeaderLength */ + 4 /* Filler */;

// BAD
const FIXED_PART_SIZE: usize = size_of::<u8>() + size_of::<u8>() + size_of::<u16>() + size_of::<u32>();
```

Please update all `FIXED_PART_SIZE` according to the new style

crates/ironrdp-dvc/src/server.rs
```diff
@@ -0,0 +1,255 @@
...
+        vc::dvc::ClientPdu::from_buffer(&mut user_data, user_data_len).map_err(|e| custom_err!("DVC client PDU", e))?;
+
+    // … | DvcData ]
+    let dvc_data = user_data;
+
+    Ok(DynamicChannelCtx { dvc_pdu, dvc_data })
+}
+
+fn encode_dvc_message(pdu: vc::dvc::ServerPdu) -> PduResult<SvcMessage> {
+    // FIXME: use PduEncode instead
```
pacmancoder, line 218: Should this `FIXME` be removed or fixed?

crates/ironrdp-pdu/src/rdp/vc/dvc/ainput.rs
```diff
@@ -0,0 +1,259 @@
+use bitflags::bitflags;
```
pacmancoder, line 1: It would be great to add a link to FreeRDP docs for AInput PDUs structures

pacmancoder (commented):
Nice work! I have some comments from my side, but nothing overly critical -- the only thing that I think may be improved is Arc<Mutex<_>> use for `RdpServerInputHandler`, maybe we could come up with something better?

crates/ironrdp-dvc/src/server.rs
elmarco: ok

crates/ironrdp-pdu/src/rdp/vc/dvc/ainput.rs
elmarco: ok

crates/ironrdp-dvc/src/server.rs
elmarco: I am working on it, but it's another branch.. tedious conversion from legacy to/from buffer to encode/decode trait...

crates/ironrdp-pdu/src/rdp/vc/dvc/ainput.rs
elmarco: ok

crates/ironrdp-server/src/server.rs
```diff
@@ -40,6 +42,41 @@ impl RdpServerSecurity {
     }
 }
 
+pub struct AInputHandler {
```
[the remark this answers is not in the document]
elmarco: Actually, it shouldn't be public. Fixed. Fwiw, I think we may want to have implementation and traits for known channels, but that can come later - it depends what API and flexibilty is exposed by the server to some degree.

crates/ironrdp-server/src/server.rs
```diff
@@ -102,7 +138,7 @@ impl RdpServerSecurity {
 /// ```
 pub struct RdpServer {
     opts: RdpServerOptions,
-    handler: Box<dyn RdpServerInputHandler>,
+    handler: Arc<Mutex<Box<dyn RdpServerInputHandler>>>,
```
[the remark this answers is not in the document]
elmarco: I don't have a good idea.. if we use a channel at this level, then maybe we should completely change the API and not use RdpServerInputHandler (& other) trait objects. But I would rather leave that choice to the implementation of the server at this point and prefer trait objects, but they need to be shareable via arc<mutex> then.

We can leave a TODO note?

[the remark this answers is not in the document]
CBenoit: If it’s too much work, I’m fine with the TODO note for now, and we can revisit later.

crates/ironrdp-pdu/src/rdp/vc/dvc.rs
```diff
@@ -11,6 +11,7 @@ use crate::PduParsing;
 #[cfg(test)]
 mod tests;
 
+pub mod ainput;
```
CBenoit, line 14: thought: Ideally, I would like to move code for specific channels in their own crates. For instance, we do this for the RDPDR static channel, for which the code is found in the `ironrdp-rdpdr` crate. In this case, a `ironrdp-ainput` crate could be created. It’s fine to do a follow-up PR for that though.

crates/ironrdp-pdu/src/lib.rs
```diff
@@ -157,6 +157,25 @@ pub trait PduEncode {
...
+    /// Returns the associated PDU name associated.
+    fn name(&self) -> &'static str {
+        "legacy-pdu-encode"
+    }
+
+    /// Computes the size in bytes for this PDU.
+    fn size(&self) -> usize {
+        self.len()
+    }
+}
```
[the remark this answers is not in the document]
elmarco: No, it's temporary. switching all the code to PduEncode is tedious.. I have other work pending, but enough in this (and other) MR at this point.

This is used in DVC server implementation, to wrap ServerPdu in SvcMessage (taking PduEncode).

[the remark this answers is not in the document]
elmarco: moved to legacy

crates/ironrdp-pdu/src/rdp/vc/dvc.rs
elmarco: done

crates/ironrdp-dvc/src/server.rs
```diff
@@ -0,0 +1,257 @@
...
+            processor: Box::new(processor),
+            data: CompleteData::new(),
+        }
+    }
+}
+/// DRDYNVC Static Virtual Channel (the Remote Desktop Protocol: Dynamic Virtual Channel Extension)
+///
+/// It adds support for dynamic virtual channels (DVC).
+pub struct DrdynvcServer {
+    dynamic_channels: Slab<DynamicChannel>,
```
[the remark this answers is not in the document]
elmarco: A slab returns the key when storing the value.  That let it manage the dynamically allocated channel ids. (and ids can be reused if a channel is replaced etc).

A vec or a btreemap or other collection don't let you do that conveniently. (Linux kernel has idr & xarray). The client doesn't need it here, has it doesn't have to do ID allocation.  

Basically slab is a convenient Vec<Option<T>> under the hood. 

slab is used by thousands of projects: https://lib.rs/crates/slab/rev

crates/ironrdp-server/src/server.rs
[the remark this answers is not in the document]
elmarco: added TODO

pacmancoder:
LGTM from my side! 🚀

pacmancoder (approved):
LGTM from my side! 🚀

CBenoit (approved):
LGTM! Great work, thank you again!

PR #357, opened 2024-01-25
Random fixes and improvements

elmarco:


crates/ironrdp-session/src/image.rs
```diff
@@ -599,7 +600,12 @@ impl DecodedImage {
                         let dst_idx = ((top + row_idx) * image_width + left + col_idx) * DST_COLOR_DEPTH;
 
                         // Copy RGB channels as is
-                        self.data[dst_idx..dst_idx + SRC_COLOR_DEPTH].copy_from_slice(src_pixel);
+                        if format == self.pixel_format {
+                            self.data[dst_idx..dst_idx + SRC_COLOR_DEPTH].copy_from_slice(src_pixel);
+                        } else {
+                            let c = format.read_color(src_pixel).unwrap();
+                            self.data[dst_idx..dst_idx + SRC_COLOR_DEPTH].copy_from_slice(&[c.r, c.g, c.b, c.a]);
+                        }
```
CBenoit, line 608: perf: style and performance considerations below.

1. `apply_rgb32_bitmap` is in the hot path when enabling the RDP6 32 bits codec
2. Special casing the case `format == self.pixel_format` like this is unlikely to help performance since we now have a branching path inside the loop. If we need the second branch as a "fallback" which always works when the format differs, it’s likely better to just keep this one…
3. Or to "[push the if up](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html)" and evaluate the condition only once instead, because `format == self.pixel_format` will always evaluate to the same value for each iteration and we don’t need to do the work each time.
4. Of course, I have no idea which one of 2. or 3. is better without looking a some numbers from a benchmark, so I’m fine with either way.

CBenoit (approved):
LGTM! Thank you!

crates/ironrdp-server/src/display.rs
```diff
@@ -35,20 +36,56 @@ pub struct BitmapUpdate {
...
 /// which will then be encoded and sent to the client
 ///
+/// See [`RdpServerDisplay`] example.
+#[async_trait::async_trait]
+pub trait RdpServerDisplayUpdates {
+    /// # Cancel safety
+    ///
+    /// This method MUST be cancellation safe because it is used in a `tokio::select!` statement.
+    /// If some other branch completes first, it MUST be guaranteed that no data is lost.
+    async fn next(&mut self) -> Option<DisplayUpdate>;
```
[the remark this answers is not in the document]
elmarco: done

crates/ironrdp-session/src/image.rs
elmarco: ok

PR #372, opened 2024-02-13
fix(web): set correct `src` for cursor proxy `Image` object

pacmancoder:
Fix for the following error message spam in `ironrdp-web`

image


CBenoit (approved):
Good catch!

PR #378, opened 2024-02-16
Experimental: add iron-vnc as optional wasm module for iron-remote-gui

irvingoujAtDevolution:
This is experimental, not intended to be merged.

Allow IronVNC as it's underlying WASM module to be loaded into iron-remote-gui

Copy and pasted the wasm-bindgen compiled output into iron-remote-gui /src, and chooses which wasm module to load at runtime when the custom element is rendered. We need to think of a better repository strucutre if we want to reuse iron-remote-gui. Maybe seperate web-client and uses git-submodule?


irvingoujAtDevolution:
image

a working screen shot

irvingoujAtDevolution:
another path to take: https://github.com/Devolutions/IronVNC/pull/319

PR #383, opened 2024-02-21
Fix DVC message length

probakowski:
Previously, when encoding DVC messages we didn't set length correctly - `dvc::ClientPdu` and `DvcData` was not included. That led to broken stream and errors.
This PR fixes that by creating additional wrapper (`DvcMessage`) that is encoded with all constituents included.

### DIFF
--- a/crates/ironrdp-session/src/legacy.rs
+++ b/crates/ironrdp-session/src/legacy.rs
@@ -1,9 +1,9 @@
-use std::io::{Read, Write};
 use ironrdp_connector::legacy::{encode_send_data_request, SendDataIndicationCtx};
 use ironrdp_pdu::rdp::vc;
+use ironrdp_pdu::rdp::vc::ChannelError;
 use ironrdp_pdu::write_buf::WriteBuf;
 use ironrdp_pdu::PduParsing;
-use ironrdp_pdu::rdp::vc::ChannelError;
+use std::io::{Read, Write};
 
 use crate::{SessionError, SessionResult};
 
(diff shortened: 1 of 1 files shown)


crates/ironrdp-session/src/legacy.rs
```diff
@@ -19,19 +21,39 @@ pub fn encode_dvc_message(
...
+    dvc_pdu: vc::dvc::ClientPdu,
+    dvc_data: &'a[u8],
+}
 
-    Ok(())
+impl PduParsing for DvcMessage<'_> {
+    type Error = ChannelError;
+
+    fn from_buffer(_: impl Read) -> Result<Self, Self::Error> where Self: Sized {
+        todo!()
```
CBenoit, line 44: suggestion: Just to be on the safe side.
```suggestion
        Err(std::io::Error::other("legacy::DvcMessage::from_buffer called – this is a bug").into())
```

crates/ironrdp-session/src/legacy.rs
CBenoit, line 1: You’ll need to run `cargo fmt`

CBenoit (approved):
Thank you! The fix itself looks good to me! I’ll enable auto-merge so it gets merged once the CI is happy.

probakowski:
Thanks @CBenoit, I pushed `cargo fmt` results and your suggestion, you'll have to set auto-merge again

crates/ironrdp-session/src/legacy.rs
```diff
@@ -19,19 +21,46 @@ pub fn encode_dvc_message(
...
+    where
+        Self: Sized,
+    {
+        Err(std::io::Error::other("legacy::DvcMessage::from_buffer called – this is a bug").into())
+    }
+
+    fn to_buffer(&self, mut stream: impl Write) -> Result<(), Self::Error> {
+        self.channel_header.to_buffer(&mut stream)?;
+        self.dvc_pdu.to_buffer(&mut stream)?;
+        stream.write(self.dvc_data)?;
```
CBenoit, line 57: issue: `write` does not guarantee processing of the entire buffer
```suggestion
        stream.write_all(self.dvc_data)?;
```

PR #388, opened 2024-02-29
fix(rdpdr): implement `Clone` and `PartialEq` on most of efs and esc

ibeckermayer:
Also includes a custom `PartialEq` implementation for `DeviceControlResponse`, ~and removes `std::io::Error::other` which appears to be a nightly-only feature for now~.

### DIFF
--- a/crates/ironrdp-session/src/legacy.rs
+++ b/crates/ironrdp-session/src/legacy.rs
@@ -48,7 +48,11 @@ impl PduParsing for DvcMessage<'_> {
     where
         Self: Sized,
     {
-        Err(std::io::Error::other("legacy::DvcMessage::from_buffer called – this is a bug").into())
+        Err(std::io::Error::new(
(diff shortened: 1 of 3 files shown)


crates/ironrdp-session/src/legacy.rs
```diff
@@ -48,7 +48,11 @@ impl PduParsing for DvcMessage<'_> {
     where
         Self: Sized,
     {
-        Err(std::io::Error::other("legacy::DvcMessage::from_buffer called – this is a bug").into())
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Other,
+            "legacy::DvcMessage::from_buffer called – this is a bug",
+        )
+        .into())
```
CBenoit, line 55: You need to update your rust toolchain, it’s stable since 1.74.0

CBenoit (approved):
LGTM!

PR #391, opened 2024-03-01
refactor(connector): compile-time verified ExtendedClientOptionalInfo builder

pacmancoder:


crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -231,15 +231,111 @@ impl ExtendedClientInfo {
     }
 }
 
+pub mod extended_client_optional_info {
```
CBenoit, line 234: suggestion: Move this module at the end of the file and call it `builder`.

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -231,15 +231,111 @@ impl ExtendedClientInfo {
...
+    pub struct BuilderStateSetPerformanceFlags;
+    pub struct BuilderStateSetReconnectCookie;
+    pub struct BuilderStateFinal;
+
+    // State machine-based builder for [`ExtendedClientOptionalInfo`].
+    //
+    // [`ExtendedClientOptionalInfo`] strictly requires to set all preceding optional fields before
+    // setting the next one, therefore we use a state machine to enforce this during the compile time.
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct Builder<State> {
```
CBenoit, line 249: suggestion: In case we add additional builders, in the future, it’s best to give a more specific name than just "Builder".
```suggestion
    pub struct ExtendedClientOptionalInfoBuilder<State> {
```

CBenoit:
note: For consistency, let’s start the conventional commit message with a lowercase (unless it’s a acronym, etc)

CBenoit (approved):
LGTM!

PR #402, opened 2024-03-07
ironrdp_svc rdpdr processing, server licenses and FastPathUpdate codes

nszetei:
This PR adds two additional `server_license` decoding functions, more commands with `FastPathUpdate::decode_with_code` and `ironrdp-svc` processing.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [ pdu_decoding, rle_decompression, bitmap_stream, cliprdr_format ]
+        target: [ pdu_decoding, rle_decompression, bitmap_stream, cliprdr_format, channel_processing.rs ]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


CBenoit (commented):
This looks good to me.

However, it seems that this PR could be split in two:

1) Extend the `pdu_decode` fuzz target
2) Add the new `channel_process` fuzz target

I also recommend you to run `cargo xtask ci` to run the same checks we enforced in our CI.

CBenoit:
> I also recommend you to run `cargo xtask ci` to run the same checks we enforced in our CI.

Also, for each improvement to the harness, you’ll likely need to add the associated fix because we also run the fuzzer briefly in our CI to catch the "most obvious bugs". Ideally I would prefer a PR per issue because it’s easier to review and easier to track regressions, but we can make a single PR with all the changes too if that’s more convenient for you.

nszetei:
> > I also recommend you to run `cargo xtask ci` to run the same checks we enforced in our CI.
> 
> Also, for each improvement to the harness/oracle, you’ll likely need to add the associated fix because we also run the fuzzer briefly in our CI to catch the "most obvious bugs". Ideally I would prefer a PR per issue because it’s easier to review and easier to track regressions, but we can make a single PR with all the changes too if that’s more convenient for you.

Thanks for your feedback. Single PR per issue could be much easier to handle. I'll close this pull request and send a new one.

PR #403, opened 2024-03-07
fixes BitmapInfoHeader::decode_with_size issue with abs()

nszetei:
Fixes the issue when the value `-2147483648` is read and passed to the `abs()` function, causing panic with `attempt to negate with overflow`. See [here](https://doc.rust-lang.org/std/primitive.i32.html#method.abs). 

Reproduction steps:
```
$ echo "CAAAAAAAAIABAAqJurq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urq6urpQTn5HTg==" | base64 -D > /tmp/testcase
$ cargo fuzz run cliprdr_format /tmp/testcase
```

### DIFF
--- a/crates/ironrdp-cliprdr-format/src/bitmap.rs
+++ b/crates/ironrdp-cliprdr-format/src/bitmap.rs
@@ -200,10 +200,12 @@ impl BitmapInfoHeader {
         let size = src.read_u32();
 
         let width = src.read_i32();
-        check_invariant(width.abs() <= 10_000).ok_or_else(|| invalid_message_err!("biWidth", "width is too big"))?;
+        check_invariant(width != i32::MIN && width.abs() <= 10_000)
+            .ok_or_else(|| invalid_message_err!("biWidth", "width is too big"))?;
 
         let height = src.read_i32();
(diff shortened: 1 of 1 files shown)


CBenoit:
Oops. It seems the same bug was fixed approximately at the same time by my colleague @pacmancoder 
https://github.com/Devolutions/IronRDP/pull/404

nszetei:
> Oops. It seems the same bug was fixed approximately at the same time by my colleague @pacmancoder in #404 See also #405 We can drop the test cases directly into the appropriate folder and we get free regression tests.

I see. I had to commit it since I wanted to pass local `cargo xtask ci` checks :) However, the [fix](https://github.com/Devolutions/IronRDP/pull/404/files) has an issue because, in both cases, `width` is checked. The second check should be `height`.

pacmancoder:
I'll fix typo right now, thanks

pacmancoder:
Thanks, I'll fix the typo

pacmancoder:
UPD: https://github.com/Devolutions/IronRDP/pull/406

PR #404, opened 2024-03-07
fix(cliprdr): panic when calling .abs() on i32::MIN

pacmancoder:


pacmancoder:
Today I learned: `.abs()` could actually panic in rust

CBenoit (approved):
LGTM!

PR #405, opened 2024-03-08
test: regression test case for cliprdr_format target

CBenoit:
This patch uniformise the regression tests for bugs found by fuzzing.
We can add all the regression test cases into the `crates/ironrdp-testsuite-core/test_data/fuzz_regression/` folder.


pacmancoder (approved):
LGTM

PR #407, opened 2024-03-08
fixing clipboard output width/height invariant to avoid crash in debug build

nszetei:
The current implementation allows passing value `10 000` to the `debug_assert!`, which causes a crash in the non-optimized builds. This fix makes the invariance testing consistent with the previous checks. I also added the regression test case.


CBenoit (approved):
Nice, thank you!!

Issue #400, opened 2024-03-07
Fuzz(channel_processing) - ironrdp-rdpdr/src/pdu/esc/rpce.rs:197:9: attempt to subtract with overflow


nszetei:
#####  IronRDP revision

[3ee44b5c9b858b003148536637476b7133c5858e](https://github.com/Devolutions/IronRDP/commit/3ee44b5c9b858b003148536637476b7133c5858e)

##### Build platform
macOS 13.6.4 / 22G513

##### Reproduction steps
- Use the following harness
```
pub fn channel_process(input: &[u8]) {
    use ironrdp_svc::SvcProcessor;

    let mut rdpdr = ironrdp_rdpdr::Rdpdr::new(Box::new(ironrdp_rdpdr::NoopRdpdrBackend), Backend".to_string()).with_smartcard(1).with_drives(None);

    let _ = rdpdr.process(input);
}
```

- Testcase:
```
$ echo ckRSSQEAAAABAAAAAAAAAA4AAADgABAAACgAACQAAAjAAAkA | base64 -D  > /tmp/poc.bin
```

##### Output

```
$ cargo fuzz run channel_processing /tmp/poc.bin

    Finished `release` profile [optimized + debuginfo] target(s) in 0.09s
     Running `target/aarch64-apple-darwin/release/channel_processing -artifact_prefix=ironrdp-testing/IronRDP/fuzz/artifacts/channel_processing/ /tmp/poc.bin`
channel_processing(31183,0x1e2632100) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 4132805750
INFO: Loaded 1 modules   (420120 inline 8-bit counters): 420120 [0x1060f82a0, 0x10615ebb8),
INFO: Loaded 1 PC tables (420120 PCs): 420120 [0x10615ebb8,0x1067c7d38),
target/aarch64-apple-darwin/release/channel_processing: Running 1 inputs 1 time(s) each.
Running: /tmp/poc.bin
thread '' panicked at ironrdp-testing/IronRDP/crates/ironrdp-rdpdr/src/pdu/esc/rpce.rs:197:9:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==31183== ERROR: libFuzzer: deadly signal
[...]

PR #408, opened 2024-03-11
fix(svc): rdpdr channel fuzzing harness and associated issues

nszetei:
Closes https://github.com/Devolutions/IronRDP/issues/400


crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1219,7 +1219,8 @@ where
     }
 
     pub fn decode(header: DeviceIoRequest, src: &mut ReadCursor<'_>) -> PduResult<Self> {
-        ensure_size!(ctx: "DeviceControlRequest", in: src, size: Self::headerless_size());
+        // Ensure the input size includes an additional 20 bytes for padding
+        ensure_size!(ctx: "DeviceControlRequest", in: src, size: Self::headerless_size() + 20);
```
CBenoit, line 1223: suggestion: I think the padding should be part of the "headerless size"

nszetei: Thank you for your suggestion! I changed it to `const`.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1214,12 +1214,14 @@ impl<T: IoCtlCode> DeviceControlRequest<T>
...
-        size_of::<u32>() * 3 // OutputBufferLength, InputBufferLength, IoControlCode
-    }
+    const HEADERLESS_SIZE: usize = 4 // OutputBufferLength
+        + 4 // InputBufferLength
+        + 4 // IoControlCode
+        + 20; // Additional 20 bytes for padding
 
     pub fn decode(header: DeviceIoRequest, src: &mut ReadCursor<'_>) -> PduResult<Self> {
-        ensure_size!(ctx: "DeviceControlRequest", in: src, size: Self::headerless_size());
+        // Ensure the input size includes an additional 20 bytes for padding
```
CBenoit, line 1223: This comment is not relevant anymore since it’s now included in the const item
```suggestion
```

crates/ironrdp-rdpdr/src/pdu/efs.rs
CBenoit: Thank you!

CBenoit (approved):
Thank you!! This is looking good to me!

I updated the title of the PR (and of the squash commit) so it’s shorter and follows the conventional commit specification.
I also updated your original message to includes the ["Closes" keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) so the issue gets closed automatically.

Issue #395, opened 2024-03-07
Fuzz(pdu_decoding) - ironrdp-pdu/src/rdp/server_license/server_license_request/cert.rs:164:23: attempt to subtract with overflow


nszetei:
#####  IronRDP revision
[3ee44b5c9b858b003148536637476b7133c5858e](https://github.com/Devolutions/IronRDP/commit/3ee44b5c9b858b003148536637476b7133c5858e)

##### Build platform
macOS 13.6.4 / 22G513

##### Reproduction steps
```
$ echo AAAABwED/wH/XVkAXSwBAAAAAAAgAgEAAAAA/Z5dXWBdXV0BBAAAAAAB/wcMAAAAAQERAAARABMAAQLeDQAIAAEAAAADAAAIAQAAAAEAAAABAAAABgD6+lJTQTEIAAAAAAAAAAAc5QY= | base64 -d > /tmp/poc.bin
```

##### Output
```
$ cargo fuzz run pdu_decoding /tmp/poc.bin
    Finished `release` profile [optimized + debuginfo] target(s) in 0.15s
     Running `target/aarch64-apple-darwin/release/pdu_decoding -artifact_prefix=ironrdp-testing/IronRDP/fuzz/artifacts/pdu_decoding/ /tmp/poc.bin`
pdu_decoding(54478,0x1e2632100) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1637312663
INFO: Loaded 1 modules   (419980 inline 8-bit counters): 419980 [0x101b1c0c0, 0x101b8294c),
INFO: Loaded 1 PC tables (419980 PCs): 419980 [0x101b82950,0x1021eb210),
target/aarch64-apple-darwin/release/pdu_decoding: Running 1 inputs 1 time(s) each.
Running: /tmp/poc.bin
thread '' panicked at ironrdp-testing/IronRDP/crates/ironrdp-pdu/src/rdp/server_license/server_license_request/cert.rs:164:23:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==54478== ERROR: libFuzzer: deadly signal
    #0 0x103381470 in __sanitizer_print_stack_trace+0x28 (librustc-nightly_rt.asan.dylib:arm64+0x59470)
[...]

PR #409, opened 2024-03-11
fix(pdu): overflow when decoding RsaPublicKey

nszetei:
Closes https://github.com/Devolutions/IronRDP/issues/395


nszetei:
> > I also considered adding another error for bit length and checking it separately, I don't know if that's more preferred.
> 
> Now that you mention it, maybe we should have another error kind for the "invalid bit length" case? Do you think you could add this before I merge?

Sure, I'll do it shortly. Thank you for confirming.

CBenoit:
Thank you! And it’s merged!

Issue #397, opened 2024-03-07
Fuzz(pdu_decoding) - ironrdp-pdu/src/basic_output/pointer.rs:75:16: attempt to calculate the remainder with a divisor of zero


nszetei:
#####  IronRDP revision

[3ee44b5c9b858b003148536637476b7133c5858e](https://github.com/Devolutions/IronRDP/commit/3ee44b5c9b858b003148536637476b7133c5858e)

##### Build platform
macOS 13.6.4 / 22G513

##### Reproduction steps
- Extend the `pdu_decode` fuzzing [harness](https://github.com/Devolutions/IronRDP/blob/3ee44b5c9b858b003148536637476b7133c5858e/crates/ironrdp-fuzzing/src/oracles/mod.rs#L48) for decoding additional update code:
```
let _ = fast_path::FastPathUpdate::decode_with_code(data, fast_path::UpdateCode::ColorPointer);
```
- Testcase:
```
$ echo ckRSSUEAslcAAAAAAAAAQwEAAAAAACnw | base64 -D  > /tmp/poc.bin
```

##### Output
```
$ cargo fuzz run pdu_decoding /tmp/poc.bin

    Finished `release` profile [optimized + debuginfo] target(s) in 0.09s
     Running `target/aarch64-apple-darwin/release/pdu_decoding -artifact_prefix=ironrdp-testing/IronRDP/fuzz/artifacts/pdu_decoding/ /tmp/poc.bin`
pdu_decoding(88023,0x1e2632100) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 360886850
INFO: Loaded 1 modules   (419980 inline 8-bit counters): 419980 [0x1061d00c0, 0x10623694c),
INFO: Loaded 1 PC tables (419980 PCs): 419980 [0x106236950,0x10689f210),
target/aarch64-apple-darwin/release/pdu_decoding: Running 1 inputs 1 time(s) each.
Running: /tmp/poc.bin
thread '' panicked at ironrdp-testing/IronRDP/crates/ironrdp-pdu/src/basic_output/pointer.rs:75:16:
attempt to calculate the remainder with a divisor of zero
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==88023== ERROR: libFuzzer: deadly signal
[...]

Issue #396, opened 2024-03-07
Fuzz(pdu_decoding) - ironrdp-pdu/src/basic_output/pointer.rs:133:35: attempt to add with overflow


nszetei:
#####  IronRDP revision

[3ee44b5c9b858b003148536637476b7133c5858e](https://github.com/Devolutions/IronRDP/commit/3ee44b5c9b858b003148536637476b7133c5858e)

##### Build platform
macOS 13.6.4 / 22G513

##### Reproduction steps
- Extend the `pdu_decode` fuzzing [harness](https://github.com/Devolutions/IronRDP/blob/3ee44b5c9b858b003148536637476b7133c5858e/crates/ironrdp-fuzzing/src/oracles/mod.rs#L48) for decoding additional update code:
```
let _ = fast_path::FastPathUpdate::decode_with_code(data, fast_path::UpdateCode::ColorPointer);
```
- Testcase:
```
$ echo ckRDQwAAjgAAADfaN3JEQw== | base64 -D  > /tmp/poc.bin
```

##### Output
```
    Finished `release` profile [optimized + debuginfo] target(s) in 0.07s
     Running `target/aarch64-apple-darwin/release/pdu_decoding -artifact_prefix=ironrdp-testing/IronRDP/fuzz/artifacts/pdu_decoding/ /tmp/poc.bin`
pdu_decoding(76633,0x1e2632100) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3635368031
INFO: Loaded 1 modules   (419980 inline 8-bit counters): 419980 [0x1026ec0c0, 0x10275294c),
INFO: Loaded 1 PC tables (419980 PCs): 419980 [0x102752950,0x102dbb210),
target/aarch64-apple-darwin/release/pdu_decoding: Running 1 inputs 1 time(s) each.
Running: /tmp/poc.bin
thread '' panicked at ironrdp-testing/IronRDP/crates/ironrdp-pdu/src/basic_output/pointer.rs:133:35:
attempt to add with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==76633== ERROR: libFuzzer: deadly signal
    #0 0x103f51470 in __sanitizer_print_stack_trace+0x28 (librustc-nightly_rt.asan.dylib:arm64+0x59470)
[...]

PR #410, opened 2024-03-11
fix(pdu): fixes three issues in pointer.rs and one in bitmap.rs

nszetei:
Closes https://github.com/Devolutions/IronRDP/issues/396
Closes https://github.com/Devolutions/IronRDP/issues/397
Closes https://github.com/Devolutions/IronRDP/issues/399

Additionally:
- Fixes a bug in `bitmap.rs` where `encoded_bitmap_data_length < CompressedDataHeader::ENCODED_SIZE` leads to overflow.
- Updates the fuzzer to detect similar problems when calling `fast_path::UpdateCode` and adds tests to prevent future regressions.


CBenoit (approved):
LGTM!

Issue #401, opened 2024-03-07
Fuzz(cliprdr_format) - OOM via ironrdp_cliprdr_format::bitmap::decode_png


nszetei:
#####  IronRDP revision

[3ee44b5c9b858b003148536637476b7133c5858e](https://github.com/Devolutions/IronRDP/commit/3ee44b5c9b858b003148536637476b7133c5858e)

##### Build platform
macOS 13.6.4 / 22G513

##### Reproduction steps
- Enable the `cliprdr_format` harness:
```
mv fuzz_targets/cliprdr_format.rs /tmp/cliprdr_format.rs
cargo fuzz add cliprdr_format
mv /tmp/cliprdr_format.rs fuzz_targets/cliprdr_format.rs
```

- Testcase:
```
$ echo iVBORw0KGgoAAAANSUhEUgAAACBaAAAgAgMAAAAOFJJnAAAAIklEQVQYPQA= | base64 -D  > /tmp/poc.bin
```

##### Output

```
$ cargo fuzz run cliprdr_format /tmp/poc.bin

    Finished `release` profile [optimized + debuginfo] target(s) in 0.08s
     Running `target/aarch64-apple-darwin/release/cliprdr_format -artifact_prefix=ironrdp-testing/IronRDP/fuzz/artifacts/cliprdr_format/ /tmp/poc.bin`
cliprdr_format(46067,0x1e2632100) malloc: nano zone abandoned due to inability to reserve vm space.
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1176174113
INFO: Loaded 1 modules   (420120 inline 8-bit counters): 420120 [0x1028d82a0, 0x10293ebb8),
INFO: Loaded 1 PC tables (420120 PCs): 420120 [0x10293ebb8,0x102fa7d38),
target/aarch64-apple-darwin/release/cliprdr_format: Running 1 inputs 1 time(s) each.
Running: /tmp/poc.bin
==46067== ERROR: libFuzzer: out-of-memory (malloc(193273532416))
   To change the out-of-memory limit use -rss_limit_mb=<N>

    #0 0x104141470 in __sanitizer_print_stack_trace+0x28 (librustc-nightly_rt.asan.dylib:arm64+0x59470)
    #1 0x10227aae0 in fuzzer::PrintStackTrace()+0x30 (cliprdr_format:arm64+0x10146eae0)
[...]

PR #411, opened 2024-03-11
fix(cliprdr): OOM via decode_png and vector allocation

nszetei:
Closes https://github.com/Devolutions/IronRDP/issues/401


CBenoit (approved):
LGTM! I wonder if we have other similar oversights in this module.

nszetei:
> LGTM! I wonder if we have other similar oversights in this module.

There was nothing obvious, but using the PNG corpus for fuzzing, I found one more bug. I can file an issue since I may need help with the fix.

CBenoit:
> > LGTM! I wonder if we have other similar oversights in this module.
> 
> There was nothing obvious, but using the PNG corpus for fuzzing, I found one more bug. I can file an issue since I may need help with the fix.

Yes, please!

PR #413, opened 2024-03-12
test(fuzz): update dependencies

nszetei:
Currently, the fuzzer uses a `png` package with the old version `0.17.11`, while the main build runs the latest `0.17.13`.

Version `0.17.11` is affected by several reproducible issues, such as https://github.com/image-rs/image-png/pull/445 or https://github.com/image-rs/image-png/issues/461. This commit updates all dependencies for the fuzzer.


CBenoit (approved):
Good catch. Thank you!

PR #416, opened 2024-03-17
feat(web): add API for cursor style override

### LINKED ISSUE #415: Override cursor image API for `iron-remote-gui`
For some special cases like headless ARD (Mac) hosts (IronVNC backend), there is no option to use client-side cursor, and the server cursor is always visible. To mitigate such cases, we need API in our web component to override the cursor image or hide it completely.

The following API should be provided:
- Completely hide the cursor when inside the remote screen area - for ARD headless case)
- Override cursor image with given data URL (e.g. base64 encoded PNG) - to have the option to increase cursor visibility when remote decides to hide the cursor both server-side and client-side.

Also, to test the API, it is advised to add some control (e.g. drop-down list) in `iron-svelte-client` to select cursor override (no override, hide the cursor, static image - crosshair)

pacmancoder:
This PR adds a new API to our `iron-remote-gui` module which allows to override cursor style (e.g. custom image or hide cursor)

Note: web-client GUI change: added "Toggle cursor kind" button to the top action bar (switched between context-defined cursor (custom VNC&RDP cursors) and static crosshair 15x15 cursor image

Closes #415


web-client/iron-svelte-client/static/crosshair.png
CBenoit, line 1: question: Where did you get this cursor? Did you draw it yourself? I want to be sure we are allowed to include and use it freely.

pacmancoder: Yep, I draw it in paint 😄

CBenoit: Perfect! :+1:

PR #418, opened 2024-03-20
feat: support for `Server Deactivate All PDU`

ibeckermayer:
The [Deactivate All PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/8a29971a-df3c-48da-add2-8ed9a05edc89)  is sent from server to client to indicate that the connection will be dropped or that a capability re-exchange using a [Deactivation-Reactivation Sequence](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/dfc234ce-481a-4674-9a5d-2a7bafb14432) will occur.

In practice, this happens when a screen resize is requested. When this PDU is received by the client, it is meant to begin executing the Deactivation-Reactivation Sequence, which is the Capability Exchange and Connection Finalization phases of the connection sequence.

What this PR does is

1. Break out the Capability Exchange and Connection Finalization phases of the connection sequence into their own logical abstraction `ConnectionActivationSequence`. This allows this execution `Sequence` to be used both during the initial connection phase, as well as be passed out as an [`x224::ProcessorOutput`](https://github.com/Devolutions/IronRDP/blob/717d02068b2cbcae804c3a5043d4f440a8ae62d8/crates/ironrdp-session/src/x224/mod.rs#L33-L37) in the event of a Deactivate All PDU.
2. Refactors the [`rfx::DecodingContext`](https://github.com/Devolutions/IronRDP/blob/717d02068b2cbcae804c3a5043d4f440a8ae62d8/crates/ironrdp-session/src/rfx.rs#L17-L21) to remove its explicit state management. This is included because, upon a successful Deactivation-Reactivation Sequence executing, the remote fx system is also reset, with a Sync being re-sent and so on. With this refactor, rather than needing to communicate this "reset" from the x224 processor over to the fast path processor and ultimately the `rfx::DecodingContext` to update its explicit [`state`](https://github.com/Devolutions/IronRDP/blob/master/crates/ironrdp-session/src/rfx.rs#L18) to the right value, we just let the server keep track of all state and simply decode based on whatever header it sends us.

In terms of existing IronRDP clients (like web and native), these changes should have no effect, since they are not hooked up to send screen resizes yet. This was tested via wiring things up in Teleport which can be seen in part starting [here](https://github.com/gravitational/teleport/blob/92929e381f95526cf7e0a2ef39462413f5fedfef/lib/srv/desktop/rdp/rdpclient/src/client.rs#L329).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
 png = "0.17"
 bitflags = "2.4"
+byteorder = "1.5"
 
 [profile.dev]
 opt-level = 1
--- a/crates/ironrdp-async/src/connector.rs
+++ b/crates/ironrdp-async/src/connector.rs
@@ -2,13 +2,13 @@ use ironrdp_connector::credssp::{CredsspProcessGenerator, CredsspSequence, Kerbe
 use ironrdp_connector::sspi::credssp::ClientState;
 use ironrdp_connector::sspi::generator::GeneratorState;
 use ironrdp_connector::{
-    custom_err, ClientConnector, ClientConnectorState, ConnectionResult, ConnectorError, ConnectorResult, Sequence,
-    ServerName, State as _, Written,
+    custom_err, ClientConnector, ClientConnectorState, ConnectionResult, ConnectorError, ConnectorResult, ServerName,
+    State as _,
 };
 use ironrdp_pdu::write_buf::WriteBuf;
 
 use crate::framed::{Framed, FramedRead, FramedWrite};
-use crate::AsyncNetworkClient;
+use crate::{single_sequence_step, AsyncNetworkClient};
 
 #[non_exhaustive]
 pub struct ShouldUpgrade;
@@ -23,7 +23,7 @@ where
     info!("Begin connection procedure");
 
     while !connector.should_perform_security_upgrade() {
-        single_connect_step(framed, connector, &mut buf).await?;
+        single_sequence_step(framed, connector, &mut buf).await?;
     }
 
     Ok(ShouldUpgrade)
@@ -73,7 +73,7 @@ where
     }
 
     let result = loop {
-        single_connect_step(framed, &mut connector, &mut buf).await?;
+        single_sequence_step(framed, &mut connector, &mut buf).await?;
 
         if let ClientConnectorState::Connected { result } = connector.state {
             break result;
@@ -174,70 +174,6 @@ where
     }
 
     connector.mark_credssp_as_done();
-
-    Ok(())
-}
-
-pub async fn single_connect_step<S>(
-    framed: &mut Framed<S>,
-    connector: &mut ClientConnector,
-    buf: &mut WriteBuf,
-) -> ConnectorResult<()>
-where
-    S: FramedWrite + FramedRead,
-{
-    buf.clear();
-    let written = single_connect_step_read(framed, connector, buf).await?;
-    single_connect_step_write(framed, buf, written).await
-}
-
-pub async fn single_connect_step_read<S>(
-    framed: &mut Framed<S>,
-    connector: &mut dyn Sequence,
-    buf: &mut WriteBuf,
-) -> ConnectorResult<Written>
-where
-    S: FramedRead,
-{
-    buf.clear();
-
-    if let Some(next_pdu_hint) = connector.next_pdu_hint() {
-        debug!(
-            connector.state = connector.state().name(),
-            hint = ?next_pdu_hint,
-            "Wait for PDU"
-        );
-
-        let pdu = framed
-            .read_by_hint(next_pdu_hint)
-            .await
-            .map_err(|e| ironrdp_connector::custom_err!("read frame by hint", e))?;
-
-        trace!(length = pdu.len(), "PDU received");
-
-        connector.step(&pdu, buf)
-    } else {
-        connector.step_no_input(buf)
-    }
-}
-
-async fn single_connect_step_write<S>(
-    framed: &mut Framed<S>,
-    buf: &mut WriteBuf,
-    written: Written,
-) -> ConnectorResult<()>
-where
-    S: FramedWrite,
-{
-    if let Some(response_len) = written.size() {
-        debug_assert_eq!(buf.filled_len(), response_len);
-        let response = buf.filled();
-        trace!(response_len, "Send response");
-        framed
-            .write_all(response)
-            .await
-            .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
-    }
 
     Ok(())
 }
--- a/crates/ironrdp-async/src/connector.rs
+++ b/crates/ironrdp-async/src/connector.rs
@@ -2,8 +2,8 @@ use ironrdp_connector::credssp::{CredsspProcessGenerator, CredsspSequence, Kerbe
 use ironrdp_connector::sspi::credssp::ClientState;
 use ironrdp_connector::sspi::generator::GeneratorState;
 use ironrdp_connector::{
-    custom_err, ClientConnector, ClientConnectorState, ConnectionResult, ConnectorError, ConnectorResult,
-    Sequence as _, ServerName, State as _,
+    custom_err, ClientConnector, ClientConnectorState, ConnectionResult, ConnectorError, ConnectorResult, Sequence,
+    ServerName, State as _, Written,
 };
 use ironrdp_pdu::write_buf::WriteBuf;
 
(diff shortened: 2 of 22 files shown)


pacmancoder:
This looks good to me 🚀 ! Let's wait for @CBenoit and merge this to unblock further PRs

PS: Clippy lints are failing (`xtask check lints`)

elmarco:
lgtm overall, however, please make it compile and run test after each commit!

crates/ironrdp-web/src/session.rs
```diff
@@ -603,6 +603,7 @@ impl Session {
                             hotspot_y,
                         })?;
                     }
+                    ActiveStageOutput::DeactivateAll(_) => todo!("DeactivateAll"),
```
CBenoit, line 606: issue: Can’t we handle this in a more graceful manner?

Cargo.toml
```diff
@@ -59,6 +59,7 @@ tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
 png = "0.17"
 bitflags = "2.4"
+byteorder = "1.5"
```
CBenoit, line 62: note: This is a legacy dependency, now that we have ReadCursor

ibeckermayer: These are marked as deprecated in the follow up PR https://github.com/Devolutions/IronRDP/pull/419/commits/9503901f066b2341480c1a2b4a35357d8c262944.

crates/ironrdp-async/src/connector.rs
```diff
@@ -187,10 +187,23 @@ where
     S: FramedWrite + FramedRead,
 {
     buf.clear();
+    let written = single_connect_step_read(framed, connector, buf).await?;
+    single_connect_step_write(framed, buf, written).await
+}
+
+pub async fn single_connect_step_read<S>(
```
CBenoit, line 194: question: Is it intended to be public?

crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs
```diff
@@ -136,6 +133,15 @@ impl PduBufferParsing<'_> for FrameBeginPdu {
             number_of_regions,
         })
     }
+}
+
+impl PduBufferParsing<'_> for FrameBeginPdu {
```
CBenoit, line 138: question: What is this trait? It seems to be a legacy one.

CBenoit:
Can you rebase on top of master to get the fuzzing fixes?

crates/ironrdp-async/src/connector.rs
ibeckermayer: It's used by Teleport for executing the deactivation-reactivation sequence. It appears to be a generally useful pattern, since I also needed it the [attempted] implementation for IronRDP web: https://github.com/Devolutions/IronRDP/pull/421/files#diff-33dc7055fcffcb2318664572070611527a75668ad24c01e2ce01ae83e0b27defR616

crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs
ibeckermayer: I don't know anything more than you about the trait, however it looks like it's defined in a `legacy` module and is still heavily used:

```
$ grep -rn 'impl PduBufferParsing'.
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:78:impl PduBufferParsing<'_> for ContextPdu {
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:138:impl PduBufferParsing<'_> for FrameBeginPdu {
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:170:impl PduBufferParsing<'_> for FrameEndPdu {
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:203:impl PduBufferParsing<'_> for RegionPdu {
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:424:impl PduBufferParsing<'_> for RfxRectangle {
crates/ironrdp-pdu/src/codecs/rfx/data_messages.rs:464:impl PduBufferParsing<'_> for Quant {
crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs:36:impl PduBufferParsing<'_> for SyncPdu {
crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs:80:impl PduBufferParsing<'_> for CodecVersionsPdu {
crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs:130:impl PduBufferParsing<'_> for ChannelsPdu {
crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs:211:impl PduBufferParsing<'_> for RfxChannel {
crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs:245:impl PduBufferParsing<'_> for CodecVersion {
```

crates/ironrdp-web/src/session.rs
ibeckermayer: Presuming [this system](https://github.com/Devolutions/IronRDP/pull/421/files#diff-33dc7055fcffcb2318664572070611527a75668ad24c01e2ce01ae83e0b27defR609-R655) is confirmed to work in web (just waiting for another test from Nefix in matrix), I think I can safely copy that into here.

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -505,3 +513,35 @@ bitflags! {
...
+///
+/// [2.2.3.1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/8a29971a-df3c-48da-add2-8ed9a05edc89
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ServerDeactivateAll;
+
+impl PduDecode<'_> for ServerDeactivateAll {
+    fn decode(src: &mut ReadCursor<'_>) -> PduResult<Self> {
+        let length_source_descriptor = src.read_u16();
+        let _ = src.read_slice(length_source_descriptor.into());
+        Ok(Self)
```
CBenoit, line 527: issue: `ensure_fixed_part_size!(in: src);` / `ensure_size!` missing (the fuzzer is failing because of this)

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -505,3 +513,35 @@ bitflags! {
...
+    }
+}
+
+impl PduEncode for ServerDeactivateAll {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> PduResult<()> {
+        // A 16-bit, unsigned integer. The size in bytes of the sourceDescriptor field.
```
CBenoit, line 533: issue:
```suggestion
        ensure_size!(in: dst, size: self.size());

        // A 16-bit, unsigned integer. The size in bytes of the sourceDescriptor field.
```

pacmancoder:
@ibeckermayer Looks like there is a problem found by fuzzing in `ShareControlPdu` pdu changes -- the following input for `pdu_decode` does not fail in master, but fails in branch: `[0, 4, 22, 0, 5, 0, 0, 0, 5, 0, 5, 40]` (see fuzzing logs)

You could just temporarily add a test in `crates\ironrdp-fuzzing\src\oracles\mod.rs` and investigate it from here

```rust
#[test]
fn test_pdu_decode() {
    const DATA: &'static [u8] = &[0, 4, 22, 0, 5, 0, 0, 0, 5, 0, 5, 40];
    pdu_decode(DATA);
}
```

cc @CBenoit

crates/ironrdp-pdu/src/rdp/headers.rs
pacmancoder: Most likely, this causing fuzzing fo fail

crates/ironrdp-async/src/connector.rs
CBenoit, line 194: naming: If I understand correctly, it’s not only about "connecting" anymore, am I right? It’s now a general-purpose helper for anything that is a `Sequence`. I suggest to move that into the `framed` module and rename to a more appropriate name. You can do the same for `single_connect_step_write`.

CBenoit (commented):
Looking good to me! Can be merged once remaining comments are addressed :+1:

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -518,18 +518,24 @@ bitflags! {
 ///
 /// [2.2.3.1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/8a29971a-df3c-48da-add2-8ed9a05edc89
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ServerDeactivateAll;
+pub struct ServerDeactivateAll {}
```
CBenoit, line 521: style (non-blocking): The style used across the codebase is the shorthand style (without `{}`)

crates/ironrdp-async/src/connector.rs
ibeckermayer: See db346b3c2aaf2be522b55d67a595f88c62cc5b29

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -505,3 +513,41 @@ bitflags! {
...
+
+impl ServerDeactivateAll {
+    const FIXED_PART_SIZE: usize = 2 /* length_source_descriptor */ + 1 /* source_descriptor */;
+}
+
+impl PduDecode<'_> for ServerDeactivateAll {
+    fn decode(src: &mut ReadCursor<'_>) -> PduResult<Self> {
+        ensure_fixed_part_size!(in: src);
+        let length_source_descriptor = src.read_u16();
+        let _ = src.read_slice(length_source_descriptor.into());
```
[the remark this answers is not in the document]
ibeckermayer: 3cd8167729697ab85a9674d347829e010bd4c566

crates/ironrdp-pdu/src/rdp/headers.rs
ibeckermayer: Oops, didn't mean to change that: 1de16664c6c7e90b0f9c5d9c3efc716f9bdbd631

pacmancoder (approved):
LGTM!

PR #420, opened 2024-03-21
fix(fuzz): fix failing fuzzing after pdu encode/decode refactoring

pacmancoder:
This PR fixed failing fuzzing on master


CBenoit (approved):
Approving now so you can merge when it’s green. Thank you!

PR #421, opened 2024-03-21
feat: implement the connection activation sequence rerun 

ibeckermayer:
upon receipt of a DeactivateAll in the web client


Cargo.toml
```diff
@@ -59,6 +59,7 @@ tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
 png = "0.17"
 bitflags = "2.4"
+byteorder = "1.5"
```
CBenoit, line 62: nitpick: Let’s try to not bring too many unrelated changes. We’ll remove this dependency eventually, I don’t think it’s worth to have it in the workspace.

crates/ironrdp-web/src/session.rs
```diff
@@ -384,6 +386,7 @@ pub struct Session {
 
 #[wasm_bindgen]
 impl Session {
+    #[allow(unused_assignments)]
```
CBenoit, line 389: question: What is the reason for suppressing this warning? Can you add a comment along it if it’s really necessary?

crates/ironrdp-client/src/rdp.rs
```diff
@@ -280,6 +280,7 @@ async fn active_session(
                 ActiveStageOutput::PointerBitmap(_) => {
                     // Not applicable, because we use the software cursor rendering.
                 }
+                ActiveStageOutput::DeactivateAll(_) => todo!("DeactivateAll"),
```
CBenoit, line 283: todo: Port the logic from ironrdp-web

Cargo.toml
ibeckermayer: This needed a rebase, this is no longer a change in this PR

crates/ironrdp-web/src/session.rs
ibeckermayer: Sort of janky, but I added a comment describing how to find the explanation alongside the actual false-positive location in the function below: 6e51da5fcce827725ddc5505459c758ab6a5ca02

Here's the lines:

https://github.com/Devolutions/IronRDP/blob/6e51da5fcce827725ddc5505459c758ab6a5ca02/crates/ironrdp-web/src/session.rs#L638-L644

crates/ironrdp-client/src/rdp.rs
ibeckermayer: https://github.com/Devolutions/IronRDP/pull/421/commits/2359d0030d88755c996f503538860011067e6ff5

crates/ironrdp-async/src/connector.rs
```diff
@@ -177,3 +177,67 @@ where
...
+        let response = buf.filled();
+        trace!(response_len, "Send response");
+        framed
+            .write_all(response)
+            .await
+            .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
+    }
+
+    Ok(())
+}
```
CBenoit, line 243: question: Is it the place we agreed to put these functions? I kind of remember having commented about this in a previous PR.

crates/ironrdp-web/src/session.rs
CBenoit: update: I don’t get the warning even if I remove the suppressing annotation:

image

Maybe it’s because of the toolchain version used?

crates/ironrdp-async/src/connector.rs
CBenoit: issue: I double checked the `framed` module. The function appears to be duplicated. Can you double check the diff again for that? Thank you!

crates/ironrdp-web/src/session.rs
ibeckermayer: Yeah, I had a goofed up local configuration. Fixed this in a888bb1954f03eb505c5844ce0fde8522eadd644

CBenoit (approved):
LGTM!

PR #422, opened 2024-03-25
feat: add a "ready" channel to the `DisplayControlClient`

ibeckermayer:
This can be used to notify callers to when the client is ready to send messages.

This is necessary, for example, in Teleport, where the user might adjust the size of their screen before the Display Control DVC is opened or before the server has sent its `DisplayControlCapabilities` pdu. Sending a resize before then would result in an error or an ignored event, respectively. Instead, Teleport wants to withhold the latest resize request, and send it when a read from this new "ready" channel indicates that the `DisplayControlClient` is ready to send messages.

This PR also adds code to automatically adjust `MonitorLayoutEntry`'s width when an odd value is passed.

### DIFF
--- a/crates/ironrdp-displaycontrol/Cargo.toml
+++ b/crates/ironrdp-displaycontrol/Cargo.toml
@@ -15,4 +15,5 @@ categories.workspace = true
 ironrdp-dvc.workspace = true
 ironrdp-pdu.workspace = true
 ironrdp-svc.workspace = true
+tokio = "1"
 tracing.workspace = true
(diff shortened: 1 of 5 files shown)


crates/ironrdp-displaycontrol/Cargo.toml
```diff
@@ -15,4 +15,5 @@ categories.workspace = true
 ironrdp-dvc.workspace = true
 ironrdp-pdu.workspace = true
 ironrdp-svc.workspace = true
+tokio = "1"
```
CBenoit, line 18: issue: Hardcoding tokio here is a bit problematic. The most obvious problem is that this is not WASM friendly. And generally, we’ve been cautious to avoid tying the core crates to a specific async runtime (and async in general). This breaks this design. Here, I think we can use a trait or a callback instead.

PR #423, opened 2024-03-25
feat(ffi): initial C# bindings (connector only)

irvingoujAtDevolution:
Created skeleton for FFI. 

Connector module is working now. Please see `ffi\dotnet\Devolutions.IronRdp.ConnectExample`
image

Note: this is a draft, we'll need to discuss more about 
1. how would we like to expose rustls.
2. how would we like to handle generics like framed and static channels.
3. should we expose async interface? or just stick with the sync one?

There are some imperfections/todo/fixme for now, my apologies.


ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            SocketAddr serverAddr = SocketAddr.LookUp(servername, 3389);
+
+            StdTcpStream stream = StdTcpStream.Connect(serverAddr);
+
+            BlockingTcpFrame tcpFrame = BlockingTcpFrame.FromTcpStream(stream);
+
+            ConfigBuilder configBuilder = ConfigBuilder.New();
+
+            // Password is wrong
```
CBenoit, line 35: question: Why this comment? I don’t understand the purpose.

ffi/justfile
CBenoit, line 1: suggestion: Can you add new commands in `xtask` instead?

ffi/src/connector/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        // FIXME: We need to create opaque for ironrdp::rdpdr::Rdpdr
+        /// Must use
+        pub fn with_static_channel_rdpdr(
+            &mut self,
+            computer_name: &str,
+            smart_card_device_id: u32,
+        ) -> Result<(), Box<IronRdpError>> {
+            let Some(connector) = self.0.take() else {
+                return Err(IronRdpErrorKind::NullPointer.into());
```
CBenoit, line 68: naming: `IronRdpErrorKind::Consumed` instead of `NullPointer`

ffi/src/pdu.rs
CBenoit, line 1: note: We’re not going to expose the PDUs. This is an implementation detail.

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
irvingoujAtDevolution: updated

ffi/justfile
irvingoujAtDevolution: updated

ffi/src/connector/mod.rs
irvingoujAtDevolution: updated

ffi/src/pdu.rs
irvingoujAtDevolution: updated

irvingoujAtDevolution:
formatter seems not working anymore, it does not tell you where the format is wrong, neither could `cargo +stable fmt --all ` fix it.

CBenoit:
> * how would we like to expose rustls.

We don’t expose rustls, we’ll use [SslStream](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslstream) in C#.
The API we expose should be sans-IO and we only push and pull bytes from a state machine.

https://sans-io.readthedocs.io/how-to-sans-io.html#why-write-i-o-free-protocol-implementations

IronRDP is already largely following this philosophy (and IronVNC too). We do expose higher level helpers to simplify things in upper layers, but this is not part of the "core crates".

> * how would we like to handle generics like framed and static channels.

For now, let’s not expose the static channels. As for "Framed", it’s not supposed to be exposed. This is a higher level helper for Rust code.

> * should we expose async interface? or just stick with the sync one?

We stick to the "core" API, which is "sans-IO" and thus do not require choosing between async or not.

However, we should provide an higher level API which may be async.
The additional  handwritten classes or addons should live in a `src/` folder like in Picky: https://github.com/Devolutions/picky-rs/tree/master/ffi/dotnet/Devolutions.Picky/src

However, let’s stick to the standard C# library in the higher level helpers. I don’t think we need much more than that and it will save us maintenance time. If a fancy dependency is required, let’s handle that downstream (in RDM).

Cargo.lock
```diff
@@ -142,9 +142,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.80"
+version = "1.0.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ad32ce52e4161730f7098c077cd2ed6229b5804ccf99e5366be1ab72a98b4e1"
+checksum = "0952808a6c2afd1aa8947271f3a60f1a6763c7b912d210184c5149b5cf147247"
```
CBenoit, line 147: nitpick: Avoid to update the Cargo.lock more than necessary. We use dependabot to update it piece by piece in a controlled way.

xtask/src/cli.rs
```diff
@@ -85,6 +87,10 @@ pub enum Action {
     WebCheck,
     WebInstall,
     WebRun,
+    FfiBuildDll {
+        debug: bool,
```
CBenoit, line 91: nitpick: Use `release` since this is what we are interested into (instead of negating the value twice).
```suggestion
        release: bool,
```

CBenoit (commented):
Can you also add a .gitattributes file like we have in Picky?
https://github.com/Devolutions/picky-rs/blob/master/.gitattributes

xtask/src/cli.rs
irvingoujAtDevolution: updated

ffi/build.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    use std::env;
+    use std::fs::File;
+    use std::io::Write;
+
+    fn generate_version_rc() -> String {
+        let output_name = "DevolutionsIronRdp";
+        let filename = format!("{}.dll", output_name);
+        let company_name = "Devolutions Inc.";
+        let legal_copyright = format!("Copyright 2019-2022 {}", company_name);
```
thenextman, line 22: Copyright date needs updating (and probably wherever this was copied from)

irvingoujAtDevolution: Thank you, both are updated, [picky.rs updated here](https://github.com/Devolutions/picky-rs/pull/259)

irvingoujAtDevolution:
> > * how would we like to expose rustls.
> 
> We don’t expose rustls, we’ll use [SslStream](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslstream) in C#. The API we expose should be sans-IO and we only push and pull bytes from a state machine. This means that as a rule of thumb, we don’t perform any I/O in the `ffi` crate (not even using `TcpStream`).
> 
> https://sans-io.readthedocs.io/how-to-sans-io.html#why-write-i-o-free-protocol-implementations
> 
> IronRDP is already largely following this philosophy (and IronVNC too). We do expose higher level helpers to simplify things in upper layers, but this is not part of the "core crates".
> 
> > * how would we like to handle generics like framed and static channels.
> 
> Let’s not expose the API for static channels like we do in Rust. I think we can safely hardcode the supported channels [like this](https://github.com/Devolutions/IronRDP/blob/a1d8bb246e69af3252dcae38497f495100775c93/crates/ironrdp-client/src/rdp.rs#L105-L109). We’re not going to implement static channels in C# either, so there is no need to care too deeply about genericity here.
> 
> As for "Framed", it’s not supposed to be exposed. This is a higher level helper for Rust code.
> 
> > * should we expose async interface? or just stick with the sync one?
> 
> We stick to the "core" API, which is "sans-IO" and thus do not require choosing between async or not.
> 
> However, we should provide an higher level API which may be async. The additional handwritten classes or addons should live in a `src/` folder like in Picky: https://github.com/Devolutions/picky-rs/tree/master/ffi/dotnet/Devolutions.Picky/src
> 
> However, let’s stick to the standard C# library in the higher level helpers. I don’t think we need much more than that and it will save us maintenance time. If a fancy dependency is required, let’s handle that downstream (in RDM).

Understood, I have updated the code align with sans-IO design. Please take a look at the example.

ffi/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+categories.workspace = true
+
+[lib]
+name = "ironrdp"
+crate-type = ["staticlib", "cdylib"]
+doc = false
+test = false
+doctest = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
CBenoit, line 19:
```suggestion
```

ffi/dotnet/Devolutions.IronRdp.ConnectExample/.vs/Devolutions.IronRdp.ConnectExample/DesignTimeBuild/.dtbcache.v2
CBenoit, line 1: question: What are these files? (.dtbcache.v2, vsidx, futdcache, etc). Can you double check you didn’t commit unwanted files?

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Frame.cs
CBenoit, line 1: naming: The name of the file doesn’t match the name of the class?

CBenoit, line 1: suggestion: How about moving it to be part of `Devolutions.IronRdp`?

ffi/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+#![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types
+pub mod connector;
+pub mod credssp;
+pub mod dvc;
+pub mod error;
+pub mod pdu;
+pub mod svc;
+pub mod utils;
+
+use tracing as _; // need this in the future
```
CBenoit, line 10: suggestion: You can remove the dependency if it’s not used. I’m not sure we really need it in the FFI crate.

ffi/src/connector/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+            };
+            self.0 = Some(
+                connector.with_static_channel(
+                    ironrdp::rdpdr::Rdpdr::new(Box::new(ironrdp::rdpdr::NoopRdpdrBackend {}), computer_name.to_owned())
+                        .with_smartcard(smart_card_device_id),
+                ),
+            );
+
+            Ok(())
+        }
```
CBenoit, line 78: suggestion: Let’s start by hardcoding the channels.

ffi/src/connector/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+            };
+            Ok(Box::new(State(connector.state())))
+        }
+    }
+
+    impl ServerName {
+        pub fn new(name: &str) -> Box<ServerName> {
+            Box::new(ServerName(ironrdp::connector::ServerName::new(name)))
+        }
+    }
```
CBenoit, line 178: suggestion: We probably don’t need to expose `ServerName` in C#.

ffi/src/connector/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+            Ok(())
+        }
+
+        pub fn is_terminal(&'a self) -> bool {
+            self.0.is_terminal()
+        }
+
+        pub fn as_any(&'a self) -> Box<crate::utils::ffi::Any<'a>> {
+            Box::new(crate::utils::ffi::Any(self.0.as_any()))
+        }
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Not useful/usable for now, removed!

ffi/src/error.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    pub fn reason(mut self, reason: &str) -> ValueConsumedError {
+        self.reason = Some(reason.to_owned());
+        self
+    }
+}
+
+impl ToString for ValueConsumedError {
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

ffi/src/utils/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+            let addr = addr.parse().map_err(|_| "Failed to parse address")?;
+            Ok(Box::new(SocketAddr(addr)))
+        }
+    }
+
+    #[diplomat::opaque]
+    pub struct VecU8(pub Vec<u8>);
+
+    impl VecU8 {
+        pub fn from_byte(bytes: &[u8]) -> Box<VecU8> {
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
```diff
@@ -0,0 +1,310 @@
...
+            connector.MarkSecurityUpgradeAsDone();
+
+            return (serverPublicKey, framedSsl);
+        }
+
+        private static async Task connect_begin(Framed<NetworkStream> framed, ClientConnector connector)
+        {
+            var writeBuf = WriteBuf.New();
+            while (!connector.ShouldPerformSecurityUpgrade())
+            {
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Frame.cs
irvingoujAtDevolution: updated

irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp.ConnectExample/.vs/Devolutions.IronRdp.ConnectExample/DesignTimeBuild/.dtbcache.v2
irvingoujAtDevolution: removed, seems like they are auto generated by dotnet or visual studio

CBenoit:
Can you double check again there is no unrequired API surface? For example: https://github.com/Devolutions/IronRDP/pull/423#discussion_r1548141349

ffi/src/utils/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+                .ok_or("Failed to resolve address")?;
+            Ok(Box::new(SocketAddr(addr)))
+        }
+
+        // named from_ffi_str to avoid conflict with std::net::SocketAddr::from_str
+        pub fn from_ffi_str(addr: &str) -> Result<Box<SocketAddr>, Box<IronRdpError>> {
+            let addr = addr.parse().map_err(|_| "Failed to parse address")?;
+            Ok(Box::new(SocketAddr(addr)))
+        }
+    }
```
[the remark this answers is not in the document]
irvingoujAtDevolution: removed socket addr

irvingoujAtDevolution:
> Can you double check again there is no unrequired API surface? For example: [#423 (comment)](https://github.com/Devolutions/IronRDP/pull/423#discussion_r1548141349)

I believe all interfaces that directly utilize network is removed. Rest of the exposed structed may not be used for now, but may be required for advanced usage in the future.

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
```diff
@@ -0,0 +1,317 @@
+﻿using System.Net;
+using System.Net.Security;
+using System.Net.Sockets;
```
CBenoit, line 3: spacing:
```suggestion
using System.Net.Sockets;

```

Cargo.lock
irvingoujAtDevolution: updated!

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return (this.stream, this.buffer);
+    }
+
+    public byte[] Peek()
+    {
+        return this.buffer.ToArray();
+    }
+
+    // read from 0 to size bytes from the front of the buffer, and remove them from the buffer,keep the rest
```
CBenoit, line 24: suggestion: Use C# doc comments.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public (S, List<byte>) GetInner()
+    {
+        return (this.stream, this.buffer);
+    }
+
+    public byte[] Peek()
+    {
+        return this.buffer.ToArray();
```
CBenoit, line 21: issue: This is copying all the elements into a new array. Return `Span<T>`.

https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay

ffi/src/connector/config.rs
```diff
@@ -0,0 +1,247 @@
...
+            self.0.height
+        }
+    }
+
+    impl ConfigBuilder {
+        pub fn new() -> Box<Self> {
+            Box::<ConfigBuilder>::default()
+        }
+
+        pub fn with_username_and_passwrord(&mut self, username: &str, password: &str) {
```
CBenoit, line 76: typo:
```suggestion
        pub fn with_username_and_password(&mut self, username: &str, password: &str) {
```

ffi/src/connector/mod.rs
```diff
@@ -0,0 +1,165 @@
+#![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types
```
CBenoit, line 1: spacing:
```suggestion
#![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types

```

ffi/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types
```
CBenoit, line 1: spacing:
```suggestion
#![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types

```

ffi/src/connector/mod.rs
CBenoit, line 1: suggestion: Remove since this is already specified in the root module.
```suggestion
```

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
```diff
@@ -16,12 +17,16 @@ public Framed(S stream)
         return (this.stream, this.buffer);
     }
 
-    public byte[] Peek()
+    public Span<byte> Peek()
     {
-        return this.buffer.ToArray();
+        return CollectionsMarshal.AsSpan(this.buffer);
     }
```
CBenoit: suggestion: Document that items should not be added or removed from the List while the Span is in use. This means no reads and no writes.

irvingoujAtDevolution: updated!

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp/src/Framed.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public (S, List<byte>) GetInner()
+    {
+        return (this.stream, this.buffer);
+    }
+
+    /// <summary>
+    /// Returns a span that represents a portion of the underlying buffer without modifying it.
+    /// </summary>
+    /// <remarks>Memory Safe:The Framed instance should not be modified (any read operations) while span is in use.</remarks>
```
CBenoit, line 23:
```suggestion
    /// <remarks>Memory safety: the Framed instance should not be modified (any read operations) while span is in use.</remarks>
```

CBenoit (approved):
LGTM! Thank you!!

PR #424, opened 2024-03-26
feat: add a callback mechanism to the `DisplayControlClient`

ibeckermayer:
which is called when the server sends the `DisplayControlCapabilities` (indicating that the client is ready for use).

(Replacement for https://github.com/Devolutions/IronRDP/pull/422)


crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -22,29 +59,12 @@ impl DvcProcessor for DisplayControlClient {
     }
 
     fn process(&mut self, _channel_id: u32, payload: &[u8]) -> PduResult<Vec<DvcMessage>> {
-        // TODO: We can parse the payload here for completeness sake,
-        // in practice we don't need to do anything with the payload.
-        debug!("Got Display PDU of length: {}", payload.len());
-        Ok(Vec::new())
+        let caps = DisplayControlCapabilities::decode(&mut ReadCursor::new(payload))?;
+        debug!("received {:?}", caps);
```
pacmancoder, line 63:
```
debug!("Received display control capabilities: {:?}", caps)
```

crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -1,14 +1,51 @@
...
+            on_capabilities_received: Box::new(|_| {
+                debug!("No capabilities received callback set, ignoring.");
+                Ok(Vec::new())
+            }),
+        }
+    }
+
+    /// Sets a callback that will be called when capabilities are received from the server.
+    #[must_use]
+    pub fn with_capabilities_received_callback<F>(mut self, callback: F) -> Self
```
pacmancoder, line 35: Lets remove this and make `callback` part of `new` signature to enforce use of callback

crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -22,29 +59,12 @@ impl DvcProcessor for DisplayControlClient {
...
-        encode_dvc_messages(channel_id, vec![Box::new(pdu)], ChannelFlags::empty())
-    }
-}
-
-impl Default for DisplayControlClient {
-    fn default() -> Self {
-        Self::new()
-    }
-}
+type OnCapabilitiesReceived = Box<dyn Fn(DisplayControlCapabilities) -> PduResult<Vec<DvcMessage>> + Send + Sync>;
```
pacmancoder, line 70: I think we could drop `Sync` trait here

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: I'm of two minds of this. In the logs, this current version looks like

```
received DisplayControlCapabilities { ... }
```

ergo in

```
Received display control capabilities: DisplayControlCapabilities { ... }
```

the "display control capabilities" is redundant. OTOH "Received display control capabilities:" makes it easier to locate from the logs where in the code we are, because it's (more or less) unique. 

However I think that ultimately we should configure the logger formatting settings to give us that, in order to avoid the messiness/redundancy of relying on the uniqueness of the debug message. In which case keeping it as is will eventually get us annoyed enough to actually fix that.

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: We can't currently, because we have

```
impl DvcClientProcessor for DisplayControlClient {}
pub trait DvcClientProcessor: DvcProcessor {}
pub trait DvcProcessor: AsAny + Send + Sync {
    /// The name of the channel, e.g. "Microsoft::Windows::RDS::DisplayControl"
    fn channel_name(&self) -> &str;

    /// Returns any messages that should be sent immediately
    /// upon the channel being created.
    fn start(&mut self, channel_id: u32) -> PduResult<Vec<DvcMessage>>;

    fn process(&mut self, channel_id: u32, payload: &[u8]) -> PduResult<Vec<DvcMessage>>;

    fn close(&mut self, _channel_id: u32) {}
}
```

I traced the lineage of the requirement of `Sync` for `DvcProcessor` back to the requirement on `DynamicVirtualChannel` in a @CBenoit PR from several months back: https://github.com/Devolutions/IronRDP/pull/183/files#diff-34003f425460df2c4178a4c2a9b61ca57ece7544ed13d42975dfa8f8e016f224R37

Unless we can recall a specific reason for it, maybe we can remove it here?

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: 41f48e6ccd89b3fd87e4b9861ecbe56628db850f

crates/ironrdp-displaycontrol/src/client.rs
CBenoit: Sounds good to me! @ibeckermayer

crates/ironrdp-displaycontrol/src/client.rs
pacmancoder: According to our [style guide](https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#logging) log message should start from the capital message.

I agree without you on this one, my suggestion adds redundancy, let it be just `Received DisplayControlCapabilities { ... }` (starting from capital later)

crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -1,14 +1,47 @@
...
-use ironrdp_pdu::PduResult;
+use ironrdp_pdu::{cursor::ReadCursor, PduDecode, PduResult};
 use ironrdp_svc::{impl_as_any, ChannelFlags, SvcMessage};
 use tracing::debug;
 
 /// A client for the Display Control Virtual Channel.
-pub struct DisplayControlClient {}
+pub struct DisplayControlClient {
+    /// A callback that will be called when capabilities are received from the server.
+    /// If no callback is set, a default (inert) callback will be used.
```
pacmancoder, line 13: I think this comment needs to be adjusted after previous refactoring, as there is no way to create the client now without explicitly setting the callback

crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -1,14 +1,47 @@
...
+    where
+        F: Fn(DisplayControlCapabilities) -> PduResult<Vec<DvcMessage>> + Send + 'static,
+    {
+        Self {
+            on_capabilities_received: Box::new(callback),
+            ready: false,
+        }
+    }
+
+    pub fn ready(&self) -> bool {
```
pacmancoder, line 35: `(optional)`

Idea: To enforce correct API usage, we could return a token here, which  is required to call `encode_monitor message` later:

```rust
#[non_exhaustive]
#[derive(Clone, Copy, Debug)]
pub struct DisplayControlReadyToken;

    pub fn poll_ready(&self) -> Option<DisplayControlReadyToken> {
        ready.then(|| DisplayControlReadyToken)
    }

    pub fn encode_monitor(&self, _token: DisplayControlReadyToken, channel_id: u32, width: u32, height: u32) -> PduResult<Vec<SvcMessage>> {
        // ...
    }
}
```

crates/ironrdp-displaycontrol/src/pdu/mod.rs
```diff
@@ -30,6 +31,18 @@ pub enum DisplayControlPdu {
 impl DisplayControlPdu {
     const NAME: &'static str = "DISPLAYCONTROL_HEADER";
     const FIXED_PART_SIZE: usize = 4 /* Type */ + 4 /* Length */;
+
+    /// Creates a new [`DisplayControlPdu::MonitorLayout`] PDU for a single monitor with the given width and height.
+    pub fn create_monitor_layout_pdu(width: u32, height: u32) -> PduResult<Self> {
```
pacmancoder, line 36: Lets move this to `DisplayControlMonitorLayout` as an alternative constructor and name it something like `pub new_single_monitor(width: u32, height: u32)`

crates/ironrdp-displaycontrol/src/client.rs
```diff
@@ -1,14 +1,47 @@
...
+        self.ready
+    }
+
+    /// Builds a [`DisplayControlPdu::MonitorLayout`] as an [`SvcMessage`] for a monitor with the given dimensions.
+    pub fn encode_monitor(&self, channel_id: u32, width: u32, height: u32) -> PduResult<Vec<SvcMessage>> {
```
pacmancoder, line 40: Let's keep client API as flexible as possible, we need to accept `DisplayControlMonitorLayout` here instead of single monitor dimensions, which could be constructed on the call site as described [here](https://github.com/Devolutions/IronRDP/pull/424/files#r1541945197)

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: More trouble than it's worth, IMO

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: I simply renamed it to reflect the new "single primary monitor" concept. This covers what I see as likely the most typical use case; we can add / refactor things going forward when we need multi-monitor support and the like.

crates/ironrdp-displaycontrol/src/pdu/mod.rs
ibeckermayer: 50c649e5f4f993979948fcad5f6b7b436b79ea7a

crates/ironrdp-displaycontrol/src/client.rs
ibeckermayer: 50c649e5f4f993979948fcad5f6b7b436b79ea7a

pacmancoder (approved):
Looks good to me!

elmarco:
For consistency with other APIs and flexibility to add more callbacks, shouldn't it use trait object instead?

CBenoit (approved):
LGTM! Thank you!

CBenoit:
> For consistency with other APIs and flexibility to add more callbacks, shouldn't it use trait object instead?

@ibeckermayer If more callbacks are to be added in the future, I also recommend a trait instead of increasing the number of parameters.

PR #419, opened 2024-03-20
Adds support for dydynvc

ibeckermayer:
Apologies for the gigantic PR, but much of this necessitated being done in one gasp. A lot of the changes here is moving PDUs that were under `crates/ironrdp-pdu/src/rdp/vc/dvc` and using the old `PduParsing` paradigm into `crates/ironrdp-dvc` and forming them to the newer `PduEncode`/`PduDecode` paradigm. 

I also added some practical facilities for working with arbitrary dynamic channels, taking inspiration from the existing patterns we have for static virtual channels.

PR #425, opened 2024-03-30
chore: dynamic resize follow up

ibeckermayer:
Follow up to https://github.com/Devolutions/IronRDP/pull/419 with some minor cleanup.


xtask/src/check.rs
```diff
@@ -14,7 +14,7 @@ const EXTRA_LINTS: &[&str] = &[
     "clippy::fn_to_numeric_cast_any",
     "clippy::ptr_cast_constness",
     // == Correctness == //
-    "unused_tuple_struct_fields",
+    "dead_code",
```
ibeckermayer, line 17: This is a Rust 1.77.0 change. It's not clear which version is expected, perhaps that should be documented somewhere (or maybe it is, but I missed it)?

ibeckermayer: Ah, I see we have a `rust-toolchain.toml`, never mind!

CBenoit (approved):
Thank you!

PR #426, opened 2024-03-30
feat: configurable performance flags

ibeckermayer:
Adds performance flags to the `connector::Config`.

Also adds a Default impl for PerformanceFlags. Teleport requires this in order to optionally use a non-default performance flag based on our user config.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/client_info.rs
+++ b/crates/ironrdp-pdu/src/rdp/client_info.rs
@@ -599,9 +599,7 @@ bitflags! {
 
 impl Default for PerformanceFlags {
     fn default() -> Self {
-        Self::DISABLE_FULLWINDOWDRAG
-            | Self::DISABLE_MENUANIMATIONS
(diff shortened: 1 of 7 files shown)


pacmancoder (approved):
Looks good, thanks! 🚀

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -597,6 +597,12 @@ bitflags! {
     }
 }
 
+impl Default for PerformanceFlags {
+    fn default() -> Self {
+        Self::DISABLE_FULLWINDOWDRAG | Self::DISABLE_MENUANIMATIONS | Self::ENABLE_FONT_SMOOTHING
+    }
+}
```
CBenoit, line 604: thought: I think the default flags should not live in the PDU crate as this is a concern which is a bit more far-reaching.

CBenoit (approved):
LGTM! Thank you!

PR #430, opened 2024-04-03
feat: hook up resize for ironrdp-client crate

ibeckermayer:
This commit adds the ability to resize the RDP session window on the fly using the DisplayControl channel.

It includes some known issues:
1. After a resize, session performance degrades.
2. Too large a resize can cause an `InvalidChannelWidth` or `InvalidChannelHeight` error.

See: https://www.loom.com/share/7cbfc5b0e02d4aff998939f54816d14b?sid=8e69fec2-aaa7-4a0a-ae67-d277a5000972


crates/ironrdp-client/src/rdp.rs
```diff
@@ -107,7 +114,9 @@ async fn connect(
 
     let mut connector = connector::ClientConnector::new(config.connector.clone())
         .with_server_addr(server_addr)
-        .with_static_channel(ironrdp::dvc::DrdynvcClient::new())
+        .with_static_channel(
+            ironrdp::dvc::DrdynvcClient::new().with_dynamic_channel(DisplayControlClient::new(|_| Ok(Vec::new()))),
```
pacmancoder, line 118: From our doc comment:
```
    /// The `callback` will be called when capabilities are received from the server.
    /// It is important to note that the channel will not be fully operational until the capabilities are received.
    /// Attempting to send messages before the capabilities are received will result in an error or a silent failure.
```

I think we need to block any dynamic resolution changes from the client until this callback (passed in new) receives some caps. Also, based on caps we should prevent setting resolution to values greater than max monitor area received in caps.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -625,7 +625,7 @@ fn create_gcc_blocks<'a>(
 
     ClientGccBlocks {
         core: ClientCoreData {
-            version: RdpVersion::V5_PLUS,
+            version: RdpVersion::V10_12,
```
pacmancoder, line 628: Hmm, can't this change break something down the line? Is this required for display control functionality?

crates/ironrdp-connector/src/lib.rs
```diff
@@ -72,6 +72,10 @@ impl Credentials {
 pub struct Config {
     /// The initial desktop size to request
     pub desktop_size: DesktopSize,
+    /// The initial desktop scale factor to request.
+    ///
+    /// This becomes the `desktop_scale_factor` in the [`TS_UD_CS_CORE`](gcc::ClientCoreOptionalData) structure.
```
pacmancoder, line 77: Maybe let's remove this option completely if currently only `0` scale factor is supported?

crates/ironrdp-displaycontrol/src/pdu/mod.rs
```diff
@@ -352,16 +374,17 @@ impl MonitorLayoutEntry {
...
     /// - `width` SHOULD be even. If it is odd, it will be adjusted
-    ///   to the nearest even number by subtracting 1.
+    ///    to the nearest even number by subtracting 1.
     fn new_impl(mut width: u32, height: u32) -> PduResult<Self> {
         if width % 2 != 0 {
             let prev_width = width;
             width = width.saturating_sub(1);
             warn!(
-                "Monitor width cannot be odd, adjusting from {} to {}",
+                "Monitor width cannot be odd, adjusting from [{}] to [{}]",
```
pacmancoder, line 383: I think I missed this part in the previous review, but now I am thinking that making silent adjustments and just printing warning in the low-level crate is not optimal. Maybe let's add a separate helper function `adjust_display_size(width, height) -> (u32, u32)`  in `ironrdp-displaycontrol` which could be called explicitly before monitor layout construction?

crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs
```diff
@@ -171,10 +171,7 @@ pub struct RfxChannelWidth(i16);
 
 impl RfxChannelWidth {
     pub fn new(value: i16) -> Result<Self, RfxError> {
-        (1..=4096)
```
pacmancoder, line 174: Are those were accidentally removed lines?

crates/ironrdp-dvc/src/lib.rs
```diff
@@ -152,34 +198,35 @@ impl DynamicChannelSet {
...
+        self.channels.insert(name, DynamicVirtualChannelInternal::new(channel))
     }
 
     fn attach_channel_id(&mut self, name: DynamicChannelName, id: DynamicChannelId) -> Option<DynamicChannelId> {
         self.channel_id_to_name.insert(id, name.clone());
-        self.name_to_channel_id.insert(name, id)
+        self.name_to_channel_id.insert(name.clone(), id);
+        let dvc = self.get_by_channel_name_mut(&name)?;
+        dvc.channel_id = Some(id);
+        Some(id)
```
pacmancoder, line 212: old id should be returned here isntead

crates/ironrdp-dvc/src/lib.rs
```diff
@@ -98,21 +98,67 @@ pub fn encode_dvc_messages(
     Ok(res)
 }
 
-pub struct DynamicVirtualChannel {
+pub struct DynamicVirtualChannel<'a, T: DvcProcessor> {
```
pacmancoder, line 101: `DynamicVirtualChannel`/`DynamicVirtualChannelInternal` separation feels a bit confusing to me, what was the reason for such refactoring?

ibeckermayer: Yeah I took a wrong turn somewhere there, fixed in 1fd4b38c34777fa52d94dd56e2a112efde4ae62e

crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs
ibeckermayer: These limits are "SHOULD"s in the [spec](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdprfx/4060f07e-9d73-454d-841e-131a93aca675#Appendix_A_Target_3)

image

and the footnote says this:

image

In practice these limits were constraining the screen to a relatively small size. I haven't found a limit that the server won't serve us, e.g. I just tried `width: 5120, height: 2774` (fullscreen on my 32" monitor without any scale factor), the performance was terrible but the server will send it. (With a proper scale factor, fullscreen works fine).

crates/ironrdp-displaycontrol/src/pdu/mod.rs
ibeckermayer: bda1a96175175ea58f02e04fea89d12bc8a6d411

crates/ironrdp-connector/src/lib.rs
ibeckermayer: All values are supported, the feature is just not hooked up in the `ironrdp-client` currently. I think it's best to leave here, since it's already properly hooked into the correct place in the Client Core Data pdu.

crates/ironrdp-connector/src/connection.rs
ibeckermayer: That was leftover from me trying something out, reverted in 4d878fbcf6608dcb19161ea62e2f36cffd8a28eb

crates/ironrdp-client/src/rdp.rs
ibeckermayer: > I think we need to block any dynamic resolution changes from the client until this callback (passed in new) receives some caps.

This is possibility is already handled here:

https://github.com/Devolutions/IronRDP/blob/4d878fbcf6608dcb19161ea62e2f36cffd8a28eb/crates/ironrdp-client/src/rdp.rs#L207-L213

I added this callback for Teleport purposes, where we do things a little differently:
1. Any resizes that occur before the channel is open are held in an internal register
2. The callback we pass here reads that register and returns the encoded resize, such that the most recent browser size gets sent as soon as the channel is ready for it.

Something similar to that could be implemented in IronRDP, but it's a substantial piece of work compared to what's there now (reconnect if the channel isn't ready).

> Also, based on caps we should prevent setting resolution to values greater than max monitor area received in caps.

Seems like a fine check but in practice it's not a big risk, I just added a TODO for this for now 1644ff717ae8294f9f155b4e510bbf9024aa2467

crates/ironrdp-client/src/gui.rs
```diff
@@ -63,9 +63,17 @@ impl GuiContext {
             match event {
                 Event::WindowEvent { window_id, event } if window_id == window.id() => match event {
                     WindowEvent::Resized(size) => {
+                        let scale_factor = (window.scale_factor() * 100.0) as u32;
+                        // TODO: it should be possible to get the physical size here, however winit doesn't make it straightforward.
+                        // FreeRDP does it based on DPI reading grabbed via [`SDL_GetDisplayDPI`](https://wiki.libsdl.org/SDL2/SDL_GetDisplayDPI):
+                        // https://github.com/FreeRDP/FreeRDP/blob/ba8cf8cf2158018fb7abbedb51ab245f369be813/client/SDL/sdl_monitor.cpp#L250-L262
+                        let (physical_width, physical_height) = (0, 0);
```
CBenoit, line 70: question: Can’t you use this? https://docs.rs/winit/latest/winit/window/struct.Window.html#method.inner_size
Is it really fine to feed in 0s?

crates/ironrdp-connector/src/connection_activation.rs
```diff
@@ -355,7 +365,10 @@ fn create_client_confirm_active(
             })),
         }])),
         CapabilitySet::FrameAcknowledge(FrameAcknowledge {
-            max_unacknowledged_frame_count: 2,
+            // TODO: Revert this to 2 per FreeRDP.
```
CBenoit, line 368:
```suggestion
            // FIXME(#447): Revert this to 2 per FreeRDP.
```

crates/ironrdp-pdu/src/codecs/rfx/header_messages.rs
CBenoit: In that case, you can remove the `Result` from the signature

crates/ironrdp-client/src/gui.rs
ibeckermayer: `inner_size` gives you a [`PhysicalSize`](https://docs.rs/winit/latest/winit/dpi/struct.PhysicalSize.html), but that only gives you a "size represented in physical pixels." However without knowing the DPI (or other relation to a unit of length), we can't calculate the "physical size" that RDP wants, which is the width/height in millimeters. 

If we hand back `0`s here, they're simply ignored [per the spec](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/ea2de591-9203-42cd-9908-be7a55237d1c):

image

pacmancoder: @ibeckermayer 
I think passing in client `(0, 0)` is not optimal (it is not obvious that they will be ignored), maybe it would be great to change `RdpInputEvent::Resize` fields to have `physical_dimensions: Option<(u32, u32)>` instead of separate u32 fields? This will keep this implementation detail as close to PDU encode/decode code as possible

pacmancoder: This could be fixed in follow-up PR, merging PR to unblock @ibeckermayer

pacmancoder (approved):
Looks good to me!

PR #431, opened 2024-04-03
server: add pointer DisplayUpdate

elmarco:
This allows server implementation to set the pointer shape.


CBenoit (approved):
Assuming removing the `main label is not causing any problem, everything looks good to me! Thank you!!

PR #432, opened 2024-04-03
server: add ServerEvent for clipboard

elmarco:
The clipboard factory/backend alone isn't enough to send back or notify of messages, adopt a similar approach as the client, using a channel to dispatch and send messages from the server loop.

(ideally, I wish the channel implementation details wouldn't be handled by the main loop, but this has pros and cons, so let's just follow the client approach for now)

### DIFF
--- a/crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
+++ b/crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
@@ -166,6 +166,10 @@ impl<'a> FormatDataResponse<'a> {
     pub fn into_owned(self) -> OwnedFormatDataResponse {
         self.into_owned_pdu()
     }
+
+    pub fn into_data(self) -> Cow<'a, [u8]> {
+        self.data
+    }
 }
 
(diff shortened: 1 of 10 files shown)


crates/ironrdp-cliprdr/src/pdu/format_data/mod.rs
```diff
@@ -166,6 +166,10 @@ impl<'a> FormatDataResponse<'a> {
     pub fn into_owned(self) -> OwnedFormatDataResponse {
         self.into_owned_pdu()
     }
+
+    pub fn into_vec(self) -> Vec<u8> {
```
CBenoit, line 170: suggestion: Return Cow<'a, [u8]> and leave the choice to the caller

elmarco: ok

CBenoit (approved):
LGTM! I’ll enable auto-merge (rebase).

elmarco:
@CBenoit CI is failing because of unrelated fmt issue.. any idea?

PR #435, opened 2024-04-05
feat: scard ascii variants

ibeckermayer:
Adds handling for the ascii variants of the currently support smartcard calls.

A couple of Teleport users have encountered these in the wild, and one of them tested this implementation.

### DIFF
--- a/crates/ironrdp-rdpdr/src/pdu/esc.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/esc.rs
@@ -12,7 +12,7 @@ use ironrdp_pdu::cursor::{ReadCursor, WriteCursor};
 use ironrdp_pdu::utils::{
     encoded_multistring_len, read_multistring_from_cursor, write_multistring_to_cursor, CharacterSet,
 };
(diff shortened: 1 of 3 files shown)


crates/ironrdp-rdpdr/src/pdu/esc.rs
```diff
@@ -1780,3 +1833,19 @@ impl rpce::HeaderlessEncode for GetReaderIconReturn {
         + self.data.len() // dst.write_slice(&self.data);
     }
 }
+
+fn expect_charset(charset: Option<CharacterSet>) -> PduResult<CharacterSet> {
+    if charset.is_none() {
+        return Err(other_err!("internal error: missing character set"));
+    }
+    Ok(charset.unwrap())
```
CBenoit, line 1841: suggestion: It’s more idiomatic (and `unwrap`-free) like this:
```suggestion
    if let Some(charset) = charset {
        Ok(charset)
    } else {
        Err(other_err!("internal error: missing character set"))
    }
```

CBenoit (approved):
LGTM!

PR #441, opened 2024-04-09
Improve .NET example Linux and macOS support (follow up to 437)

CBenoit:
Follow up to https://github.com/Devolutions/IronRDP/pull/437

This PR is improving Linux (and macOS) support of the C# bindings. (Especially, build system and example.)

Can you review? @irvingoujAtDevolution


irvingoujAtDevolution (approved):
Thank you!

PR #444, opened 2024-04-22
fix(connector): fix cursor rendering in Windows 2019

probakowski:
This change fixes cursor rendering in Windows 2019 and older.
Previously only text cursor was rendered, all others defaulted to native cursor from client.

Set of flags matches what is used in [FreeRDP](https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/settings.c#L541)


CBenoit (approved):
Thank you!!

PR #449, opened 2024-04-29
refactor(displaycontrol): make `physical_size` an `Option`

ibeckermayer:
Follows up on #430 to make `physical_size` an `Option`

### DIFF
--- a/crates/ironrdp-web/src/image.rs
+++ b/crates/ironrdp-web/src/image.rs
@@ -3,7 +3,7 @@
 use ironrdp::pdu::geometry::{InclusiveRectangle, Rectangle as _};
 use ironrdp::session::image::DecodedImage;
 
-pub fn extract_partial_image(image: &DecodedImage, region: InclusiveRectangle) -> (InclusiveRectangle, Vec<u8>) {
+pub(crate) fn extract_partial_image(image: &DecodedImage, region: InclusiveRectangle) -> (InclusiveRectangle, Vec<u8>) {
     // PERF: needs actual benchmark to find a better heuristic
     if region.height() > 64 || region.width() > 512 {
         extract_whole_rows(image, region)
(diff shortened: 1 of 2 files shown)


crates/ironrdp-web/src/image.rs
```diff
@@ -3,7 +3,7 @@
 use ironrdp::pdu::geometry::{InclusiveRectangle, Rectangle as _};
 use ironrdp::session::image::DecodedImage;
 
-pub(crate) fn extract_partial_image(image: &DecodedImage, region: InclusiveRectangle) -> (InclusiveRectangle, Vec<u8>) {
+pub fn extract_partial_image(image: &DecodedImage, region: InclusiveRectangle) -> (InclusiveRectangle, Vec<u8>) {
```
pacmancoder, line 6: Why do we need `extract_partial_image` and a whole `mod image` to be public?
`ironrdp-web` crate compiles to wasm module, there should not be a public rust interface AFIK 🤔

ibeckermayer: I had this function copied in Teleport and in doing some cleanup work there, figured it would be more straightforward to just import it from IronRDP instead

- https://github.com/gravitational/teleport/pull/41028/files#diff-361bdd6758a56dbfc37f2acf44d79ba8da64212ae87bf32a221d66559bdda3adL345-L422
- https://github.com/gravitational/teleport/pull/41028/files#diff-361bdd6758a56dbfc37f2acf44d79ba8da64212ae87bf32a221d66559bdda3adR34

However if that's treading on your preferred design I can just revert that.

pacmancoder: Yes, I would rather keep `ironrdp-web` crate without public rust API, lets revert that

pacmancoder (approved):
LGTM

PR #452, opened 2024-05-01
feat(ffi): API for enabling tracing logs

irvingoujAtDevolution:


ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/MainWindow.axaml.cs
```diff
@@ -30,6 +30,7 @@ public MainWindow()
 
     private void OnOpened(object? sender, EventArgs e)
     {
+        Log.InitLogWithEnv();
```
CBenoit, line 33:
```suggestion
        Log.InitLogWithEnv();

```

ffi/src/log.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    #[diplomat::opaque]
+    pub struct Log;
+
+    impl Log {
+        pub fn init_log_with_env() {
+            if LOG_INITED.load(std::sync::atomic::Ordering::Relaxed) {
+                return;
+            }
+            LOG_INITED.store(true, std::sync::atomic::Ordering::Relaxed);
```
CBenoit, line 18: suggestion: Use [`Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) instead

ffi/src/log.rs
```diff
@@ -0,0 +1,33 @@
...
+    use std::{fs::File, path::PathBuf};
+
+    use super::LOG_INITED;
+    use std::fs;
+
+    #[diplomat::opaque]
+    pub struct Log;
+
+    impl Log {
+        pub fn init_log_with_env() {
```
CBenoit, line 14: naming:
```suggestion
        pub fn init_with_env() {
```

Will give `Log.InitWithEnv()`

ffi/src/log.rs
```diff
@@ -0,0 +1,33 @@
...
+            if LOG_INITED.load(std::sync::atomic::Ordering::Relaxed) {
+                return;
+            }
+            LOG_INITED.store(true, std::sync::atomic::Ordering::Relaxed);
+
+            let filepath: PathBuf = std::env::var("IRONRDP_LOG_FILE")
```
CBenoit, line 20: suggestion: Use `IRONRDP_LOG`, like in `ironrdp-client`

ffi/src/log.rs
```diff
@@ -0,0 +1,33 @@
...
+                .unwrap_or_else(|_| "/tmp/ironrdp.log".to_owned())
+                .try_into()
+                .expect("Failed to parse log file path");
+
+            fs::create_dir_all(filepath.parent().expect("Path has no parent")).expect("Failed to create log directory");
+            let file = File::create(filepath).unwrap();
+            tracing_subscriber::fmt::SubscriberBuilder::default()
+                .with_env_filter(tracing_subscriber::EnvFilter::from_env("IRONRDP_LOG"))
+                .with_writer(file)
+                .init();
```
CBenoit, line 30: suggestion: Reuse the same logic as in `ironrdp-client`

https://github.com/Devolutions/IronRDP/blob/083b5044ad2b961581cf25255aeaf01872edb647/crates/ironrdp-client/src/main.rs#L87-L129

ffi/src/log.rs
irvingoujAtDevolution: updated

ffi/src/log.rs
irvingoujAtDevolution: updated

ffi/src/log.rs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM! Thank you!

PR #457, opened 2024-05-03
chore(ffi):allow Devolutions.IronRdp to be build and packed with Nuget with the dll on Windows

irvingoujAtDevolution:
First step in building FFI, allow it to be used simply by `dotnet add package` command by our consumers without worrying about loading the .dll

Only support windows x64 for now, and used solely in local development, not suppose to be pushed to any Nuget registry.
Other platform support will be added later.

---

step to build and publish
1. run `cargo xtask ffi bindings` and `cargo xtask ffi build --release` in IronRdp project root
2. (optional) remove the /bin/ directory under `Devolutions.IronRdp` 
3. run `dotnet build -r  win-x64` to build the dotnet project as well as packages it
4. run `dotnet nuget push ".\bin\Release\Devolutions.IronRdp.{version}.nupkg" --source "{your_local_nuget_server_source}"`

the other end can download with 
`dotnet add package "Devolutions.IronRdp" --source "{your_local_nuget_server_source}"`

reference: https://learn.microsoft.com/en-us/nuget/create-packages/native-files-in-net-packages

### DIFF
--- a/ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
+++ b/ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
@@ -7,9 +7,9 @@ namespace Devolutions.IronRdp.ConnectExample
     {
         static async Task Main(string[] args)
         {
-            var arguments = ParseArguments(args);
-
             Log.InitWithEnv();
+
+            var arguments = ParseArguments(args);
 
             if (arguments == null)
             {
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Build artifacts
 /target
-
+/dependencies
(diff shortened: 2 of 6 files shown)


ffi/dotnet/Devolutions.IronRdp.ConnectExample/Program.cs
```diff
@@ -7,10 +7,10 @@ class Program
     {
         static async Task Main(string[] args)
         {
-            var arguments = ParseArguments(args);
-
             Log.InitWithEnv();
```
irvingoujAtDevolution, line 10: This will invoke dll not found sooner

thenextman (approved):

PR #459, opened 2024-05-08
feat(ffi): expose cliprdr to dotnet bindings

irvingoujAtDevolution:
Example: 
https://github.com/Devolutions/IronRDP/assets/139169536/4b58fb99-7fcc-4960-9be8-367d4574c761

1. Expose Windows specific clipboard interface to dotnet
2. Added clipboard to example


ffi/src/error.rs
```diff
@@ -92,6 +105,8 @@ pub mod ffi {
         AccessDenied,
         #[error("Incorrect rust enum type")]
         IncorrectEnumType,
+        #[error("Clipboard error")]
+        ClipboardError,
```
CBenoit, line 109: naming: Do not use the "Error" suffix in variant names, this is redundant with the enum name

ffi/src/clipboard/message.rs
```diff
@@ -0,0 +1,65 @@
+#[diplomat::bridge]
+pub mod ffi {
+
+    #[diplomat::opaque]
+    pub struct ClipboardMessage(pub ironrdp::cliprdr::backend::ClipboardMessage);
+
+    impl ClipboardMessage {
+        pub fn get_enum_type(&self) -> ClipboardMessageType {
```
CBenoit, line 8: naming: get_type

The returned data structure being an "enum" is not relevant

ffi/src/clipboard/message.rs
```diff
@@ -0,0 +1,65 @@
...
+        SendFormatData,
+        SendInitiatePaste,
+        Error,
+    }
+
+    #[diplomat::opaque]
+    pub struct ClipboardFormatIterator(pub Vec<ironrdp::cliprdr::pdu::ClipboardFormat>);
+
+    #[diplomat::opaque]
+    pub struct OwndFormatDataResponse(pub Option<ironrdp::cliprdr::pdu::OwnedFormatDataResponse>);
```
CBenoit, line 61: naming: remove the Owned prefix, it's not relevant for C#

ffi/src/clipboard/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+            let proxy = FfiClipbarodMessageProxy { sender };
+
+            let clipboard = unsafe { ironrdp_cliprdr_native::WinClipboard::new(HWND(hwnd.0), proxy) }?;
+
+            Ok(Box::new(WinCliprdr { clipboard, receiver }))
+        }
+
+        pub fn next_clipboard_message(&self) -> Option<Box<ClipboardMessage>> {
+            self.receiver.try_recv().ok().map(ClipboardMessage).map(Box::new)
```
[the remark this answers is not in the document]
pacmancoder: I think adding a blocking call would be ok, that way the client code will have a choice of how to handle the clipboard - via polling or by blocking wait.

ffi/src/clipboard/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+            let cliprdr = ironrdp::cliprdr::Cliprdr::new(backend);
+            Box::new(Cliprdr(Some(cliprdr)))
+        }
+    }
+
+    #[diplomat::opaque]
+    pub struct Cliprdr(pub Option<ironrdp::cliprdr::Cliprdr<Client>>);
+
+    #[diplomat::opaque]
+    pub struct CliprdrReference<'a>(pub &'a ironrdp::cliprdr::Cliprdr<Client>);
```
pacmancoder, line 77: Oof, I don't like the idea of passing the Rust reference to dotnet garbage-collected object and keeping it as a public API -- we may make a shoot in our legs with this someday 😢.

Could we avoid this? maybe by moving `CliprdrReference` methods to `Connection` instead without construction or intermediate "Reference" object?

cc @irvingoujAtDevolution @CBenoit

CBenoit: @pacmancoder This should be fine theoretically. It’s currently not sound, but I plan to fix the C# backend for keeping track of such borrows. Diplomat model allows because this can be made safe. They’ve done it for their JavaScript backend for example.

ffi/src/clipboard/message.rs
irvingoujAtDevolution: GetType is a C# reserved function for all classes.

CBenoit: In that case, `get_message_type` may be a better alternative

ffi/src/clipboard/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+            let clipboard = unsafe { ironrdp_cliprdr_native::WinClipboard::new(HWND(hwnd.0), proxy) }?;
+
+            Ok(Box::new(WinCliprdr { clipboard, receiver }))
+        }
+
+        pub fn next_clipboard_message(&self) -> Option<Box<ClipboardMessage>> {
+            self.receiver.try_recv().ok().map(ClipboardMessage).map(Box::new)
+        }
+
+        pub fn backend_facotry(&self) -> Box<CliprdrBackendFactory> {
```
CBenoit, line 36: typo: factory

ffi/src/clipboard/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+                .map_err(|_| "cannot convert from u64 input".into())
+        }
+
+        #[cfg(target_pointer_width = "64")]
+        pub fn new(hwnd: u64) -> Result<Box<Hwnd>, Box<IronRdpError>> {
+            isize::try_from(hwnd)
+                .map(Hwnd)
+                .map(Box::new)
+                .map_err(|_| "cannot convert from u64 input".into())
+        }
```
CBenoit, line 59: issue: Diplomat does not know about `cfg(target_pointer_width)`.

ffi/src/clipboard/mod.rs
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/Devolutions.IronRdp.AvaloniaExample.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Avalonia.Fonts.Inter" Version="11.0.10" />
     <!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
     <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="Avalonia.Diagnostics" Version="11.0.10" />
+    <PackageReference Include="Avalonia.Win32" Version="11.0.10" />
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated, I cannot yet verify if it runs on linux machine, I currently only have wsl, but I made sure it compiles on Linux.

ffi/src/error.rs
irvingoujAtDevolution: updated

ffi/src/clipboard/message.rs
irvingoujAtDevolution: updated

thenextman:
There is a typo throughout the PR: `Facotry` -> "Factory"

ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/Devolutions.IronRdp.AvaloniaExample.csproj
```diff
@@ -16,7 +16,12 @@
...
     <!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
     <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="Avalonia.Diagnostics" Version="11.0.10" />
-    <PackageReference Include="Avalonia.Win32" Version="11.0.10" />
     <ProjectReference Include="../Devolutions.IronRdp/Devolutions.IronRdp.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0-windows' ">
+    <PackageReference Include="Avalonia.Win32" Version="11.0.10" />
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

ffi/src/clipboard/mod.rs
```diff
@@ -1,63 +1,15 @@
 pub mod message;
 
+
+#[cfg(target_os = "windows")]
+pub mod windows;
```
CBenoit, line 4: issue: It’s problematic to conditionally compile Rust code while still generating the bindings for functions that may or may not exist. Let’s just check dynamically and return an error dynamically when the system is not supported. Do not try to add non-diplomat cfgs in Diplomat bindings.

ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/Devolutions.IronRdp.AvaloniaExample.csproj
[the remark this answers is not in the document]
CBenoit: I’ll give it a try tomorrow morning then!

ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/Devolutions.IronRdp.AvaloniaExample.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net8.0;</TargetFrameworks>
```
CBenoit, line 4:
```suggestion
    <TargetFramework>net8.0;</TargetFramework>
```

CBenoit (changes requested):
issue: It seems like you forgot to commit the new file. I still have a concern about the windows handle API so I’m blocking the PR for now.

ffi/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![allow(clippy::unnecessary_box_returns)] // Diplomat requires returning Boxed types
 #![allow(clippy::should_implement_trait)] // Implementing extra traits is not useful for FFI
+#![allow(clippy::needless_lifetimes)]
```
CBenoit, line 3: nitpick: Add a comment to explain why the lint is disabled.

irvingoujAtDevolution:
Yes, give me one second, I am resolving these issues

ffi/src/session/mod.rs
```diff
@@ -66,6 +67,27 @@ pub mod ffi {
...
+
+        pub fn get_svc_processor_cliprdr<'a>(
+            &'a mut self,
+        ) -> Result<Option<Box<CliprdrReference<'a>>>, Box<IronRdpError>> {
+            let cliprdr_ref = self.0.get_svc_processor::<ironrdp::cliprdr::CliprdrClient>();
+            cliprdr_ref
+                .map(|cliprdr| Box::new(CliprdrReference(cliprdr)))
+                .map(Some)
+                .map_or(Ok(None), Ok)
+        }
```
CBenoit, line 79: issue: Unfortunately, the `CliprdrReference` abstraction is problematic for other reasons… I suggest its removal. We have very high risks of violating the aliasing model of Rust: we have a shared reference and at the same time it’s possible to mutate the ActiveStage structure with one of the `&mut self` functions.

I suggest to remove both `get_svc_processor_cliprdr` and `process_svc_processor_message_cliprdr`. Remove all the related boilerplate. Add `initiate_clipboard_copy`, `initiate_clipboard_paste` and `submit_clipboard_format_data` directly on `ActiveStage`. Process the message immediately and make the functions return `VecU8` directly.

This will be much safer to use, and requires less code.

irvingoujAtDevolution: Will update 👍

ffi/src/lib.rs
irvingoujAtDevolution: added

ffi/dotnet/Devolutions.IronRdp.AvaloniaExample/Devolutions.IronRdp.AvaloniaExample.csproj
irvingoujAtDevolution: updated

ffi/src/clipboard/mod.rs
irvingoujAtDevolution: updated

ffi/src/clipboard/mod.rs
irvingoujAtDevolution: Updated, we now hide the reference behind the active stage.

ffi/src/clipboard/mod.rs
irvingoujAtDevolution: updated, now checks are dynamic. Generated bindings and build output interfaces are matched.

ffi/src/session/mod.rs
irvingoujAtDevolution: updated

irvingoujAtDevolution:
Update Summary:

1. No conditional compilation inside FFI modules
2. All incorrect invocation error is throwed/returned at runtime.

ffi/src/clipboard/mod.rs
```diff
@@ -73,40 +21,40 @@ pub mod ffi {
...
+    // #[diplomat::opaque]
+    // pub struct CliprdrReference<'a>(pub &'a ironrdp::cliprdr::Cliprdr<Client>);
+
+    // impl CliprdrReference<'_> {
+    //     pub fn initiate_copy(
+    //         &self,
+    //         formats: &ClipboardFormatIterator,
+    //     ) -> Result<Box<ClipboardSvgMessage>, Box<IronRdpError>> {
+    //         let result = self.0.initiate_copy(&formats.0)?;
+    //         let message = ClipboardSvgMessage(Some(result));
```
pacmancoder, line 33: leftover commented code?

ffi/src/clipboard/windows.rs
```diff
@@ -0,0 +1,134 @@
...
+    use crate::error::ffi::IronRdpError;
+    use crate::error::{PointWidth, WrongPointWidthError};
+
+    use super::WinCliprdrInner;
+
+    #[diplomat::opaque]
+    pub struct WinCliprdr(WinCliprdrInner);
+
+    impl WinCliprdr {
+        pub fn new_32bit(hwnd: u32) -> Result<Box<WinCliprdr>, Box<IronRdpError>> {
```
pacmancoder, line 34: could we keep passing `hwnd` as `isize` here instead of two separate methods for 32/64 bits?

ffi/src/clipboard/mod.rs
irvingoujAtDevolution: updated

ffi/src/clipboard/windows.rs
irvingoujAtDevolution: it actually can be isize ! I checked C# documentation and diplomat source code, yes, we don't have to worry about the pointer  width anymore.

irvingoujAtDevolution: updated

ffi/src/clipboard/windows.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        let (sender, receiver) = std::sync::mpsc::channel();
+
+        let proxy = crate::clipboard::FfiClipbarodMessageProxy { sender };
+
+        // SAFETY: `hwnd` must be a valid window handle
+        let clipboard = unsafe { ironrdp_cliprdr_native::WinClipboard::new(HWND(hwnd), proxy) }?;
+
+        Ok(WinCliprdrInner { clipboard, receiver })
+    }
```
CBenoit, line 105: issue (non-blocking): The `SAFETY:` comment above the unsafe usage should explain why the preconditions are upheld. This comment is explaining what is the precondition, but there is no guarantee it is upheld as the value of `hwnd` is arbitrary. In this case, the function should be made unsafe and the comment be part of the safety section documenting the function. If Diplomat does not allow the function to be marked unsafe, let’s at least keep the comment at the right place.

irvingoujAtDevolution: Added comments, I get errors from compiler with unsafe function in FFI block.

PR #464, opened 2024-05-21
feat(ffi): add graceful shutdown API

irvingoujAtDevolution:
Allow ffi active stage to gracefully shut down.

1. Added FFI function
2. Updated example


CBenoit (approved):
LGTM!

PR #465, opened 2024-05-22
feat(ffi): deployable Devolutions.IronRdp package

CBenoit:


ffi/dotnet/Devolutions.IronRdp/Devolutions.IronRdp.Build.props
```diff
@@ -0,0 +1,108 @@
+<Project>
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;Xamarin.iOS10</TargetFrameworks>
```
thenextman, line 3: I would drop the Xamarin target at this point

ffi/dotnet/Devolutions.IronRdp/Devolutions.IronRdp.csproj
```diff
@@ -1,28 +1,45 @@
...
+    <NativeLibPath_ios_x64>$(RuntimesPath)/ios-x64/native/libDevolutionsIronRdp.dylib</NativeLibPath_ios_x64>
+    <NativeLibPath_ios_arm64>$(RuntimesPath)/ios-arm64/native/libDevolutionsIronRdp.dylib</NativeLibPath_ios_arm64>
+    <NativeLibPath_ios_universal>$(RuntimesPath)/ios-universal/native/libDevolutionsIronRdp.dylib</NativeLibPath_ios_universal>
+    <NativeLibPath_ios_framework>$(RuntimesPath)/ios-universal/native/libDevolutionsIronRdp.framework</NativeLibPath_ios_framework>
   </PropertyGroup>
 
+  <Import Condition="!$(PackageId.EndsWith('iOS'))" Project="Devolutions.IronRdp.Build.props "/>
+  <Import Project="Devolutions.IronRdp.Build.iOS.props "/>
+
+  <!--
```
thenextman, line 37: Seems like this is an artifact (it belongs, and exists in, the .targets file)

thenextman:
I'm not sure the separate iOS package is needed, what do you think?

It's only relevant for cases where we need an iOS specific attribute to decorate callbacks from native to managed code. I don't see any of that in there. 

It would simplify if we removed it, we could add it back at a later time if needed.

I'm also happy to just leave it as-is 🤷‍♂️

ffi/dotnet/Devolutions.IronRdp/Devolutions.IronRdp.Build.props
CBenoit: Sounds good!

ffi/dotnet/Devolutions.IronRdp/Devolutions.IronRdp.csproj
CBenoit: I’ll remove it!

thenextman (approved):

PR #471, opened 2024-06-05
fix(web): fix wrong type assertion caused error

irvingoujAtDevolution:
The value should be typed IronRdpError instead. We perhaps need to publish a small new patch for this.

DPS-10099

### DIFF
--- a/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
+++ b/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
@@ -193,11 +193,11 @@ export class WasmBridgeService {
                             });
                             return of(err);
                         }),
-                        map((termination_info: SessionTerminationInfo) => {
+                        map((termination_info: IronRdpError) => {
                             this.setVisibility(false);
                             this.raiseSessionEvent({
                                 type: SessionEventType.TERMINATED,
-                                data: 'Session was terminated: ' + termination_info.reason() + '.',
+                                data: 'Session was terminated: ' + termination_info.backtrace() + '.',
                             });
                         }),
                     )
--- a/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
+++ b/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
@@ -191,13 +191,13 @@ export class WasmBridgeService {
                                 type: SessionEventType.TERMINATED,
                                 data: 'Session was terminated.',
                             });
-                            return of(err);
+                            throw err;
                         }),
-                        map((termination_info: IronRdpError) => {
+                        map((termination_info: SessionTerminationInfo) => {
                             this.setVisibility(false);
                             this.raiseSessionEvent({
                                 type: SessionEventType.TERMINATED,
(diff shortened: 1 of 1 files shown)


CBenoit (changes requested):
I’m surprised by this. Isn’t that supposed to be the value returned on success? It is supposed to be `IronRdpError` in the `catchError`, and `SessionTerminationInfo` in the `map`…

irvingoujAtDevolution:
> I’m surprised by this. Isn’t that supposed to be the value returned on success? It is supposed to be `IronRdpError` in the `catchError`, and `SessionTerminationInfo` in the `map`…

I think the logic goes like this:
``` rust
pub async fn run(&self) -> Result<SessionTerminationInfo, IronRdpError>
```
Rust async function returns a result, then wasm-bindgen maps it to `Promise<SessionTerminationInfo>` with the possibility to reject it with type `IronRdpError`

`from(session.run())` convert it to an observable, when returns/resolves to `SessionTerminationInfo`, it emits value like regular observable, but when reject/throw error, it emits error and caught by `catchError` operator. 

At the end, catchError returns `of(err)`, which passes down to map function.

You are right, after written this down, I realize it is possible for map to receive two values, it should be either `IronRdpError` or `SessionTerminationInfo`

CBenoit:
@irvingoujAtDevolution Yes, I think the code was originally written assuming only the `Ok` value would be injected into the `map` closure. We need two distinct code paths: success or error, and the error should not be injected into the success code path. That would be the correct fix I think.

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -191,7 +191,7 @@ export class WasmBridgeService {
                                 type: SessionEventType.TERMINATED,
                                 data: 'Session was terminated.',
                             });
-                            return of(err);
+                            throw err;
```
CBenoit, line 194: I read this code again, and I’m wondering if we need to re-throw the error at all.
```suggestion
```

irvingoujAtDevolution: we need, as the signature requires either to return an observable or break the control flow with throw.
`export declare function catchError<T, O extends ObservableInput<any>>(selector: (err: any, caught: Observable<T>) => O): OperatorFunction<T, T | ObservedValueOf<O>>;`

irvingoujAtDevolution: alternative is to use `tap` operator instead.

irvingoujAtDevolution: we may return a observable with nothing, like the `EMPTY` observable constant. which is practically the same in this case but throw an error really align better with the semantics RxJS provides.

CBenoit: Sounds good

CBenoit (approved):
Thank you for fixing this!

irvingoujAtDevolution:
@CBenoit Let me know when there's a new build, I can update the version in DVLS

PR #472, opened 2024-06-07
feat(ffi): enable dynamic resize

irvingoujAtDevolution:


ffi/Cargo.toml
```diff
@@ -19,7 +19,7 @@ doctest = false
 [dependencies]
 diplomat = "0.7.0"
 diplomat-runtime = "0.7.0"
-ironrdp = { workspace = true, features = ["connector", "dvc", "svc", "rdpdr", "rdpsnd", "graphics", "input", "cliprdr"] }
+ironrdp = { workspace = true, features = ["connector", "dvc", "svc", "rdpdr", "rdpsnd", "graphics", "input", "cliprdr","displaycontrol"] }
```
CBenoit, line 22:
```suggestion
ironrdp = { workspace = true, features = ["connector", "dvc", "svc", "rdpdr", "rdpsnd", "graphics", "input", "cliprdr", "displaycontrol"] }
```

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
```diff
@@ -23,17 +23,38 @@ public static class Connection
...
     }
+    
+    public static async Task<Written> SingleSequenceStepRead<TStream>(Framed<TStream> frame, ConnectionActivationSequence sequence, WriteBuf buf)
+    where TStream: Stream
+    {
+        buf.Clear();
+
+        var pudHint = sequence.NextPduHint();
+        
+        // Don't remove, NextPduHint is generated and can return null
```
CBenoit, line 46: question: What do you mean by "NextPduHint is generated"?

irvingoujAtDevolution: The function NextPduHint() is generated, the function signature suggest that pudHint cannot be null but the actual implementation does return null. It is generated signifies that the function signature is misleading. Rider will read the signature and produce warning about pdduHint cannot be null.

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
```diff
@@ -23,17 +23,38 @@ public static class Connection
...
+        {
+
+            var pdu = await frame.ReadByHint(pudHint);
+
+            return sequence.Step(pdu, buf);
+        }
+
+
+        return sequence.StepNoInput(buf);
+    }
```
CBenoit, line 57: thought: Is this code duplicated? We may be able to avoid duplication by defining a C# interface.

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
```diff
@@ -218,7 +239,7 @@ static async Task<NetworkStream> CreateTcpConnection(String servername, int port
         }
         catch (FormatException)
         {
-            IPHostEntry ipHostInfo = await Dns.GetHostEntryAsync(servername);
+            IPHostEntry ipHostInfo = await Dns.GetHostEntryAsync(servername).WaitAsync(TimeSpan.FromSeconds(5));
```
CBenoit, line 242: question: Any specific reason for the 5 seconds timeout? We generally use 10 seconds elsewhere.

ffi/src/utils/mod.rs
```diff
@@ -79,4 +79,21 @@ pub mod ffi {
...
+        }
+
+        pub fn get_first(&self) -> u32 {
+            self.0 .0
+        }
+
+        pub fn get_second(&self) -> u32 {
+            self.0 .1
+        }
+    }
```
CBenoit, line 98: question: I don’t see where this is actually used. Can you confirm this is not a leftover, please?

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
irvingoujAtDevolution: Yeah, but the sequences are generated and does not share interfaces. We certainly could use reflection but that would be harder to maintain and debug in the future.

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
irvingoujAtDevolution: updatedd

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
CBenoit: I see, thank you! The comment should be modified to clarify what is being meant.

```
// FIXME: The NextPduHint() function signature is incorrectly generated: the return value is nullable, so this check is necessary.

ffi/src/utils/mod.rs
irvingoujAtDevolution: removed

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM! Good job!

ffi/dotnet/Devolutions.IronRdp/src/Connection.cs
CBenoit: See https://github.com/Devolutions/IronRDP/pull/475

PR #473, opened 2024-06-10
fix(pdu): fix preamble_size in LicensingErrorMessage regression

elmarco:
According to
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/73170ca2-5f82-4a2d-9d1b-b439f3d8dadc

wMsgSize is "the size in bytes of the licensing packet (including the size of the preamble)". It must thus exclude BASIC_SECURITY_HEADER_SIZE.

Fixes mstsc connection to IronRDP server/acceptor.

Fixes 5c42ade597fa442 ("fix: make license parsing and protocol more resilient (#436)")


CBenoit (approved):
Good catch, thank you!

PR #482, opened 2024-06-25
Handle the set keyboard indicators PDU

zmb3:
This message doesn't require a response of any kind, but handling it here will prevent an unknown PDU error which kills the session.

### DIFF
--- a/crates/ironrdp-session/src/x224/mod.rs
+++ b/crates/ironrdp-session/src/x224/mod.rs
@@ -106,6 +106,7 @@ impl Processor {
                         debug!("Got Session Save Info PDU: {session_info:?}");
                         Ok(Vec::new())
                     }
+                    // FIXME: workaround fix to not terminate the session on "unhandled PDU: Set Keyboard Indicators PDU"
                     ShareDataPdu::SetKeyboardIndicators(data) => {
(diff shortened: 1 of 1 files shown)


crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -106,6 +106,10 @@ impl Processor {
                         debug!("Got Session Save Info PDU: {session_info:?}");
                         Ok(Vec::new())
                     }
+                    ShareDataPdu::SetKeyboardIndicators(data) => {
+                        debug!("Got Keyboard Indicators PDU: {data:?}");
+                        Ok(Vec::new())
+                    }
```
CBenoit, line 112:
```suggestion
                    // FIXME: workaround fix to not terminate the session on "unhandled PDU: Set Keyboard Indicators PDU"
                    ShareDataPdu::SetKeyboardIndicators(data) => {
                        debug!("Got Keyboard Indicators PDU: {data:?}");
                        Ok(Vec::new())
                    }
```

CBenoit (approved):
I think we want a more general fix to this problem (which could happen for many other messages we didn’t consider), so I added a FIXME comment to not forget considering this

Issue #180, opened 2023-08-22
Scancodes for native `ironrdp-client` incorrect on MacOS


ibeckermayer:
MacOS uses [different scancodes](https://eastmanreference.com/complete-list-of-applescript-key-codes) from [windows](https://superuser.com/a/1454198) and thus the native client requires a translation layer for MacOS.

CBenoit:
I’m thinking about a `HashMap` mapping from macOS to Windows scancodes.
@thenextman any input on this?

thenextman:
@CBenoit IIRC FreeRDP does a two-step mapping: they go from "system" key code to "virtual" key code, and from "virtual" key codes to scan codes. They should already have the tables etc. Check the [input](https://github.com/FreeRDP/FreeRDP/tree/master/winpr/libwinpr/input) module in winpr to start.

CBenoit:
Thank you for the pointer!

PR #486, opened 2024-07-02
fix: scancodes for native ironrdp-client incorrect on MacOS

zmrush:
Fixes https://github.com/Devolutions/IronRDP/issues/180

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,3 +154,18 @@ jobs:
 
       - name: Lock files
         run: cargo xtask check locks -v
+
+  success:
+    name: Success
+    runs-on: ubuntu-latest
+    if: ${{ success() }}
+    needs:
+      - formatting
+      - typos
+      - checks
+      - fuzz
+      - web
+
+    steps:
+      - name: CI succeeded
+        run: exit 0


zmrush:
@elmarco @sandeepbansal @CBenoit can u review this pr for bug fix?

CBenoit (changes requested):
Thank you for attempting to fix this!
Unfortunately, using virtual keycodes is not the right approach here.
In this case, the QWERTY layout is assumed, but everything breaks as soon as you use something else.
There are also other shortcomings related to special characters.

Actually, I was working on a fix in this branch: https://github.com/Devolutions/IronRDP/tree/scancode-mapping

In essence, what we need is a big `match` for mapping Apple scancodes to Windows scancodes.

thenextman:
Also to note that even for Windows, you wouldn't use a static map of VKCode to scancode. `MapVirtualKey` and `MapVirtualKeyEx` achieve this according to the current keyboard layout.

zmrush:
@CBenoit  i am sorry for reply u so late. I am very interesting for this. And I am working on solving this according to your advice, based on freerdp and I almost finished it. In addition,  i am also doing work for file uploading/downloading,  and i test a lot, and i think i almost finish it. I expect i will give the pr, next week maybe. thank u so much.

PR #487, opened 2024-07-02
Allow smart card logins with CredSSP enabled

probakowski:
This PR makes it possible to log into AD joined machine with CredSSP/NLA enabled using smart card credentials.

Depends on https://github.com/Devolutions/sspi-rs/pull/259 that's why it points to my fork for the time being


crates/ironrdp-connector/src/lib.rs
```diff
@@ -48,7 +49,7 @@ pub struct BitmapConfig {
 #[derive(Debug, Clone)]
 pub enum Credentials {
     UsernamePassword { username: String, password: String },
-    SmartCard { pin: String },
+    SmartCard { pin: String, config: Option<SmartCardIdentity> },
```
CBenoit, line 52: issue: This requires types from another crate into the public API of `ironrdp-connector`.

suggestion: Re-export the required types from this module using `pub use` or expose a wrapper struct that we internally convert into the sspi type later.

CBenoit (commented):
Thank you for this!

crates/ironrdp-connector/src/lib.rs
probakowski: I created wrapper structure, seems to be cleaner this way, `cargo xtask ci` is green locally for me

Cargo.toml
```diff
@@ -58,7 +58,7 @@ expect-test = "1"
 png = "0.17"
 proptest = "1.4"
 rstest = "0.18"
-sspi = "0.11"
+sspi = { git = "https://github.com/Devolutions/sspi-rs", rev = "7dc730d9d5269a356531810fd38b4a86d55881f8" }
```
CBenoit, line 61: We’ll wait until sspi 0.12 is released before merging this PR

probakowski: Makes sense, any timeline for that?

probakowski:
Closing in favor of https://github.com/Devolutions/IronRDP/pull/513

PR #495, opened 2024-07-17
feat(web)!: use JS primitives instead of RxJS Observable in public API

irvingoujAtDevolution:


CBenoit:
Do we need to keep rxjs as a dependency in our package.json?

web-client/iron-remote-gui/src/services/PublicAPI.ts
```diff
@@ -24,9 +23,9 @@ export class PublicAPI {
         desktopSize?: DesktopSize,
         preConnectionBlob?: string,
         kdc_proxy_url?: string,
-    ): Observable<NewSessionInfo> {
+    ): Promise<NewSessionInfo> {
         loggingService.info('Initializing connection.');
-        return this.wasmService.connect(
+        const resultObservable = this.wasmService.connect(
```
CBenoit, line 28: suggestion: Is it possible to make the connect function return a Promise as well? Let’s just get rid of RxJS completely, I don’t think we are really benefiting a lot from it.

irvingoujAtDevolution: I think we actually benefiting from it, at least for the connect function. This is clean and well-structured, easier to read compare to procedural TypeScript.
``` typescript
        return from(sessionBuilder.connect()).pipe(
            catchError((err: IronRdpError) => {
                this.raiseSessionEvent({
                    type: SessionEventType.ERROR,
                    data: {
                        backtrace: () => err.backtrace(),
                        kind: () => err.kind() as number as UserIronRdpErrorKind,
                    },
                });
                return of(err);
            }),
            filter(isSession),
            map((session: Session) => {
                from(session.run())
                    .pipe(
                        catchError((err) => {
                            this.setVisibility(false);
                            this.raiseSessionEvent({
                                type: SessionEventType.ERROR,
                                data: err.backtrace(),
                            });
                            this.raiseSessionEvent({
                                data: 'Session was terminated.',
                            });
                            throw err;
                        }),
                        map((termination_info: SessionTerminationInfo) => {
                            this.setVisibility(false);
                            this.raiseSessionEvent({
                                data: 'Session was terminated: ' + termination_info.reason() + '.',
                            });
                        }),
                    )
                    .subscribe();
                return session;
            }),
            map((session: Session) => {
                loggingService.info('Session started.');
                this.session = session;
                this._resize.next({
                    desktop_size: session.desktop_size(),
                    session_id: 0,
                });
                this.raiseSessionEvent({
                    type: SessionEventType.STARTED,
                    data: 'Session started',
                });
                return {
                    session_id: 0,
                    initial_desktop_size: session.desktop_size(),
                    websocket_port: 0,
                };
            }),
        );
```

CBenoit: I’m not really convinced we need `Observer`s given that all functions provided by the WASM module are simply returning `Promise`s / `Future`s. We are using an abstraction for returning many values, but we ever only return one single value. I believe we could just use `Promise` combinators when appropriate, and it wouldn’t change the style too much. I wouldn’t be surprised we could even improve over this actually.

CBenoit (approved):

PR #496, opened 2024-07-18
map macos scan code to windows scancode

zmrush:
map macos scancode to windows', as mentioned [Scancodes Mapping](https://github.com/Devolutions/IronRDP/issues/180).
@CBenoit  can u check this? only  macos/EVDEV  are supported for now.

### DIFF
--- a/crates/ironrdp-input/Cargo.toml
+++ b/crates/ironrdp-input/Cargo.toml
@@ -16,6 +16,6 @@ doctest = false
 test = false
 
 [dependencies]
-ironrdp-pdu.workspace = true
+ironrdp-pdu = { workspace = true, features = ["std"] }
 bitvec = "1.0"
 smallvec = "1.13"
--- a/crates/ironrdp-input/src/keycode.rs
+++ b/crates/ironrdp-input/src/keycode.rs
@@ -1,0 +1,1236 @@
+/**
+ * Key Flags
+ */
+
+pub(crate) const KBDEXT: u16 = 0x0100;
+/*
+ * Virtual Key Codes (Windows):
+ * http://msdn.microsoft.com/en-us/library/windows/desktop/dd375731/
+ * http://msdn.microsoft.com/en-us/library/ms927178.aspx
+ */
+
+/* Mouse buttons */
+pub(crate) const VK_LBUTTON: u16 = 0x01; /* Left mouse button */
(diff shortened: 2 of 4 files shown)


zmrush:
@thenextman cc.

CBenoit:
Hi! Sorry for the delay here. I’ll be busy for a while, but I think I’ll be able to look into it on Friday or next Monday!

Michael9127:
Im also looking forward to this getting merged! Would it be possible for this to get reviewed soon?

crates/ironrdp-input/Cargo.toml
```diff
@@ -16,6 +16,6 @@ doctest = false
 test = false
 
 [dependencies]
-ironrdp-pdu.workspace = true
+ironrdp-pdu = { workspace = true, features = ["std"] }
```
CBenoit, line 19: It seems that it’s not necessary to enable this feature here.

CBenoit (commented):
I still need to review more in depth, but it seems the PR is not complete. The mapping logic is not actually used anywhere. I think this is fine to do the integration in a separate PR if that is the intention though.
Overall things are looking reasonable, but some cleanup is required as some warnings are suppressed without a comment explaining why.

glamberson:
@zmrush Circling back on this one. The automated review has flagged the same set of issues twice now, four days apart, so I want to name them plainly rather than leave this as a vague rebase request. The keyboard-type branch is inverted, so the IbmEnhanced case, which is the overwhelming majority of real keyboards, selects the wrong table. The extended-key flag uses the wrong constant, so it gets silently dropped during composition, which is why Left Arrow ends up mapping to Numpad 4. Right Control maps to the Windows key by mistake. The platform selection uses a compile-time `cfg(target_os)` check for something that needs to be a runtime property, and that breaks the web client entirely. There is also no consumer anywhere in the tree for any of this code, and no tests.

This is a real and useful thing to have mapped correctly, issue #180 has been open a long time. But this PR needs the branch inversion, the extended-key constant, and the RightControl mapping fixed, a real caller wired up, the platform check made runtime instead of compile-time, and at least basic test coverage before it is close to mergeable. If you are still interested in taking this on, I am glad to look again once those are addressed.

CBenoit:
Thank you for the investigation. Closing in favor of https://github.com/Devolutions/IronRDP/pull/1929

Issue #110, opened 2023-05-01
Support for file upload


CBenoit:
[TODO comment in code](https://github.com/Devolutions/IronRDP/blob/ccdd3825b08c64d7736dd3207c3d7281d0445b30/crates/client/src/gui.rs#LL72C1-L72C1)

CBenoit:
FYI @ibeckermayer I assigned you since you are working on it

PR #497, opened 2024-07-18
feat(rdpdr): simple file backend for file upload/download

zmrush:
As mentiond in [file upload support](https://github.com/Devolutions/IronRDP/issues/110). i implemented a simple file backend,which supports file upload/download. @CBenoit @ibeckermayer

### DIFF
--- a/crates/ironrdp-rdpdr-native/Cargo.toml
+++ b/crates/ironrdp-rdpdr-native/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "ironrdp-rdpdr-native"
+version = "0.1.0"
+readme = "README.md"
+description = "Native RDPDR static channel backend implementations for IronRDP"
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+authors.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[lib]
+doctest = false
+test = false
+
+[dependencies]
+tracing.workspace = true
+
+[target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
+ironrdp-pdu.workspace = true
+ironrdp-svc.workspace = true
+ironrdp-rdpdr.workspace = true
+nix = { version = "0.29", features = ["fs"] }
+libc = "0.2"
--- a/crates/ironrdp-rdpdr-native/Cargo.toml
+++ b/crates/ironrdp-rdpdr-native/Cargo.toml
@@ -15,8 +15,6 @@ categories.workspace = true
 doctest = false
 test = false
 
-[dependencies]
-tracing.workspace = true
 
 [target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
 ironrdp-pdu.workspace = true
@@ -24,3 +22,4 @@ ironrdp-svc.workspace = true
 ironrdp-rdpdr.workspace = true
 nix = { version = "0.29", features = ["fs"] }
 libc = "0.2"
+tracing.workspace = true
--- a/crates/ironrdp-rdpdr-native/Cargo.toml
+++ b/crates/ironrdp-rdpdr-native/Cargo.toml
@@ -20,6 +20,5 @@ test = false
 ironrdp-pdu.workspace = true
 ironrdp-svc.workspace = true
 ironrdp-rdpdr.workspace = true
-nix = { version = "0.29", features = ["fs"] }
-libc = "0.2"
+nix = { version = "0.29", features = ["fs", "dir"] }
 tracing.workspace = true
--- a/crates/ironrdp-rdpdr-native/README.md
+++ b/crates/ironrdp-rdpdr-native/README.md
@@ -1,0 +1,3 @@
+# IronRDP RDPDR native backends
+
+Native RDPDR backend implementations. Currently only *nix systems are supported.
--- a/crates/ironrdp-rdpdr-native/src/lib.rs
+++ b/crates/ironrdp-rdpdr-native/src/lib.rs
@@ -1,0 +1,20 @@
+#![doc = include_str!("../README.md")]
+#![warn(unsafe_op_in_unsafe_fn)]
+#![warn(invalid_reference_casting)]
+#![warn(pointer_structural_match)]
+#![warn(clippy::undocumented_unsafe_blocks)]
+#![warn(clippy::multiple_unsafe_ops_per_block)]
+#![warn(clippy::transmute_ptr_to_ptr)]
+#![warn(clippy::as_ptr_cast_mut)]
+#![warn(clippy::cast_ptr_alignment)]
+#![warn(clippy::fn_to_numeric_cast_any)]
+#![warn(clippy::ptr_cast_constness)]
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+#[macro_use]
+extern crate tracing;
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+mod nix;
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+pub use nix::simple;
--- a/crates/ironrdp-rdpdr-native/src/lib.rs
+++ b/crates/ironrdp-rdpdr-native/src/lib.rs
@@ -17,4 +17,4 @@ extern crate tracing;
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 mod nix;
 #[cfg(any(target_os = "macos", target_os = "linux"))]
-pub use nix::simple;
+pub use nix::backend;
--- a/crates/ironrdp-rdpdr-native/src/nix/mod.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/mod.rs
@@ -1,0 +1,1 @@
+pub mod simple;
--- a/crates/ironrdp-rdpdr-native/src/nix/mod.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/mod.rs
@@ -1,1 +1,1 @@
-pub mod simple;
+pub mod backend;
--- a/crates/ironrdp-rdpdr-native/src/nix/simple.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/simple.rs
@@ -412,13 +412,15 @@ pub(crate) fn set_information(
                 FileInformationClass::EndOfFile(info) => {
                     let mut set_end_res = -1;
                     // SAFETY: it is safe to call on *nix platform
+                    // SAFETY: `file` is a regular file that we opened with write access.
                     unsafe {
                         if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
                             set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
                         }
                     }
                     if set_end_res < 0 {
-                        warn!("set file end error, code is:{}", set_end_res);
+                        let error = nix::errno::Errno::last(); // However, only call this if `ftruncate` was called.
+                        warn!(%error, "Failed to set end of file");
                         let res = RdpdrPdu::ClientDriveSetInformationResponse(ClientDriveSetInformationResponse::new(
                             &req_inner,
                             NtStatus::UNSUCCESSFUL,
--- a/crates/ironrdp-rdpdr-native/src/nix/simple.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/simple.rs
@@ -410,24 +410,24 @@ pub(crate) fn set_information(
                     }
                 }
                 FileInformationClass::EndOfFile(info) => {
-                    // SAFETY: the file must has been opend with write access in the last steps, since rdp prepares to set information. In addition it is a regular file.
-                    unsafe {
-                        if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
-                            if nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file) < 0 {
-                                let error = nix::errno::Errno::last();
-                                warn!(%error, "Failed to set end of file");
-                                let res = RdpdrPdu::ClientDriveSetInformationResponse(
-                                    ClientDriveSetInformationResponse::new(&req_inner, NtStatus::UNSUCCESSFUL)?,
-                                );
-                                return Ok(vec![SvcMessage::from(res)]);
-                            }
-                        } else {
-                            warn!("no such file");
+                    if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
+                        // SAFETY: the file must has been opend with write access in the last steps, since rdp prepares to set information. In addition it is a regular file.
+                        let set_end_res = unsafe { nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file) };
+                        if set_end_res < 0 {
+                            let error = nix::errno::Errno::last();
+                            warn!(%error, "Failed to set end of file");
                             let res = RdpdrPdu::ClientDriveSetInformationResponse(
-                                ClientDriveSetInformationResponse::new(&req_inner, NtStatus::NO_SUCH_FILE)?,
+                                ClientDriveSetInformationResponse::new(&req_inner, NtStatus::UNSUCCESSFUL)?,
                             );
                             return Ok(vec![SvcMessage::from(res)]);
                         }
+                    } else {
+                        warn!("no such file");
+                        let res = RdpdrPdu::ClientDriveSetInformationResponse(ClientDriveSetInformationResponse::new(
+                            &req_inner,
+                            NtStatus::NO_SUCH_FILE,
+                        )?);
+                        return Ok(vec![SvcMessage::from(res)]);
                     }
                 }
                 _ => {
--- a/crates/ironrdp-rdpdr-native/src/nix/simple.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/simple.rs
@@ -1,33 +1,33 @@
 use ironrdp_pdu::PduResult;
-use ironrdp_svc::{impl_as_any, SvcMessage};
-use libc::ftruncate;
-use nix::libc::{closedir, opendir, readdir, DIR};
-
 use ironrdp_rdpdr::pdu::efs::*;
 use ironrdp_rdpdr::pdu::esc::{ScardCall, ScardIoCtlCode};
 use ironrdp_rdpdr::pdu::RdpdrPdu;
 use ironrdp_rdpdr::RdpdrBackend;
-use std::ffi::{CStr, CString};
+use ironrdp_svc::{impl_as_any, SvcMessage};
+use nix::dir::{Dir, OwningIter};
+use std::ffi::CString;
 use std::io::Read;
 use std::io::{Seek, SeekFrom, Write};
 use std::os::fd::{AsFd, AsRawFd};
 use std::os::unix::fs::MetadataExt;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct SimpleRdpdrBackend {
-    pub file_id: u32,
-    pub file_base: String,
-    pub file_map: std::collections::HashMap<u32, std::fs::File>,
-    pub file_path_map: std::collections::HashMap<u32, String>,
-    pub file_dir_map: std::collections::HashMap<u32, DirBox>,
+    file_id: u32,
+    file_base: String,
+    file_map: std::collections::HashMap<u32, std::fs::File>,
+    file_path_map: std::collections::HashMap<u32, String>,
+    file_dir_map: std::collections::HashMap<u32, OwningIter>,
 }
 
-#[derive(Debug)]
-pub struct DirBox {
-    pub dir: *mut DIR,
+impl SimpleRdpdrBackend {
+    pub fn new(file_base: String) -> Self {
+        Self {
+            file_base,
+            ..Default::default()
+        }
+    }
 }
-// SAFETY: it is safe to send *DIR to a different thread, and it is only accessed by one thread at one time.
-unsafe impl Send for DirBox {}
 
 impl_as_any!(SimpleRdpdrBackend);
 
@@ -123,6 +123,7 @@ pub(crate) fn write_device(
         Ok(length)
     }
 }
+
 pub(crate) fn read_device(
     backend: &mut SimpleRdpdrBackend,
     req_inner: DeviceReadRequest,
@@ -165,18 +166,14 @@ pub(crate) fn read_device(
         Ok(buf)
     }
 }
+
 pub(crate) fn close_device(
     backend: &mut SimpleRdpdrBackend,
     req_inner: DeviceCloseRequest,
 ) -> PduResult<Vec<SvcMessage>> {
     backend.file_map.remove(&req_inner.device_io_request.file_id);
     backend.file_path_map.remove(&req_inner.device_io_request.file_id);
-    if let Some(dir_box) = backend.file_dir_map.remove(&req_inner.device_io_request.file_id) {
-        // SAFETY: it is safe to close dir when we close device
-        unsafe {
-            closedir(dir_box.dir);
-        }
-    }
+    backend.file_dir_map.remove(&req_inner.device_io_request.file_id);
     let res = RdpdrPdu::DeviceCloseResponse(DeviceCloseResponse {
         device_io_response: DeviceIoResponse::new(req_inner.device_io_request, NtStatus::SUCCESS),
     });
@@ -379,6 +376,7 @@ pub(crate) fn query_volume_information(
         }
     }
 }
+
 pub(crate) fn set_information(
     backend: &mut SimpleRdpdrBackend,
     req_inner: ServerDriveSetInformationRequest,
@@ -416,7 +414,7 @@ pub(crate) fn set_information(
                     // SAFETY: it is safe to call on *nix platform
                     unsafe {
                         if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
-                            set_end_res = ftruncate(file.as_raw_fd(), info.end_of_file);
+                            set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
                         }
                     }
                     if set_end_res < 0 {
@@ -446,6 +444,7 @@ pub(crate) fn set_information(
         ClientDriveSetInformationResponse::new(&req_inner, NtStatus::SUCCESS)?,
     ))])
 }
+
 // in fact, it is time in secs which is very small
 #[allow(clippy::arithmetic_side_effects)]
 pub(crate) fn transform_to_filetime(time_in_secs: i64) -> i64 {
@@ -453,6 +452,7 @@ pub(crate) fn transform_to_filetime(time_in_secs: i64) -> i64 {
     time += 116444736000000000;
     time
 }
+
 pub(crate) fn get_file_attributes(meta: &std::fs::Metadata, file_name: &str) -> FileAttributes {
     let mut file_attribute = FileAttributes::empty();
     if meta.is_dir() {
@@ -470,6 +470,7 @@ pub(crate) fn get_file_attributes(meta: &std::fs::Metadata, file_name: &str) -> 
     }
     file_attribute
 }
+
 pub(crate) fn make_query_dir_resp(
     find_file_name: Option<String>,
     device_io_request: DeviceIoRequest,
@@ -540,6 +541,7 @@ pub(crate) fn make_query_dir_resp(
         }
     }
 }
+
 pub(crate) fn query_directory(
     backend: &mut SimpleRdpdrBackend,
     req_inner: ServerDriveQueryDirectoryRequest,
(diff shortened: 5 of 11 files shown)


CBenoit (commented):
Thank you for moving this forward!!

Could you please put the backend implementation in a separate crate named `ironrdp-rdpdr-native`?

See for instance [`ironrdp-cliprdr-native`](https://github.com/Devolutions/IronRDP/tree/abc0ee0617fcdcda3f0ace3ffd292251979d1910/crates/ironrdp-cliprdr-native)

Also, try to run `cargo xtask ci` to run the various checks we are enforcing in our CI.

zmrush:
> Thank you for moving this forward!!
> 
> Could you please put the backend implementation in a separate crate named `ironrdp-rdpdr-native`?
> 
> See for instance [`ironrdp-cliprdr-native`](https://github.com/Devolutions/IronRDP/tree/abc0ee0617fcdcda3f0ace3ffd292251979d1910/crates/ironrdp-cliprdr-native)
> 
> Also, try `cargo xtask ci` to run the various checks we are enforcing in our CI.

ok,i will try it. thank u

crates/ironrdp-rdpdr-native/README.md
```diff
@@ -0,0 +1,3 @@
+# IronRDP RDPDR native backends
+
+Native CLIPRDR backend implementations. Currently only *nix systems are supported.
```
CBenoit, line 3:
```suggestion
Native RDPDR backend implementations. Currently only *nix systems are supported.
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+    pub file_base: String,
+    pub file_map: std::collections::HashMap<u32, std::fs::File>,
+    pub file_path_map: std::collections::HashMap<u32, String>,
+    pub file_dir_map: std::collections::HashMap<u32, DirBox>,
+}
+#[derive(Debug)]
+pub struct DirBox {
+    pub dir: *mut DIR,
+}
+unsafe impl Send for DirBox {}
```
CBenoit, line 28: issue: This need to be documented with an `SAFETY` comment.

crates/ironrdp-rdpdr-native/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[cfg(any(target_os = "macos", target_os = "linux"))]
```
CBenoit, line 1: suggestion: Can you add this at the top of the file and fix the new warnings?

```rust
#![doc = include_str!("../README.md")]
#![warn(unsafe_op_in_unsafe_fn)]
#![warn(invalid_reference_casting)]
#![warn(pointer_structural_match)]
#![warn(clippy::undocumented_unsafe_blocks)]
#![warn(clippy::multiple_unsafe_ops_per_block)]
#![warn(clippy::transmute_ptr_to_ptr)]
#![warn(clippy::as_ptr_cast_mut)]
#![warn(clippy::cast_ptr_alignment)]
#![warn(clippy::fn_to_numeric_cast_any)]
#![warn(clippy::ptr_cast_constness)]
```

crates/ironrdp-rdpdr-native/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+test = false
+
+[dependencies]
+ironrdp-error.workspace = true
+ironrdp-pdu.workspace = true
+ironrdp-svc.workspace = true
+ironrdp-rdpdr.workspace = true
+tracing.workspace = true
+bitflags.workspace = true
+nix = { version = "0.29" ,features = ["fs"]}
```
CBenoit, line 25: nitpick:
```suggestion
nix = { version = "0.29", features = ["fs"] }
```

crates/ironrdp-rdpdr-native/src/nix/mod.rs
```diff
@@ -0,0 +1,2 @@
+#[allow(clippy::unnecessary_cast)]
```
CBenoit, line 1: question: Why did you allow this warning? Could you add a comment explaining why directly in the code?

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+
+use ironrdp_rdpdr::pdu::efs::*;
+use ironrdp_rdpdr::pdu::esc::{ScardCall, ScardIoCtlCode};
+use ironrdp_rdpdr::pdu::RdpdrPdu;
+use ironrdp_rdpdr::RdpdrBackend;
+use std::ffi::{CStr, CString};
+use std::io::{Seek, SeekFrom, Write};
+use std::os::fd::{AsFd, AsRawFd};
+use std::os::unix::fs::MetadataExt;
+use tracing::{debug, warn};
```
CBenoit, line 14: suggestion: Put this in lib.rs so we don’t need to `use` the logging macros. Usually I don’t recommend to do that expect for tracing.

```rust
#[macro_use]
extern crate tracing;
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+                    warn!("write len:{} is not equal to {}", length, req_inner.write_data.len());
+                    let res = RdpdrPdu::DeviceWriteResponse(DeviceWriteResponse {
+                        device_io_reply: DeviceIoResponse::new(request, NtStatus::UNSUCCESSFUL),
+                        length: 0_u32,
+                    });
+                    Ok(vec![SvcMessage::from(res)])
+                }
+            }
+            Err(e) => {
+                warn!("write error:{:?}", e);
```
CBenoit, line 110: style: See [this section](https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/STYLE.md#logging)
```suggestion
            Err(error) => {
                warn!(%error, "Write error");
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+}
+pub(crate) fn read_device(
+    backend: &mut SimpleRdpdrBackend,
+    req_inner: DeviceReadRequest,
+) -> PduResult<Vec<SvcMessage>> {
+    let read_inner = |file: &mut std::fs::File, offset, length| -> std::io::Result<Vec<u8>> {
+        let sf = SeekFrom::Start(offset);
+        file.seek(sf)?;
+        let mut buf = vec![0; length];
+        use std::io::Read;
```
CBenoit, line 128: suggestion: I’m all for putting local `use` inside functions, but this should be at the top of the outermost function for clarity.

```rust
fn read_device() {
    use std::io::Read as _;

    fn helper() {
        // Use std::io::Read
    }
}
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+    backend: &mut SimpleRdpdrBackend,
+    req_inner: DeviceWriteRequest,
+) -> PduResult<Vec<SvcMessage>> {
+    let write_inner = |file: &mut std::fs::File, offset, write_data| -> std::io::Result<usize> {
+        let sf = SeekFrom::Start(offset);
+        file.seek(sf)?;
+        let length = file.write(write_data)?;
+        file.flush()?;
+        Ok(length)
+    };
```
CBenoit, line 80: style: If capturing environment is not required, [this is the style we follow](https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/STYLE.md#local-helper-functions)

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+                    let io_response = DeviceIoResponse::new(req_inner.device_io_request, NtStatus::UNSUCCESSFUL);
+                    let res = RdpdrPdu::DeviceCreateResponse(DeviceCreateResponse {
+                        device_io_reply: io_response,
+                        file_id,
+                        information: Information::empty(),
+                    });
+                    return Ok(vec![SvcMessage::from(res)]);
+                }
+                //这个我们先不执行，我们看看能不能统一
+                //我们可以统一到文件打开的条件中去
```
CBenoit, line 667: issue (non-blocking): Use English in the comments, because very few of IronRDP contributor can read Chinese 😆

Something like this? I’ll leave you adjust accordingly.
```suggestion
                // Let's not execute this for now, let's see if we can unify it.
                // This can be unified with the condition for opening the file.
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+                            backend.file_path_map.insert(file_id, path.clone());
+                            return make_create_drive_resp(
+                                req_inner.device_io_request,
+                                req_inner.create_disposition,
+                                file_id,
+                            );
+                        }
+                        Err(e) => {
+                            warn!("open file dir error:{:?}", e);
+                            //return by downside
```
CBenoit, line 700: style: See [this](https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/STYLE.md#inline-code-comments-are-proper-sentences)

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+            });
+            Ok(vec![SvcMessage::from(res)])
+        },
+        |file, request| match write_inner(file, req_inner.offset, &req_inner.write_data) {
+            Ok(length) => {
+                if length == req_inner.write_data.len() {
+                    Ok(vec![SvcMessage::from(RdpdrPdu::DeviceWriteResponse(
+                        DeviceWriteResponse {
+                            device_io_reply: DeviceIoResponse::new(request, NtStatus::SUCCESS),
+                            length: req_inner.write_data.len() as u32,
```
CBenoit, line 97: suggestion: `as` conversions should be avoided when possible. We prefer using `u32::from` or `u32::try_from`. It’s fine to use `u32::try_from(…).unwrap()` in most cases, to at least crash when the conversion is unexpectedly lossy.

rationale: `as` conversions will perform many kinds of conversions, including silently lossy conversions and dangerous coercions.

crates/ironrdp-rdpdr/src/backend/mod.rs
```diff
@@ -7,10 +7,11 @@ use ironrdp_svc::AsAny;
...
 use crate::pdu::efs::{DeviceControlRequest, ServerDeviceAnnounceResponse, ServerDriveIoRequest};
 use crate::pdu::esc::{ScardCall, ScardIoCtlCode};
+use ironrdp_svc::SvcMessage;
 
 /// OS-specific device redirection backend interface.
 pub trait RdpdrBackend: AsAny + fmt::Debug + Send {
     fn handle_server_device_announce_response(&mut self, pdu: ServerDeviceAnnounceResponse) -> PduResult<()>;
     fn handle_scard_call(&mut self, req: DeviceControlRequest<ScardIoCtlCode>, call: ScardCall) -> PduResult<()>;
-    fn handle_drive_io_request(&mut self, req: ServerDriveIoRequest) -> PduResult<()>;
+    fn handle_drive_io_request(&mut self, req: ServerDriveIoRequest) -> PduResult<Vec<SvcMessage>>;
```
CBenoit, line 16: @ibeckermayer This is slightly changing the API for `RdpdrBackend`. Does that look good to you?

CBenoit (commented):
Great work!! Thank you a lot

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,765 @@
...
+    match backend.file_path_map.get(&req_inner.device_io_request.file_id) {
+        Some(parent_pos_for_next) => {
+            let mut find_file_name = None;
+            if req_inner.initial_query > 0 {
+                if req_inner.path.ends_with('*') {
+                    unsafe {
+                        let mut parent = backend.file_base.clone();
+                        let query_path = req_inner.path.replace('\\', "/");
+                        let len = query_path.len();
+                        parent.push_str(&query_path[0..len - 1]);
```
CBenoit, line 538: suggestion: FYI, we are using the `clippy::arithmetic_side_effects` to warn us about potential arithmetic side-effect bugs. Typically, we’ll `allow` it and explain why there is no problem with a comment. Alternatively, it’s possible to use checked arithmetic operations.

For instance:

https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/crates/ironrdp-cliprdr-format/src/bitmap.rs#L497-L501

https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/crates/ironrdp-web/src/session.rs#L548-L550

See also: https://github.com/Devolutions/IronRDP/blob/abc0ee0617fcdcda3f0ace3ffd292251979d1910/STYLE.md#explain-non-obvious-assumptions-by-referencing-the-invariants

CBenoit:
Modified the title of the PR to follow the conventional commits specification: https://www.conventionalcommits.org/en/v1.0.0/

Also, please push new commits when applying review suggestions. It’s easier for me to review incrementally only what is changed. Of course, it’s fine to clean the history using rebase before or after the review.

crates/ironrdp-rdpdr-native/src/lib.rs
zmrush: ok, i add these contents

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, i modify according to our style

crates/ironrdp-rdpdr-native/src/nix/mod.rs
zmrush: because statvfs.blocks_available() return u32/u64 on different platform(for example, on my computer its u32), but volume_serial_number is u32, so i need to use ('as u32' or u32::try_from()),but it emit warning 'unnecessary_cast', so i add this

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok,this seems good. i modify the code

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, i am sorry for forgetting to modify this comments

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, it seems these warning will emit after i add clippy warns. And therefore i modify the code

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, i modify according to our style

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, i modify according to our style

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: yes,it seems reasonable. And i modify it

zmrush:
> Modified the title of the PR to follow the conventional commits specification: https://www.conventionalcommits.org/en/v1.0.0/
> 
> Also, please push new commits when applying review suggestions. It’s easier for me to review incrementally only what is changed. Of course, it’s fine to clean the history using rebase before or after the review.

ok,i modify the the title of the PR. but i do not know how to 'Also, please push new commits when applying review suggestions'. i only reply u on this 'Conversation' page and 'Files Changed' page

crates/ironrdp-rdpdr-native/src/nix/mod.rs
zmrush: ok, i added this explation in the newest commit. But when i change my code from 'as u32' to 'u32::try_from', it emit 'useless_conversion' warning, so i use '#[allow(clippy::useless_conversion)]' in the simple.rs and i add explanation.

CBenoit:
>  but i do not know how to 'Also, please push new commits when applying review suggestions'. i only reply u on this 'Conversation' page and 'Files Changed' page

My suggestion was to create a new commit and push it normally instead of, e.g., amending it and force-push it. That way I can check the new commits instead, it’s much much easier to review.

zmrush:
> > but i do not know how to 'Also, please push new commits when applying review suggestions'. i only reply u on this 'Conversation' page and 'Files Changed' page
> 
> My suggestion was to create a new commit and push it normally instead of, e.g., amending the initial one and force-pushing it each time. That way I can check the new commits instead, it’s much much easier to review.
> 
> When force-pushing, it appears like this in GitHub UI: image
> 
> I can squash all the commits into a single commit when merging, so it’s fine to push incremental review commits.

ok, i am sorry. i got it.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+    pub file_id: u32,
+    pub file_base: String,
+    pub file_map: std::collections::HashMap<u32, std::fs::File>,
+    pub file_path_map: std::collections::HashMap<u32, String>,
+    pub file_dir_map: std::collections::HashMap<u32, DirBox>,
+}
+
+#[derive(Debug)]
+pub struct DirBox {
+    pub dir: *mut DIR,
```
CBenoit, line 27: naming:
```suggestion
    pub ptr: *mut DIR,
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+}
+// SAFETY: it is safe to send *DIR to a different thread, and it is only accessed by one thread at one time.
```
CBenoit, line 29: issue: It’s easy to break this assumption with the current API, the `dir` field is `pub` and the user can do whatever he wants with it:

```rust
let dir1 = DirBox { /* … */ }
let dir2 = DirBox { ptr: dir1.ptr) }
```

At this point, if you send `dir2` to a different thread, we are breaking the thread-safety. I suggest to not leave the field public and encapsulate it better.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+                            if dir_name_in_rs == "." || dir_name_in_rs == ".." {
+                                continue;
+                            }
+                            parent.push_str(&dir_name_in_rs);
+                            find_file_name = Some(parent);
+                            break;
+                        }
+                        backend
+                            .file_dir_map
+                            .insert(req_inner.device_io_request.file_id, DirBox { dir: dirp });
```
CBenoit, line 585: question: What happens when we keep around a DirBox with a null pointer? Is it intended? Shouldn’t we use an `Option<DirBox>` instead? Do we need to keep track of this at all when it failed?

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+                    let len = query_path.len();
+                    // path ends with *, so its len > 0
+                    #[allow(clippy::arithmetic_side_effects)]
+                    parent.push_str(&query_path[0..len - 1]);
+                    let dirp = if let Ok(real_parent_c) = CString::new(parent.clone()) {
+                        // SAFETY: it is safe to call this on *nix platform
+                        unsafe { opendir(real_parent_c.as_ptr()) }
+                    } else {
+                        std::ptr::null::<DIR>().cast_mut()
+                    };
```
CBenoit, line 563: suggestion: I suggest to encapsulate this logic in `BoxDir`, in a `new` function.

```rust
impl BoxDir {
    fn new(path: &Path) -> Result<Self> {
        /* … */    
    }
}
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+                            // SAFETY: it is safe to call readdir on *nix platform
+                            let first = unsafe { readdir(dirp) };
+                            if first.is_null() {
+                                break;
+                            }
+                            // SAFETY: it is safe to call this for rdp
+                            let dir_name = unsafe { &(*first).d_name };
+                            // SAFETY: it is safe to call this for rdp
+                            let dir_name_in_rs =
+                                unsafe { CStr::from_ptr(dir_name as *const i8).to_string_lossy().into_owned() };
```
CBenoit, line 575: question: What do you mean by "for rdp"? This seems unrelated to the RDP protocol.

suggestion: In this case, it’s a FFI structure and we can only assume that the library is well-behaved.
```suggestion
                            // SAFETY: When a non-null value is returned, the pointed struct should be properly initialized by the library.
                            let dir_name = unsafe { &(*first).d_name };

                            // SAFETY: DIR::name field contains the null terminated filename.
                            // etc etc
                            let dir_name_in_rs =
                                unsafe { CStr::from_ptr(dir_name as *const i8).to_string_lossy().into_owned() };
```

Also check the safety section of `CStr::from_ptr`: https://doc.rust-lang.org/std/ffi/struct.CStr.html#safety

suggestion: You can also likely encapsulate this in `BoxDir` directly.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+                make_query_dir_resp(
+                    find_file_name,
+                    req_inner.device_io_request,
+                    req_inner.file_info_class_lvl,
+                    true,
+                )
+            } else {
+                if let Some(dir_box) = backend.file_dir_map.get_mut(&req_inner.device_io_request.file_id) {
+                    // SAFETY: it is safe to call readdir in multiple threads
+                    let next = unsafe { readdir(dir_box.dir) };
```
CBenoit, line 602: question: You probably didn’t mean that. `readdir` is [not thread-safe](https://man7.org/linux/man-pages/man3/readdir.3.html).
image

But you didn’t implement `Sync` on `DirBox`, so we’re not supposed to be sharing references across several threads.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+            let res = RdpdrPdu::DeviceCreateResponse(DeviceCreateResponse {
+                device_io_reply: io_response,
+                file_id,
+                information: Information::empty(),
+            });
+            Ok(vec![SvcMessage::from(res)])
+        }
+    }
+}
+pub(crate) fn process_dependent_file(
```
CBenoit, line 780: style: Use some newlines to separate functions/types definitions.
```suggestion
}

pub(crate) fn process_dependent_file(
```

CBenoit (commented):
`SAFETY` comments should be based on the documentation of the C functions. You can find that at different places, like:

- https://www.gnu.org/software/libc/manual/html_node/Reading_002fClosing-Directory.html
- https://man7.org/linux/man-pages/man3/readdir.3.html

You’ll find important information such as thread-safety, preconditions, postconditions, etc, etc. When there is nothing in particular which needs our attention, you can just justify with something like `FFI call without precondition`. When accessing fields of a struct returned by a C function, you can often use the postconditions of the function found in the doc and assume that the function is well-behaved.

crates/ironrdp-rdpdr-native/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+doctest = false
+test = false
+
+
+[target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
+ironrdp-pdu.workspace = true
+ironrdp-svc.workspace = true
+ironrdp-rdpdr.workspace = true
+nix = { version = "0.29", features = ["fs"] }
+libc = "0.2"
```
CBenoit, line 24: suggestion: You don’t need to specify this dependency at all since it’s re-exported by `nix`.

```suggestion
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+pub struct SimpleRdpdrBackend {
+    pub file_id: u32,
+    pub file_base: String,
+    pub file_map: std::collections::HashMap<u32, std::fs::File>,
+    pub file_path_map: std::collections::HashMap<u32, String>,
+    pub file_dir_map: std::collections::HashMap<u32, DirBox>,
+}
+
+#[derive(Debug)]
+pub struct DirBox {
```
CBenoit, line 26: question: Maybe we can replace most of the unsafe code by using [`nix::dir::Dir`](https://docs.rs/nix/latest/nix/dir/struct.Dir.html). Did you consider it?

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,790 @@
...
+    req_inner: DeviceCloseRequest,
+) -> PduResult<Vec<SvcMessage>> {
+    backend.file_map.remove(&req_inner.device_io_request.file_id);
+    backend.file_path_map.remove(&req_inner.device_io_request.file_id);
+    if let Some(dir_box) = backend.file_dir_map.remove(&req_inner.device_io_request.file_id) {
+        // SAFETY: it is safe to close dir when we close device
+        unsafe {
+            closedir(dir_box.dir);
+        }
+    }
```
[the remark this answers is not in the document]
zmrush: yes, add a drop function is on my schedule, but i am working on windows build and forgot it.  and as u suggest, nix::dir::Dir is better,  i modify the code

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,778 @@
...
+                        let res = RdpdrPdu::ClientDriveSetInformationResponse(ClientDriveSetInformationResponse::new(
+                            &req_inner,
+                            NtStatus::UNSUCCESSFUL,
+                        )?);
+                        return Ok(vec![SvcMessage::from(res)]);
+                    }
+                }
+                FileInformationClass::EndOfFile(info) => {
+                    let mut set_end_res = -1;
+                    // SAFETY: it is safe to call on *nix platform
```
CBenoit, line 414: question: This is not really documenting why the unsafe operation is indeed safe. Per the documentation, it would be something like that:

```suggestion
                    // SAFETY: `file` is a regular file that we opened with write access.
```

However, I see `file_map` may contain files with `read` permission only. (See line 699.)
Can you explain how we are never calling `ftruncate` on these?

CBenoit (commented):
Thank you! We can merge the PR once the last two comments are addressed! :slightly_smiling_face: 

Also: why `SimpleRdpdrBackend`? Do you think about implementing several backends for various use cases?

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok, i got it

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,778 @@
...
+                    unsafe {
+                        if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
+                            set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
+                        }
+                    }
+                    if set_end_res < 0 {
+                        warn!("set file end error, code is:{}", set_end_res);
```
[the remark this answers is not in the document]
zmrush: ok

zmrush:
> Thank you! We can merge the PR once the last two comments are addressed! 🙂
> 
> Also: why `SimpleRdpdrBackend`? Do you think about implementing several backends for various use cases?

because, i has not implements all the features. In addition, i can implements more features. for example 'query directory information', the file information is return one by one, which is not efficient.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,780 @@
...
+                    let mut set_end_res = -1;
+                    // SAFETY: it is safe to call on *nix platform
+                    // SAFETY: `file` is a regular file that we opened with write access.
+                    unsafe {
+                        if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
+                            set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
+                        }
+                    }
+                    if set_end_res < 0 {
+                        let error = nix::errno::Errno::last(); // However, only call this if `ftruncate` was called.
```
CBenoit, line 422: issue: The comment wasn’t intended to be actually included. You should modify the code to not call `Errono::last()` unless we actually called `ftruncate`.

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,780 @@
...
+                            &req_inner,
+                            NtStatus::UNSUCCESSFUL,
+                        )?);
+                        return Ok(vec![SvcMessage::from(res)]);
+                    }
+                }
+                FileInformationClass::EndOfFile(info) => {
+                    let mut set_end_res = -1;
+                    // SAFETY: it is safe to call on *nix platform
+                    // SAFETY: `file` is a regular file that we opened with write access.
```
CBenoit, line 415: issue: The first comment is not about safety so you can remove it. Also you forgot to explain why we know that the file is actually opened with write access. I see that `file_map` may contain files with read permission only. (See line 699.) Can you explain how we are never calling ftruncate on these?

```suggestion
                    // SAFETY: `file` is a regular file that we opened with write access.
                    // We know that `file` is opened with write access because… [explain why]
```

crates/ironrdp-rdpdr-native/src/nix/simple.rs
```diff
@@ -0,0 +1,780 @@
...
+                    unsafe {
+                        if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
+                            set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
+                        }
+                    }
```
CBenoit, line 420: suggestion: Keep unsafe blocks as small as possible.
```suggestion
                    if let Some(file) = backend.file_map.get(&req_inner.device_io_request.file_id) {
                         unsafe {
                            set_end_res = nix::libc::ftruncate(file.as_raw_fd(), info.end_of_file);
                        }
                    }
```

CBenoit:
> > Thank you! We can merge the PR once the last two comments are addressed! 🙂
> > Also: why `SimpleRdpdrBackend`? Do you think about implementing several backends for various use cases?
> 
> because, i has not implements all the features. In addition, i can implements more features. for example 'query directory information', the file information is return one by one, which is not efficient.

Even if not all features are implemented now, we could implement them in the future. I’m fine with dropping the `Simple` part given that we’ll likely just extend the current code, instead of creating a separate backend. Or do you think we should keep a "simple" backend that is not supporting specific features?

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: i am sorry for not understanding your suggestion. and i modify it

crates/ironrdp-rdpdr-native/src/nix/simple.rs
zmrush: ok,i got it and modify it

zmrush:
> > > Thank you! We can merge the PR once the last two comments are addressed! 🙂
> > > Also: why `SimpleRdpdrBackend`? Do you think about implementing several backends for various use cases?
> > 
> > 
> > because, i has not implements all the features. In addition, i can implements more features. for example 'query directory information', the file information is return one by one, which is not efficient.
> 
> Even if not all features are implemented now, we could implement them in the future. I’m fine with dropping the `Simple` part given that we’ll likely just extend the current code, instead of creating a separate backend. Or do you think we should keep a "simple" backend that is not supporting specific features?

ok,i also think it is fine, i modify the name.

CBenoit (approved):
Thank you for this work!

PR #502, opened 2024-07-23
chore(release): prepare for iron-remote-gui v0.12.0

CBenoit:


web-client/iron-remote-gui/package.json
```diff
@@ -9,7 +9,7 @@
     "Zacharia Ellaham"
   ],
   "description": "Web Component providing agnostic implementation for Iron Wasm base client",
-  "version": "0.11.6",
+  "version": "0.0.0",
```
CBenoit, line 12: We don’t use the version specified in this file, so I set it to 0.0.0 so that the intent is clear.

pacmancoder (approved):

PR #503, opened 2024-07-23
Move lints to workspace

elmarco:
Let's move the lints settings to the workspace, so all tools actually can use them (not just during xtask check)

CI check all targets.

Fix the visible warnings.


crates/ironrdp-testsuite-core/tests/input/smoke.rs
```diff
@@ -98,24 +98,21 @@ fn smoke_keyboard() {
                         (None, None) => {}
                         (None, Some(_)) => unreachable!(),
                         (Some(pressed_packet), None) => {
-                            if let FastPathInputEvent::KeyboardEvent(flags, scancode) = pressed_packet {
+                            if let FastPathInputEvent::KeyboardEvent(flags, _scancode) = pressed_packet {
                                 ensure!(!flags.contains(KeyboardFlags::RELEASE));
-                                ensure!(scancode == u8::from(scancode))
```
CBenoit, line 103: I believe the intention was to ensure that the scancode from `Operation::KeyPressed(scancode)` was the same value as the scancode from the `KeyboardEvent` PDU. Maybe a refactor went wrong, but we should keep the check, changing the second `scancode` for `other_scancode` and something like `ensure!(other_scancode == u8::from(scancode))`.

CBenoit (approved):
Thank you!! I’ve been wanting to make this change for a while now. You can address the comment before I merge, otherwise I’ll just send a follow up PR later :slightly_smiling_face:

crates/ironrdp-pdu/src/rdp/server_license/client_platform_challenge_response.rs
```diff
@@ -179,7 +179,7 @@ enum LicenseDetailLevel {
 }
 
 #[derive(Debug, PartialEq)]
-pub(crate) struct PlatformChallengeResponseData {
+pub struct PlatformChallengeResponseData {
     client_type: ClientType,
     license_detail_level: LicenseDetailLevel,
     challenge: Vec<u8>,
```
CBenoit, line 185: I’m not sure it’s very useful to export this type with making the fields public too

crates/ironrdp-pdu/src/rdp/server_license/client_platform_challenge_response.rs
```diff
@@ -241,7 +241,7 @@ impl<'de> PduDecode<'de> for PlatformChallengeResponseData {
 }
 
 #[derive(Debug, PartialEq, Eq)]
-pub(crate) struct ClientHardwareIdentification {
+pub struct ClientHardwareIdentification {
     pub(crate) platform_id: u32,
     pub(crate) data: Vec<u8>,
 }
```
CBenoit, line 247: Same problem here

crates/ironrdp-pdu/src/rdp/server_license/client_platform_challenge_response.rs
elmarco: ok

CBenoit (commented):
Thank you!

PR #513, opened 2024-07-30
feat(connector): allow smart card logins with CredSSP enabled

probakowski:
This PR makes it possible to log into AD joined machine with CredSSP/NLA enabled using smart card credentials.

Supersedes https://github.com/Devolutions/IronRDP/pull/487


crates/ironrdp-connector/src/credssp.rs
```diff
@@ -96,18 +99,42 @@ impl CredsspSequence {
...
+                        certificate: cert,
+                        reader_name: config.reader_name.clone(),
+                        card_name: None,
+                        container_name: config.container_name.clone(),
+                        csp_name: config.csp_name.clone(),
+                        pin: pin.as_bytes().to_vec().into(),
+                        private_key_file_index: None,
+                        private_key: Some(key.into()),
+                    };
+                    sspi::Credentials::SmartCard(Box::new(identity))
```
CBenoit, line 132: thought: I think we should not expose picky types in sspi-rs public API. I may change that before releasing sspi v0.12.0 :thinking:

Cargo.toml
```diff
@@ -58,7 +58,7 @@ expect-test = "1"
 png = "0.17"
 proptest = "1.4"
 rstest = "0.18"
-sspi = "0.11"
+sspi = "0.13.0"
```
pacmancoder, line 61: Let's keep consistent versioning in this file and omit patch version.
```diff
- sspi = "0.13.0"
+ sspi = "0.13"
```

pacmancoder (approved):
Looks like CI is failing due to not updated `Cago.lock`

Other than that, PR looks good to me, I'll merge it after CI issues will be resolved 👍

probakowski:
@pacmancoder fixed locks problem by merging master into this branch

PR #514, opened 2024-08-01
fix(ironrdp-error): fix build with `alloc` feature

pacmancoder:


CBenoit (approved):
Good catch!

elmarco:
warning: unnecessary qualification
  --> /home/elmarco/src/IronRDP/crates/ironrdp-error/src/lib.rs:51:32
   |
51 |             this.source = Some(alloc::boxed::Box::new(source));
   |                                ^^^^^^^^^^^^^^^^^^^^^^


I think we should simply "use alloc::boxed::Box".

PR #518, opened 2024-08-05
RFC: feat(server): flush server events, avoid audio buffer filling up

elmarco:
When display updates and server events are competing, tokio::select!() will randomly pick one branch and process one event.

This is good for fairness, except when the display updates are big, and delay other messages for example wave data. We may end up in a situation where the server events queue grows and audio data gets delayed.

To avoid this situation, let's flush all server events and drop audio frames when they are queuing up.

This introduce some bias for server events, since all pending server events will be processed before other updates/inputs.

There are many different ways to tackle the problem, this simple approach seems to work fine atm.

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -381,72 +381,84 @@ impl RdpServer {
         Ok(false)
     }
 
-    async fn dispatch_server_event<S>(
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
Sounds reasonable. Let’s investigate more advanced approaches in the future. Thank you!

PR #519, opened 2024-08-05
fix!: remove dependency on ClientConnector in CredsspSequence

msabansal:
Refactored CredsspSequence to remove direct dependency on ClientConnector. This allows CredSSPSequence to be reused in cases where one doesn't necessarily create a clientconnector. As  an example i am building a rust POC that performs a MITM for an RDP connection and don't need to create a client connector directly.

### DIFF
--- a/crates/ironrdp-blocking/src/connector.rs
+++ b/crates/ironrdp-blocking/src/connector.rs
@@ -5,8 +5,8 @@ use ironrdp_connector::sspi::credssp::ClientState;
 use ironrdp_connector::sspi::generator::GeneratorState;
 use ironrdp_connector::sspi::network_client::NetworkClient;
 use ironrdp_connector::{
-    ClientConnector, ClientConnectorState, ConnectionResult, ConnectorError, ConnectorResult, Sequence as _,
-    ServerName, State as _,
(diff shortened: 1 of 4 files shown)


CBenoit:
Thank you! I'm currently on vacation, I'll review next week!

crates/ironrdp-blocking/src/connector.rs
```diff
@@ -125,8 +125,19 @@ where
 {
     assert!(connector.should_perform_credssp());
 
-    let (mut sequence, mut ts_request) =
-        CredsspSequence::init(connector, server_name, server_public_key, kerberos_config)?;
+    let selected_protocol = match connector.state {
+        ClientConnectorState::Credssp { selected_protocol, .. } => selected_protocol,
+        _ => unreachable!(),
```
CBenoit, line 130: suggestion: Can you replace the `unreachable!` with something like
```rust
return Err(general_err!("invalid connector state for CredSSP sequence"));
```
Like we previously had? This should indeed never happen unless there is a bug, but I prefer not failing with an unrecoverable error. IronRDP is also used in backends and should kill threads.

CBenoit (commented):
Can you run rustfmt? The CI is failing on formatting

CBenoit (approved):
LGTM! Thank you!

PR #524, opened 2024-08-07
fix(client): let address resolve to either ipv4 or ipv6

elmarco:
Let TcpStream::connect() handle address resolution, so that multiple addresses (typically localhost -> ipv4 127.0.0.1 or ipv6 ::1) are attempted.


CBenoit (approved):
Good catch!

PR #525, opened 2024-08-09
Improve Display impl for ConnectorErrorKind

zmb3:
Prior to this change, a CredSSP connector error would display as `[CredSSP] CredSSP` - not very helpful.

After the change it looks something like this: `[CredSSP] NoAuthenticatingAuthority: Could not contact KDC`.

I'm sure there's more we can do here, I've only modified cases where I've been able to induce real errors.

### DIFF
--- a/crates/ironrdp-connector/src/lib.rs
+++ b/crates/ironrdp-connector/src/lib.rs
@@ -19,6 +19,7 @@ mod server_name;
 
 use core::any::Any;
 use core::fmt;
+use std::error::Error;
 
 pub use channel_connection::{ChannelConnectionSequence, ChannelConnectionState};
 pub use connection::{encode_send_data_request, ClientConnector, ClientConnectorState, ConnectionResult};
@@ -260,17 +261,24 @@ impl fmt::Display for ConnectorErrorKind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &self {
             ConnectorErrorKind::Pdu(_) => write!(f, "PDU error"),
-            ConnectorErrorKind::Credssp(_) => write!(f, "CredSSP"),
-            ConnectorErrorKind::Reason(description) => write!(f, "reason: {description}"),
-            ConnectorErrorKind::AccessDenied => write!(f, "access denied"),
-            ConnectorErrorKind::General => write!(f, "general error"),
-            ConnectorErrorKind::Custom => write!(f, "custom error"),
+            ConnectorErrorKind::Credssp(e) => write!(f, "{:?}: {}", e.error_type, e.description),
+            ConnectorErrorKind::Reason(description) => write!(f, "Reason: {description}"),
+            ConnectorErrorKind::AccessDenied => write!(f, "Access Denied"),
+            ConnectorErrorKind::General => write!(f, "General Error"),
+            ConnectorErrorKind::Custom => {
+                write!(f, "Custom Error")?;
+                if let Some(src) = self.source() {
+                    write!(f, " ({})", src)
+                } else {
+                    Ok(())
+                }
+            }
         }
     }
 }
 
-impl std::error::Error for ConnectorErrorKind {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+impl Error for ConnectorErrorKind {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
         match &self {
             ConnectorErrorKind::Pdu(e) => Some(e),
             ConnectorErrorKind::Credssp(e) => Some(e),
@@ -290,7 +298,7 @@ pub trait ConnectorErrorExt {
     fn reason(context: &'static str, reason: impl Into<String>) -> Self;
     fn custom<E>(context: &'static str, e: E) -> Self
     where
-        E: std::error::Error + Sync + Send + 'static;
+        E: Error + Sync + Send + 'static;
 }
 
 impl ConnectorErrorExt for ConnectorError {
@@ -308,7 +316,7 @@ impl ConnectorErrorExt for ConnectorError {
 
     fn custom<E>(context: &'static str, e: E) -> Self
     where
-        E: std::error::Error + Sync + Send + 'static,
+        E: Error + Sync + Send + 'static,
     {
         Self::new(context, ConnectorErrorKind::Custom).with_source(e)
     }
@@ -320,7 +328,7 @@ pub trait ConnectorResultExt {
     #[must_use]
     fn with_source<E>(self, source: E) -> Self
     where
-        E: std::error::Error + Sync + Send + 'static;
+        E: Error + Sync + Send + 'static;
 }
 
 impl<T> ConnectorResultExt for ConnectorResult<T> {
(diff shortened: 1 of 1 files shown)


CBenoit (changes requested):
This is intentional: there are two possible avenues when it comes to reporting nested errors.

- Print the underlying error in the `Display` implementation like you did in this PR, **but** _do not propagate the source_.
- Do not print the underlying error in the `Display` implementation, and propagate the source. This is what we do in IronRDP.

I think the first is fine, but I consider the second one superior. It’s more powerful and leave the final formatting decision to the error reporter instead of imposing it. However, with this change, we are both formatting the underlying error in the `Display` implementation _and_ propagating the source. This will cause weird formatting when a proper reporter is used.

For example, if I enter the wrong password, with this patch I get this error:

```
Connection error: [CredSSP] InvalidToken: CredSSP server returned an error status, caused by: InvalidToken: CredSSP server returned an error status; status is STATUS_LOGON_FAILURE [0xc000006d]
```

The description is duplicated because the reporter already crawls the sources. See also how the formatted error is missing the last error message.

While with master I get this:

```
Connection error: [CredSSP] CredSSP, caused by: InvalidToken: CredSSP server returned an error status; status is STATUS_LOGON_FAILURE [0xc000006d]
```

Which is the intended behavior.

I would however consider this patch an improvement over what we already have:

```diff
--- a/crates/ironrdp-connector/src/lib.rs
+++ b/crates/ironrdp-connector/src/lib.rs
@@ -260,7 +260,7 @@ impl fmt::Display for ConnectorErrorKind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &self {
             ConnectorErrorKind::Pdu(_) => write!(f, "PDU error"),
-            ConnectorErrorKind::Credssp(_) => write!(f, "CredSSP"),
+            ConnectorErrorKind::Credssp(e) => write!(f, "{e}"),
             ConnectorErrorKind::Reason(description) => write!(f, "reason: {description}"),
             ConnectorErrorKind::AccessDenied => write!(f, "access denied"),
             ConnectorErrorKind::General => write!(f, "general error"),
@@ -273,7 +273,7 @@ impl std::error::Error for ConnectorErrorKind {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match &self {
             ConnectorErrorKind::Pdu(e) => Some(e),
-            ConnectorErrorKind::Credssp(e) => Some(e),
+            ConnectorErrorKind::Credssp(e) => e.source(),
             ConnectorErrorKind::Reason(_) => None,
             ConnectorErrorKind::AccessDenied => None,
             ConnectorErrorKind::Custom => None,
```

Giving this:

```
Connection error: [CredSSP] InvalidToken: CredSSP server returned an error status; status is STATUS_LOGON_FAILURE [0xc000006d]
```

FYI:

- We are providing [a basic error reporter](https://github.com/Devolutions/IronRDP/blob/0f238097e0fd9780b017bde473dc495a759906c1/crates/ironrdp-error/src/lib.rs#L124) that you can easily get by calling [`report`](https://github.com/Devolutions/IronRDP/blob/0f238097e0fd9780b017bde473dc495a759906c1/crates/ironrdp-error/src/lib.rs#L81) on the error. See how we are using it here:  https://github.com/Devolutions/IronRDP/blob/0f238097e0fd9780b017bde473dc495a759906c1/crates/ironrdp-client/src/app.rs#L285-L305

- `anyhow::Error` also provides some "reporting" facilities [by using the "alternate selector" or the Debug implementation](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations).

- We are not using `thiserror` here, but you can see that’s how it’s done in [this example](https://docs.rs/thiserror/latest/thiserror/#example). The source error is not formatted in the error message of the variant holding it.

- There is a `Reporter` type that is being worked on in the standard library: https://doc.rust-lang.org/std/error/struct.Report.html

I suggest that you either:

- Use our reporter
- Use `anyhow`’s error
- Write your own reporter if you would rather have your own

crates/ironrdp-connector/src/lib.rs
```diff
@@ -260,17 +261,24 @@ impl fmt::Display for ConnectorErrorKind {
...
-            ConnectorErrorKind::Reason(description) => write!(f, "reason: {description}"),
-            ConnectorErrorKind::AccessDenied => write!(f, "access denied"),
-            ConnectorErrorKind::General => write!(f, "general error"),
-            ConnectorErrorKind::Custom => write!(f, "custom error"),
+            ConnectorErrorKind::Credssp(e) => write!(f, "{:?}: {}", e.error_type, e.description),
+            ConnectorErrorKind::Reason(description) => write!(f, "Reason: {description}"),
+            ConnectorErrorKind::AccessDenied => write!(f, "Access Denied"),
+            ConnectorErrorKind::General => write!(f, "General Error"),
+            ConnectorErrorKind::Custom => {
+                write!(f, "Custom Error")?;
```
CBenoit, line 269: See also [this section](https://github.com/Devolutions/IronRDP/blob/0f238097e0fd9780b017bde473dc495a759906c1/STYLE.md#formatting-of-error-messages) of our style doc about error formatting

CBenoit:
> I would however consider this patch an improvement over what we already have:
> 
> ```diff
> --- a/crates/ironrdp-connector/src/lib.rs
> +++ b/crates/ironrdp-connector/src/lib.rs
> @@ -260,7 +260,7 @@ impl fmt::Display for ConnectorErrorKind {
>      fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
>          match &self {
>              ConnectorErrorKind::Pdu(_) => write!(f, "PDU error"),
> -            ConnectorErrorKind::Credssp(_) => write!(f, "CredSSP"),
> +            ConnectorErrorKind::Credssp(e) => write!(f, "{e}"),
>              ConnectorErrorKind::Reason(description) => write!(f, "reason: {description}"),
>              ConnectorErrorKind::AccessDenied => write!(f, "access denied"),
>              ConnectorErrorKind::General => write!(f, "general error"),
> @@ -273,7 +273,7 @@ impl std::error::Error for ConnectorErrorKind {
>      fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
>          match &self {
>              ConnectorErrorKind::Pdu(e) => Some(e),
> -            ConnectorErrorKind::Credssp(e) => Some(e),
> +            ConnectorErrorKind::Credssp(e) => e.source(),
>              ConnectorErrorKind::Reason(_) => None,
>              ConnectorErrorKind::AccessDenied => None,
>              ConnectorErrorKind::Custom => None,
> ```
> 
> Giving this:
> 
> ```
> Connection error: [CredSSP] InvalidToken: CredSSP server returned an error status; status is STATUS_LOGON_FAILURE [0xc000006d]
> ```

Actually, I’m not convinced anymore this is an improvement either, because this would prevent the CredSSP error itself from being reached using the `source()` method. This is sometimes useful when inspecting nested errors (I myself needed to do this in the past for various reasons).

I’ll close as I’m not convinced there is really something we can do to improve anything as-is, but feel free to re-open if you revisit!

PR #530, opened 2024-08-13
Server "resize" support

elmarco:
Add deactivation-reactivation support to the server.

Add BitmapUpdate "stride" support, which helps/fixes qemu-rdp server for odd dimensions.

PR #531, opened 2024-08-14
refactor: follow up to PR 530

CBenoit:
Applying suggestions from https://github.com/Devolutions/IronRDP/pull/530

cc @elmarco


elmarco:
lgtm

would be nice to have some checks for that :)

CBenoit:
True! We'll have to build a small exception list for acronyms, but it's something we could do.

awakecoding (approved):

PR #532, opened 2024-08-14
Server resize follow-up

elmarco:
I forgot to squash BitmapEncoder support in #530.

The main part is handling pending incoming client message during deact-react;  queuing and replaying them after acceptation.

### DIFF
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -230,12 +230,13 @@ pub async fn single_sequence_step<S>(
     framed: &mut Framed<S>,
     sequence: &mut dyn Sequence,
     buf: &mut WriteBuf,
+    unmatched: Option<&mut Vec<Bytes>>,
 ) -> ConnectorResult<()>
 where
     S: FramedWrite + FramedRead,
 {
     buf.clear();
-    let written = single_sequence_step_read(framed, sequence, buf).await?;
+    let written = single_sequence_step_read(framed, sequence, buf, unmatched).await?;
     single_sequence_step_write(framed, buf, written).await
 }
 
@@ -243,6 +244,7 @@ pub async fn single_sequence_step_read<S>(
     framed: &mut Framed<S>,
     sequence: &mut dyn Sequence,
     buf: &mut WriteBuf,
+    unmatched: Option<&mut Vec<Bytes>>,
 ) -> ConnectorResult<Written>
 where
     S: FramedRead,
@@ -257,7 +259,7 @@ where
         );
 
         let pdu = framed
-            .read_by_hint(next_pdu_hint, None)
+            .read_by_hint(next_pdu_hint, unmatched)
             .await
             .map_err(|e| ironrdp_connector::custom_err!("read frame by hint", e))?;
 
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -164,14 +164,25 @@ where
     /// `tokio::select!` statement and some other branch
     /// completes first, then it is safe to drop the future and re-create it later.
     /// Data may have been read, but it will be stored in the internal buffer.
-    pub async fn read_by_hint(&mut self, hint: &dyn PduHint) -> io::Result<Bytes> {
+    pub async fn read_by_hint(
+        &mut self,
+        hint: &dyn PduHint,
+        mut unmatched: Option<&mut Vec<Bytes>>,
+    ) -> io::Result<Bytes> {
         loop {
             match hint
                 .find_size(self.peek())
                 .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?
             {
-                Some((_matched, length)) => {
-                    return Ok(self.read_exact(length).await?.freeze());
+                Some((matched, length)) => {
+                    let bytes = self.read_exact(length).await?.freeze();
+                    if matched {
+                        return Ok(bytes);
+                    } else if let Some(ref mut unmatched) = unmatched {
+                        unmatched.push(bytes);
+                    } else {
+                        warn!("Received and lost an unexpected PDU");
+                    }
                 }
                 None => {
                     let len = self.read().await?;
@@ -246,7 +257,7 @@ where
         );
 
         let pdu = framed
-            .read_by_hint(next_pdu_hint)
+            .read_by_hint(next_pdu_hint, None)
             .await
             .map_err(|e| ironrdp_connector::custom_err!("read frame by hint", e))?;
 
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -170,7 +170,7 @@ where
                 .find_size(self.peek())
                 .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?
             {
-                Some(length) => {
+                Some((_matched, length)) => {
                     return Ok(self.read_exact(length).await?.freeze());
                 }
                 None => {
--- a/crates/ironrdp-server/src/encoder/bitmap.rs
+++ b/crates/ironrdp-server/src/encoder/bitmap.rs
@@ -3,7 +3,7 @@ use ironrdp_graphics::rdp6::{ABgrChannels, ARgbChannels, BgrAChannels, BitmapStr
 use ironrdp_pdu::bitmap::{self, BitmapData, BitmapUpdateData, Compression};
 use ironrdp_pdu::cursor::WriteCursor;
 use ironrdp_pdu::geometry::InclusiveRectangle;
-use ironrdp_pdu::{PduEncode, PduError};
+use ironrdp_pdu::{invalid_message_err, PduEncode, PduError};
 
 use crate::{BitmapUpdate, PixelOrder};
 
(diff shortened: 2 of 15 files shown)


CBenoit:
@elmarco Thank you!
Is it ready for review?

elmarco:
> @elmarco Thank you! Is it ready for review?

yes, thanks

crates/ironrdp-server/src/encoder/bitmap.rs
```diff
@@ -20,7 +20,13 @@ impl BitmapEncoder {
     }
 
     pub(crate) fn encode(&mut self, bitmap: &BitmapUpdate, output: &mut [u8]) -> Result<usize, PduError> {
-        let row_len = usize::from(bitmap.width.get()) * usize::from(bitmap.format.bytes_per_pixel());
+        if bitmap.width.get() % 4 != 0 {
+            // FIXME
```
CBenoit, line 24: suggestion: It would be helpful to document such things with an inline code comment in addition to the commit message which is almost only read when reviewing the patch, but not, i.e., one year later.

```rust
// FIXME: support non-multiple of 4 widths.
// 
// It’s not clear how to achieve that yet, but generally, server uses multiple of 4 widths,
// and client has surface capabilities, so this path is unlikely.
```

ffi/src/connector/mod.rs
```diff
@@ -158,9 +158,19 @@ pub mod ffi {
...
+            matched: &mut bool,
+        ) -> Result<Box<crate::utils::ffi::OptionalUsize>, Box<IronRdpError>> {
             let pdu_hint = self.0;
-            let size = pdu_hint.find_size(bytes)?;
+            let size = if let Some((pdu_matched, size)) = pdu_hint.find_size(bytes)? {
+                *matched = pdu_matched;
+                Some(size)
+            } else {
+                *matched = false;
+                None
```
CBenoit, line 172: issue?: This is suspicious.

crates/ironrdp-web/src/session.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(non_snake_case)]
+
```
CBenoit, line 2: question: I don’t have any warning when I open this file or when I try to compile manually. Do you understand why the warning is emitted on your side?

suggestion: Add a comment explaining why the lint is silenced. We don’t really want to silent that forever, so at some point it would be nice to remember we didn’t intend to so in the first place.

CBenoit (commented):
There are two commits with the same summary: [feat(async): teach single_sequence_step() to keep unmatched PDUs](https://github.com/Devolutions/IronRDP/pull/532/commits/0b2547f3bd69efe27069bf3c734786e50fbc3d8b)

Did you intend to change the scope?

CBenoit (commented):
Refactor code into something like this:

```diff
diff --git a/crates/ironrdp-async/src/framed.rs b/crates/ironrdp-async/src/framed.rs
index 2597f644..fa4cf600 100644
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -60,10 +60,6 @@ impl<S> Framed<S> {
     pub fn peek(&self) -> &[u8] {
         &self.buf
     }
-
-    pub fn leftover_mut(&mut self) -> &mut BytesMut {
-        &mut self.buf
-    }
 }

 impl<S> Framed<S>
diff --git a/crates/ironrdp-server/src/server.rs b/crates/ironrdp-server/src/server.rs
index 63888d3d..e15deacb 100644
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -17,6 +17,7 @@ use ironrdp_pdu::{self, decode, encode_vec, mcs, nego, rdp, Action, PduResult};
 use ironrdp_svc::{impl_as_any, server_encode_svc_messages, StaticChannelId, StaticChannelSet, SvcProcessor};
 use ironrdp_tokio::{Framed, FramedRead, FramedWrite, TokioFramed};
 use rdpsnd::server::{RdpsndServer, RdpsndServerMessage};
+use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::{mpsc, Mutex};
 use tokio::task;
@@ -797,25 +798,29 @@ impl RdpServer {
         }
     }

-    async fn accept_finalize<S>(&mut self, mut framed: Framed<S>, mut acceptor: Acceptor) -> Result<()>
+    async fn accept_finalize<S>(&mut self, mut framed: TokioFramed<S>, mut acceptor: Acceptor) -> Result<()>
     where
-        S: FramedRead + FramedWrite,
+        S: AsyncRead + AsyncWrite + Sync + Send + Unpin,
     {
         let mut other_pdus = None;
+
         loop {
             let (new_framed, result) = ironrdp_acceptor::accept_finalize(framed, &mut acceptor, other_pdus.as_mut())
                 .await
                 .context("failed to accept client during finalize")?;

-            framed = new_framed;
+            let (stream, mut leftover) = new_framed.into_inner();

             debug!(other_pdus = other_pdus.as_ref().map_or(0, |o| o.len()));
             if let Some(pdus) = other_pdus.take() {
-                let mut leftover: Vec<_> = pdus.iter().flat_map(|p| p.iter().copied()).collect();
-                leftover.extend_from_slice(&framed.leftover_mut().split()[..]);
-                framed.leftover_mut().extend_from_slice(&leftover);
+                let unmatched_frames = pdus.into_iter().flatten();
+                let previous_leftover = leftover.split();
+                leftover.extend(unmatched_frames);
+                leftover.extend_from_slice(&previous_leftover);
             }

+            framed = TokioFramed::new_with_leftover(stream, leftover);
+
             match self.client_accepted(&mut framed, result).await? {
                 RunState::Continue => {
                     unreachable!();
```

crates/ironrdp-async/src/framed.rs
```diff
@@ -164,23 +168,22 @@ where
     /// `tokio::select!` statement and some other branch
     /// completes first, then it is safe to drop the future and re-create it later.
     /// Data may have been read, but it will be stored in the internal buffer.
-    pub async fn read_by_hint(&mut self, hint: &dyn PduHint) -> io::Result<Bytes> {
+    pub async fn read_by_hint(&mut self, hint: &dyn PduHint) -> io::Result<(Bytes, Vec<Bytes>)> {
```
CBenoit, line 171: suggestion: I feel it would be better to follow the same pattern everywhere, and optionally provide the accumulator like this.
```suggestion
    pub async fn read_by_hint(&mut self, hint: &dyn PduHint, unmatched: Option<&mut Vec<Bytes>>) -> io::Result<Bytes> {
```

We can just drop the unmatched frame when the accumulator is `None`.

ffi/dotnet/Devolutions.IronRdp/Generated/RawPduHint.cs
```diff
@@ -17,7 +17,7 @@ public partial struct PduHint
     private const string NativeLib = "DevolutionsIronRdp";
 
     [DllImport(NativeLib, CallingConvention = CallingConvention.Cdecl, EntryPoint = "PduHint_find_size", ExactSpelling = true)]
-    public static unsafe extern ConnectorFfiResultBoxOptionalUsizeBoxIronRdpError FindSize(PduHint* self, byte* bytes, nuint bytesSz);
+    public static unsafe extern ConnectorFfiResultBoxOptionalUsizeBoxIronRdpError FindSize(PduHint* self, byte* bytes, nuint bytesSz, bool* matched);
```
CBenoit, line 20: issue: Can you revert this for now? At least the C# NuGet is only used on client-side so we probably don’t need to break the ABI for that just now.

crates/ironrdp-server/src/encoder/bitmap.rs
elmarco: ack, thanks

crates/ironrdp-web/src/session.rs
elmarco: I added a link to an issue I opened for wasm-bindgen.

crates/ironrdp-async/src/framed.rs
```diff
@@ -195,6 +198,16 @@ where
...
+
+    /// Reads from stream and fills internal buffer, returning how many bytes were read.
+    /// If the stream is empty, returns `io::ErrorKind::UnexpectedEof`.
+    async fn read_check_eof(&mut self) -> Result<(), io::Error> {
+        let len = self.read().await?;
+        if len == 0 {
+            return Err(io::Error::new(io::ErrorKind::UnexpectedEof, "not enough bytes"));
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
elmarco: agree, thanks

crates/ironrdp-async/src/framed.rs
```diff
@@ -60,6 +60,10 @@ impl<S> Framed<S> {
     pub fn peek(&self) -> &[u8] {
         &self.buf
     }
+
+    pub fn leftover_mut(&mut self) -> &mut BytesMut {
+        &mut self.buf
+    }
```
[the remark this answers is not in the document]
elmarco: thanks a lot, indeed better

ffi/dotnet/Devolutions.IronRdp/Generated/RawPduHint.cs
elmarco: ok

CBenoit:
@elmarco By the way, are you getting these client messages from mstsc as well? I’m wondering if the real fix should not be about modifying the IronRDP client to not send messages unrelated to the connection sequence before the connection sequence is over.

elmarco:
> @elmarco By the way, are you getting these client messages from mstsc as well? I’m wondering if the real fix should not be about modifying the IronRDP client to not send messages unrelated to the connection sequence before the connection sequence is over.

Yes. The reactivation sequence is triggered by the server, so there can always be client messages in flight.

CBenoit:
> > @elmarco By the way, are you getting these client messages from mstsc as well? I’m wondering if the real fix should not be about modifying the IronRDP client to not send messages unrelated to the connection sequence before the connection sequence is over.
> 
> Yes. The reactivation sequence is triggered by the server, so there can always be client messages in flight.

Oh right. It becomes a problem during the reactivation sequence. Sounds good.

crates/ironrdp-async/src/framed.rs
elmarco: ok

CBenoit (approved):
LGTM!

Issue #498, opened 2024-07-22
Planning 0.5 release


elmarco:
Hi,

What' s the plan for the next development release? Is there any blocker?

CBenoit:
Hi! I’ll come up with a detailed roadmap in the next few days and let you know about it. Mainly, I want to adjust a few last things in the public API, but you already helped a lot with the effort of migrating to the new traits.

CBenoit:
I’m sorry, I didn’t get enough bandwidth to write that down, but let me elaborate the major things I want to get done before publishing a new version.

One of is to get most traits such as `Encode` / `Decode` out of `ironrdp-pdu` and put that into a crate called `ironrdp-core`. `ironrdp-pdu` contains lots of code that we don’t actually need in other crates such as the virtual channels. I started this in this branch: https://github.com/Devolutions/IronRDP/tree/7d46aa26e4a8892c0eb17b00c82cea46f9f7d88f/crates/ironrdp-common/src
Just moving the traits without changing any API (unlike I did in my branch) would be a good enough first step for `ironrdp 0.5`.

Then, I wanted to address the limitations I raised in this PR: https://github.com/Devolutions/devolutions-gateway/pull/916

> - `RdpServer` should implement `Send`.
> - `RdpServer` should support multiple listener (bind) addresses.
> - `RdpServer` should support graceful shutdown.

Of which you already addressed the last one in a previous PR I think.

FYI, I also intend to rework the API of `ironrdp-server` so that it’s possible to use something else than `tokio` and a custom main loop, but I’m fine with pushing all of that in a future `ironrdp 0.6` version, let’s just publish more frequently!

elmarco:
@CBenoit are you ok with an initial RC release?

fwiw, `cargo workspaces version custom 0.5.0-rc.0` seems handy to update versions and tags, and deps. (alas, cargo nightly has some support for packaging workspaces, but not publishing yet)

`cargo release --workspace` is not handling rc properly apparently.

It seems it will have to be manual for now.

CBenoit:
Done! I also pushed `ironrdp-v0.5.0` and `ironrdp-v0.6.0` in addition to `v0.5.0` and `v0.6.0`. I’ll push tags for all the other crates as well, so we’ll need to specify the same version number for different crates at different times.

PR #537, opened 2024-08-27
Refactoring for 0.5

elmarco:
Move common traits from pdu to core, as requested by Benoit:
https://github.com/Devolutions/IronRDP/issues/498#issuecomment-2269190786


CBenoit (approved):
This is looking good to me! Thank you for the massive work here!

PR #547, opened 2024-09-04
Save the disconnect reason from the set error info PDU

zmb3:
> The Set Error Info PDU is sent as a precursor to a server-side
disconnect and informs the client of the reason for the disconnection
which will follow. Once this PDU has been processed, the client MUST
store the error code so that the reason for the server disconnect
which will follow can be accurately reported to the user.

### DIFF
--- a/crates/ironrdp-session/src/active_stage.rs
+++ b/crates/ironrdp-session/src/active_stage.rs
@@ -265,14 +265,17 @@ impl TryFrom<x224::ProcessorOutput> for ActiveStageOutput {
     fn try_from(value: x224::ProcessorOutput) -> Result<Self, Self::Error> {
         match value {
             x224::ProcessorOutput::ResponseFrame(frame) => Ok(Self::ResponseFrame(frame)),
-            x224::ProcessorOutput::Disconnect(reason) => {
-                let reason = match reason {
-                    mcs::DisconnectReason::UserRequested => GracefulDisconnectReason::UserInitiated,
-                    mcs::DisconnectReason::ProviderInitiated => GracefulDisconnectReason::ServerInitiated,
-                    other => GracefulDisconnectReason::Other(other.description()),
+            x224::ProcessorOutput::Disconnect(desc) => {
+                let desc = match desc {
+                    x224::DisconnectDescription::McsDisconnect(reason) => match reason {
+                        mcs::DisconnectReason::ProviderInitiated => GracefulDisconnectReason::ServerInitiated,
+                        mcs::DisconnectReason::UserRequested => GracefulDisconnectReason::UserInitiated,
+                        other => GracefulDisconnectReason::Other(other.description().to_owned()),
+                    },
+                    x224::DisconnectDescription::ErrorInfo(info) => GracefulDisconnectReason::Other(info.description()),
                 };
 
-                Ok(Self::Terminate(reason))
+                Ok(Self::Terminate(desc))
             }
             x224::ProcessorOutput::DeactivateAll(cas) => Ok(Self::DeactivateAll(cas)),
         }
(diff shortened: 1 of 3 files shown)


crates/ironrdp-session/src/active_stage.rs
```diff
@@ -281,25 +284,25 @@ impl TryFrom<x224::ProcessorOutput> for ActiveStageOutput {
...
 impl GracefulDisconnectReason {
-    pub fn description(&self) -> &'static str {
+    pub fn description(&self) -> String {
         match self {
-            GracefulDisconnectReason::UserInitiated => "user initiated disconnect",
-            GracefulDisconnectReason::ServerInitiated => "server initiated disconnect",
-            GracefulDisconnectReason::Other(description) => description,
+            GracefulDisconnectReason::UserInitiated => "user initiated disconnect".to_owned(),
+            GracefulDisconnectReason::ServerInitiated => "server initiated disconnect".to_owned(),
+            GracefulDisconnectReason::Other(description) => description.clone(),
```
zmb3, line 299: Technically, many of these are _user-initiated disconnects_, but using `Other` is the easiest way to get a custom description..

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -16,14 +16,26 @@ pub enum ProcessorOutput {
...
     /// [Deactivation-Reactivation Sequence]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/dfc234ce-481a-4674-9a5d-2a7bafb14432
     DeactivateAll(Box<ConnectionActivationSequence>),
 }
 
+#[derive(Debug, Clone)]
+pub enum DisconnectDescription {
+    /// Includes the reason from the MCS Disconnect Provider Ultimatum.
+    /// This is the least-specific disconnect reason and is only used
+    /// when a more specific disconnect code is not available.
+    McsDisconnect(DisconnectReason),
```
pacmancoder, line 32: [suggestion]: Naming - I think `McsDisconnect` is too specific about protocol details, I suggest the following naming. WDYT?:
```rust 
pub enum DisconnectKind {
    Graceful(DisconnectReason), 
    Failure(ErrorInfo),
}
```

crates/ironrdp-session/src/active_stage.rs
CBenoit: What are the other user-initiated disconnects?

crates/ironrdp-session/src/x224/mod.rs
pacmancoder: Ah, right, `Graceful` is not fitting name here, not all `DisconnectReason` variants are graceful disconnects

CBenoit: Then I guess it’s good as-is?

crates/ironrdp-session/src/active_stage.rs
zmb3: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/a21a1bd9-2303-49c1-90ec-3932435c248c

A lot of these might be considered user-initiated:
- ERRINFO_DISCONNECTED_BY_OTHERCONNECTION
- ERRINFO_RPC_INITIATED_DISCONNECT_BYUSER
- ERRINFO_LOGOFF_BY_USER
- ERRINFO_SERVER_SHUTDOWN
- ERRINFO_SERVER_REBOOT

That said, I think we'd be debating semantics, the goal is to give the user as specific a disconnect reason as possible, and I think we're accomplishing that as-is.

CBenoit (approved):
Thank you!

PR #548, opened 2024-09-10
Improve server responsiveness & usability

elmarco:
The current server implementation processes events one at a time.

This creates issues when doing large IO or CPU intensive task, such as rfx encoding of large surfaces.

This series split read & write halfes, so they can be processed in parallel. It moves bitmap encoding to a blocking task thread, and finally processes the various server tasks concurrently, rather than via one tokio:select!() branch at a time. 

This makes qemu-rdp a lot more usable under some stressing conditions (playing video, large screen etc)

### DIFF
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -209,10 +209,14 @@ where
     }
 }
 
-impl<S> Framed<S>
+impl<S> FramedWrite for Framed<S>
 where
     S: FramedWrite,
 {
+    type WriteAllFut<'write> = S::WriteAllFut<'write>
+    where
+        Self: 'write;
+
     /// Attempts to write an entire buffer into this `Framed`’s stream.
     ///
     /// # Cancel safety
@@ -222,8 +226,8 @@ where
     /// branch completes first, then the provided buffer may have been
     /// partially written, but future calls to `write_all` will start over
     /// from the beginning of the buffer.
-    pub async fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
-        self.stream.write_all(buf).await
+    fn write_all<'a>(&'a mut self, buf: &'a [u8]) -> Self::WriteAllFut<'a> {
+        self.stream.write_all(buf)
     }
 }
 
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -10,7 +10,7 @@ use ironrdp::session::{fast_path, ActiveStage, ActiveStageOutput, GracefulDiscon
 use ironrdp::{cliprdr, connector, rdpdr, rdpsnd, session};
 use ironrdp_core::WriteBuf;
 use ironrdp_rdpsnd_native::cpal;
-use ironrdp_tokio::{single_sequence_step_read, split_tokio_framed};
+use ironrdp_tokio::{single_sequence_step_read, split_tokio_framed, FramedWrite};
 use rdpdr::NoopRdpdrBackend;
 use smallvec::SmallVec;
 use tokio::net::TcpStream;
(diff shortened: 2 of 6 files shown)


CBenoit:
I see you refactored quite a bit the async code for `framed`, `connector` and `acceptor` modules. However, I believe it was already possible to split the writer and the writer given that we know that the tokio implementation is used.

```rust
fn split_tokio_framed<S>(framed: TokioFramed<S>) -> (TokioFramed<ReadHalf<S>>, TokioFramed<WriteHalf<S>>)
where
    S: Sync + Sync + Unpin + AsyncRead + AsyncWrite,
{
    let stream = framed.into_inner_no_leftover();
    let (read_half, write_half) = tokio::io::split(stream);
    (TokioFramed::new(read_half), TokioFramed::new(write_half))
}
```

This way, the unsafe bits are left to the tokio crate.

Actually, we already do something similar to this in `ironrdp-web`:

https://github.com/Devolutions/IronRDP/blob/3d3d9f2c5643360a61f02ea221defb99a03cbcb0/crates/ironrdp-web/src/session.rs#L304-L337

elmarco:
The split in `ironrdp-web` is not on `Framed` but on `WebSocket`. It creates only a read-only `LocalFuturesFramed`. 

Should we limit the API to tokio only? 

What guarantee is there that there is no leftover during a split? I'd need to review the call places.

Yes, it introduces one `unsafe`, which is the same as in tokio to handle the lock.

CBenoit:
> The split in `ironrdp-web` is not on `Framed` but on `WebSocket`. It creates only a read-only `LocalFuturesFramed`.

That’s correct. I’m also suggesting to create two `Framed`s respectively wrapping `ReadHalf` and `WriteHalf` as the inner stream as shown in the snippet of my initial comment. We effectively get a `FramedRead` and `FrameWrite` without changing the API. (That’s actually why the functions are [implemented](https://github.com/Devolutions/IronRDP/blob/3d3d9f2c5643360a61f02ea221defb99a03cbcb0/crates/ironrdp-async/src/framed.rs#L101-L102) [conditionally](https://github.com/Devolutions/IronRDP/blob/3d3d9f2c5643360a61f02ea221defb99a03cbcb0/crates/ironrdp-async/src/framed.rs#L213-L214).)

> Should we limit the API to tokio only?

I’m suggesting we change nothing in all the other crates, as we’ll always read and write sequentially during the acceptation or connection sequence.
`ironrdp-server` itself is already limited to tokio as of today, so I think it’s okay to rely on that.

> What guarantee is there that there is no leftover during a split? I'd need to review the call places.

If you suspect we may have to handle leftovers, it’s possible to reuse the buffer like this:

```rust
let (stream, leftover) = framed.into_inner();
let (read_half, write_half) = tokio::io::split(stream);
let framed_read = TokioFramed::new_with_leftover(read_half, leftover);
let framed_write = TokioFramed::new(write_half);
```

> Yes, it introduces one `unsafe`, which is the same as in tokio to handle the lock.

I indeed recognize that it’s similar to what `tokio` is doing internally.
My argument is more that we could avoid caring about this on our side.
It’s still easier to modify, review and trust code free from unsafe blocks.

I also have a concern about the way the `poll_read` and `poll_write_all` methods are implemented.

For instance:

```rust
    fn poll_write_all<'a>(self: Pin<&'a mut Self>, cx: &mut Context<'_>, buf: &'a [u8]) -> Poll<io::Result<()>> {
        use futures_util::io::AsyncWriteExt as _;
        use std::future::Future;

        let f = async {
            let inner = &mut self.get_mut().inner;
            inner.write_all(buf).await?;
            inner.flush().await?;

            Ok(())
        };
        let mut pin = Box::pin(f);
        pin.as_mut().poll(cx)
    }
```

My understanding is that we create a future using an `async` block that we poll only once. This future is dropped and created again each time we call `poll_write_all`. Are you sure the future returned by `flush()` is ever called? `write_all` is also returning a stateful future that is not cancellation safe.

As such, I also have concerns about the correctness of the new code, and I feel that it would be safer and simpler to use `tokio::split` and `TokioFramed` directly in the `ironrdp-server` (which is tokio-specific anyway), keeping `ironrdp-async` as it is today on master.

If you still believe it would be better to refactor all of that, I’m open, but I’ll have to look more in depth into that.

elmarco:
@CBenoit updated

CBenoit (approved):
LGTM! Thank you!!

PR #550, opened 2024-09-17
feat(web): allow dynamic resize for web

irvingoujAtDevolution:
Allow IronRdp Web to use display control for smart resizing.
Add example to use popup window for connection

Demo:
https://github.com/user-attachments/assets/d9f94937-3ac8-4788-b70e-6292cc8ecde0

### DIFF
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -887,15 +887,15 @@ async fn writer_task(rx: mpsc::UnboundedReceiver<Vec<u8>>, rdp_writer: WriteHalf
     }
 }
 
-pub(crate) struct ConnectParams {
-    pub(crate) ws: WebSocket,
-    pub(crate) config: connector::Config,
-    pub(crate) proxy_auth_token: String,
-    pub(crate) destination: String,
-    pub(crate) pcb: Option<String>,
-    pub(crate) kdc_proxy_url: Option<String>,
-    pub(crate) clipboard_backend: Option<WasmClipboardBackend>,
-    pub(crate) use_display_control: bool,
+struct ConnectParams {
+    ws: WebSocket,
+    config: connector::Config,
+    proxy_auth_token: String,
+    destination: String,
+    pcb: Option<String>,
+    kdc_proxy_url: Option<String>,
+    clipboard_backend: Option<WasmClipboardBackend>,
+    use_display_control: bool,
 }
 
 async fn connect(
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -316,18 +316,16 @@ impl SessionBuilder {
 
         let use_display_control = self.0.borrow().use_display_control;
 
-        let (connection_result, ws) = connect(
-            ConnectParams{
-                ws,
-                config,
-                proxy_auth_token:auth_token,
-                destination,
-                pcb,
-                kdc_proxy_url,
-                clipboard_backend: clipboard.as_ref().map(|clip| clip.backend()),
-                use_display_control,
-            }
-        )
+        let (connection_result, ws) = connect(ConnectParams {
+            ws,
+            config,
+            proxy_auth_token: auth_token,
+            destination,
+            pcb,
+            kdc_proxy_url,
+            clipboard_backend: clipboard.as_ref().map(|clip| clip.backend()),
+            use_display_control,
+        })
         .await?;
 
         info!("Connected!");
@@ -810,8 +808,7 @@ impl Session {
                 width,
                 height,
                 scale_factor,
-                physical_size: physical_width
-                    .and_then(|width| physical_height.map(|height| (width, height))),
+                physical_size: physical_width.and_then(|width| physical_height.map(|height| (width, height))),
             })
             .expect("send resize event to writer task");
     }
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -516,8 +516,7 @@ impl Session {
                             if width == 0 || height == 0 {
                                 warn!("Resize event ignored: width or height is zero");
                                 Vec::new()
-                            }
-                            else if let Some(response_frame) = active_stage.encode_resize(width, height, scale_factor, physical_size) {
+                            } else if let Some(response_frame) = active_stage.encode_resize(width, height, scale_factor, physical_size) {
                                 self.render_canvas.set_width(width);
                                 self.render_canvas.set_height(height);
                                 gui.resize(NonZeroU32::new(width).unwrap(), NonZeroU32::new(height).unwrap());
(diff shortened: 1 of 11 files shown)


crates/ironrdp-web/src/session.rs
```diff
@@ -832,14 +888,28 @@ async fn writer_task(rx: mpsc::UnboundedReceiver<Vec<u8>>, rdp_writer: WriteHalf
...
+
 async fn connect(
-    ws: WebSocket,
-    config: connector::Config,
-    proxy_auth_token: String,
-    destination: String,
-    pcb: Option<String>,
-    kdc_proxy_url: Option<String>,
-    clipboard_backend: Option<WasmClipboardBackend>,
+    ConnectParams {
```
irvingoujAtDevolution, line 903: Too much parameters, hence use struct

crates/ironrdp-web/src/session.rs
```diff
@@ -489,6 +511,22 @@ impl Session {
...
+                            if width == 0 || height == 0 {
+                                warn!("Resize event ignored: width or height is zero");
+                                Vec::new()
+                            }
+                            else if let Some(response_frame) = active_stage.encode_resize(width, height, scale_factor, physical_size) {
+                                self.render_canvas.set_width(width);
+                                self.render_canvas.set_height(height);
+                                gui.resize(NonZeroU32::new(width).unwrap(), NonZeroU32::new(height).unwrap());
+                                vec![ActiveStageOutput::ResponseFrame(response_frame?)]
+                            }else {
```
CBenoit, line 525:
```suggestion
                            } else {
```

crates/ironrdp-web/src/session.rs
```diff
@@ -489,6 +511,22 @@ impl Session {
                             active_stage.process_fastpath_input(&mut image, &events)
                                 .context("fast path input events processing")?
                         }
+                        RdpInputEvent::Resize { width, height, scale_factor, physical_size } => {
+                            debug!(width, height, scale_factor, "Resize event received");
+                            if width == 0 || height == 0 {
+                                warn!("Resize event ignored: width or height is zero");
+                                Vec::new()
+                            }
+                            else if let Some(response_frame) = active_stage.encode_resize(width, height, scale_factor, physical_size) {
```
CBenoit, line 520:
```suggestion
                            } else if let Some(response_frame) = active_stage.encode_resize(width, height, scale_factor, physical_size) {
```

crates/ironrdp-web/src/session.rs
```diff
@@ -832,14 +888,28 @@ async fn writer_task(rx: mpsc::UnboundedReceiver<Vec<u8>>, rdp_writer: WriteHalf
...
+pub(crate) struct ConnectParams {
+    pub(crate) ws: WebSocket,
+    pub(crate) config: connector::Config,
+    pub(crate) proxy_auth_token: String,
+    pub(crate) destination: String,
+    pub(crate) pcb: Option<String>,
+    pub(crate) kdc_proxy_url: Option<String>,
+    pub(crate) clipboard_backend: Option<WasmClipboardBackend>,
+    pub(crate) use_display_control: bool,
+}
```
CBenoit, line 900:
```suggestion
struct ConnectParams {
    ws: WebSocket,
    config: connector::Config,
    proxy_auth_token: String,
    destination: String,
    pcb: Option<String>,
    kdc_proxy_url: Option<String>,
    clipboard_backend: Option<WasmClipboardBackend>,
    use_display_control: bool,
}
```

web-client/iron-remote-gui/src/interfaces/UserInteraction.ts
```diff
@@ -31,4 +32,6 @@ export interface UserInteraction {
     setCursorStyleOverride(style: string | null): void;
 
     onSessionEvent(callback: (event: SessionEvent) => void): void;
+
+    resizeDynamic(width: number, height: number, scale?: number): void;
```
CBenoit, line 36: naming: Why "Dymanic"? What about naming it `resize`?

e.g.: `session.resize(width, height)`

web-client/iron-remote-gui/src/iron-remote-gui.svelte
```diff
@@ -445,6 +445,11 @@
             scaleSession(s);
         });
 
+        wasmService.dynamicResize.subscribe((evt) => {
+            loggingService.info('Dynamic resize!');
```
CBenoit, line 449: Leftover?

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -57,6 +57,11 @@ export class WasmBridgeService {
     sessionObserver: Observable<SessionEvent> = this.sessionEvent.asObservable();
     scaleObserver: Observable<ScreenScale> = this.scale.asObservable();
 
+    dynamicResize = new Subject<{
+        width: number;
+        height: number;
+    }>();
```
CBenoit, line 63: suggestion: Any reason for using a `Subject` instead of an `Observable` here?

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -131,6 +136,7 @@ export class WasmBridgeService {
         desktopSize?: IDesktopSize,
         preConnectionBlob?: string,
         kdc_proxy_url?: string,
+        use_display_control = false,
```
CBenoit, line 139: suggestion: What about using `true` as the default value?

CBenoit (commented):
Nice work!

General comment: you don’t need to put the word "dynamic" everywhere "resize" is used. I think it only makes sense as an option name like in "enable dynamic resizing".

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
irvingoujAtDevolution: Observable constructor require a callback, where with subject I could simply do dynamicResize.next.

web-client/iron-remote-gui/src/interfaces/UserInteraction.ts
irvingoujAtDevolution: will update for public API

web-client/iron-remote-gui/src/iron-remote-gui.svelte
pacmancoder: [suggestion]: Maybe add width/height values to this log message?

pacmancoder (approved):
LGTM! 🚀

web-client/iron-remote-gui/src/iron-remote-gui.svelte
CBenoit: I also want to suggest the `debug` or `trace` level instead of `info`. It’s typically not something we need to log at the `info` level.

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
CBenoit: I see!! Can you add that as an in-code comment, please?

CBenoit (approved):
LGTM too. Can you address the two unresolved comments and merge?

web-client/iron-remote-gui/src/iron-remote-gui.svelte
irvingoujAtDevolution: updated, IronRDP does not have debug logging for now, only info and error.

I'll update to create a level logging in another PR

PR #558, opened 2024-10-09
feat(server): credentials (with credssp)

elmarco:
Modify Acceptor to take optionally Credentials at construct time.

The credentials can be changed via ServerEvent::SetCredentials.

Unauthorized connections are no longer accepted.

However, I don't know how to report the error to the client correctly. For now it returns a SetErrorInfo/DeniedConnection, but that doesn't seem appropriate during the connection sequence.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,9 +78,8 @@ num-traits = "0.2"
 # Safer unsafe
 unsafe_op_in_unsafe_fn = "warn"
 invalid_reference_casting = "warn"
-pointer_structural_match = "warn"
 # Correctness
-unused_tuple_struct_fields = "warn"
+dead_code = "warn"
 # Style, readability
 # https://quinedot.github.io/rust-learning/dont-hide.htm
 elided_lifetimes_in_paths = "warn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ expect-test = "1"
 png = "0.17"
 proptest = "1.4"
 rstest = "0.18"
-sspi = "0.13"
+sspi = "0.14.1"
 tracing = { version = "0.1", features = ["log"] }
 thiserror = "1.0"
 windows = "0.58"
--- a/crates/ironrdp-acceptor/src/connection.rs
+++ b/crates/ironrdp-acceptor/src/connection.rs
@@ -27,7 +27,7 @@ const IO_CHANNEL_ID: u16 = 1003;
 const USER_CHANNEL_ID: u16 = 1002;
 
 pub struct Acceptor {
-    state: AcceptorState,
+    pub(crate) state: AcceptorState,
     security: nego::SecurityProtocol,
     io_channel_id: u16,
     user_channel_id: u16,
@@ -35,7 +35,7 @@ pub struct Acceptor {
     server_capabilities: Vec<CapabilitySet>,
     static_channels: StaticChannelSet,
     saved_for_reactivation: AcceptorState,
-    creds: Option<Credentials>,
+    pub(crate) creds: Option<Credentials>,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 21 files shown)


Cargo.toml
```diff
@@ -78,9 +78,8 @@ num-traits = "0.2"
 # Safer unsafe
 unsafe_op_in_unsafe_fn = "warn"
 invalid_reference_casting = "warn"
-pointer_structural_match = "warn"
 # Correctness
-unused_tuple_struct_fields = "warn"
+dead_code = "warn"
```
CBenoit, line 82: nitpick: dead_code is warn-by-default, so we can remove it from our Cargo.toml

Cargo.toml
```diff
@@ -61,7 +61,7 @@ expect-test = "1"
 png = "0.17"
 proptest = "1.4"
 rstest = "0.18"
-sspi = "0.13"
+sspi = "0.14.1"
```
CBenoit, line 64: suggestion: Do not specify the patch version to avoid over-constraining. It’s technically correct here since we don’t really want to accept 0.14.0, but dependabot will keep the patch number around forever otherwise. 
```suggestion
sspi = "0.14"
```

crates/ironrdp/examples/server.rs
```diff
@@ -106,10 +133,23 @@ fn setup_logging() -> anyhow::Result<()> {
     Ok(())
 }
 
-fn acceptor(cert_path: &str, key_path: &str) -> anyhow::Result<TlsAcceptor> {
+fn acceptor(cert_path: &str, key_path: &str) -> anyhow::Result<(TlsAcceptor, Vec<u8>)> {
```
CBenoit, line 136: nitpick: Maybe use a struct to name the values. `Vec<u8>` on its own is not telling much.

```rust
struct AcceptResult {
    tls_acceptor: TlsAcceptor,
    public_key: Vec<u8>,
}

fn acceptor(cert_path: &str, key_path: &str) -> anyhow::Result<AcceptResult> { … }
```

or

```rust
struct PublicKey(Vec<u8>);

fn acceptor(cert_path: &str, key_path: &str) -> anyhow::Result<(TlsAcceptor, PublicKey)> { … }
```

But it’s not too much of a big deal since it’s an example. Feel free to address or not.

crates/ironrdp-server/src/server.rs
```diff
@@ -44,13 +44,16 @@ pub struct RdpServerOptions {
 pub enum RdpServerSecurity {
     None,
     Tls(TlsAcceptor),
+    // for both hybrid + hybrid-ex
```
CBenoit, line 47: suggestion: Could be part of the public documentation.
```suggestion
    /// Used for both hybrid + hybrid-ex
```

crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -438,6 +521,22 @@ impl Sequence for Acceptor {
 
                 debug!(message = ?client_info, "Received");
 
+                if !protocol.intersects(nego::SecurityProtocol::HYBRID | nego::SecurityProtocol::HYBRID_EX) {
+                    let creds = client_info.client_info.credentials;
+
+                    if self.creds.as_ref().map_or(true, |srv_creds| srv_creds != &creds) {
+                        // how authorization should be denied with standard RDP security?
```
CBenoit, line 528: nitpick: Maybe we should emphasize this way?
```suggestion
                        // FIXME: How authorization should be denied with standard RDP security?
                        // Since standard RDP security is not a priority, we just send a ServerDeniedConnection ServerSetErrorInfo PDU.
```

CBenoit (commented):
Looking good to me! Just waiting for `sspi v0.14.1` so we can update the Cargo.lock file with it :)

CBenoit:
@elmarco I think you need to run cargo build/check/clippy once to update the Cargo.lock :)

pacmancoder (approved):

CBenoit:
I’m fixing the macOS warnings, and will merge the PR as soon as they are resolved.

CBenoit (approved):

elmarco:
thanks

PR #563, opened 2024-10-25
fix(acceptor): raise credssp error to caller

elmarco:


crates/ironrdp-acceptor/src/lib.rs
```diff
@@ -192,6 +192,8 @@ where
             .await
             .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
     }
+
+    result?;
```
CBenoit, line 196:
```suggestion
    result?;

```

crates/ironrdp-acceptor/src/lib.rs
```diff
@@ -172,16 +172,16 @@ where
...
+    let result = credssp_loop(framed, acceptor, buf, client_computer_name, public_key, kerberos_config).await;
 
     if protocol.intersects(nego::SecurityProtocol::HYBRID_EX) {
         trace!(?result, "HYBRID_EX");
+
+        let result = if result.is_ok() {
+            EarlyUserAuthResult::Success
+        } else {
+            EarlyUserAuthResult::AccessDenied
+        };
```
CBenoit, line 184:
```suggestion
        };

```

CBenoit (approved):
Good catch! Thank you!

PR #565, opened 2024-10-28
Various README & docs improvements

elmarco:
Like other keys, readme is inherited via {key}.workspace = true


CBenoit (commented):
I think we can’t do that, because it will attempt to take the root README.md instead of the README.md in the folder of the crate.

```shell
IronRDP/crates/ironrdp-core $ cargo package
warning: readme `../../README.md` appears to be a path outside of the package, but there is already a file named `README.md` in the root of the package. The archived crate will contain the copy in the root of the package. Update the readme to point to the path relative to the root of the package to remove this warning.
```

CBenoit:
> Ah indeed, I didn't realize we had per-crate README.md. But not always, for example ironrdp-error.

Oh you are right. The intention is rather to have one README.md per crate, so do you think you could instead replace the workspace inheritance by a basic README.md file for each of these crates?

CBenoit (approved):
Thank you very much!

PR #570, opened 2024-11-04
fix(server): pick one selected protocol

elmarco:
The spec says we should have "the" selected protocol, and mstsc expects that.

### DIFF
--- a/crates/ironrdp-acceptor/src/connection.rs
+++ b/crates/ironrdp-acceptor/src/connection.rs
@@ -8,6 +8,7 @@ use ironrdp_connector::{
 use ironrdp_core::decode;
 use ironrdp_core::WriteBuf;
 use ironrdp_pdu as pdu;
+use ironrdp_pdu::nego::SecurityProtocol;
 use ironrdp_pdu::x224::X224;
 use ironrdp_svc::{StaticChannelSet, SvcServerProcessor};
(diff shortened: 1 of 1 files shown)


crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -310,7 +311,16 @@ impl Sequence for Acceptor {
...
-                let protocol = requested_protocol & self.security;
+                let protocols = requested_protocol & self.security;
+                let protocol = if protocols.intersects(SecurityProtocol::HYBRID_EX) {
+                    SecurityProtocol::HYBRID_EX
+                } else if protocols.intersects(SecurityProtocol::HYBRID) {
+                    SecurityProtocol::HYBRID
+                } else if protocols.intersects(SecurityProtocol::SSL) {
+                    SecurityProtocol::SSL
+                } else {
+                    return Err(ConnectorError::general("Failed to negotiate security protocol"));
```
CBenoit, line 322: style: https://github.com/Devolutions/IronRDP/blob/3e738a96ed5952553bfa8705c7242d83989a9276/STYLE.md#formatting-of-error-messages
```suggestion
                    return Err(ConnectorError::general("failed to negotiate security protocol"));
```

CBenoit (approved):
Makes sense! Looking good to me!

PR #571, opened 2024-11-04
Speed up RemoteFX encoding

elmarco:
- introduce rfx encoding benchmark with criterion
- speed up a bit DWT with const generics
- use rayon to parallel encode each tile
- speed up a bit rgb->yuv conversion

Unfortunately, it's not simple to use hand-written assembly from a project like yuvutils, because RDP uses odd bias and precision. Something left for the another day.

### DIFF
--- a/crates/ironrdp-graphics/src/color_conversion.rs
+++ b/crates/ironrdp-graphics/src/color_conversion.rs
@@ -29,28 +29,44 @@ where
     }
 }
 
-fn pixel_format_to_rgb_fn(format: PixelFormat) -> fn(&[u8]) -> Rgb {
+fn xrgb_to_rgb(pixel: &[u8]) -> Rgb {
+    Rgb {
+        r: pixel[1],
+        g: pixel[2],
+        b: pixel[3],
+    }
+}
+
+fn xbgr_to_rgb(pixel: &[u8]) -> Rgb {
+    Rgb {
+        b: pixel[1],
+        g: pixel[2],
+        r: pixel[3],
+    }
+}
+
+fn bgrx_to_rgb(pixel: &[u8]) -> Rgb {
+    Rgb {
+        b: pixel[0],
+        g: pixel[1],
+        r: pixel[2],
+    }
+}
+
+fn rgbx_to_rgb(pixel: &[u8]) -> Rgb {
+    Rgb {
+        r: pixel[0],
+        g: pixel[1],
+        b: pixel[2],
+    }
+}
+
+const fn pixel_format_to_rgb_fn(format: PixelFormat) -> fn(&[u8]) -> Rgb {
     match format {
-        PixelFormat::ARgb32 | PixelFormat::XRgb32 => |pixel: &[u8]| Rgb {
-            r: pixel[1],
-            g: pixel[2],
-            b: pixel[3],
-        },
-        PixelFormat::ABgr32 | PixelFormat::XBgr32 => |pixel: &[u8]| Rgb {
-            b: pixel[1],
-            g: pixel[2],
-            r: pixel[3],
-        },
-        PixelFormat::BgrA32 | PixelFormat::BgrX32 => |pixel: &[u8]| Rgb {
-            b: pixel[0],
-            g: pixel[1],
-            r: pixel[2],
-        },
-        PixelFormat::RgbA32 | PixelFormat::RgbX32 => |pixel: &[u8]| Rgb {
-            r: pixel[0],
-            g: pixel[1],
-            b: pixel[2],
-        },
+        PixelFormat::ARgb32 | PixelFormat::XRgb32 => xrgb_to_rgb,
+        PixelFormat::ABgr32 | PixelFormat::XBgr32 => xbgr_to_rgb,
+        PixelFormat::BgrA32 | PixelFormat::BgrX32 => bgrx_to_rgb,
+        PixelFormat::RgbA32 | PixelFormat::RgbX32 => rgbx_to_rgb,
     }
 }
 
--- a/crates/ironrdp-graphics/src/color_conversion.rs
+++ b/crates/ironrdp-graphics/src/color_conversion.rs
@@ -135,15 +135,12 @@ pub fn to_64x64_ycbcr_tile(
     height: usize,
     stride: usize,
     format: PixelFormat,
-    y: &mut [i16],
-    cb: &mut [i16],
-    cr: &mut [i16],
+    y: &mut [i16; 64 * 64],
+    cb: &mut [i16; 64 * 64],
+    cr: &mut [i16; 64 * 64],
 ) {
     assert!(width <= 64);
     assert!(height <= 64);
-    assert_eq!(y.len(), 64 * 64);
-    assert_eq!(cb.len(), 64 * 64);
-    assert_eq!(cr.len(), 64 * 64);
 
     let to_rgb = pixel_format_to_rgb_fn(format);
     let bpp = format.bytes_per_pixel() as usize;
--- a/crates/ironrdp-graphics/src/color_conversion.rs
+++ b/crates/ironrdp-graphics/src/color_conversion.rs
@@ -17,7 +17,7 @@ pub fn ycbcr_to_bgra(input: YCbCrBuffer<'_>, mut output: &mut [u8]) -> io::Resul
 
 fn iter_to_ycbcr<'a, I, C>(input: I, y: &mut [i16], cb: &mut [i16], cr: &mut [i16], conv: C)
 where
-    I: IntoIterator<Item = &'a [u8]>,
+    I: ExactSizeIterator<Item = &'a [u8]>,
     C: Fn(&[u8]) -> Rgb,
 {
     for (i, pixel) in input.into_iter().enumerate() {
@@ -141,6 +141,12 @@ impl<'a> Iterator for TileIterator<'a> {
         }
 
         Some(&self.slice[pos..pos + self.bpp])
+    }
+}
+
+impl ExactSizeIterator for TileIterator<'_> {
+    fn len(&self) -> usize {
+        64 * 64
     }
 }
 
--- a/crates/ironrdp-graphics/src/color_conversion.rs
+++ b/crates/ironrdp-graphics/src/color_conversion.rs
@@ -254,24 +254,33 @@ impl From<Rgb> for YCbCr {
         // terms need to be scaled by << 5 we simply scale the final
         // sum by >> 10
         const DIVISOR: f32 = (1 << 15) as f32;
+        const Y_R: i32 = (0.299 * DIVISOR) as i32;
+        const Y_G: i32 = (0.587 * DIVISOR) as i32;
+        const Y_B: i32 = (0.114 * DIVISOR) as i32;
+        const CB_R: i32 = (0.168_935 * DIVISOR) as i32;
+        const CB_G: i32 = (0.331_665 * DIVISOR) as i32;
+        const CB_B: i32 = (0.500_59 * DIVISOR) as i32;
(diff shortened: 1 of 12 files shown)


crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -1,5 +1,7 @@
-pub(crate) mod bitmap;
-pub(crate) mod rfx;
+#![allow(unreachable_pub)]
```
CBenoit, line 1: suggestion: Can you add a comment explaining the reason for silencing this lint? In this case, we need these private items to be public for re-exporting via the bench feature.

crates/ironrdp-graphics/Cargo.toml
```diff
@@ -11,6 +11,9 @@ authors.workspace = true
 keywords.workspace = true
 categories.workspace = true
 
+[features]
+bench = []
```
CBenoit, line 15: suggestion: Make it clear that this feature is not part of the public API of the crate.
```suggestion
# Internal (PRIVATE!) features. Do not rely on these whatsoever. They may disappear at anytime.
__bench = []
```

crates/ironrdp-graphics/src/color_conversion.rs
```diff
@@ -15,11 +15,12 @@ pub fn ycbcr_to_bgra(input: YCbCrBuffer<'_>, mut output: &mut [u8]) -> io::Resul
     Ok(())
 }
 
-fn iter_to_ycbcr<'a, I, C>(input: I, y: &mut [i16], cb: &mut [i16], cr: &mut [i16], conv: C)
+fn iter_to_ycbcr<'a, const HINT: usize, I, C>(input: I, y: &mut [i16], cb: &mut [i16], cr: &mut [i16], conv: C)
 where
-    I: IntoIterator<Item = &'a [u8]>,
+    I: ExactSizeIterator<Item = &'a [u8]>,
     C: Fn(&[u8]) -> Rgb,
```
CBenoit, line 21: question: Can you explain as a in-code comment the purpose of the `HINT` generic parameter? I believe it's a to help diagnostics problems, but is there any rationale for the value it must be set to? Would it make sense to use `debug_assert!` instead?

crates/ironrdp-graphics/src/color_conversion.rs
```diff
@@ -235,30 +259,40 @@ impl From<YCbCr> for Rgb {
 }
 
 impl From<Rgb> for YCbCr {
+    #[inline(always)]
```
CBenoit, line 262: question: Can you add a brief comment explaining that this is backed by benchmarks and the performance improvement that was measured if you still have the numbers.

crates/ironrdp-server/src/encoder/rfx.rs
```diff
@@ -1,6 +1,9 @@
+#![allow(unreachable_pub)]
```
CBenoit, line 1: Likewise, please accompany with a comment

crates/ironrdp-server/src/lib.rs
```diff
@@ -28,3 +28,23 @@ pub use handler::*;
...
+}
+
+#[macro_export]
+macro_rules! time_warn {
+    ($context:expr, $threshold:expr, $op:expr) => {{
+        let start = std::time::Instant::now();
+        let result = $op;
+        let duration = start.elapsed().as_millis();
+        if duration > $threshold {
+            ::tracing::warn!("{} took {} ms!", $context, duration);
```
CBenoit, line 46: question: Should we try to backoff in case the server is weak or a regression is shipped?

Example:
```rust
#[macro_export]
macro_rules! time_warn {
    ($context:expr, $threshold:expr, $op:expr) => {{
        #[cold]
        fn warn_log(context: &str, duration: Duration) {
            static COUNT: AtomicUsize = AtomicUsize::new(0);
            let current_count = COUNT.fetch_add(1);
            if current_count < 50 || current_count % 100 == 0 {
                ::tracing::warn!("{context} took {duration} ms! (count: {current_count})");
            }
        }

        let start = std::time::Instant::now();
        let result = $op;
        let duration = start.elapsed().as_millis();
        if duration > $threshold {
            warn_log($context, duration);
        }
        result
    }};
}
```

Generating a unique function per `time_warn!` is intended so the counter is not shared.

CBenoit (commented):
Nice work!! Do you have some numbers from criterion? How much the performance was improved?

suggestion: Put the benchmarks in a separate crate `ironrdp-benchmarks` or `benchmarks`.

CBenoit:
> fwiw, I opened [awxkee/yuvutils-rs#3](https://github.com/awxkee/yuvutils-rs/issues/3)

Thank you for following up on this. I'll watch the issue.

crates/ironrdp-server/src/encoder/mod.rs
elmarco: yes, I couldn't find a better way. Do you have another idea?

crates/ironrdp-graphics/Cargo.toml
elmarco: Right.. do you think there is a standard for that? _ or x- prefix is fine for me.

crates/ironrdp-graphics/src/color_conversion.rs
elmarco: I tried to teach the iterator how many times it can iterate.. assert input.len() == HINT. That didn't help. I should probably remove that hint.

crates/ironrdp-server/src/lib.rs
elmarco: sounds good to me. There is probably a crate for that already ;)

crates/ironrdp-graphics/Cargo.toml
CBenoit: Prefixing with `__` (two underscores) is what I've seen so far as the de-facto standard.

reqwest is doing that:
https://github.com/seanmonstar/reqwest/blob/6e214133ab45ca2e4de493f13e37bb246d05b71a/Cargo.toml#L87-L99

And we followed this style in sspi-rs:
https://github.com/Devolutions/sspi-rs/blob/00158fb8e1425a131960b08469f26f04bfe74d60/Cargo.toml#L71-L75

stylish:
https://github.com/Nullus157/stylish-rs/blob/20da78181ab037993da3291b2736b5ad8908cec3/Cargo.toml#L36

By the way, features prefixed by `_` (even a single one) are considered private and hidden by docs.rs:
https://github.com/rust-lang/docs.rs/blob/a7f23414a35f951c96a1059a858c99ce36bcfe45/src/db/types.rs#L16

So either `_bench` or `__bench`.

The proper solution is still undergoing the RFC process: https://github.com/rust-lang/rfcs/pull/3487

crates/ironrdp-server/src/lib.rs
CBenoit: If you find something suitable we can consider using it :slightly_smiling_face:

crates/ironrdp-server/src/encoder/mod.rs
CBenoit: In this case, you could provide a proxy function using the private items for you while keeping them hidden.
This is not always practical depending on what you are trying to do, but for the UpdateEncoder it works decently well.

```rust
// lib.rs
#[cfg(feature = "__bench")]
pub mod bench {
    pub mod encoder {
        pub fn update_encoder<’a>(
            bitmap: &BitmapUpdate,
            quant: rfx::Quant,
            entropy_algorithm: rfx::EntropyAlgorithm,
            tile_x: usize,
            tile_y: usize,
            buf: &'a mut [u8]
        ) -> Result<EncodedTile<'a>, RlgrError> {
            let enc = rfx::UpdateEncoder::new(&bitmap, quant, algo);
            enc.encode(0, 0, &mut buf)
        }
    }
}

// bench.rs
    let mut buf = vec![0; 64 * 64 * 3];
    c.bench_function("rfxenc", |b| {
        b.iter(|| {
            let _ = update_encoder(&bitmap, quant, algo, 0, 0, &mut buf);
        })
    });
```

elmarco:
> Nice work!! Do you have some numbers from criterion? How much the performance was improved?


Well, criterion is huge time saver depending on your HW, but it still uses same amount of compute:
rfx_enc                 time:   [6.4386 ms 6.4813 ms 6.5296 ms]
                        change: [-85.105% -85.013% -84.895%] (p = 0.00 < 0.05)
                        Performance has improved.


> suggestion: Put the benchmarks in a separate crate `ironrdp-benchmarks` or `benchmarks`.

what's the rationale to put tests and benchmark in various crates? We try to ensure a common framework for the various crates that way?

crates/ironrdp-server/src/encoder/rfx.rs
elmarco: gone

crates/ironrdp-server/src/lib.rs
elmarco: there is https://lib.rs/crates/logging_timer, but it's not quite the same :)

crates/ironrdp-server/src/encoder/rfx.rs
elmarco: gone

crates/ironrdp-server/src/encoder/mod.rs
elmarco: done

crates/ironrdp-graphics/src/color_conversion.rs
elmarco: It's already inlined by rust by default. not sure if we want to be explicit about it here. My hope is that we can leverage yuvutils some time later anyway.

elmarco: dropped the patch for now

CBenoit:
> > Nice work!! Do you have some numbers from criterion? How much the performance was improved?
> 
> Well, criterion is huge time saver depending on your HW, but it still uses same amount of compute: rfx_enc time: [6.4386 ms 6.4813 ms 6.5296 ms] change: [-85.105% -85.013% -84.895%] (p = 0.00 < 0.05) Performance has improved.

Good numbers! By "criterion" here, you mean "rayon" for parallelizing the encoding?

> > suggestion: Put the benchmarks in a separate crate `ironrdp-benchmarks` or `benchmarks`.
> 
> what's the rationale to put tests and benchmark in various crates? We try to ensure a common framework for the various crates that way?

For the tests, the rationale is explained at several places:
- ARCHITECTURE.md: https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#testing
- tests/main.rs: https://github.com/Devolutions/IronRDP/blob/58f31b88e06209a0cf24bb17213ae081438e3e57/crates/ironrdp-testsuite-core/tests/main.rs#L3-L12

I'll also add less objective opinions and personal tastes:
- The lint `unused_crate_dependencies` will detect less false-positives.
- We can put code useful across tests along in the library part of the crate, for instance the macros: https://github.com/Devolutions/IronRDP/blob/58f31b88e06209a0cf24bb17213ae081438e3e57/crates/ironrdp-testsuite-core/src/macros.rs
- The Cargo.toml is kept free from test only elements which are not relevant for the user (although I think it's typically fine)

For the benchmarks, in addition to the above arguments:
- It's easier for someone not super familiar with the codebase to discover about the benchmarks
- It's possible to see the list of all the benchmarks in the Cargo.toml
- As for the tooling, you have the autocompletion with `cargo bench --bench ^<TAB>` which also returns you the list of targets. (Assuming you have completers for that in your shell.)

We also have `ironrdp-fuzz` in the `fuzz` folder at the root of the workspace for similar reasons. It's also easier to have a single README.md documenting things at one easy to discover place.

That's about it for the rationale behind this suggestion.

CBenoit (approved):
LGTM! Really nice work on the performance.

PR #575, opened 2024-11-05
docs(now-proto): add missing NOW-PROTO docs

pacmancoder:
Somehow NOW-proto docs gone missing in `devolutions-gateway` repository 🤔 
I restored last docs state from https://github.com/Devolutions/devolutions-gateway/pull/913 (thank god GitHub keeps raw files from PRs)

Now, that `now-proto` crate has been moved to IronRDP in https://github.com/Devolutions/devolutions-gateway/pull/941/files , we could place documentation in IronRDP repo instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # Log files
 *.log
 
+# Coverage
+/docs/coverage
+
 # Editor/IDE files
 *~
 /tags
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,6 @@
 # Log files
 *.log
 
-# Coverage
-/docs
-
 # Editor/IDE files
 *~
 /tags
--- a/docs/NOW-spec.md
+++ b/docs/NOW-spec.md
@@ -1,0 +1,1601 @@
+
+# NOW-PROTO
+
+<style>
+    .byte-layout {
+        width: 100%;
+        table-layout: fixed;
+    }
+    .byte-layout th {
+        colspan: 1;
+        text-align: center;
+        vertical-align: bottom;
+    }
+    .byte-layout td {
+        text-align: center;
+    }
+</style>
+
+
+[[_TOC_]]
+
+
+# Messages
+
+## Transport
+
+The NOW virtual channel protocol use an RDP dynamic virtual channel ("Devolutions::Now::Agent") as a transport type.
+
+## Message Syntax
+
+The following sections specify the NOW protocol message syntax. Unless otherwise specified, all fields defined in this document use the little-endian format.
+
+### Common Structures
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -7,9 +7,6 @@
 # Log files
 *.log
 
-# Coverage
```
pacmancoder, line 10: Question (@CBenoit): Is coverage using `docs` directory for some reason, or is this an accidental .gitignore item? 🤔

CBenoit: Not accidental, this folder is created by the coverage task: https://github.com/Devolutions/IronRDP/blob/9470d2a7c2835e634f0a907db43d1f7165faccb3/xtask/src/cov.rs#L204
That's why there is a "Coverage" comment I think :slightly_smiling_face: 

Try to run the task and see if we can ignore `docs/coverage` only instead? Probably some other files are placed into the `docs/` folder so I'm unsure.
Alternatively, use something else than `docs/` for the coverage, and ignore that instead.

docs/NOW-spec.md
```diff
@@ -0,0 +1,1601 @@
+
```
CBenoit, line 1:
```suggestion
```

docs/NOW-spec.md
```diff
@@ -0,0 +1,1601 @@
...
+    }
+    .byte-layout th {
+        colspan: 1;
+        text-align: center;
+        vertical-align: bottom;
+    }
+    .byte-layout td {
+        text-align: center;
+    }
+</style>
```
CBenoit, line 17: Does this actually takes effects? Not sure if GitHub allows styling.

image

Or are you viewing from somewhere else?

docs/NOW-spec.md
```diff
@@ -0,0 +1,1601 @@
...
+
+
+[[_TOC_]]
```
CBenoit, line 20: Likewise, this have no effect on GitHub, and there is already a table of contents by clicking on the three dots at the top right of the markdown page
image

docs/NOW-spec.md
pacmancoder: Yep, It should be viewed in Markdown Monster, @awakecoding has written initial document here (see [this message](  https://devolutions.slack.com/archives/GVCJYNXBJ/p1718913383781139?thread_ts=1718744469.277199&cid=GVCJYNXBJ))

.gitignore
pacmancoder: I have tested `grcov` command, the docs are only placed inside "docs/coverage", so I added that folder to `.gitignore` instead of the whole `docs`.

CBenoit (approved):

PR #586, opened 2024-11-21
fix(connector): do not send a Cookie when using a smart card

CBenoit:


crates/ironrdp-connector/src/lib.rs
```diff
@@ -71,15 +71,15 @@ pub enum Credentials {
     },
     SmartCard {
         pin: String,
-        config: Option<Box<SmartCardIdentity>>,
+        config: Option<SmartCardIdentity>,
```
CBenoit, line 74: I also noticed this was (I believe) unnecessarily boxed.

PR #592, opened 2024-11-28
Some fixes and changes for example code 

elmarco:
- Bump some deps
- Fix build of screenshot
- Replace "bmp" with "image" crate

### DIFF
--- a/crates/ironrdp-connector/Cargo.toml
+++ b/crates/ironrdp-connector/Cargo.toml
@@ -30,9 +30,9 @@ rand_core = { version = "0.6", features = [
 sspi.workspace = true
 tracing.workspace = true
 url = "2.5"
-picky-asn1-der = "0.5.0"
-picky-asn1-x509 = "0.13.0"
-picky = "7.0.0-rc.10"
+picky-asn1-der = "0.5.1"
+picky-asn1-x509 = "0.14.1"
+picky = "7.0.0-rc.11"
 
 [lints]
 workspace = true
--- a/crates/ironrdp/Cargo.toml
+++ b/crates/ironrdp/Cargo.toml
@@ -56,6 +56,7 @@ async-trait = "0.1"
 bmp = "0.5"
 pico-args = "0.5"
 x509-cert = { version = "0.2", default-features = false, features = ["std"] }
+sspi = { workspace = true, features = ["network_client"] }
 tracing.workspace = true
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 tokio-rustls = "0.26"
@@ -68,7 +69,7 @@ all-features = true
 [[example]]
 name = "screenshot"
 doc-scrape-examples = true
-required-features = ["session"]
+required-features = ["session", "connector", "graphics"]
 
 [[example]]
 name = "server"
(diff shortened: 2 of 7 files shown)


crates/ironrdp/Cargo.toml
```diff
@@ -53,9 +53,10 @@ ironrdp-blocking.workspace = true
 ironrdp-cliprdr-native.workspace = true
 anyhow = "1"
 async-trait = "0.1"
-bmp = "0.5"
+image = { version = "0.25.5", default-features = false, features = ["png"] }
```
CBenoit, line 56: `image` is a rather big crate that I would rather not use just for outputting an image file in an example. (Even if we don’t enable all the features.) What is the reason for switching to `image` and outputting a png instead?

`bmp` may not be updated anymore, but it only depends on a single dependency which is not outdated (no new major bump), and it’s just about outputting a bmp file so I wouldn’t surprised if the crate was considered complete.

If you really want to switch to png images, what about using the [`png` crate](https://lib.rs/crates/png) directly as we don’t really need all the features of the `image` crate?

crates/ironrdp-connector/Cargo.toml
```diff
@@ -30,9 +30,9 @@ rand_core = { version = "0.6", features = [
 sspi.workspace = true
 tracing.workspace = true
 url = "2.5"
-picky-asn1-der = "0.5.0"
-picky-asn1-x509 = "0.13.0"
-picky = "7.0.0-rc.10"
+picky-asn1-der = "0.5.1"
+picky-asn1-x509 = "0.14.1"
```
CBenoit, line 34: suggestion: To avoid dependency over-constraining:
```suggestion
picky-asn1-der = "0.5"
picky-asn1-x509 = "0.14"
```

CBenoit (approved):
Thank you! I’ll merge now!

crates/ironrdp-connector/Cargo.toml
elmarco: It's not a so strict constaint, cargo will update to >=0.5.1 <0.6, but ok

elmarco: "image" is likely to be already downloaded by a rust dev & available in various distros. otoh, bmp is an outdated format.

It's true, that image is a bigger crate (cargo bloat):

 0.1%   0.3%   26.5KiB image
 0.0%   0.0%    3.4KiB bmp

But it's so marginal compared to:

 7.3%  16.8%    1.5MiB sspi
 4.8%  10.9% 1013.8KiB aws_lc_sys
 4.0%   9.2%  857.7KiB std
 3.1%   7.1%  663.3KiB reqwest
 2.1%   4.8%  441.8KiB rustls
 1.8%   4.2%  390.2KiB regex_syntax
 1.5%   3.5%  325.7KiB h2
 1.4%   3.3%  304.6KiB ironrdp_pdu
 1.4%   3.1%  290.3KiB ironrdp_connector
 1.1%   2.5%  235.2KiB hickory_proto
 1.1%   2.5%  232.9KiB picky
...

CBenoit: > It's not a so strict constaint, cargo will update to >=0.5.1 <0.6, but ok

Sure, but let’s avoid accumulate this kind of unnecessary constraints. I already got compilation errors due to crates specifying the patch version for every single dependency.

PR #596, opened 2024-12-04
docs(now-proto): now-proto improvements

pacmancoder:
This PR proposes `now-proto` improvements based on missing features found during my initial DevolutionsAgent/C# lib implementation. 

Rendered PDF: 
[NOW-spec.pdf](https://github.com/user-attachments/files/18028473/NOW-spec.pdf)


docs/NOW-spec.md
```diff
@@ -39,82 +38,6 @@ The following sections specify the NOW protocol message syntax. Unless otherwise
 
 Signed and unsigned integer encoding structures of various sizes.
 
-##### NOW_VARU16
```
pacmancoder, line 42: Removed unused core types

docs/NOW-spec.md
```diff
@@ -396,50 +123,33 @@ The NOW_STRING16 structure is used to represent variable-length strings of up to
...
-            <td colspan="32">...</td>
-        </tr>
-    </tbody>
-</table>
+**msgClass (1 byte)**: The message class.
 
-**len (1 byte)**: An unsigned 8-bit number containing the string length, excluding the null terminator. The maximum value is 31.
+| Flag                            | Meaning              |
+|---------------------------------|----------------------|
+| NOW_NEGOTIATION_MSG_CLASS_ID<br>0x10 | Protocol negotiation class |
```
pacmancoder, line 142: diff decided to break here (NOW_HEADER), but basically, new separate message class was added for the capabilities message (exec caps message was removed in favor of global capabilities)

docs/NOW-spec.md
```diff
@@ -396,50 +123,33 @@ The NOW_STRING16 structure is used to represent variable-length strings of up to
...
+| NOW_EXEC_MSG_CLASS_ID<br>0x13 | Exec message class |
 
-**str (variable)**: The UTF-8 encoded string excluding the null terminator.
+**msgType (1 byte)**: The message type, specific to the message class.
 
-##### NOW_STRING64
+**msgFlags (2 bytes)**: The message flags, specific to the message type and class.
 
-The NOW_STRING64 structure is used to represent variable-length strings of up to 63 characters that can easily fit within a fixed-size buffer of 64 bytes.
+#### NOW_STATUS
```
pacmancoder, line 151: Diff looks messy here, it is better to view `NOW_STATUS` changes in raw file/rendered PDF, however, here is the quick list of changes:
- Simplified NOW_STATUS overall structure, its not based on HRESULT now. 
- It contains a 32-bit code field instead of a 16-bit one; now it could fit the process exit code from any OS.
- Defines common error kinds - now-proto-specific errors, win32 errors, unix errors. Error kinds could be extended.
- Provides optional string message field for errors
- Defines more now-proto error kinds based on DVC server implementation experience
- Removed severity. Not its either success or error status.

docs/NOW-spec.md
```diff
@@ -538,134 +271,79 @@ The NOW_VARBUF structure is used to represent variable-length buffers.
...
-| NOW_SESSION_MSG_CLASS_ID<br>0x12 | Session message class |
-| NOW_EXEC_MSG_CLASS_ID<br>0x13 | Exec message class |
+**msgType (1 byte)**: The message type (NOW_NEGOTIATION_CAPSET_MSG_ID).
 
-**msgType (1 byte)**: The message type, specific to the message class.
+**msgFlags (2 bytes)**: Message flags.
 
-**msgFlags (2 bytes)**: The message flags, specific to the message type and class.
+| Flag | Meaning |
+|-------|---------|
```
pacmancoder, line 309: Capabilities were moved to a separate message class. Main changes:
- Added capsets for session & system commands
- Added major & minor protocol version fields
- Added possibility to specify session heartbeat timeout (new feature)

docs/NOW-spec.md
```diff
@@ -538,134 +271,79 @@ The NOW_VARBUF structure is used to represent variable-length buffers.
...
 
-**type (8 bits)**: The status type.
+| Flag | Meaning |
+|-------|---------|
+| NOW_CAP_EXEC_STYLE_RUN<br>0x00000001 | Generic "Run" execution style. |
+| NOW_CAP_EXEC_STYLE_PROCESS<br>0x00000002 | CreateProcess() execution style. |
+| NOW_CAP_EXEC_STYLE_SHELL<br>0x00000004 | System shell (.sh) execution style. |
+| NOW_CAP_EXEC_STYLE_BATCH<br>0x00000008 | Windows batch file (.bat) execution style. |
+| NOW_CAP_EXEC_STYLE_WINPS<br>0x00000010 | Windows PowerShell (.ps1) execution style. |
+| NOW_CAP_EXEC_STYLE_PWSH<br>0x00000020 | PowerShell 7 (.ps1) execution style. |
```
pacmancoder, line 340: removed AppleScipt style variant until its implemented in next protocol version.

docs/NOW-spec.md
```diff
@@ -538,134 +271,79 @@ The NOW_VARBUF structure is used to represent variable-length buffers.
...
+| NOW_CAP_SESSION_MSGBOX<br>0x00000004 | Message box command support. |
 
-**class (4 bits)**: Reserved. MUST be set to zero.
+**execCapset (4 bytes)**: Remote execution capabilities set.
 
-**type (8 bits)**: The status type.
+| Flag | Meaning |
+|-------|---------|
+| NOW_CAP_EXEC_STYLE_RUN<br>0x00000001 | Generic "Run" execution style. |
+| NOW_CAP_EXEC_STYLE_PROCESS<br>0x00000002 | CreateProcess() execution style. |
```
pacmancoder, line 336: [?] Removed `CMD` execution style, as the same operation could be achieved via `BATCH` style

docs/NOW-spec.md
```diff
@@ -910,7 +588,7 @@ The NOW_SESSION_MSGBOX_REQ_MSG is used to show a message box in the user session
 
 | Flag                                | Meaning                                 |
 |-------------------------------------|-----------------------------------------|
-| NOW_MSGBOX_FLAG_TITLE<br>0x00000001 | The title field contains a non-default value |
+| NOW_MSGBOX_FLAG_TITLE<br>0x00000001 | Set if `title` field is present |
```
pacmancoder, line 591: Guarded some optional fields via flags in several messages, so they could be skipped from serialization/deserialization

docs/NOW-spec.md
```diff
@@ -951,7 +629,7 @@ The NOW_SESSION_MSGBOX_RSP_MSG is a message sent in response to NOW_SESSION_MSGB
             <td colspan="32">requestId</td>
         </tr>
         <tr>
-            <td colspan="32">response</td>
+            <td colspan="32">status</td>
```
pacmancoder, line 632: Use new NOW_STATUS instead of simple response value to report errors if message box request has failed

docs/NOW-spec.md
```diff
@@ -1597,8 +1265,52 @@ The NOW_EXEC_PWSH_MSG message is used to execute a remote PowerShell 7 (pwsh) co
...
-**executionPolicy (variable)**: A NOW_VARSTR structure, same as with NOW_EXEC_WINPS_MSG.
+**command (variable)**: A NOW_VARSTR structure containing the command to execute.
 
-**configurationName (variable)**: A NOW_VARSTR structure, same as with NOW_EXEC_WINPS_MSG.
+**executionPolicy (variable, optional)**: A NOW_VARSTR structure, same as with NOW_EXEC_WINPS_MSG.
 
-**command (variable)**: A NOW_VARSTR structure containing the command to execute.
+**configurationName (variable, optional)**: A NOW_VARSTR structure, same as with NOW_EXEC_WINPS_MSG.
+
+#### NOW_EXEC_HEARTBEAT_MSG
```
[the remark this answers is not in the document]
CBenoit: Do we really need an heartbeat for that? Since this is a virtual channel via RDP transport, my understanding is that the (RDP) session is necessarily alive. Am I missing something?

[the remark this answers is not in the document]
pacmancoder: Hmm, I think you are right, a _per-session_ periodical heartbeat indeed is unnecessary. My reasoning for adding it was to detect when the specific session processing has failed unexpectedly without sending ExecResult, to avoid unresponsive sessions, but as sessions are run as separate threads, it should be possible to detect when a thread fails and send ExecResult right away.

However, what if the _whole DVC_ has become unresponsive? e.g. dvc processing logic crashed without closing channel. Maybe I am overthinking it, and this corner case could be ignored.

Also, we still need to know that the agent has registered the exec session and that execution has started (Currecntly there is no "success" response after ExecSession request). Therefore we still need to send some message after initial exec request to start the session was sent. Maybe something like `ExecSessionStartedRsp` message with session_id will be sufficient. We need this message to show the session state correctly on the client. (Starting -> Running -> Finished)

@awakecoding thoughts?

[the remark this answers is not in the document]
CBenoit: I see! Sounds like a heartbeat message could be good then. At least if the DVC is crashed, we have a way to detect that.

docs/NOW-spec.md
```diff
@@ -480,52 +233,32 @@ The NOW_STRING128 structure is used to represent variable-length strings of up t
...
+| NOW_NEGOTIATION_CAPSET_MSG_ID<br>0x01 | NOW_NEGOTIATION_CAPSET_MSG |
 
-#### NOW_BUFFER
+#### NOW_NEGOTIATION_CAPSET_MSG
 
-##### NOW_VARBUF
+This message is first set by the server side, to advertise capabilities.
 
-The NOW_VARBUF structure is used to represent variable-length buffers.
+Received server message should be downgraded by the client (remove non-intersecting capabilities) and sent back to the server at the start of DVC channel communications. DVC channel should be closed if protocol
```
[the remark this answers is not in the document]
CBenoit: I would suggest the client like in  TLS (ClientHello) and RDP itself, but it's typically not super relevant as other protocols like VNC do the other way around.

docs/NOW-spec.pdf
CBenoit, line 1: I suggest we don’t upload the PDF to the repository, given that [the markdown file is already showing not bad at all in GitHub natively](https://github.com/Devolutions/IronRDP/blob/proto/now-proto-improvements/docs/NOW-spec.md). It’s also possible to read the document using Markdown Monster. It’s one binary file less to maintain up to date.

pacmancoder: Agreed 👍

docs/NOW-spec.md
pacmancoder: @awakecoding Is my reasoning correct, or do we still need a separate CMD execution style? If yes - what is the difference between CMD and BATCH (besides running commands directly from cmd arguments or separate `.bat` file)

docs/NOW-spec.md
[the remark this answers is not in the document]
pacmancoder: Removed session-specific heartbeat, added heartbeat for a whole channel; added separate ExecStarted message

awakecoding (approved):

pacmancoder:
@CBenoit FYI: I made some modifications to proto after discussion with Marc-Andre in slack yesterday. Changes:
- Removed concept of "optionally serializable" fields. If the field is optional (soecufued by flag in header), it will contain a default value instead of skipping serialization.
NOW_STATUS now does not contain an operation-specific success value. The code contains either an error code or 0 on a successful operation.
- Variable-length fields are always specified last in message.

I am ready to merge this PR and move this doc to a separate repo.

CBenoit (approved):
All of this sounds good!

My last suggestion would be to use `xtask` instead of `make.ps1`, but you could modify that when you move the document to the `now-proto` repository!

PR #597, opened 2024-12-04
fix(session): avoid panics when First fragmentation is missing

probakowski:
When we hit `Last` fragmentation PDU and there was no `First` before `unwrap` will cause panic.
With this change we will simply skip invalid/incomplete frames the same way we do when we encounter `Single` after `First` or `Next` without `Last`.


CBenoit (approved):
Thank you!

PR #623, opened 2024-12-17
feat(server): make TlsIdentityCtx accept PEM files

elmarco:
This is in general more convenient than DER files.


CBenoit (approved):
Indeed! Thank you!

elmarco:
argh, and I broke the build with features=helper.. my bad.. I will try to fix it asap, but I am off for a few days..

CBenoit:
> argh, and I broke the build with features=helper.. my bad.. I will try to fix it asap, but I am off for a few days..

Oops! We should update xtask to check for that properly. No problem, this can be wait as you are currently the only consumer of ironrdp-server.

PR #630, opened 2025-01-07
fix(server): drop unexpected PDUs during deactivation-reactivation

elmarco:
The current behaviour of handling unmatched PDUs in fn read_by_hint() isn't good enough. An unexpected PDUs may be received and fail to be decoded during Acceptor::step().

Change the code to simply drop unexpected PDUs (as opposed to attempting to replay the unmatched leftover, which isn't clearly needed) and make single_sequence_step_read() resilient to decode issues when ignore_unexpected_pdu. This require Acceptor::step() to be idempotent on error, so we have to clone the state fields, instead of taking the state.


CBenoit (commented):
Thank you!

> Change the code to simply drop unexpected PDUs (as opposed to attempting to replay the unmatched leftover, which isn't clearly needed)

Sounds reasonable!

> This require Acceptor::step() to be idempotent on error, so we have to clone the state fields, instead of taking the state.

To be honest, I’m not very inclined into making this change, as I believe there is a better approach.

Before going further, I want to check my understanding at this place: https://github.com/Devolutions/IronRDP/pull/630/files#r1905600306

crates/ironrdp-async/src/framed.rs
```diff
@@ -236,21 +230,29 @@ pub async fn single_sequence_step<S>(
...
+    let written = match single_sequence_step_read(framed, sequence, buf).await {
+        Ok(written) => written,
+        Err(err) => {
+            return if ignore_unexpected_pdu && matches!(err.kind(), ConnectorErrorKind::Decode(_)) {
+                Ok(())
+            } else {
+                Err(err)
+            };
+        }
+    };
```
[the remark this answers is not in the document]
elmarco: Yes, the received PDU matches the expected hint but fails to be decoded. This will break the connection during deact-react. I have no better idea how we could flush/sync/fence the client PDUs before.

crates/ironrdp-async/src/framed.rs
```diff
@@ -179,8 +175,6 @@ where
                     let bytes = self.read_exact(length).await?.freeze();
                     if matched {
                         return Ok(bytes);
-                    } else if let Some(ref mut unmatched) = unmatched {
-                        unmatched.push(bytes);
                     } else {
                         warn!("Received and lost an unexpected PDU");
```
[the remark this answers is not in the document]
elmarco: ok

elmarco:
I added a few other deact-react fixes to this PR

crates/ironrdp-dvc/src/server.rs
```diff
@@ -163,6 +163,7 @@ impl SvcProcessor for DrdynvcServer {
                 let channel_id = data.channel_id();
                 let c = self.channel_by_id(channel_id).map_err(|e| decode_err!(e))?;
                 if c.state != ChannelState::Opened {
+                    debug!("invalid channel state {channel_id}:{state:?}", state = c.state);
```
CBenoit, line 166: style: https://github.com/Devolutions/IronRDP/blob/25bbb2682c95419fc18a61cab81e29a0ab9f23d4/STYLE.md#logging

CBenoit:
> I added a few other deact-react fixes to this PR

I’m not sure what deact-react means in this context, but all of these new commits seems good to be merged. I still need some time to wrap my head around the first one.

crates/ironrdp-dvc/src/server.rs
elmarco: oops, ok

crates/ironrdp-async/src/framed.rs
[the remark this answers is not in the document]
elmarco: @CBenoit any idea for a better approach? maybe we can leave a FIXME/TODO ?

CBenoit (commented):
Hi, sorry for the delay.

I gave a bit more of thought to this, and it’s indeed not so simple to do better as I initially thought.

Let me get over the problem we are solving, and reformulate my concerns and questions I have with the way it’s implemented now.

-----

We are trying to be as resilient as possible by ignoring anything we don’t expect for the current accept sequence state. This means:
- Things that are not matching the provided hint.
- Things that can’t be decoded into the expected structure.

This happens because as I understand it, the client does not stop sending active session packets (e.g.: FastPathInput PDUs, etc) even though the deact-react sequence is already started.

Ideally, we would rely on some synchronization logic, a "fence", to agree with the client on when packets relevant for the reactivation (~= connection/accept) logic should be send and received.

In the absence of such synchronization logic, we need to run the sequence state machine, without consuming its state even if we failed at decoding the expected structure from the provided frame. This results into having to clone everything because of the way we are consuming the current state to turn it into the next one.

-----

Here are some of my concerns:

- It’s not very clean to clone our way through this. I acknowledge that this is a possible fix, and I would be okay to merge this as a stopgap solution if you think it’s the best we can do with minimal efforts.

- Typically, servers are just killing the connection brutally when the client is behaving in some weird way. Microsoft’s official server is doing this. I feel we are doing something wrong if we need to accept what the official server would usually reject.

- From the above: instead of terminating the connection as soon as we get something weird, we ignore the problem, and the session may stay hanging around, doing nothing useful for a while.

-----

I investigated the RDP deact-react sequence more in depth, to better understand the domain and possible reasons that may be leading to this problem in the first place.

Normally, the [deact-react sequence](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/dfc234ce-481a-4674-9a5d-2a7bafb14432) works as follows:

- The server initiates the deactivation. When the server needs to tear down the current share or reconfigure it, by [sending](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/b307d88b-167d-4e1a-9d41-3ae39d657239) a [Deactivate All PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/8a29971a-df3c-48da-add2-8ed9a05edc89).

- Upon reception of the Deactivate All PDU, the client _must_ stop sending active-share graphics updates (both FastPath updates and SlowPath updates).

- The server begins reactivation by sending a new [Demand Active PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/a07abad1-38bb-4a1a-96c9-253e3d5440df), which carries the new share settings.

- The client then responds with Confirm Active (and possibly other PDUs like Synchronize, Control, Font List, etc.).

The key point is that once the server has sent Deactivate All, the client is expected to stop sending further PDUs refering to the old context, and wait for the new handshake. _In principle, it should be fine to just fail upon reception of an unexpected payload_.

-----

At this point, I suspect we may be hitting a race condition depending on the network latency with client’s send buffer hitting the wire even after the server decided to deactivate, and by the time these packets arrive, the server has already started the reactivation process. Or, there is a client-side bug. For instance, I’m not sure if the IronRDP client is properly adhering to the deactivation order sent by the server.

-----

**Best approach:**

- Fix IronRDP client if it does not properly follow the logic to quiesce old PDUs on Deactivate All.
- Send the Deactivate All and Demande Active PDUs to the client, and wait for the Confirm Active PDU. Any PDU received in between should be discarded.

The key point to this approach is that we just need to parse received frames until a Confirm Active PDU is received, and then we can drive the sequence as usual.

In pseudo-code:

```rust
let frame = loop {
    let frame = read_frame_by_hint();
    
    match pdu::decode(frame) {
        Ok(_) => break frame,
        Err(DecodError) => {}
        Err(error) => return Err(error),
    }
};

// Feed the frame, which is just the PDU that is expected for the next step.
// We decode the frame twice, but from this point on we don’t need to check for the decode result again.
accept_sequence.step(frame);

// At this point, drive the sequence as usual.
```

=> We pump and discard all the messages exchanged between the Deactivate All and the Confirm Active PDUs.

**Less ideal approach:**

If we really need the ability of handling unexpected PDUs at any point in the accept sequence, then maybe a better approach is to try decoding the next expected PDU for the current state before actually executing the step. We may need to add an extra trait in the ironrdp-acceptor crate for this purpose.

**Less ideal approach 2:**

Move data out of the state variants, and use `Option`s directly at the `ClientConnector` so that we can set and take the various states without cloning everything, and it’s possible to leave the state machine untouched until the expected PDU is received.

-----

Do you think the "best approach" (as I call it) would work?
Would you prefer we still merge this PR as-is and improve on it afterwards?

I’m open to any of your suggestions at this point.

crates/ironrdp-async/src/framed.rs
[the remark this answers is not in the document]
CBenoit: Let’s discuss about this in the main thread: https://github.com/Devolutions/IronRDP/pull/630#pullrequestreview-2553558242

elmarco:
not tested very carefuly yet, it would be worth to write a test for it.

crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -95,9 +97,10 @@ impl Acceptor {
             io_channel_id: consumed.io_channel_id,
             desktop_size,
             server_capabilities: consumed.server_capabilities,
-            static_channels: StaticChannelSet::new(),
+            static_channels: consumed.static_channels,
```
CBenoit, line 100:
```suggestion
            /// No description of such behavior was found in the specification, but apparently,
            /// we must keep the channel state as they were during reactivation.
            /// This fixes various state issues during client resize.
            static_channels: consumed.static_channels,
```

crates/ironrdp-acceptor/src/finalization.rs
```diff
@@ -6,7 +6,7 @@ use pdu::rdp;
 
 use crate::util::{self, wrap_share_data};
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
CBenoit, line 9:
```suggestion
#[derive(Debug)]
```

crates/ironrdp-acceptor/src/finalization.rs
```diff
@@ -15,7 +15,7 @@ pub struct FinalizationSequence {
     pub input_events: Vec<Vec<u8>>,
 }
 
-#[derive(Default, Debug)]
+#[derive(Clone, Default, Debug)]
```
CBenoit, line 18:
```suggestion
#[derive(Default, Debug)]
```

crates/ironrdp-async/src/framed.rs
```diff
@@ -236,39 +230,44 @@ pub async fn single_sequence_step<S>(
...
+            .read_by_hint(next_pdu_hint)
             .await
-            .map_err(|e| ironrdp_connector::custom_err!("read frame by hint", e))?;
+            .map_err(|e| ironrdp_connector::custom_err!("read frame by hint", e));
 
+        if ignore_pdu_error && pdu.is_err() {
+            return Ok(Written::Nothing);
+        }
+
+        let pdu = pdu?;
```
CBenoit, line 270: question: I’m not sure we need the new `ignore_pdu_error` thing?
`read_by_hint` is not decoding PDUs, and unexpected frames (based on the "next PDU hint"), are already ignored.
The `DecodeError` itself is handled inside the acceptor sequence. Am I missing something?

CBenoit:
> not tested very carefuly yet, it would be worth to write a test for it.

Understood! I’m fine with merging before the test is written.

My last concern is about the extended interface for the Sequence trait: https://github.com/Devolutions/IronRDP/pull/630#discussion_r1921221343

If that’s necessary I’m happy to go that way, but I failed at understanding why it became necessary given that the previous version of this PR was not doing anything similar.

crates/ironrdp-async/src/framed.rs
elmarco: It would raise an error, either during read_by_hint or later DecodeError. This aborts accept_finalize().

CBenoit: Do you know why? What kind of payload are we receiving here which may cause accept_finalize() to fail?

elmarco: That can be pretty much any PDU inflight during reactivation-reactivation. Server sends DemandActive, but anything can be in flight.

elmarco:
still working on some tests, do not merge :)

crates/ironrdp-async/src/framed.rs
CBenoit: We still need to be framing packets properly, otherwise we wouldn’t know where to start reading the next relevant packet. And for that matter, all active sharing packets are [either X224 (SlowPath) or FastPath](https://github.com/Devolutions/IronRDP/blob/dd221bf22401c4635798ec012724cba7e6d503b2/crates/ironrdp-pdu/src/lib.rs#L93-L118).
The ConfirmActive PDU which is sent by the client after stopping sending any other active share PDU is one of such packets (X224?).

elmarco: It can/does fail during single_sequence_step(). However it's true that PduHint::find_size() shouldn't fail since all incoming packets should be either x224 or FP. Thus we could probably remove the ignore_pdu_error(). I think I have seen it happen though, not sure how/when. If you feel strongly about it, I'll drop it.

CBenoit: `single_sequence_step()` has multiple point of failure other than `read_by_hint()`, including `sequence.step()`.
I suspect the real culprit of the failure was `sequence.step()`, not `read_by_hint()`.
Now it is handling unexpected PDUs until a safe point.
Yes, I would rather drop the new API surface to keep the code as simple as possible.
I do agree that defensive patterns are useful when dealing with network, but I don’t think it’s worth being that defensive here.
Thank you for all the efforts!

CBenoit (approved):
LGTM! Thank you!!

PR #632, opened 2025-01-09
fix(web): allow clipboard callback to be set to null explicitly to disable clipboard

irvingoujAtDevolution:
Sometimes in DVLS we want to be able to disable clipboard.

https://devolutions.slack.com/archives/C1FL0GS03/p1734031803626829

### DIFF
--- a/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
+++ b/web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
@@ -43,7 +43,7 @@ export class WasmBridgeService {
     private keyboardActive: boolean = false;
     private keyboardUnicodeMode: boolean = false;
     private backendSupportsUnicodeKeyboardShortcuts: boolean | undefined = undefined;
-    private onRemoteClipboardChanged?: OnRemoteClipboardChanged;
+    private onRemoteClipboardChanged?: OnRemoteClipboardChanged | null;
     private onRemoteReceivedFormatList?: OnRemoteReceivedFormatsList;
     private onForceClipboardUpdate?: OnForceClipboardUpdate;
     private cursorHasOverride: boolean = false;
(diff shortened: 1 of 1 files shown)


web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -43,7 +43,7 @@ export class WasmBridgeService {
     private keyboardActive: boolean = false;
     private keyboardUnicodeMode: boolean = false;
     private backendSupportsUnicodeKeyboardShortcuts: boolean | undefined = undefined;
-    private onRemoteClipboardChanged?: OnRemoteClipboardChanged;
+    private onRemoteClipboardChanged?: OnRemoteClipboardChanged | null;
```
CBenoit, line 46: There is a `?` on `onRemoteClipboardChanged`, so I think this is redundant.

web-client/iron-remote-gui/src/services/wasm-bridge.service.ts
```diff
@@ -75,7 +75,7 @@ export class WasmBridgeService {
     }
 
     /// Callback to set the local clipboard content to data received from the remote.
-    setOnRemoteClipboardChanged(callback: OnRemoteClipboardChanged) {
+    setOnRemoteClipboardChanged(callback: OnRemoteClipboardChanged | null) {
         this.onRemoteClipboardChanged = callback;
     }
```
CBenoit, line 80: I think the right way to do what you need on DVLS side, is to not call this function at all. Do you think this is problematic?

irvingoujAtDevolution:
Sorry, the problem is not here, the setOnRemoteClipboardChanged is private and not exposed to the consumer. I'll update the PR and reopen it for review.

PR #634, opened 2025-01-15
feat: support license caching

probakowski:
This PR adds support for license caching by storing the license obtained from [SERVER_UPGRADE_LICENSE](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpele/e8339fbd-1fe3-42c2-a599-27c04407166d) message and sending [CLIENT_LICENSE_INFO](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpele/9407b2eb-f180-4827-9488-cdbff4a5d4ea) if we have license requested by the server stored.


awakecoding:
👀 thanks for improving licensing support, one of the trickiest areas of RDP

crates/ironrdp-client/Cargo.toml
```diff
@@ -77,6 +78,7 @@ reqwest = "0.12"
 url = "2.5"
 raw-window-handle = "0.6.2"
 ironrdp-core = { workspace = true, features = ["alloc"] }
+uuid = { version = "1.11.0"}
```
CBenoit, line 81:
```suggestion
uuid = { version = "1.11"}
```

crates/ironrdp-client/src/config.rs
```diff
@@ -316,6 +317,8 @@ impl Config {
                 whoami::Platform::Android => MajorPlatformType::ANDROID,
                 _ => MajorPlatformType::UNSPECIFIED,
             },
+            hardware_id: Uuid::default(),
+            license_cache: Arc::new(NoopLicenseCache {}),
```
CBenoit, line 321: nitpick:
```suggestion
            license_cache: Arc::new(NoopLicenseCache),
```

CBenoit (commented):
Many thanks for working on this!!

crates/ironrdp-connector/src/lib.rs
```diff
@@ -161,6 +163,7 @@ pub struct Config {
     pub dig_product_id: String,
     pub client_dir: String,
     pub platform: capability_sets::MajorPlatformType,
+    pub hardware_id: Uuid,
```
CBenoit, line 166: question: Another question: what exactly is an hardware ID? Is it really an UUID? Should we wrap `[u32; 5]` or some variation of this into a new type instead? In UUIDs, some bits are used to encode format’s variant and version, which may not be appropriate if an hardware ID is defined as something else. The docs: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpele/95ada355-e25e-4601-83fa-a4b00392a352

crates/ironrdp-client/Cargo.toml
probakowski: I removed the `uuid` crate altogether

crates/ironrdp-connector/src/lib.rs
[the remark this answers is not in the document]
probakowski: Yeah, that makes sense, I changed both of them to `Option`

probakowski: I moved it to `[u32; 4]` - `PlatformID` part we care about so it doesn't makes sense to leave it for the user to provide. The user should provide the 4 parts of the data identifying (somehow 🤷 ) the computer (we will use `UUID` of the node but you're right that it doesn't have to be this way)

crates/ironrdp-connector/src/license_exchange.rs
```diff
@@ -107,52 +137,100 @@ impl Sequence for LicenseExchangeSequence {
...
+                                    version: license_request.product_info.version,
+                                    scope: scope.0.clone(),
+                                    company_name: license_request.product_info.company_name.clone(),
+                                    product_id: license_request.product_info.product_id.clone(),
+                                    license_info: vec![],
+                                })
+                            })
+                            .find(|res| res.is_err() || res.as_ref().is_ok_and(Option::is_some))
+                            .transpose()?
+                            .flatten();
```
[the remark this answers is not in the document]
probakowski: Yep, that is much better, thanks!

crates/ironrdp-connector/src/lib.rs
```diff
@@ -161,6 +162,7 @@ pub struct Config {
     pub dig_product_id: String,
     pub client_dir: String,
     pub platform: capability_sets::MajorPlatformType,
+    pub hardware_id: Option<[u32; 4]>,
```
CBenoit, line 165:
```suggestion
    /// Unique identifier for the computer
    ///
    ///  Each 32-bit integer contains client hardware-specific data helping the server uniquely identify the client.
    pub hardware_id: Option<[u32; 4]>,
```

CBenoit (approved):
LGTM! I just added a little bit of documentation in the public API. (Although I didn’t write anything extensive, anything interesting you could add will be greatly appreciated.)
Thank you again for moving this forward!

PR #640, opened 2025-01-21
Verifications de vulns

inter0hm:
J'ai remarquer que bcp de vos softwares ne font pas les requirement pour compliance. - Les analyse vont servir pour des futures references.


.github/dependabot.yml
```diff
@@ -12,6 +12,7 @@ updates:
         patterns:
           - "md-5"
           - "md5"
+          - "sha512"
```
CBenoit, line 15: issue: This crate is not in our dependency graph.

web-client/iron-svelte-client/package.json
CBenoit, line 1: question: Did you try to run the project with these dependencies updated? I’m really not sure it even compiles anymore with these changes.

FYI, this project is just a test client for testing purposes. Unless there is a virus in the dependency tree there is nothing critical here.

SECURITY.md
CBenoit, line 1: note: We could add this file although we don’t have any "support version" for security updates as we are still developing actively the project and the API did not reach the required maturity yet. That being said, we should create this file ourselves, as you can’t know what should go inside at this point.

CBenoit:
I double checked, we already have security policy configured at the organization level, and applied to all our repositories including IronRDP: https://github.com/Devolutions/IronRDP/securitym

Closing as nothing meaningful is included in the PR, and the author does not seem trustworthy.

PR #653, opened 2025-01-28
fix(connector): make LicenseCache RefUnwindSafe

elmarco:
This fixes commit dd221bf ("feat: support license caching (#634)") and semver-checks is now happy:

    type ProcessorOutput is no longer UnwindSafe, in /tmp/.tmppi9kAf/IronRDP/crates/ironrdp-session/src/x224/mod.rs:15


CBenoit (approved):
I see, thank you!

PR #654, opened 2025-01-29
fix: use 'OR' for SPDX license expression

elmarco:
Replacing deprecated '/' operator in SPDX license expression with 'OR' https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/#d42-disjunctive-or-operator


CBenoit (approved):
Good catch, thank you!

PR #670, opened 2025-02-13
WIP: add QOI compression

elmarco:
I recently found the QOI image format (https://qoiformat.org/) and was excited by its potential for desktop remoting.

I hacked up a few things to actually benchmark it compared to the other codecs. 

Based on a raw recording of a typical Fedora desktop usage (terminal/web), for 730 4k frames:

None (lossless):  5s user CPU, 0% compression
Bitmap (lossless):  74s user CPU, 92.5% compression
RemoteFx (lossy): 201s user CPU,  96.72% compression
QOI (lossless): 10s user CPU, 96.20% compression

There is still a lot of various optimizations to be desired regardless of the codecs, but this will hopefully help. Also, we should start thinking how QOI could be adapted for streaming.


awakecoding:
Custom codecs not supported by the Microsoft RDP implementation have rarely been used, but since we're building both the client and server in IronRDP, we're free to experiment! QOI is an excellent choice, it is very simple, and easier to optimize due to its low complexity. IIRC QOI uses RGB, not YUV like most codecs. Color conversion and chroma subsampling alone can take up a lot of the processing time if not properly optimized.

There is just one custom codec extension I've seen in the past: JPEG in RDP. XRDP supports it in the server, and FreeRDP supports it in the client as an optional build feature. JPEG has many advantages: it is by far the most widely available image codec, but it is also one for which the most effort has been put into optimizing the implementation (think libjpeg-turbo). For an RDP web client, this also makes it possible to leverage built-in browser support for JPEG decoding, which is a huge advantage over porting and optimizing a custom codec in WASM.

In the past, we've designed our own remote desktop protocol (Wayk Now) at Devolutions, and we experimented with many different ways to deal with the codecs more efficiently than RDP. We had [GFWX](https://www.gfwx.org/) ported in Rust: https://github.com/Devolutions/gfwx-rs

GFWX, unlike QOI, is a simplified *wavelet* codec, which would make it closer to RemoteFX. It is generic and can accept a wide variety of YUV color formats (RGB wouldn't perform well, you're better off doing the YUV conversion to avoid the color channel correlation issue of RGB). It is one of the rare codecs that would enable really nice things like lossless encoding in which you could trim part of the encoded output for lossy encoding. In our implementation, we've used a reversible color conversion (YCoCg-R) due to its reduced complexity (it's only integer operations, not floating point). Unfortunately, we found out JPEG still performed better than GFWX. In theory GFWX offers ways to beat JPEG, but libjpeg-turbo is just hard to beat.

If you're curious about the color transformations we've experimented with, we still have them here:
https://github.com/Devolutions/cadeau/tree/master/libxpp

I'm open to experimenting with various custom codecs, the ones in RDP are good, but not necessarily special or great. We can definitely beat the standard protocol if we go custom :)

elmarco:
@awakecoding I added some jpegturbo benchmarks above. For lossless, desktop usage, QOI is very good. Performance with wasm should be ok (rust-qoi compiles for wasm target)

 We used to have some heuristics in Spice server to detect video zones, and used jpeg aggressively iirc (when not using whole frame gpu video encoding).

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -221,10 +358,19 @@ pub(crate) struct UpdateFragmenterOwned {
     len: usize,
 }
 
+#[cfg_attr(feature = "__bench", visibility::make(pub))]
 pub(crate) struct UpdateFragmenter<'a> {
     code: UpdateCode,
     index: usize,
-    data: &'a [u8],
+    pub data: &'a [u8],
```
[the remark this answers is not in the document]
elmarco: hmm, it doesn't work on fields: error: expected non-macro attribute, found attribute macro `visibility::make`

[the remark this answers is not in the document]
CBenoit: I see! Then, could you add something like this:
```rust
#[doc(hidden)] // not part of the public API, used by benchmarks
```

Just for easier reviewing in the future

elmarco:
@CBenoit I think we should be good to start reviewing this series. I can split various preliminary patches in different MR if you prefer?

CBenoit:
> @CBenoit I think we should be good to start reviewing this series. I can split various preliminary patches in different MR if you prefer?

Sounds great! Yes, I would prefer multiple MRs. I’ll do my best to be as responsive as possible!

CBenoit:
Hi @elmarco 
Do you want me to review this now?

CBenoit:
Sounds great! I’m starting the review.

First, this commit is not following the convention: [benches:: fix could not find time in tokio](https://github.com/Devolutions/IronRDP/pull/670/commits/c9362d4c66debf7fa60f7fe5b0f5d0d33c374a27)
I guess this should be `chore(bench):` or `chore(benches)` (tool change, do not go into production)

CBenoit (commented):
issue: The commit `feat(session): add apply_rgb24() to apply non-inverted bitmap` does not add any feature by itself. It does not change any existing behavior, nor it exposes any new item. Looks like internal groundwork for one of the next commit. Should be `refactor(session)`. Note that we are using the commits for auto-generating the changelog, and the idea here is that this commit should not be part of the changelog (we ignore refactor commits for the changelog).

crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs.rs
```diff
@@ -715,3 +715,51 @@ List of codecs:
 
     Ok(BitmapCodecs(codecs))
 }
+
+///
```
CBenoit: style: empty first documentation line

CBenoit (commented):
issue: Invalid commit type for `feature(server)!: replace with_remote_fx option with codecs`, use `feat`

suggestion: Could be a follow up PR, but I think that the crate `ironrdp-cfg` may be a better place for `client_codecs_capabilities` and `server_codecs_capabilities`. It feels off to me to have CLI-related code in the protocol layer.

crates/ironrdp-server/src/builder.rs
```diff
@@ -155,17 +156,17 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
-    pub fn with_remote_fx(mut self, enabled: bool) -> Self {
-        self.state.with_remote_fx = enabled;
-        self
+    pub fn with_codecs_config(mut self, config: &[&str]) -> Result<Self, String> {
+        self.state.codecs = server_codecs_capabilities(config)?;
+        Ok(self)
     }
```
CBenoit: issue: This API takes an array of strings to configure the bitmap codecs, which feels a bit unidiomatic for Rust. Typically, the layering goes: CLI / UI (stringly typed) -> strong types -> internal library APIs. Since this function is meant to be called from Rust, I think it would be more appropriate and safer to take a `BitmapCodecs` value directly here. That way, string parsing and validation can remain at the boundary (e.g., CLI layer), and this builder stays in the strongly typed domain. The function also remains infaillible.

crates/ironrdp-server/src/server.rs
```diff
@@ -38,7 +38,22 @@ use crate::{builder, capabilities, SoundServerFactory};
 pub struct RdpServerOptions {
     pub addr: SocketAddr,
     pub security: RdpServerSecurity,
-    pub with_remote_fx: bool,
+    pub codecs: BitmapCodecs,
+}
+impl RdpServerOptions {
```
CBenoit: style: missing new line

crates/ironrdp-session/Cargo.toml
```diff
@@ -25,7 +29,7 @@ ironrdp-graphics = { path = "../ironrdp-graphics", version = "0.4" } # public
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.5", features = ["std"] } # public
 ironrdp-displaycontrol = { path = "../ironrdp-displaycontrol", version = "0.3" }
 tracing = { version = "0.1", features = ["log"] }
+qoicoubeh = { version = "0.5", optional = true }
```
CBenoit, line 33: That’s a funny name, very current :joy:

crates/ironrdp-session/src/fast_path.rs
```diff
@@ -9,6 +9,8 @@ use ironrdp_pdu::codecs::rfx::FrameAcknowledgePdu;
 use ironrdp_pdu::fast_path::{FastPathHeader, FastPathUpdate, FastPathUpdatePdu, Fragmentation};
 use ironrdp_pdu::geometry::{InclusiveRectangle, Rectangle as _};
 use ironrdp_pdu::pointer::PointerUpdateData;
+#[cfg(feature = "qoi")]
+use ironrdp_pdu::rdp::capability_sets::CODEC_ID_QOI;
```
CBenoit: suggestion: I would recommend to fully qualify at the use site, just so we don’t have cfgs in the imports.

crates/ironrdp-web/Cargo.toml
```diff
@@ -18,8 +18,9 @@ test = false
 crate-type = ["cdylib", "rlib"]
 
 [features]
-default = ["panic_hook"]
+default = ["panic_hook", "qoi"]
```
CBenoit: thought: I’m not sure I want to enable qoi by default just yet :thinking:

crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs.rs
```diff
@@ -620,12 +638,14 @@ pub struct CodecId(u8);
 
 pub const CODEC_ID_NONE: CodecId = CodecId(0);
 pub const CODEC_ID_REMOTEFX: CodecId = CodecId(3);
+pub const CODEC_ID_QOI: CodecId = CodecId(0x0A);
```
CBenoit, line 659: question: Any rationale for choosing 0x0A? I assume it’s mostly arbitrary? Do you know about other similar extensions? (JPEG?)

benches/Cargo.toml
CBenoit, line 1: note: I think it’s fine to enable qoi unconditionally in the benches crate, it’s not going to production

crates/ironrdp-web/Cargo.toml
```diff
@@ -18,9 +18,11 @@ test = false
 crate-type = ["cdylib", "rlib"]
 
 [features]
+# qoiz is not enabled by default, because MacOS github runner is lacking clang wasm target
```
CBenoit: note: I don’t think the lack of support in the macOS GitHub runner is a strong reason to exclude this from the default set. That’s more of an infrastructure limitation than a product decision. Since this is still experimental, I’d suggest keeping both disabled for now.

crates/ironrdp-web/Cargo.toml
elmarco: ok, do you think we need a test period?

crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs.rs
elmarco: It's arbitrary, this is just hard-coded values used by the client to reference a bitmap codec (the server doesn't care)

crates/ironrdp-session/src/fast_path.rs
elmarco: ok

crates/ironrdp-server/src/server.rs
elmarco: done

crates/ironrdp-server/src/builder.rs
elmarco: indeed

crates/ironrdp-web/Cargo.toml
CBenoit: From what I understand it seems fine to enable it by default, but I would still prefer we give a test period so we can decide more comfortably whether to enable it by default a bit later with Marc-André

CBenoit (approved):
LGTM! Thanks!!

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -408,6 +448,101 @@ impl BitmapUpdateHandler for RemoteFxHandler {
...
+            .compress_stream2(
+                &mut outb,
+                &mut inb,
+                zstd_safe::zstd_sys::ZSTD_EndDirective::ZSTD_e_flush,
+            )
+            .map_err(zstd_safe::get_error_name)
+            .unwrap();
+        if res != 0 {
+            return Err(anyhow!("Failed to zstd compress"));
+        }
```
CBenoit, line 520: According to the doc of the function `compress_stream2`:

> An lower bound for the amount of data that still needs to be flushed out.
>
> This is useful when flushing or ending the frame: you need to keep calling this function
> until it returns 0.

Why is it fine to call it only once in our case?

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -408,6 +448,101 @@ impl BitmapUpdateHandler for RemoteFxHandler {
...
+
+        let mut zctxt = self.zctxt.lock().unwrap();
+        let res = zctxt
+            .compress_stream2(
+                &mut outb,
+                &mut inb,
+                zstd_safe::zstd_sys::ZSTD_EndDirective::ZSTD_e_flush,
+            )
+            .map_err(zstd_safe::get_error_name)
+            .unwrap();
```
CBenoit, line 517: question: Why is it fine to call unwrap at this place?

elmarco: you are right, I looked over it as it doesn't happen in my tests. I'll fix it.

PR #672, opened 2025-02-16
fix(pdu): make AddressFamily parsing resilient

Perlmint:
I found this value while trying to connect to ironrdp/server example with Windows App for iOS.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/client_info.rs
+++ b/crates/ironrdp-pdu/src/rdp/client_info.rs
@@ -234,7 +234,7 @@ impl ExtendedClientInfo {
     fn encode(&self, dst: &mut WriteCursor<'_>, character_set: CharacterSet) -> EncodeResult<()> {
         ensure_size!(in: dst, size: self.size(character_set));
 
-        dst.write_u16(self.address_family.to_u16());
+        dst.write_u16(self.address_family.as_u16());
         // // + size of null terminator, which will write in the write_string function
         dst.write_u16(string_len(self.address.as_str(), character_set) + character_set.to_u16().unwrap());
         utils::write_string_to_cursor(dst, self.address.as_str(), character_set, true)?;
@@ -622,7 +622,7 @@ impl AddressFamily {
         Self(val)
     }
 
-    pub fn to_u16(self) -> u16 {
+    pub fn as_u16(self) -> u16 {
         self.0
     }
 }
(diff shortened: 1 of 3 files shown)


CBenoit:
Hello!

Thank you for the fix. It looks good to me as-is, but I think it would be even better to apply the advice documented at this place: https://github.com/Devolutions/IronRDP/tree/5555c7b9dd2bacc543496251c55a42953a44e88f/crates/ironrdp-pdu#enumeration-like-types-should-allow-resilient-parsing

Is it something you are willing to look into?

Perlmint:
Oh, I couldn't know about the documentation you mentioned. I tried to apply it. :)

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -609,12 +609,12 @@ impl Default for PerformanceFlags {
...
 }
 
-#[repr(u16)]
+#[repr(transparent)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, FromPrimitive, ToPrimitive)]
-pub enum AddressFamily {
-    Invalid = 0x0000, // value given by MS Remote Desktop for Android 10.0.18.1251
-    INet = 0x0002,
-    INet6 = 0x0017,
+pub struct AddressFamily(u16);
```
CBenoit, line 614:
```suggestion
pub struct AddressFamily(u16);

```

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -609,12 +609,13 @@ impl Default for PerformanceFlags {
     }
 }
 
-#[repr(u16)]
+#[repr(transparent)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, FromPrimitive, ToPrimitive)]
```
CBenoit, line 613: issue?: Just noticed about that: we can probably remove these traits.

```suggestion
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```

Instead provide functions such as `from_u16` and `as_u16`

Perlmint:
I just noticed that test was broken, and commited fixing it.

CBenoit:
Thank you! Can you also address this? https://github.com/Devolutions/IronRDP/pull/672#discussion_r1957674463

crates/ironrdp-pdu/src/rdp/client_info.rs
Perlmint: I removed traits and added implementation infallible `from_u16` and `to_u16`.

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -610,14 +609,24 @@ impl Default for PerformanceFlags {
...
     pub const INET: Self = Self(0x0002);
     pub const INET_6: Self = Self(0x0017);
 }
 
+impl AddressFamily {
+    pub fn from_u16(val: u16) -> Self {
+        Self(val)
+    }
+
+    pub fn to_u16(self) -> u16 {
```
CBenoit, line 625:
```suggestion
    pub fn as_u16(self) -> u16 {
```

CBenoit (approved):
LGTM! Thank you!!

PR #685, opened 2025-03-04
fix(web):allow keyboard input even if the mouse is off the area of canvas

irvingoujAtDevolution:
https://github.com/user-attachments/assets/426b8b9d-dd6d-4ab5-903d-3f22f00e838d


CBenoit (approved):
Good job fixing this!

PR #687, opened 2025-03-07
fix(pdu): fix FastPathHeader minimal size

elmarco:
The minimal_size() logic didn't properly take into account the overall PDU size.

This fixes random error/disconnect in client.


CBenoit (approved):
Good catch!

PR #693, opened 2025-03-12
fix(client): set back control flow to wait

elmarco:
After resize event is sent, we should switch back to a normal mode, instead of wait-duration(0)/polling.


CBenoit (approved):
Thank you!

PR #694, opened 2025-03-12
fix(pdu): TS_RFX_CHANNELT width/height SHOULD be within range

elmarco:
According to the specification, the value does not need to be in the range: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdprfx/4060f07e-9d73-454d-841e-131a93aca675

(the ironrdp-server can send larger values)

### DIFF
--- a/crates/ironrdp-pdu/src/codecs/rfx.rs
+++ b/crates/ironrdp-pdu/src/codecs/rfx.rs
@@ -15,9 +15,7 @@ pub use self::data_messages::{
     ContextPdu, EntropyAlgorithm, FrameBeginPdu, FrameEndPdu, OperatingMode, Quant, RegionPdu, RfxRectangle, Tile,
     TileSetPdu,
 };
-pub use self::header_messages::{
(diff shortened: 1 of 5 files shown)


CBenoit (commented):
I agree that we should not hardcode this check in the PDU parsing code.
However, I think it would be better to log a warning at the place where we actually decode and use the RfxChannel PDU. Do you know where that would be?

elmarco:
It's also in pdu, src/rfx.rs DecodingContext::decode() I think.

elmarco:
Or perhaps session DecodingContext::process_frame()

CBenoit:
I see… what about just removing the checks, and the logging along? I think this SHOULD is because some previous version of RemoteFX is performing the validation on client side, but I don’t see how it matters for us today 🤔

image

CBenoit (approved):
LGTM!

CBenoit (commented):
Oh, should we remove `RfxChannelWidth` and `RfxChannelHeight` types altogether as well? Sounds like they do not serve any purpose anymore.

PR #680, opened 2025-02-28
chore(web): migrate to Svelte 5 and Vite 6

irvingoujAtDevolution:
Tested on DVLS, works as usual! 
Only need to change import umd.js to just import the package.
Prepare for later fix to transfer focus in and out iron-remote-gui

image

PR #708, opened 2025-03-19
fix(web): fix firefox selection issue

irvingoujAtDevolution:
Though I tested it a bit myself, but I'm not 100% confident in this change, would appericate someone can checkout to this branch and test it a liitle.

### DIFF
--- a/web-client/iron-remote-gui/src/iron-remote-gui.svelte
+++ b/web-client/iron-remote-gui/src/iron-remote-gui.svelte
@@ -690,14 +690,14 @@
 <div bind:this={inner}>
     <div
         bind:this={wrapper}
-        class="screen-wrapper scale-{scale}"
+        class="iron-rdp-screen-wrapper scale-{scale}"
         class:hidden={!isVisible}
         class:capturing-inputs={capturingInputs}
         style={wrapperStyle}
     >
         <div
             bind:this={screenViewer}
-            class="screen-viewer"
+            class="iron-rdp-screen-viewer"
             style={viewerStyle}
             contenteditable={isFirefox}
             onpaste={ffOnPasteHandler}
@@ -716,7 +716,11 @@
                 }}
                 oncontextmenu={(event) => event.preventDefault()}
                 onwheel={mouseWheel}
+                onselectstart={(event) => {
+                    event.preventDefault();
+                }}
                 id="renderer"
+                class="iron-rdp-canvas"
                 tabindex="0"
             ></canvas>
         </div>
@@ -724,21 +728,25 @@
 </div>
 
 <style>
-    .screen-wrapper {
+    .iron-rdp-screen-wrapper {
         position: relative;
     }
 
-    .capturing-inputs {
+    .iron-rdp-capturing-inputs {
         outline: 1px solid rgba(0, 97, 166, 0.7);
         outline-offset: -1px;
     }
 
-    canvas {
+    .iron-rdp-canvas {
         width: 100%;
         height: 100%;
     }
 
-    .screen-wrapper.hidden {
+    ::selection {
+        background-color: transparent;
+    }
+
+    .iron-rdp-screen-wrapper.hidden {
         pointer-events: none !important;
         position: absolute !important;
         visibility: hidden;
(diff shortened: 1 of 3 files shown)


web-client/iron-svelte-client/src/lib/remote-screen/remote-screen.svelte
```diff
@@ -60,27 +61,44 @@
...
+            </button>
+            <button on:click={() => toggleCursorKind()}>Toggle cursor kind</button>
+            <button on:click={() => uiService.shutdown()}>Terminate Session</button>
+            <label style="color: white;">
+                <input on:click={(e) => onUnicodeModeChange(e)} type="checkbox" />
+                Unicode keyboard mode
+            </label>
+        </div>
+
+        {#if showDebugPanel}
```
irvingoujAtDevolution, line 89: debug panel is for testing 
1. if input focus is correct
2. if ::selection selector is applied for element outside of iron-remote-gui

web-client/iron-remote-gui/src/iron-remote-gui.svelte
```diff
@@ -704,29 +704,37 @@
...
+                class="iron-rdp-canvas"
                 tabindex="0"
             ></canvas>
         </div>
     </div>
 </div>
 
 <style>
-    .screen-wrapper {
+    .iron-rdp-screen-wrapper {
```
irvingoujAtDevolution, line 719: since we do not use shadow-root anymore, in order to reduce css naming conflict, we use iron-rdp prefix

CBenoit:
Thank you, I’ll try on my side too.

web-client/iron-remote-gui/src/iron-remote-gui.svelte
CBenoit: I’m a bit surprised. When did we stop using a shadow-root?? Was it part of https://github.com/Devolutions/IronRDP/pull/680 ?
This sounds like a bad idea. Are things working as expected when opening multiple sessions in the same tab?

irvingoujAtDevolution: https://github.com/user-attachments/assets/edef4ec2-1e40-458f-8938-17ba8895acbd

irvingoujAtDevolution: Works for me!

irvingoujAtDevolution: image

irvingoujAtDevolution: I went over the code, each time a new <iron-remote-gui> is created, a new instance of session will be created as well. sperated in memory. From my understanding there's no need for shadow root. However, we can roll it back if you think this is not the practice we want.

CBenoit: Do you mean "a new instance of ironrdp_web::Session"?
I guess we are fine because we are not trying to fetch the canvas by using its HTML ID.
If it works, then it’s good enough, but my understanding is that using a shadow root is definitely a better practice for a web component.
I don’t think we should have to give prefix to our CSS classes to avoid conflicts with global style sheets… Shadow DOM is what allows proper encapsulation of custom elements.
What is the reason for going away from that in the first place?

irvingoujAtDevolution: https://github.com/Devolutions/IronRDP/pull/717

I rolled back the shadow root problem in the PR above, I'll update this one accordingly.

irvingoujAtDevolution: Sorry, the \< needs to be escaped in comment, I meant to say \<iron-remote-gui\>

irvingoujAtDevolution: I sugguest we merge that one first, then I could rebase and test to see if there's anything needs fix.

CBenoit (approved):
Confirmed working on my side too!

PR #722, opened 2025-03-25
refactor(web-client): refactor `iron-remote-gui` into `iron-remote-desktop`

RRRadicalEdward:
This is the first part of [the "Modularize the remote desktop WebComponent" ticket](https://devolutions.slack.com/docs/T0Z2XMSRZ/F08C5PC7C2C).
PR changes the `web-client` of the _IronRDP_:
* Renamed some of the structures on the _Rust_ side
* Removed the `iron-remote-gui` package
* Added the `iron-remote-desktop` package with the _TypeScript_ interfaces and exposed more interfaces from the _WASM_ bindings
 * Added `iron-remote-desktop-rdp` package which combines _WASM_ bindings and the _TypeScript_ interfaces from the `iron-remote-desktop` to create an` iron-remote-desktop` _HTML_ tag which is responsible for the _RDP_ connection.
 * Updated `xtask`


web-client/iron-remote-desktop-rdp/src/services/logging.service.ts
RRRadicalEdward, line 1: May services be considered a part of the iron-remote-desktop API and be moved into the `iron-remote-desktop` package?
This may help to reduce the amount of code duplication in _IronVNC_

web-client/iron-remote-desktop-rdp/src/services/wasm-bridge.service.ts
RRRadicalEdward, line 1: If we want to port this code to _IronVNC_, we will have a lot of duplicated code. The real difference would be only in the `connect` function.
To reduce code duplication we may:
1. Move this service into the `iron-remote-desktop` crate
2. Use dynamic imports to abstract away from the actual implementation logic.
3. Refactor the `connect` function to take a map/list of extensions -> iterate over them and insert them into the session builder

This approach would allow this service in both IronRDP and IronVNC since it moves the responsibility of the import and the correctness of the provided extensions to the caller.
What do you think?

web-client/iron-remote-desktop-rdp/src/iron-remote-desktop.svelte
RRRadicalEdward, line 1: As with the `wasm-bridge.service.ts`, if we want to port this code to _IronVNC_ in order to create an _HTML_ tag with _IronVNC_'s logic, we will end up with a lot of code duplication.
To prevent this situation I think we may as well move this code to `iron-remote-desktop` and use dynamic imports to use the desired by the caller functionality.
What do you think?

web-client/iron-remote-desktop-rdp/src/services/PublicAPI.ts
RRRadicalEdward, line 1: As I proposed in the `wasm-bridge.service.ts`, we can change the _API_ to take the list/map of the extension. So, eventually this file may also be moved to the iron-remote-desktop package to avoid code duplication.

crates/ironrdp-web/src/error.rs
RRRadicalEdward, line 1: This file and _sessions.rs_ contain a good amount of duplicates with _IronVNC_ logic.
Does it make sense to add a new crate(like `iron-remote-desktop-rs`) that would describe the interface of the Rust part of the project? It would include all the sharable structures and logic between _IronVNC_ and _IronRDP_ parts.
The components chain would look like this way:
* VNC:
_UI_ (TS) -> `iron-remote-desktop-vnc` (TS) -> `iron-remote-desktop` (TS) -> `iron-remote-desktop-rs` (Rust) -> `ironvnc-web` (Rust)
* RDP:
 _UI_ (TS) -> `iron-remote-desktop-rdp` (TS) -> `iron-remote-desktop` (TS) -> `iron-remote-desktop-rs` (Rust) -> `ironrdp-web` (Rust)

crates/ironrdp-web/src/session.rs
```diff
@@ -729,13 +740,14 @@ impl Session {
         Ok(())
     }
 
+    // QUESTION: in VNC we cannot synchronize lock keys, so do we need to remove this from the Session API?
```
RRRadicalEdward, line 743: How should we approach it? Should it be part of the extensions API?

web-client/iron-svelte-client/src/services/server-bridge.service.ts
```diff
@@ -1,14 +1,13 @@
...
 export interface ServerRect {
-    free(): void;
-
     clone_buffer(): Uint8Array;
 
     bottom: number;
     left: number;
     right: number;
     top: number;
+    free(): void;
```
CBenoit, line 10: style: I’m not sure what is the convention here, but the properties are now in-between two functions :thinking:

CBenoit:
I think you are aware of this yourself as far as I can see in your own comments, but this is not really what I imagined when I talked about dependency injection.

To clarify: the new `iron-remote-desktop` package should contain ALL the logic that is found in the previous `iron-remote-gui` package.
The `iron-remote-desktop-rdp` package should only re-export the WASM module API, and nothing more. No additional logic. (Mostly?)
The architectural invariant is that these packages do not contain any common logic, and no code is duplicated.

I’m not very good at JavaScript/TypeScript, so let me illustrate what I mean using Rust:

```rust
// -- iron-remote-desktop --
// Implements all the backend-agnostic logic, based on a given "remote desktop interface".

trait RemoteDesktopApi {
    fn remote_desktop_init(&self);
}

fn run(api: Box<dyn RemoteDesktopApi>) {
    // The logic is calling the remote desktop API, and does not know whether this is RDP or VNC.
    api.remote_desktop_init();
}

// -- ironrdp-web --
// This is what actually implements the interface.

pub struct Api;

impl RemoteDesktopApi for Api {
    fn remote_desktop_init(&self) {
        // do stuff
    }
}

// -- iron-remote-desktop-rdp --
// The purpose is to inline the WASM module in base64,
// but essentially, it exposes the exact same things as ironrdp-web.

pub use ironrdp_web::Api;

// -- iron-svelte-client --
// Injects the dependency (IronRDP) into the web component (iron-remote-desktop), and uses it.
// It knows that its IronRDP behind the scenes, and may inject additional extensions (future improvement).

iron_remote_desktop::run(Box::new(iron_remote_desktop_rdp::Api));
```

Because of the way JavaScript/TypeScript works, we don’t need `iron-remote-desktop-rdp` to know about the interface defined in `iron-remote-desktop`.

Of course, it’s not possible to directly translate my Rust code into TypeScript, and you may not be able to use a TypeScript interface for everything.

My understanding is that when building ironrdp_web using wasmpack, we get a **module** exporting a few items such as `SessionBuilder`. `SessionBuilder` in turns implements a specific interface (as defined in the Slack canvas).

My expectation is that `iron-remote-desktop-rdp` should basically end up being this:

```typescript
export * as default from './path/to/wasm/module'
```

= Re-export all the interface of the WASM module.

If you hit a problem, let me know in Slack so we can think through this together. (I never implemented such a thing myself, so I don’t know all the problems you may encounter beforehand.)

crates/ironrdp-web/src/session.rs
```diff
@@ -216,13 +205,28 @@ impl SessionBuilder {
...
+                Extension::Pcb(pcb) => self.0.borrow_mut().pcb = Some(pcb),
+                Extension::DisplayControl(use_display_control) => {
+                    self.0.borrow_mut().use_display_control = use_display_control
+                }
+            },
+            Err(err) => error!("Provided JsValue is not a valid extension value: {err:?}"),
+        }
+
         self.clone()
     }
```
CBenoit, line 221: thought: I’m not sure this is exactly what we want. At least it’s defined differently in the design document. But it does not look bad either. I need to think about it.

crates/ironrdp-web/src/session.rs
CBenoit: If it’s not part of the common API, yes, we should turn that into an extension! :+1:

crates/ironrdp-web/src/error.rs
CBenoit: I agree that we could provide a companion crate to avoid further code duplication! :+1: 
That would be a good follow up!
Let’s call this crate `iron-remote-desktop`, and make it live in the IronRDP repository.
Is it possible to export WASM functions from a dependency? My expectation is that it should be possible?
We could also provide a trait + macro helpers to define proper interfaces. Let’s discuss that on Slack!

web-client/iron-remote-desktop-rdp/src/services/wasm-bridge.service.ts
CBenoit: > Use dynamic imports to abstract away from the actual implementation logic

Are you thinking about something similar to what I described here?
https://github.com/Devolutions/IronRDP/pull/722#issuecomment-2752805936

web-client/iron-remote-desktop-rdp/src/iron-remote-desktop.svelte
CBenoit: I think I agree! Can you compare with what I wrote here? https://github.com/Devolutions/IronRDP/pull/722#issuecomment-2752805936

web-client/iron-svelte-client/src/services/server-bridge.service.ts
RRRadicalEdward: My bad. Let me revert

RRRadicalEdward:
> https://github.com/Devolutions/IronRDP/pull/722#issuecomment-2752805936

Let's clarify:  iron-remote-desktop-rdp has to re-export WASM and provide extension types. The next step is to modify the UserInteraction to accept the "backend" which will be either RDP or VNC and implements an interface (I have really missed the fact that JS/TS has a duck types). Maybe, the next step will be the refactor of the connect function into a builder (or I will leave it to a follow-up PR if there won't be a direct need in it right now).

web-client/iron-remote-desktop-rdp/src/iron-remote-desktop.svelte
RRRadicalEdward: I don't think the issue is actual after you discussed your idea in https://github.com/Devolutions/IronRDP/pull/722#issuecomment-2752805936, but I will keep it mind since the need for it might come

web-client/iron-remote-desktop-rdp/src/services/PublicAPI.ts
RRRadicalEdward: I don't think the issue is actual after the idea in #722 (comment), but I will keep it mind since the need for it might come

web-client/iron-remote-desktop-rdp/src/services/logging.service.ts
RRRadicalEdward: It will be a part of the single crate/packet `iron-remote-desktop`

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Indeed. It's not the same you shared in the design document:
```rust
struct Extension<'ident> {
    ident: &'ident str,
    value: JsValue,
}
```
I think that instead we can encapsulate the identifier in a type. Rust and TS allow us to do this and this approach is more reliable, since we do not need to do manual matching of `ident` to the provided `JsValue`, and therefore this approach is less error-prone

CBenoit:
> > [#722 (comment)](https://github.com/Devolutions/IronRDP/pull/722#issuecomment-2752805936)
> 
> Let's clarify: iron-remote-desktop-rdp has to re-export WASM and provide extension types. The next step is to modify the UserInteraction to accept the "backend" which will be either RDP or VNC and implements an interface (I have really missed the fact that JS/TS has a duck types). The next step should be the refactoring of the `connect` function into a builder (or I will leave it to a follow-up PR if there won't be a direct need in it right now).

Exactly!
For this PR, really focus on the core architectural refactoring.
Anything that can be a follow up should be a follow up :slightly_smiling_face:

crates/ironrdp-web/src/session.rs
CBenoit: I like that!

web-client/iron-remote-desktop-rdp/pre-build.js
```diff
@@ -6,7 +6,7 @@ let run = async function (command, cwd) {
             stdio: 'pipe',
             shell: true,
             cwd: cwd,
-            env: { ...process.env, RUSTFLAGS: '-Ctarget-feature=+simd128' },
+            env: { ...process.env, RUSTFLAGS: '-Ctarget-feature=+simd128,+bulk-memory' },
```
CBenoit, line 9: note: I think we have a similar env variable in xtask::web. Can you also update at this place?

RRRadicalEdward: Done

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -16,12 +16,16 @@
...
+    import { type RemoteDesktopModule, WasmBridgeService } from './services/wasm-bridge.service';
     import { LogType } from './enums/LogType';
     import type { ResizeEvent } from './interfaces/ResizeEvent';
     import { PublicAPI } from './services/PublicAPI';
     import { ScreenScale } from './enums/ScreenScale';
-    import { ClipboardContent, ClipboardTransaction } from '../../../crates/ironrdp-web/pkg/ironrdp_web';
+    import type { ClipboardTransaction as IClipboardTransaction } from './interfaces/ClipboardTransaction';
+    import { ClipboardContent, ClipboardTransaction } from './../../iron-remote-desktop-rdp/src/main';
+    import * as remote_desktop from './../../iron-remote-desktop-rdp/src/main';
+    // import { ClipboardContent, ClipboardTransaction } from './../../../../ironvnc/web-client/iron-remote-desktop-vnc/src/main';
```
RRRadicalEdward, line 27: Here as you can see, we can import VNC instead of an RDP and this will inject selected module into a WasmBridgeService

web-client/iron-svelte-client/src/services/server-bridge.service.ts
RRRadicalEdward: Reverted

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Done

crates/ironrdp-web/src/error.rs
RRRadicalEdward: Decided to do it in a follow-up PR

web-client/iron-remote-desktop-rdp/src/services/PublicAPI.ts
RRRadicalEdward: Moved to the iron-remote-desktop

web-client/iron-remote-desktop-rdp/src/iron-remote-desktop.svelte
RRRadicalEdward: Moved to the iron-remote-desktop

web-client/iron-remote-desktop-rdp/src/services/wasm-bridge.service.ts
RRRadicalEdward: Move this service into the `iron-remote-desktop` and we will refactor `connect` in a follow-up PR

web-client/iron-remote-desktop-rdp/src/services/logging.service.ts
RRRadicalEdward: Moved `iron-remote-desktop`

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
CBenoit: Wonderful!

CBenoit: Wait, I was expecting this logic to be in iron-svelte-client, not iron-remote-desktop.
Is it possible to choose between injecting IronRDP or injecting IronVNC from iron-svelte-client?

ARCHITECTURE.md
```diff
@@ -168,7 +168,13 @@ WebAssembly high-level bindings targeting web browsers.
...
-#### [`web-client/iron-remote-gui`](./web-client/iron-remote-gui)
+#### [`web-client/iron-remote-desktop`](./web-client/iron-remote-desktop)
+
+TypeScript interfaces exposed by WebAssembly bindings from `ironrdp-web` and used by `iron-remote-desktop-rdp`.
+
+This crate is an **API Boundary**.
+
+#### [`web-client/iron-remote-desktop-rdp`](./web-client/iron-remote-desktop-rdp)
 
 Core frontend UI used by `iron-svelte-client` as a Web Component.
```
CBenoit, line 180: note: This needs to be updated

crates/ironrdp-web/src/clipboard/transaction.rs
```diff
@@ -19,7 +19,7 @@ impl ClipboardTransaction {
 
 #[wasm_bindgen]
 impl ClipboardTransaction {
-    pub fn new() -> Self {
+    pub fn construct() -> Self {
```
CBenoit, line 22: question: Is there a reason for not using `new` here?

crates/ironrdp-web/src/session.rs
```diff
@@ -216,13 +205,28 @@ impl SessionBuilder {
...
+    pub fn extension(&self, value: JsValue) -> SessionBuilder {
+        match serde_wasm_bindgen::from_value::<Extension>(value) {
+            Ok(value) => match value {
+                Extension::KdcProxyUrl(kdc_proxy_url) => self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url),
+                Extension::Pcb(pcb) => self.0.borrow_mut().pcb = Some(pcb),
+                Extension::DisplayControl(use_display_control) => {
+                    self.0.borrow_mut().use_display_control = use_display_control
+                }
+            },
+            Err(err) => error!("Provided JsValue is not a valid extension value: {err:?}"),
```
CBenoit, line 217: https://github.com/Devolutions/IronRDP/blob/401cedf01014d8400b460911764d9ebc837900fc/STYLE.md#logging

```suggestion
            Err(error) => error!(%error, "Provided JsValue is not a valid extension value"),
```

crates/ironrdp-web/src/session.rs
```diff
@@ -297,11 +301,11 @@ impl SessionBuilder {
         loop {
             match ws.state() {
                 websocket::State::Closing | websocket::State::Closed => {
-                    return Err(IronRdpError::from(anyhow::anyhow!(
+                    return Err(RemoteDesktopError::from(anyhow::anyhow!(
                         "Failed to connect to {proxy_address} (WebSocket is `{:?}`)",
```
CBenoit, line 305: Not something you added, but this should be:
```suggestion
                        "failed to connect to {proxy_address} (WebSocket is `{:?}`)",
```

crates/ironrdp-web/src/session.rs
```diff
@@ -818,6 +829,33 @@ impl Session {
         // plain scancode events are allowed to function correctly).
         false
     }
+
+    pub fn extension_call(&self, value: JsValue) -> Result<JsValue, RemoteDesktopError> {
+        match serde_wasm_bindgen::from_value::<SessionExtensionCall>(value)
+            .map_err(|err| anyhow::anyhow!("provided JsValue is not a valid extension call: {err:?}"))?
```
CBenoit, line 835: style: Can be simplified:

```suggestion
        match serde_wasm_bindgen::from_value::<SessionExtensionCall>(value)
            .context("provided JsValue is not a valid extension call")?
```

crates/ironrdp-web/src/session.rs
```diff
@@ -1039,8 +1077,8 @@ where
                 } => (x224_connection_response, server_cert_chain, server_addr),
                 ironrdp_rdcleanpath::RDCleanPath::Err(error) => {
                     return Err(
-                        IronRdpError::from(anyhow::anyhow!("received an RDCleanPath error: {error}"))
-                            .with_kind(IronRdpErrorKind::RDCleanPath),
+                        RemoteDesktopError::from(anyhow::anyhow!("received an RDCleanPath error: {error}"))
+                            .with_kind(RemoteDesktopErrorKind::RDCleanPath),
```
CBenoit, line 1081: note: Again, not your code, but it’s better to do this:

```suggestion
                        RemoteDesktopError::from(anyhow::Error::new(error).context("received an RDCleanPath error"))
                            .with_kind(RemoteDesktopErrorKind::RDCleanPath),
```

web-client/README.md
```diff
@@ -2,7 +2,7 @@
 
 IronRDP also supports the web browser as a first class target.
 
-See the [iron-remote-gui](./iron-remote-gui) for the reusable Web Component, and [iron-svelte-client](./iron-svelte-client) for a demonstration.
+See the [iron-remote-desktop-rdp](./iron-remote-desktop-rdp) for the reusable Web Component, and [iron-svelte-client](./iron-svelte-client) for a demonstration.
```
CBenoit, line 5: issue: The reusable web component is `iron-remote-desktop`.
```suggestion
See the [iron-remote-desktop](./iron-remote-desktop) for the reusable Web Component, and [iron-svelte-client](./iron-svelte-client) for a demonstration.
```
`iron-remote-desktop-rdp` would be the RDP backend based on IronRDP!

web-client/iron-remote-desktop-rdp/README.md
CBenoit, line 1: This should be the README of `iron-remote-desktop`, not `iron-remote-desktop-rdp` I think

web-client/iron-remote-desktop-rdp/pre-build.js
CBenoit, line 1: issue: I think pre-build.js should be in `iron-remote-desktop`, not `iron-remote-desktop-rdp`

web-client/iron-svelte-client/README.md
```diff
@@ -1,7 +1,7 @@
 # SvelteKit UI for IronRDP
 
 Web-based frontend using [`SvelteKit`](https://kit.svelte.dev/) and [`Material`](https://material.io) frameworks.
-This is a simple wrapper around the `iron-remote-gui` Web Component demonstrating how to use the API.
+This is a simple wrapper around the `iron-remote-desktop-rdp` Web Component demonstrating how to use the API.
```
CBenoit, line 4: issue: Needs update. The web component is from `iron-remote-desktop`. You can also mention `iron-remote-desktop-rdp` as the backend. My expectation is that `iron-svelte-client` has a dependency on both `iron-remote-desktop` and `iron-remote-desktop-rdp`.

xtask/src/web.rs
```diff
@@ -1,6 +1,7 @@
 use crate::prelude::*;
 
-const IRON_REMOTE_GUI_PATH: &str = "./web-client/iron-remote-gui";
+const IRON_REMOTE_DESKTOP_PATH: &str = "./web-client/iron-remote-desktop";
+const IRON_REMOTE_DESKTOP_PATH_RDP: &str = "./web-client/iron-remote-desktop-rdp";
```
CBenoit, line 4:
```suggestion
const IRON_REMOTE_DESKTOP_PATH: &str = "./web-client/iron-remote-desktop";
const IRON_REMOTE_DESKTOP_RDP_PATH: &str = "./web-client/iron-remote-desktop-rdp";
```

CBenoit:
Good work! I like the direction!

Let me clarify my expectations further:

- I want to publish three NPM packages: `iron-remote-desktop`, `iron-remote-desktop-rdp` and `iron-remote-desktop-vnc`
- `iron-svelte-client` should a dependency on both `iron-remote-desktop` and `iron-remote-desktop-rdp`.
  - Dependency injection happens in `iron-svelte-client`, not in `iron-remote-desktop`. `iron-remote-desktop` does not know about `iron-remote-desktop-rdp` or `iron-remote-desktop-vnc`.
  - We could even rename `iron-svelte-client` into `ironrdp-svelte-client`, to clarify that it is only used as a testing web client for the IronRDP backend. (In a follow up PR.)
- In the Devolutions Gateway standalone, we have a dependency on all the NPM packages: `iron-remote-desktop`, `iron-remote-desktop-rdp` and `iron-remote-desktop-vnc`.
  - We inject either `iron-remote-desktop-rdp` or `iron-remote-desktop-vnc` depending on whether we want to connect to a RDP server, or a VNC server.
  - Only the consumer knows about which backend is loaded, and can use extensions as appropriate.

web-client/iron-remote-desktop/src/services/wasm-bridge.service.ts
```diff
@@ -426,12 +442,14 @@ export class WasmBridgeService {
...
-            syncKanaModeActive,
-        );
+        this.session?.extension_call({
+            SynchronizeLockKeys: {
+                scroll_lock: syncScrollLockActive,
+                num_lock: syncNumsLockActive,
+                caps_lock: syncCapsLockActive,
+                kana_lock: syncKanaModeActive,
+            },
+        });
```
CBenoit, line 452: thought: Maybe it’s simpler to keep `synchronize_lock_keys` in the common API, the same way as previously. It’s fine if it’s a no-op in `ironvnc-web`.

web-client/iron-remote-desktop/src/interfaces/DeviceEvent.ts
```diff
@@ -0,0 +1,11 @@
+export interface DeviceEvent {
+    new_mouse_button_pressed(button: number): DeviceEvent;
+    new_mouse_button_released(button: number): DeviceEvent;
+    new_mouse_move(x: number, y: number): DeviceEvent;
+    new_wheel_rotations(vertical: boolean, rotation_units: number): DeviceEvent;
+    new_key_pressed(scancode: number): DeviceEvent;
+    new_key_released(scancode: number): DeviceEvent;
+    new_unicode_pressed(unicode: string): DeviceEvent;
+    new_unicode_released(unicode: string): DeviceEvent;
+    free(): void;
```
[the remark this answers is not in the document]
RRRadicalEdward: We don't. I will remove it

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Done

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Done

crates/ironrdp-web/src/lib.rs
```diff
@@ -23,7 +23,7 @@ mod session;
 use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
-pub fn ironrdp_init(log_level: &str) {
+pub fn remote_desktop_init(log_level: &str) {
```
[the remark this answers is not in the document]
RRRadicalEdward: Done

web-client/README.md
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/interfaces/DeviceEvent.ts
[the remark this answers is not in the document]
RRRadicalEdward: Removed

xtask/src/web.rs
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/services/wasm-bridge.service.ts
RRRadicalEdward: Reverted

crates/ironrdp-web/src/clipboard/transaction.rs
RRRadicalEdward: We can't have `new` in an interface

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Done

web-client/iron-remote-desktop-rdp/README.md
RRRadicalEdward: Done

web-client/iron-svelte-client/README.md
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: Change the logic to do dependency injection in  iron-svelte-client

web-client/iron-remote-desktop-rdp/pre-build.js
RRRadicalEdward, line 1: `wasm-pack` generates the` ironrdp_web.js` that contains URL to the WASM module: 
```js
module_or_path = new URL('ironrdp_web_bg.wasm', import.meta.url);
```
However, during a build process for `iron-remote-desktop-rdp`, Vite automatically inlines the `.wasm` file:
`new URL('data:application/wasm;base64,AGFzbQ...', import.meta.url);`
 
The problem is, when we are importing the builded `iron-remote-dekstop-rdp` module in the `remote-desktop.svelte`, the Vite treats this URL as file path.
In result, we have incorrect URL which block loading the WASM module:
```js
new URL('http:localhost:5173/static/iron-remote-desktop-rdp/data:application/wasm;base64,AGFzbQ...', import.meta.url);
```
The solution would be to disable inlining, but Vite does not provide options to do this for our case.
The issue reference: https://github.com/vitejs/vite/issues/4454.

To work around it, I added _pre-build.js_ and _post-build.js_ files:
1. prebuild: rename _ironrdp_web_bg.wasm_ to _irondrp_web_bg1.wasm_ to prevent inlining
2. Then, the package get built
3. postbuild: rename the WASM file back and copy it to `/dist` folder.

RRRadicalEdward: _pre-build_ for `iron-remote-dekstop-rdp` runs the `wasm-pack` for `ironrdp_web`, so all is well here.
For `iron-remote-desktop`, I removed the `pre-build` file because it also runs the `wasm-pack` (duplicate functionality)

CBenoit: That’s a problem… we are relying on it being inlined, and I would like to keep it that way. I don’t know how to fix the problem you encountered on the top of the head, but let’s find a way.

RRRadicalEdward: The issue is that `wasm-pack` generates _ironrdp_web.js_ file. Vite,when uses it, sees the `new URL(url, import.meta.url)`, and treats the `url` as a file path even if it's inlined URL.
 
So, I propose to resolve this by replacing the `new URL` with the inlined URL string directly.
 
We have two options to implement this:
1. Post-build in the `iron-remote-desktop-rdp`: modify the resulting JS file to implement needed fix.
2. Modify the generated JS file right after `wasm-pack` stage. We could use cargo xtask for this. 
This way, the WASM NPM package would contain modified JS file that we than can use then in `iron-remote-desktop-rdp`.
 
I believe the second option is better since it avoids the need to modify the file after every build and keeps the WASM NPM package self-contained.

CBenoit: Sounds good!

crates/ironrdp-web/src/session.rs
```diff
@@ -216,13 +205,28 @@ impl SessionBuilder {
...
+    pub fn extension(&self, value: JsValue) -> SessionBuilder {
+        match serde_wasm_bindgen::from_value::<Extension>(value) {
+            Ok(value) => match value {
+                Extension::KdcProxyUrl(kdc_proxy_url) => self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url),
+                Extension::Pcb(pcb) => self.0.borrow_mut().pcb = Some(pcb),
+                Extension::DisplayControl(use_display_control) => {
+                    self.0.borrow_mut().use_display_control = use_display_control
+                }
+            },
+            Err(error) => error!(%error, "Provided JsValue is not a valid extension value"),
```
CBenoit, line 217:
```suggestion
            Err(error) => error!(%error, "Unsupported extension value"),
```

crates/ironrdp-web/src/session.rs
```diff
@@ -216,13 +205,28 @@ impl SessionBuilder {
...
+        }
+
+        self.clone()
+    }
+
+    pub fn extension_call(&self, _ident: String, _call: js_sys::Function) -> SessionBuilder {
+        warn!("Extension calls are not supported yet");
+
         self.clone()
     }
```
CBenoit, line 227: question: In the design I had, `extension_call` was only part of `Session`, not `SessionBuilder`. Do you have a use case for `extension_call` on `SessionBuilder`?

crates/ironrdp-web/src/error.rs
```diff
@@ -3,7 +3,7 @@ use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
 #[derive(Clone, Copy)]
-pub enum IronRdpErrorKind {
+pub enum RemoteDesktopErrorKind {
```
CBenoit, line 6: thought: Should we rename to `IronErrorKind`/`IronError`? Since we already have `iron_init`. Just discussing the idea, no need to rename in this PR.

web-client/iron-remote-desktop/src/interfaces/DeviceEvent.ts
```diff
@@ -0,0 +1,10 @@
+export interface DeviceEvent {
+    new_mouse_button_pressed(button: number): DeviceEvent;
+    new_mouse_button_released(button: number): DeviceEvent;
+    new_mouse_move(x: number, y: number): DeviceEvent;
+    new_wheel_rotations(vertical: boolean, rotation_units: number): DeviceEvent;
+    new_key_pressed(scancode: number): DeviceEvent;
+    new_key_released(scancode: number): DeviceEvent;
+    new_unicode_pressed(unicode: string): DeviceEvent;
+    new_unicode_released(unicode: string): DeviceEvent;
```
CBenoit, line 9: thought: Maybe we should remove the `new_` prefix.

web-client/iron-remote-desktop/src/interfaces/DesktopSize.ts
```diff
@@ -1,4 +1,6 @@
 ﻿export interface DesktopSize {
     width: number;
     height: number;
+
+    construct(width: number, height: number): DesktopSize;
```
CBenoit, line 5: note: It seems the position of this item is inconsistent across the interfaces.

web-client/iron-svelte-client/README.md
CBenoit, line 4: Leftover?
```suggestion
This is a simple wrapper around the `iron-remote-desktop` Web Component demonstrating how to use the API.
```

web-client/iron-svelte-client/README.md
```diff
@@ -111,13 +111,13 @@ If you have a Rust toolchain available, you can use the [`tokengen`][tokengen] t
 
 ## Run in development mode
 
-First, run `npm install` in the [iron-remote-gui](../iron-remote-gui/) folder, and then `npm install` in [iron-svelte-client](./) folder.
+First, run `npm install` in the [iron-remote-desktop-rdp](../iron-remote-desktop-rdp/) folder, and then `npm install` in [iron-svelte-client](./) folder.
```
CBenoit, line 114: Should this be changed to this?
```suggestion
First, run `npm install` in [iron-remote-desktop](../iron-remote-desktop/) and [iron-remote-desktop-rdp](../iron-remote-desktop-rdp/) folders, and then `npm install` in [iron-svelte-client](./) folder.
```

web-client/iron-remote-desktop/src/services/wasm-bridge.service.ts
```diff
@@ -23,14 +12,21 @@ import { SpecialCombination } from '../enums/SpecialCombination';
...
+import type { DeviceEvent } from '../interfaces/DeviceEvent';
+import type { SessionTerminationInfo } from '../interfaces/SessionTerminationInfo';
+import type { RemoteDesktopModule } from '../interfaces/RemoteDesktopModule';
 
 type OnRemoteClipboardChanged = (transaction: ClipboardTransaction) => void;
 type OnRemoteReceivedFormatsList = () => void;
 type OnForceClipboardUpdate = () => void;
 
 export class WasmBridgeService {
+    private importedModule: RemoteDesktopModule;
```
CBenoit, line 29: naming: `rdModule`
```suggestion
    // Remote Desktop implementation.
    private rdModule: RemoteDesktopModule;
```

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -364,11 +373,11 @@
 
                     file.arrayBuffer().then((buffer: ArrayBuffer) => {
                         const strict_buffer = new Uint8Array(buffer);
-                        let content = ClipboardContent.new_binary(mime, strict_buffer);
+                        let content = wasmService.constructClipboardContentFromBinary(mime, strict_buffer);
                         transaction.add_content(content);
 
                         if (!transaction.is_empty()) {
-                            wasmService.onClipboardChanged(transaction);
+                            wasmService.onClipboardChanged(transaction as ClipboardTransaction);
```
CBenoit, line 380: question: Is the cast really required?

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -51,8 +55,7 @@
 
     let viewerStyle = $state('');
     let wrapperStyle = $state('');
-
-    let wasmService = new WasmBridgeService();
+    let wasmService = new WasmBridgeService(module);
```
CBenoit, line 58: thought: `WasmBridgeService` should probably be renamed since we are not wrapping a WASM necessarily anymore. `RemoteDesktopService` may be appropriate? What do you think?

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -129,18 +132,24 @@
...
+
     // This function is required to covert `ClipboardTransaction` to a object that can be used
     // with `ClipboardItem` API.
     function clipboardTransactionToRecord(transaction: ClipboardTransaction): Record<string, Blob> {
         let result = {} as Record<string, Blob>;
 
         for (const item of transaction.content()) {
-            if (!(item instanceof ClipboardContent)) {
+            if (!isContent(item)) {
                 continue;
```
CBenoit, line 147: question: Do we really need to check this?

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -129,18 +132,24 @@
         return (evt.ctrlKey && evt.code === 'KeyV') || evt.code == 'Paste';
     }
 
+    function isContent(content: ClipboardContent) {
+        // Check whether function exists. To make it more robust we can check every method.
+        return content.mime_type() !== undefined || content.value() !== undefined;
+    }
```
CBenoit, line 138: thought: It looks like we are checking whether a value implementing the `ClipboardContent` interface, is implementing the `ClipboardContent` interface :thinking:

web-client/iron-remote-desktop-rdp/pre-build.js
RRRadicalEdward: @CBenoit 
Pushed a commit with the fix - https://github.com/Devolutions/IronRDP/pull/722/commits/ecd0c60db560497b420e73fea1eff0a8212cae0b

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
```diff
@@ -0,0 +1,89 @@
+import type { Session } from './Session';
+import type { DesktopSize } from './DesktopSize';
+import type { ClipboardTransaction } from './ClipboardTransaction';
+export interface SessionBuilder {
```
CBenoit, line 4:
```suggestion
import type { ClipboardTransaction } from './ClipboardTransaction';

export interface SessionBuilder {
```

web-client/iron-remote-desktop/src/interfaces/RemoteDesktopModule.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { SessionTerminationInfo } from './SessionTerminationInfo';
+import type { ClipboardTransaction } from './ClipboardTransaction';
+import type { ClipboardContent } from './ClipboardContent';
+
+export interface RemoteDesktopModule {
+    init: () => Promise<unknown>;
+    DesktopSize: DesktopSize;
+    DeviceEvent: DeviceEvent;
+    InputTransaction: InputTransaction;
+    iron_init: (logLevel: string) => void;
```
CBenoit, line 16:
```suggestion
    init: () => Promise<unknown>;
    iron_init: (logLevel: string) => void;
    DesktopSize: DesktopSize;
    DeviceEvent: DeviceEvent;
    InputTransaction: InputTransaction;
```

web-client/iron-remote-desktop/README.md
```diff
@@ -91,4 +92,4 @@ connect(
 > `setScale(scale: ScreenScale)`
 >
 > Sets the scale behavior of the canvas.
-> See the [ScreenScale](./src/services/user-interaction-service.ts) enum for possible values.
+> See the [ScreenScale](../iron-remote-desktop/src/services/user-interaction-service.ts) enum for possible values.
```
CBenoit, line 95: Leftover?

web-client/iron-remote-desktop/README.md
```diff
@@ -41,14 +42,14 @@ Other advanced functionalities (sharing / copy past...) are not implemented yet.
...
+You can add some parameters for default initialization on the component `<iron-remote-desktop-rdp />`.
 
 > Note that due to a limitation of the framework all parameters need to be lower-cased.
 
 - `scale`: The scaling behavior of the distant screen. Can be `fit`, `real` or `full`. Default is `real`;
-- `verbose`: Show logs from `iron-remote-gui`. `true` or `false`. Default is `false`.
+- `verbose`: Show logs from `iron-remote-desktop-rdp`. `true` or `false`. Default is `false`.
 - `debugwasm`: Show debug info from web assembly. Can be `"OFF"`, `"ERROR"`, `"WARN"`, `"INFO"`, `"DEBUG"`, `"TRACE"`. Default is `"OFF"`.
-- `flexcentre`: Helper to force `iron-remote-gui` a flex and centering the content automatically. Otherwise, you need to manage manually. Default is `true`.
+- `flexcentre`: Helper to force `iron-remote-desktop-rdp` a flex and centering the content automatically. Otherwise, you need to manage manually. Default is `true`.
```
CBenoit, line 52:
```suggestion
- `flexcentre`: Helper to force `iron-remote-desktop` a flex and centering the content automatically. Otherwise, you need to manage manually. Default is `true`.
```

web-client/iron-remote-desktop-rdp/README.md
```diff
@@ -0,0 +1,23 @@
+# Iron Remote Desktop RDP
+
+This is implementation of `RemoteDesktopModule` interface from [iron-remote-desktop](../iron-remote-desktop) for RDP connection.
+
+## Development
+
+Make you modification in the source code then use [iron-svelte-client](../iron-svelte-client) to test.
```
CBenoit, line 7:
```suggestion
Make your modifications in the source code then use [iron-svelte-client](../iron-svelte-client) to test.
```

web-client/iron-remote-desktop-rdp/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@devolutions/iron-remote-desktop-rdp",
+  "author": "Nicolas Girot",
+  "email": "ngirot@devolutions.net",
+  "contributors": [
+    "Benoit Cortier",
+    "Irving Ou",
+    "Vladislav Nikonov",
+    "Zacharia Ellaham"
+  ],
```
CBenoit, line 10: note: Feel free to add your name to the list of contributors.

xtask/src/web.rs
```diff
@@ -36,22 +42,35 @@ pub fn check(sh: &Shell) -> anyhow::Result<()> {
...
+    // Modify the js file to get rid of the `URL` object.
+    // Vite doesn't work properly with inlined urls in `new URL(url, import.meta.url)`.
+    ironrdp_web_js_content = format!(
+        "import wasmUrl from './ironrdp_web_bg.wasm?url';\n\n{}",
+        ironrdp_web_js_content
+    );
+    ironrdp_web_js_content =
+        ironrdp_web_js_content.replace("new URL('ironrdp_web_bg.wasm', import.meta.url)", "wasmUrl");
+
+    fs::write(&ironrdp_web_js_file_path, ironrdp_web_js_content)?;
```
CBenoit, line 71: suggestion: More idiomatic to not use `mut` here. You can shadow the previous binding.

```rust
    let ironrdp_web_js_content = fs::read_to_string(&ironrdp_web_js_file_path)?;

    // Modify the js file to get rid of the `URL` object.
    // Vite doesn't work properly with inlined urls in `new URL(url, import.meta.url)`.
    ironrdp_web_js_content = format!(
        "import wasmUrl from './ironrdp_web_bg.wasm?url';\n\n{}",
        ironrdp_web_js_content
    );
    let ironrdp_web_js_content =
        ironrdp_web_js_content.replace("new URL('ironrdp_web_bg.wasm', import.meta.url)", "wasmUrl");

    fs::write(&ironrdp_web_js_file_path, ironrdp_web_js_content)?;
```

CBenoit (commented):
Really good work so far. Remaining comments are "minor" :slightly_smiling_face: 
That was a big refactoring, but we are almost done!

crates/ironrdp-web/src/session.rs
RRRadicalEdward: Done

web-client/iron-svelte-client/src/lib/popup-screen/popup-screen.svelte
```diff
@@ -1,7 +1,8 @@
 <script lang="ts">
     import { onMount } from 'svelte';
     import { setCurrentSessionActive, userInteractionService } from '../../services/session.service';
-    import type { UserInteraction } from '../../../static/iron-remote-gui';
+    import type { UserInteraction } from '../../../static/iron-remote-desktop';
+    import RemoteDesktop from '../../../static/iron-remote-desktop-rdp';
```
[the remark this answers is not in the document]
RRRadicalEdward: Done

web-client/iron-svelte-client/README.md
RRRadicalEdward: Yes. Fixed

web-client/iron-remote-desktop/src/services/wasm-bridge.service.ts
RRRadicalEdward: Renamed to `module`

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: Renamed to `RemoteDesktopService`

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
RRRadicalEdward: Done

web-client/iron-svelte-client/README.md
RRRadicalEdward: Fixed

web-client/iron-remote-desktop/src/interfaces/RemoteDesktopModule.ts
RRRadicalEdward: Done

web-client/iron-remote-desktop-rdp/README.md
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -55,7 +54,7 @@
 
     let viewerStyle = $state('');
     let wrapperStyle = $state('');
-    let wasmService = new WasmBridgeService(module);
+    let wasmService = new RemoteDesktopService(module);
```
CBenoit, line 57: `wasmService` should be renamed as well

web-client/iron-svelte-client/vite.config.ts
```diff
@@ -6,6 +6,11 @@ import topLevelAwait from 'vite-plugin-top-level-await';
 const config: UserConfig = {
     mode: 'process.env.MODE' || 'development',
     plugins: [sveltekit(), wasm(), topLevelAwait()],
+    server: {
+        fs: {
+            allow: ['./static'],
+        },
+    },
```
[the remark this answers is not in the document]
RRRadicalEdward: It's necessary to allow a server to access the ./static folder, where the` iron-remote-desktop` and `iron-resmote-desktop-rdp` packages will be after the build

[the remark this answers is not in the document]
CBenoit: Do you know why this was not necessary until now?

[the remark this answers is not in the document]
RRRadicalEdward: There were no imports of files from the static folder earlier.
Only TS types were imported, which is a TS-specific thing and does not take place in compiled JS code.

web-client/iron-svelte-client/src/models/desktop-size.ts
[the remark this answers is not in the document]
RRRadicalEdward: No, it's not possible to use `DesktopSize` exported from iron-remote-desktop-rdp. That's because the `construct` function in this class is static, so the resulting instance would not have this function.
So, we need to add another `DesktopSize` class that will have the construct function in its instance (so this function is not static)

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: Done

xtask/src/web.rs
RRRadicalEdward: Done

web-client/iron-remote-desktop-rdp/package.json
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/interfaces/DesktopSize.ts
RRRadicalEdward: Fixed

web-client/iron-remote-desktop/src/interfaces/DeviceEvent.ts
RRRadicalEdward: Done

web-client/iron-svelte-client/src/models/desktop-size.ts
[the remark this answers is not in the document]
CBenoit: Wait, isn’t that just what we need? If you are constructing an instance, a static function is all that is needed, and you can’t use an instance method in the first place, isn’t it? I don’t understand the need for a non-static `constructor` function on `DesktopSize`.

web-client/iron-svelte-client/vite.config.ts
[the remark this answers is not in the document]
CBenoit: Help me to follow what is happening. I understand from what you say that this PR is modifying how things were previously working at this place, but I don’t understand why this was necessary. Can you clarify the changes you made for this to become necessary?

CBenoit:
Summing up.

Remaining questions to address before merging:
- https://github.com/Devolutions/IronRDP/pull/722#discussion_r2037009735
- https://github.com/Devolutions/IronRDP/pull/722#discussion_r2039344544
- https://github.com/Devolutions/IronRDP/pull/722#discussion_r2039333244

Suggestions to address in follow up PRs:
- https://github.com/Devolutions/IronRDP/pull/722#discussion_r2037086903
- https://github.com/Devolutions/IronRDP/pull/722#discussion_r2011785510

web-client/iron-svelte-client/vite.config.ts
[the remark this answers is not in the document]
RRRadicalEdward: Right now, we are importing the `IronRdp` module from _static/iron-remote-desktop-rdp_:
https://github.com/Devolutions/IronRDP/blob/c39055fc113c26c17cb53c2cfec0b398933fcf5e/web-client/iron-svelte-client/src/lib/remote-screen/remote-screen.svelte#L6

That's why we need to tell _Vite_ that the static folder is safe and that we can import from there.
 
Previously, we didn't import anything from the static folder because there was no Dependency Injection in the `iron-remote-desktop` component.  `import type {}` is not considered an import since it will not be in the code after compilation.

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
```diff
@@ -0,0 +1,90 @@
...
+     * Optional
+     */
+    remote_received_format_list_callback(callback: RemoteReceiveForwardListCallback): SessionBuilder;
+    /**
+     * Optional
+     */
+    force_clipboard_update_callback(callback: ForceClipboardUpdateCallback): SessionBuilder;
+    extension(value: unknown): SessionBuilder;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+    extension_call(_ident: string, _call: Function): SessionBuilder;
```
CBenoit, line 67: Leftover?

CBenoit (approved):
LGTM! Excellent work. Let’s follow up with small incremental changes in separate PRs :slightly_smiling_face:

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
RRRadicalEdward: Yes. Let me fix

web-client/iron-svelte-client/src/models/desktop-size.ts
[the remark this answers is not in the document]
RRRadicalEdward: The `DesktopSize` declared in _ironrdp_web.js_(auto generated from Rust WASM) in the following way:
```js
export class DesktopSize {
  private constructor();
  free(): void;
  static construct(width: number, height: number): DesktopSize;
  width: number;
  height: number;
}
```
The `constructor` function is static because it is auto-generated from the `contruct` in Rust, which is an associated function. So, the instance of the `DesktopSize` won't have this function. This class is re-exported in the `iron-remote-desktop-rdp`: https://github.com/Devolutions/IronRDP/blob/0ff1ed8de5a22ffe34a6a492425dc74728202fbe/web-client/iron-remote-desktop-rdp/src/main.ts#L17
 
If we create the instance of this exported `DesktopSize`, like this `IronRdp.DesktopSize.construct()`, the resulting instance won't contain the `construct` function, meaning that it doesn't implement the `DesktopSize` interface:
https://github.com/Devolutions/IronRDP/blob/0ff1ed8de5a22ffe34a6a492425dc74728202fbe/web-client/iron-remote-desktop/src/interfaces/DesktopSize.ts#L1-L6

So, we are using the [redefined DesktopSize](https://github.com/Devolutions/IronRDP/blob/0ff1ed8de5a22ffe34a6a492425dc74728202fbe/web-client/iron-svelte-client/src/models/desktop-size.ts#L1-L14) to create an instance that implements [the DesktopSize interface](https://github.com/Devolutions/IronRDP/blob/0ff1ed8de5a22ffe34a6a492425dc74728202fbe/web-client/iron-remote-desktop/src/interfaces/DesktopSize.ts#L1-L6): https://github.com/Devolutions/IronRDP/blob/0ff1ed8de5a22ffe34a6a492425dc74728202fbe/web-client/iron-svelte-client/src/lib/login/login.svelte#L17

Later in _iron-remote-desktop_, it's able to cast `DesktopSize` from `iron-remote-desktop-rdp` to the `DesktopSize` interface. 

It's my understanding of what happens.

web-client/iron-svelte-client/vite.config.ts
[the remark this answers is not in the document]
RRRadicalEdward: @CBenoit  let me know if the comment above didn't clarify the point to you

web-client/iron-svelte-client/src/models/desktop-size.ts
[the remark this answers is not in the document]
RRRadicalEdward: This could be resolved by splitting the interface into two parts: one that defines the instance fields and another that defines the `constructor` function.
 
It produces a greater issue that makes this fix not relevant: the WASM module is not initialized in the `login.svelte`. Because of this, there will be an error when calling the `construct` function, which internally calls `wasm.desktopsize_construct` (this function calls Rust code, but it is undefined because the module isn't initialized at this point).
The WASM module itself is initialized in the `iron-remote-desktop` component. We would have a double initialization error if we initialized it in `login.svelte` as well

crates/ironrdp-web/src/clipboard/transaction.rs
RRRadicalEdward: Fixed in https://github.com/Devolutions/IronRDP/pull/747

crates/ironrdp-web/src/error.rs
RRRadicalEdward: Fixed in https://github.com/Devolutions/IronRDP/pull/747

web-client/iron-remote-desktop-rdp/src/services/wasm-bridge.service.ts
RRRadicalEdward: >  we will refactor connect in a follow-up PR

#749

crates/ironrdp-web/src/error.rs
RRRadicalEdward: Done in #755

PR #729, opened 2025-03-29
feat(cliprdr)!: add on_ready() callback

elmarco:
Give a hint to the backend when the channel is actually connected & ready to process messages.


CBenoit (approved):
LGTM!

PR #730, opened 2025-03-29
Preliminary refactoring patches for QOI

elmarco:
subset of commits taken from #670

### DIFF
--- a/crates/ironrdp-bench/benches/bench.rs
+++ b/crates/ironrdp-bench/benches/bench.rs
@@ -10,8 +10,8 @@ pub fn rfx_enc_tile_bench(c: &mut Criterion) {
     let quant = rfx::Quant::default();
     let algo = rfx::EntropyAlgorithm::Rlgr3;
     let bitmap = BitmapUpdate {
-        top: 0,
-        left: 0,
+        x: 0,
+        y: 0,
         width: NonZero::new(64).unwrap(),
         height: NonZero::new(64).unwrap(),
         format: ironrdp_server::PixelFormat::ARgb32,
@@ -25,8 +25,8 @@ pub fn rfx_enc_bench(c: &mut Criterion) {
     let quant = rfx::Quant::default();
     let algo = rfx::EntropyAlgorithm::Rlgr3;
     let bitmap = BitmapUpdate {
-        top: 0,
-        left: 0,
+        x: 0,
+        y: 0,
         width: NonZero::new(2048).unwrap(),
         height: NonZero::new(2048).unwrap(),
         format: ironrdp_server::PixelFormat::ARgb32,
--- a/crates/ironrdp-bench/benches/bench.rs
+++ b/crates/ironrdp-bench/benches/bench.rs
@@ -15,7 +15,7 @@ pub fn rfx_enc_tile_bench(c: &mut Criterion) {
         width: NonZero::new(64).unwrap(),
         height: NonZero::new(64).unwrap(),
         format: ironrdp_server::PixelFormat::ARgb32,
-        data: vec![0; 64 * 64 * 4],
+        data: vec![0; 64 * 64 * 4].into(),
         stride: 64 * 4,
     };
     c.bench_function("rfx_enc_tile", |b| b.iter(|| rfx_enc_tile(&bitmap, &quant, algo, 0, 0)));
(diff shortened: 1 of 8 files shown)


CBenoit:
For API breaking refactors, can you use `fix!` or `feat!`? `refactor` commits are typically not part of the changelog (I may adjust things)

crates/ironrdp-server/src/display.rs
```diff
@@ -67,10 +68,35 @@ pub struct BitmapUpdate {
...
+            top,
+            left,
+            width,
+            height,
+            format,
+            data: data.into(),
+            stride,
+        }
+    }
+}
```
CBenoit: suggestion: I would suggest we don’t add this function. I think it’s much clearer to use the struct initialization syntax, so we need to name the fields. With a function, it’s easier to confuse the parameter orders, especially since the types are u16 or NonZeroU16.

crates/ironrdp-server/src/display.rs
```diff
@@ -56,6 +56,44 @@ pub struct ColorPointer {
     pub xor_mask: Vec<u8>,
 }
 
+pub struct Framebuffer {
+    pub width: NonZeroU16,
+    pub height: NonZeroU16,
+    pub format: PixelFormat,
+    pub data: BytesMut,
```
CBenoit, line 63: question: What is the benefit of using `BytesMut` at this place?

crates/ironrdp-bench/benches/bench.rs
```diff
@@ -10,8 +10,8 @@ pub fn rfx_enc_tile_bench(c: &mut Criterion) {
     let quant = rfx::Quant::default();
     let algo = rfx::EntropyAlgorithm::Rlgr3;
     let bitmap = BitmapUpdate {
-        top: 0,
-        left: 0,
+        x: 0,
+        y: 0,
```
CBenoit, line 14: question: I think I’m fine with this change if you really want to, but I’m not sure whether this is really more idiomatic and less confusing. For instance, `x` and `y` could denote the center of the rectangle, while `top` and `left` and much more unambiguous.

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -114,7 +114,7 @@ impl UpdateEncoder {
     }
 
     pub(crate) fn bitmap(&mut self, bitmap: BitmapUpdate) -> Result<UpdateFragmenter<'_>> {
-        self.bitmap_updater.handle(bitmap, &mut self.pdu_encoder)
+        self.bitmap_updater.handle(&bitmap, &mut self.pdu_encoder)
     }
```
CBenoit: question: Any reason to not take BitmapUpdate by reference here too?

CBenoit: question: Any reason to not take BitmapUpdate by reference here too?

crates/ironrdp-server/src/display.rs
elmarco: ok

crates/ironrdp-server/src/display.rs
elmarco: it's not nessarily a Vec<u8>, we may have shared memory. Although I realize shared memory may be problematic too... Perhaps a whole copy is inavoidable..

crates/ironrdp-server/src/encoder/mod.rs
elmarco: There is more refactoring on this area with the following patches, but I can include that here too I guess.

crates/ironrdp-bench/benches/bench.rs
elmarco: To me, "top/left" do not clearly denote coordinates. It also bothers me a bit to have "foo.x + bar.left" for example. I would rather use the same field name. But if we rather keep mixing the two, fine.

crates/ironrdp-server/src/encoder/mod.rs
elmarco: actually, the patch right after requires to pass the update.

crates/ironrdp-server/src/display.rs
elmarco: User has to be careful that shared memory will be owned by the server. But we should still allow shared memory to avoid copy.

crates/ironrdp-bench/benches/bench.rs
CBenoit: Okay, let’s go with x, y

crates/ironrdp-server/src/display.rs
CBenoit: I assume that at some point in the future `freeze()` is called, and memory is actually shared then?
This sounds good to me, but as you point out, let’s be careful about how we use this: we experienced issues in the past where the way BytesMut was used caused too much memory was being allocated and freed, resulting in both performance and memory usage being poor.
I assume you are measuring the CPU and memory usage, and have a good insight into what you are doing.

PR #733, opened 2025-04-01
refactor(server): make UpdateEncoder::update() an iterator 

elmarco:
A single display update can now result in multiple commands / update
code (FastPathUpdate).

The update dispatching and bitmap encoding is now done by the
UpdateEncoder itself.

(preliminary patches from #670)

### DIFF
--- a/crates/ironrdp-server/Cargo.toml
+++ b/crates/ironrdp-server/Cargo.toml
@@ -13,7 +13,6 @@ categories.workspace = true
 
 [lib]
 doctest = true
-test = false
 
 [features]
 default = ["rayon"]
--- a/crates/ironrdp-server/Cargo.toml
+++ b/crates/ironrdp-server/Cargo.toml
@@ -13,6 +13,7 @@ categories.workspace = true
 
 [lib]
 doctest = true
+test = false
 
 [features]
 default = ["rayon"]
--- a/crates/ironrdp-server/src/encoder/mod.rs
+++ b/crates/ironrdp-server/src/encoder/mod.rs
@@ -2,7 +2,7 @@ use core::fmt;
 
 use anyhow::{Context, Result};
 use ironrdp_acceptor::DesktopSize;
-use ironrdp_core::{Encode, WriteCursor};
+use ironrdp_pdu::encode_vec;
 use ironrdp_pdu::fast_path::UpdateCode;
 use ironrdp_pdu::geometry::ExclusiveRectangle;
 use ironrdp_pdu::pointer::{ColorPointerAttribute, Point16, PointerAttribute, PointerPositionAttribute};
@@ -30,7 +30,6 @@ pub(crate) struct UpdateEncoder {
     desktop_size: DesktopSize,
     // FIXME: draw updates on the framebuffer
     framebuffer: Option<Framebuffer>,
-    pdu_encoder: PduEncoder,
     bitmap_updater: BitmapUpdater,
 }
 
@@ -44,7 +43,6 @@ impl fmt::Debug for UpdateEncoder {
 
 impl UpdateEncoder {
     pub(crate) fn new(desktop_size: DesktopSize, surface_flags: CmdFlags, remotefx: Option<(EntropyBits, u8)>) -> Self {
-        let pdu_encoder = PduEncoder::new();
         let bitmap_updater = if !surface_flags.contains(CmdFlags::SET_SURFACE_BITS) {
             BitmapUpdater::Bitmap(BitmapHandler::new())
         } else if remotefx.is_some() {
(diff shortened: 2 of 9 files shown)


crates/ironrdp-server/Cargo.toml
```diff
@@ -13,7 +13,6 @@ categories.workspace = true
 
 [lib]
 doctest = true
-test = false
```
CBenoit, line 16: issue: Let’s keep the tests in the testsuite crates.
https://github.com/Devolutions/IronRDP/blob/aeb1193674641846ae1873def8c84a62a59213d5/ARCHITECTURE.md#test-at-the-boundaries-test-features-not-code

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -85,11 +83,11 @@ impl UpdateEncoder {
             xor_bpp: 32,
             color_pointer,
         };
-        let buf = self.pdu_encoder.encode(ptr)?;
-        Ok(UpdateFragmenter::new(UpdateCode::NewPointer, buf))
+        Ok(UpdateFragmenter::new(UpdateCode::NewPointer, encode_vec(&ptr)?))
     }
 
-    pub(crate) fn color_pointer(&mut self, ptr: ColorPointer) -> Result<UpdateFragmenter<'_>> {
+    #[allow(clippy::unused_self)]
```
CBenoit, line 96: nitpick: I think it’s clearer to define the method as an associated function or a free standing function instead of an instance method if it doesn’t require self. Exception: avoid API breaking changes in the future ("maybe self will be used in the future"). But this API is private so it’s not a concern.

crates/ironrdp-server/Cargo.toml
elmarco: How can we test internal code ? I understand the will to focus on external boundaries, but why not allow internal stuff to be tested?

CBenoit: To enumerate a few reasons:
- Tests on internal API often ends up being "technical debt" slowing down future refactors.
  - Somewhat related [good code is easy to delete](https://programmingisterrible.com/post/139222674273/write-code-that-is-easy-to-delete-not-easy-to), and such tests represent an investment on the existing code, making it harder to delete.
- Enabling Cargo "unit tests" slows down our CI. The explanation is written down in `ironrdp-testsuite-core`’s main.rs file IIRC.

The solution to all of this is to test at the boundaries instead.
My expectation is that it should be possible to test the same code by calling the public API of `ironrdp-server`.
Maybe the design of `ironrdp-server` is not making this straightforward though, since `ironrdp-server` is not following the sans-IO approach like the rest of the codebase.

I feel mixed because I don’t want the tests you wrote to just be deleted, but at the same time I really don’t want us to start writing this kind of tests everywhere, and moving away from this after the facts is difficult.

Maybe we can add a feature flag to expose a few high value internal APIs that we really want to test directly if it’s impractical to do it the "ideal way".

What is your opinion here?

To not further block you, I would be okay with merging now and addressing this point in a follow up.

CBenoit: > We could add an "extra-tests" cfg, but that won't help with CI time if we enable it.

What I’m suggesting is to expose the internal API for testing purposes, not to enable tests conditionally.
This is helping with CI time, because we don’t have to build and link extra binaries, since all the tests are kept in the test suite.

The trick you are using in the last commit is also interesting though.
It feels as a hack, but does not require an extra feature flag.
Instead of creating a symlink, maybe you can use the [`path` attribute for the module](https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute)?

elmarco: "path" would not get the file packaged. I know we don't publish the testsuite, but this will probably have to change.

CBenoit: I don’t think we’ll ever package the test suite, unless you have a use case for that?

CBenoit (approved):
LGTM! Merging now, as there is no blocker anymore.

crates/ironrdp-server/Cargo.toml
elmarco: I wouldn't call it a "use-case" :) but often distributions (for ex) help finding issues early on various environements, especially when package have tests.  The more tests you can ship, the better. Kind of CI at a larger scale.

CBenoit: This sounds nice! But tests are not going to be part of individual packages for IronRDP. How would you go about this? I feel that it will be necessary to rely on something else than a "crate", and you would test the IronRDP "crate collection" as a whole. It’s not really clear to me what would be the most convenient for you, but I can tell that for now I’m a bit unwilling to publish all our internal packages to crates.io.

PR #734, opened 2025-04-01
feat(tokio): add reqwest feature

elmarco:
Move the client ReqwestNetworkClient to ironrdp-tokio, so other clients can optionally use the implementation.


CBenoit (approved):
Sounds reasonable! Thank you!

PR #742, opened 2025-04-08
fix(rdpsnd): send client formats that match server

elmarco:
Windows seems to be confused if the client replies with more formats, or unknown formats (opus).

### DIFF
--- a/crates/ironrdp-rdpsnd/src/client.rs
+++ b/crates/ironrdp-rdpsnd/src/client.rs
@@ -94,6 +94,8 @@ impl Rdpsnd {
     }
 
     pub fn client_formats(&mut self) -> PduResult<RdpsndSvcMessages> {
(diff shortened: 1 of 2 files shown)


crates/ironrdp-rdpsnd/src/client.rs
```diff
@@ -93,10 +94,20 @@ impl Rdpsnd {
     }
 
     pub fn client_formats(&mut self) -> PduResult<RdpsndSvcMessages> {
+        let server_format: HashSet<_> = self
```
CBenoit, line 97:
```suggestion
        // Windows seems to be confused if the client replies with more formats, or unknown formats (e.g.: opus).
        // We ensure to only send supported formats in common with the server.
        let server_format: HashSet<_> = self
```

CBenoit (approved):
LGTM!

PR #749, opened 2025-04-14
refactor(web-client): introduce `ConfigBuilder` for a better connection API

RRRadicalEdward:
Refactored `connect` method
https://github.com/Devolutions/IronRDP/blob/fe676eeac5bb2383bc6cd571621eb718881693e9/web-client/iron-remote-desktop/src/services/PublicAPI.ts#L16-L27
 as discussed here - https://github.com/Devolutions/IronRDP/pull/722#discussion_r2011775982. 

Follow-up to https://github.com/Devolutions/IronRDP/pull/722.

### DIFF
--- a/web-client/iron-remote-desktop-rdp/src/interfaces/Extension.ts
+++ b/web-client/iron-remote-desktop-rdp/src/interfaces/Extension.ts
@@ -1,0 +1,28 @@
+type ExtensionValue = { Pcb: string } | { KdcProxyUrl: string } | { DisplayControl: boolean };
+
+export class Extension {
+    static init(ident: string, value: unknown): ExtensionValue {
+        switch (ident) {
+            case 'Pcb':
+                if (typeof value === 'string') {
+                    return { Pcb: value };
+                } else {
+                    throw new Error('Pcb must be a string');
+                }
+            case 'KdcProxyUrl':
+                if (typeof value === 'string') {
+                    return { KdcProxyUrl: value };
+                } else {
+                    throw new Error('KdcProxyUrl must be a string');
+                }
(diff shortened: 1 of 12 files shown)


CBenoit (approved):
General review:
- I like the new builder API.
- It’s well documented, thank you!
- But as for the extensions, it seems to me that we are basically back to something that is very close to the initial extension API proposal, but slightly worse because we can’t easily hide the string keys.

About my second point:

Here is what we have in this PR:
```js
configBuilder
    .withExtension('DisplayControl', true)
    .withExtension('Pcb', pcb)
    .withExtension('KdcProxyUrl', kdc_proxy_url);
```

Here is what we would have using the `Extension` type defined in the initial proposal:
```js
configBuilder
    .withExtension(displayControlExtension(true))
    .withExtension(pcbExtension(pcb))
    .withExtension(kdcProxyUrlExtension(kdc_proxy_url));
```

I’m open to dropping the `Extension` suffix from the helper functions above and aim for something like this:
```js
configBuilder
    .withExtension(displayControl(true))
    .withExtension(preconnectionBlob(pcb))
    .withExtension(kdcProxyUrl(kdc_proxy_url));
```

In this second version, you can remove the `Extension::init` factory and associated interface.

The benefit is that it’s possible to remove the fragile string-based keys from the public API.
At this point, it’s also not necessary anymore to keep the serde-based dependencies that were previously introduced, and we can reduce the size of the resulting WASM module to what it used to be.

Requested follow ups:
- Update the `README.md` file for `iron-remote-desktop` to document the changes.
- Change the extension API to what was initially suggested as it seems to play better with the generic API exposed to the user.

PR #756, opened 2025-04-16
fix(server): use desktop size for RFX channel size

elmarco:
The CodecChannel should only be changed when the desktop size changes.

(from #670)


CBenoit (approved):
LGTM!

PR #758, opened 2025-04-17
fix(client): swap transposed width/height parameters for dynamic resizes

gabrielbauman:
Dynamic resizes were setting height to width and width to height, which ended up distorting things rather badly. Here's the fix.


gabrielbauman:
(hello from Cloudflare!)

awakecoding (approved):

awakecoding:
Good catch, thanks!

PR #761, opened 2025-04-18
refactor(web-client): improve `Extension` API

RRRadicalEdward:
I improved the API based on the feedback in https://github.com/Devolutions/IronRDP/pull/749


crates/ironrdp-web/Cargo.toml
```diff
@@ -42,12 +42,11 @@ iron-remote-desktop.path = "../iron-remote-desktop"
 # WASM
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+wasm-bindgen-derive = "0.3"
```
RRRadicalEdward, line 45: I have tried to use `wasm-bindgen` `convert` module to be able to downcast from ` JsValue` to an extension, but I have found that this module is unstable and gives different results each call (first call - fine, everything else - errors). This is a known issue: https://github.com/rustwasm/wasm-bindgen/issues/2231.
This crate fixes the issue.

crates/ironrdp-web/src/extension.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Debug, Clone)]
+pub(crate) enum Extension {
+    DisplayControl(bool),
+    Pcb(String),
+    KdcProxyUrl(String),
+}
+
+impl Extension {
+    pub(crate) fn try_from_js_value(value: JsValue) -> Result<Self, anyhow::Error> {
+        if let Ok(Some(display_control)) = try_from_js_option::<DisplayControl>(value.clone()) {
```
RRRadicalEdward, line 68: We have to `clone` every time since `wasm_bindgen_derive` doesn't have an API to take `JsValue` by reference

CBenoit (commented):
This is looking much more complex than necessary to me. Here is what I had in mind: https://github.com/Devolutions/IronRDP/pull/762

Can you check this PR instead?

RRRadicalEdward:
Done in #762

PR #762, opened 2025-04-21
refactor: rework extension API for iron-remote-desktop

CBenoit:
Rework the extension API for iron-remote-desktop as per https://github.com/Devolutions/IronRDP/pull/749#pullrequestreview-2768950469

I used this PR as basis: https://github.com/Devolutions/IronRDP/pull/761


crates/ironrdp-web/src/session.rs
```diff
@@ -207,16 +203,12 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
...
-                }
-            },
-            Err(error) => error!(%error, "Unsupported extension value"),
+    fn extension(&self, ext: Extension) -> Self {
+        iron_remote_desktop::extension_match! {
+            match ext;
+            |pcb: String| { self.0.borrow_mut().pcb = Some(pcb) };
+            |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
+            |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
         }
```
CBenoit, line 212: We directly use `JsValue`.
No extra dependency required for that.
For now there is no extension requiring to pass an "object", but if we had one, we could use the [`Reflect` module of `js_sys`](https://docs.rs/js-sys/latest/js_sys/Reflect/index.html) (that we already depend on)

crates/iron-remote-desktop/src/extension.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    ( match $ext:ident ; $( | $value:ident : $ty:ident | $operation:block ; )* ) => {
+        let ident = $ext.ident();
+
+        match ident {
+            $( stringify!($value) => $crate::extension_match!( @ $ext.into_value(), $value, $ty, $operation ), )*
+            unknown_extension => ::tracing::warn!("Unknown extension: {unknown_extension}"),
+        }
+    };
+}
```
CBenoit, line 40: Helper macro to extract extension values. This is not strictly needed, but it’s less boilerplate in the consumer code.

RRRadicalEdward (approved):
LGTM

PR #763, opened 2025-04-21
refactor(web): iron_init/setup logic

CBenoit:
- `iron_init` is renamed to `setup`, so that
  - `init` is the function for initializing the WASM module.
  - `setup` is the function taking parameters for logger and performing other setting up operations.
  - Since `setup` may be called after `init` is called, it’s possible to use some functions of the WASM module before calling `setup` if necessity arises.

- Common initializion code is moved to an hidden "internal" module of iron-remote-desktop crate, that is thus not part of the public API.

- Restored the "IronRDP is ready" debug log when `setup` is called.

### DIFF
--- a/crates/iron-remote-desktop/src/lib.rs
+++ b/crates/iron-remote-desktop/src/lib.rs
@@ -25,41 +25,6 @@ impl DesktopSize {
     }
 }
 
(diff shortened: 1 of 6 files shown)


RRRadicalEdward (approved):
Clean 🧹

PR #770, opened 2025-04-22
feat(server): add perfenc example

elmarco:
Measure the performance of the IronRDP server encoder, given a raw RGBX video input.


crates/ironrdp-testsuite-core/Cargo.toml
CBenoit, line 1: question: This file seems to have been modified by mistake?

crates/ironrdp-server/examples/perfenc.rs
CBenoit, line 1: suggestion: Instead of an example, what about a dedicated crate in a `/benches/` folder?

crates/ironrdp-testsuite-core/Cargo.toml
elmarco: no, it's because we link fast_path.rs.. I tried to explain in the commit message.

crates/ironrdp-server/examples/perfenc.rs
elmarco: you mean a top-level benches/ folder?

crates/ironrdp-testsuite-core/Cargo.toml
CBenoit: Sorry I missed that. But in general, when it’s relevant, I think it’s better to put this kind of information as an in-code comment. It’s much less convenient than digging the git history (though some tools are making this easier).

elmarco: done

crates/ironrdp-server/examples/perfenc.rs
elmarco: done

CBenoit (approved):
LGTM

PR #771, opened 2025-04-22
feat(server): find & send the damaged tiles

elmarco:
Keep a framebuffer and tile-diff against it, to save from
    encoding/sending the same bitmap data regions.

(from #670)

### DIFF
--- a/crates/ironrdp-server/src/display.rs
+++ b/crates/ironrdp-server/src/display.rs
@@ -3,6 +3,7 @@ use core::num::NonZeroU16;
 use anyhow::Result;
 use bytes::{Bytes, BytesMut};
 use ironrdp_displaycontrol::pdu::DisplayControlMonitorLayout;
+use ironrdp_graphics::diff;
 use ironrdp_pdu::pointer::PointerPositionAttribute;
 
 #[rustfmt::skip]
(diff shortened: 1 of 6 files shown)


crates/ironrdp-server/src/display.rs
```diff
@@ -231,3 +293,43 @@ pub trait RdpServerDisplay: Send {
         debug!(?layout, "Requesting layout")
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{BitmapUpdate, Framebuffer};
+    use core::num::NonZeroU16;
+    use ironrdp_graphics::{diff::Rect, image_processing::PixelFormat};
```
CBenoit, line 301: issue: Same as https://github.com/Devolutions/IronRDP/pull/733#discussion_r2022990632

elmarco: yeah.. it's a tough requirement.. this is unit-testing the internal framebuffer state after an update. Since tests are disabled this isn't even built.

CBenoit: If you want, we can merge as-is, and I’ll look into that

elmarco: sure, wfm -- there are certainly many things to be improved :) thanks

PR #775, opened 2025-04-23
refactor(web-client): refactor WASM interfaces to remove static consructor functions

RRRadicalEdward:


web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
```diff
@@ -65,14 +64,7 @@ export class RemoteDesktopService {
     }
 
     createClipboardData(): ClipboardData {
```
CBenoit, line 66: note: For myself. Verify if we still need this function.

CBenoit (approved):
LGTM!

PR #779, opened 2025-04-25
style(ironrdp-web): improve naming of clipboard variables

RRRadicalEdward:


CBenoit (approved):
Good catch! Thank you!

PR #782, opened 2025-04-29
Some preliminary refactoring for #670

elmarco:
From #670


elmarco:
@CBenoit ping

CBenoit (approved):
I’m not sure ironrdp-pdu is the appropriate place to share this code, but right now I don’t have a better suggestion to make. I’ll merge now to not block you any longer.

PR #787, opened 2025-05-08
refactor(iron-remote-desktop): use `PartialObserver` instead of single callback for `onSessionEvent`

RRRadicalEdward:
I've refactored the `onSessionEvent` method to use` PartialObserver` instead of a single callback.
This change gives the ability to pass `next` and `error` callbacks instead of only `next` in places like [this](https://github.com/Devolutions/devolutions-gateway/blob/e90f06f7bab3267eedb0586eb167e43d409f7f82/webapp/src/client/app/modules/web-client/ard/web-client-ard.component.ts#L336-L350) in Devolutions Gateway Standalone

### DIFF
--- a/web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
+++ b/web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
@@ -3,6 +3,7 @@ import type { NewSessionInfo } from './NewSessionInfo';
 import type { SessionEvent } from './session-event';
 import { ConfigBuilder } from '../services/ConfigBuilder';
 import type { Config } from '../services/Config';
+import type { PartialObserver } from 'rxjs';
 
 export interface UserInteraction {
     setVisibility(state: boolean): void;
@@ -23,7 +24,7 @@ export interface UserInteraction {
 
     setCursorStyleOverride(style: string | null): void;
 
-    onSessionEvent(callback: (event: SessionEvent) => void): void;
+    onSessionEvent(partialObserver: PartialObserver<SessionEvent>): void;
 
     resize(width: number, height: number, scale?: number): void;
 
--- a/web-client/iron-remote-desktop/src/services/PublicAPI.ts
+++ b/web-client/iron-remote-desktop/src/services/PublicAPI.ts
@@ -68,8 +68,8 @@ export class PublicAPI {
             configBuilder: this.configBuilder.bind(this),
             connect: this.connect.bind(this),
             setScale: this.setScale.bind(this),
-            onSessionEvent: (callback) => {
-                this.remoteDesktopService.sessionObserver.subscribe(callback);
+            onSessionEvent: (partialObserver) => {
+                this.remoteDesktopService.sessionObserver.subscribe(partialObserver);
             },
             ctrlAltDel: this.ctrlAltDel.bind(this),
             metaKey: this.metaKey.bind(this),
--- a/web-client/iron-svelte-client/src/lib/login/login.svelte
+++ b/web-client/iron-svelte-client/src/lib/login/login.svelte
@@ -30,20 +30,22 @@
     });
 
     const initListeners = () => {
-        userInteraction.onSessionEvent((event) => {
-            if (event.type === 2) {
-                console.log('Error event', event.data);
+        userInteraction.onSessionEvent({
+            next: (event) => {
+                if (event.type === 2) {
+                    console.log('Error event', event.data);
 
-                toast.set({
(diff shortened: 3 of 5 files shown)


CBenoit:
Before we move ahead with this change, I’d like to resurface an idea I mentioned in [this PR comment](https://github.com/Devolutions/IronRDP/pull/495#discussion_r1682383229): can we drop RxJS from `iron-remote-desktop` altogether?

Rationale:
- Leaner dependency graph: fewer transitive deps, less maintenance work tracking major RxJS releases.
- No implicit requirement for consumers: users don’t have to bundle RxJS or fight version conflicts.
- Smaller bundle size: RxJS isn't huge, but every KB matters for a web component.
- Limited upside here: our event flow looks simple enough that native EventTarget, callback function or a tiny custom emitter should suffice.

In our context, is RxJS used mainly for error propagation? `SessionEvent` is currently `IronError | string`; are we emitting any other error shapes? If we need richer diagnostics, could we extend SessionEvent (with a discriminated union) instead of reaching for RxJS?

I’m all for better diagnostics, but unless RxJS solves a real pain point, zero dependencies feels like the bigger win.
If you see clear benefits (operators we can’t replicate, planned features that need RxJS semantics, etc.), I’m open to being convinced. Let’s weigh the trade-offs explicitly before committing.

If you end up agreeing with me, do you think you could work on removing RxJS instead?

Thanks!

CBenoit:
> > Before we move ahead with this change, I’d like to resurface an idea I mentioned in [this PR comment](https://github.com/Devolutions/IronRDP/pull/495#discussion_r1682383229): can we drop RxJS from `iron-remote-desktop` altogether?
> 
> > 
> 
> > Rationale:
> 
> > 
> 
> >     * Leaner dependency graph: fewer transitive deps, less maintenance work tracking major RxJS releases.
> 
> > 
> 
> >     * No implicit requirement for consumers: users don’t have to bundle RxJS or fight version conflicts.
> 
> > 
> 
> >     * Smaller bundle size: RxJS isn't huge, but every KB matters for a web component.
> 
> > 
> 
> >     * Limited upside here: our event flow looks simple enough that native EventTarget, callback function or a tiny custom emitter should suffice.
> 
> > 
> 
> > 
> 
> > In our context, is RxJS used mainly for error propagation? `SessionEvent` is currently `IronError | string`; are we emitting any other error shapes? If we need richer diagnostics, could we extend `SessionEvent` (with a discriminated union) instead of reaching for RxJS?
> 
> > 
> 
> > I’m all for better diagnostics, but unless RxJS solves a real pain point, zero dependencies feels like the bigger win. If you see clear benefits (operators we can’t replicate, etc.), I’m open to being convinced. Let’s weigh the trade-offs explicitly before committing.
> 
> > 
> 
> > If you end up agreeing with me, do you think you could work on removing RxJS instead?
> 
> > 
> 
> > Thanks!
> 
> 
> 
> Hello. I think it should be possible to rewrite it to use `Promise` instead
> 
> 
> 
> > In our context, is RxJS used mainly for error propagation?
> 
> 
> 
> Not only error propagation, RxJS is also used to introduce `Observable`, like in the `connect`: https://github.com/Devolutions/IronRDP/blob/master/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts#L122
> 
> But I can try to replace it with `Promise`.
> 

That's actually something I'm mentioning in my original comment:

> I’m not really convinced we need `Observer`s given that all functions provided by the WASM module are simply returning `Promise`s / `Future`s. We are using an abstraction for returning many values, but we ever only return one single value. I believe we could just use `Promise` combinators when appropriate, and it wouldn’t change the style too much. I wouldn’t be surprised we could even improve over this actually.

It does not really make sense to me to use an Observable (~ AsyncIterator) because there is only a single value ever returned at this place (this is accurately represented by Future/Promise).

I think it's then possible to remove the code heavy with RxJS combinators (akin to iterator combinators, for operation on multiple values) to a more straightforward imperative code using the await keyword.

But even if we really need something like Observable (listener/observer pattern), our usage seems so minor that the native EventTarget seems enough.


> I already made local changes to Devolutions Gateway Standalone, considering changes made by this PR. 
> 
> Can we merge this PR and release a new version of `iron-remote-desktop`? I will come back to removing depending on RxJS once I'm done with integrating the new packages to Devolutions Gateway standalone and finishing with https://github.com/Devolutions/IronVNC/issues/811 (Irving's waiting for it). Does it sound good? 

Sounds good to me!

CBenoit (approved):

PR #791, opened 2025-05-16
feat(dvc): add DVC named pipe proxy support

pacmancoder:
This PR adds DVC proxy implementation based on named pipes (currently windows-only), a separate `fifio` Unix implementation is needed.

How to test:
- Run IronRDP pipe with NowProto pipe proxy:
    `cargo run -p ironrdp-client -- -u <USER> -p <PASS> <IP> --dvc-proxy Devolutions::Now::Agent=now-proto-GLOBAL`
- Run NowDvcApp project from [now-proto](https://github.com/Devolutions/now-proto) repo.
- 
image

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -5,8 +5,10 @@ use ironrdp::displaycontrol::client::DisplayControlClient;
 use ironrdp::displaycontrol::pdu::MonitorLayoutEntry;
 use ironrdp::graphics::image_processing::PixelFormat;
 use ironrdp::pdu::input::fast_path::FastPathInputEvent;
+use ironrdp::pdu::{pdu_other_err, PduResult};
 use ironrdp::session::image::DecodedImage;
 use ironrdp::session::{fast_path, ActiveStage, ActiveStageOutput, GracefulDisconnectReason, SessionResult};
+use ironrdp::svc::SvcMessage;
 use ironrdp::{cliprdr, connector, rdpdr, rdpsnd, session};
 use ironrdp_core::WriteBuf;
 use ironrdp_rdpsnd_native::cpal;
@@ -20,6 +22,7 @@ use tokio::sync::mpsc;
 use winit::event_loop::EventLoopProxy;
 
 use crate::config::{Config, RDCleanPathConfig};
+use ironrdp_dvc_pipe_proxy::DvcNamedPipeProxy;
 
 #[derive(Debug)]
 pub enum RdpOutputEvent {
@@ -43,6 +46,10 @@ pub enum RdpInputEvent {
     FastPath(SmallVec<[FastPathInputEvent; 2]>),
     Close,
     Clipboard(ClipboardMessage),
+    SendDvcMessages {
+        channel_id: u32,
+        messages: Vec<SvcMessage>,
+    },
 }
 
 impl RdpInputEvent {
@@ -51,18 +58,50 @@ impl RdpInputEvent {
     }
 }
 
+pub struct DvcPipeProxyFactory {
+    rdp_input_sender: mpsc::UnboundedSender<RdpInputEvent>,
+}
+
+impl DvcPipeProxyFactory {
+    pub fn new(rdp_input_sender: mpsc::UnboundedSender<RdpInputEvent>) -> Self {
+        Self { rdp_input_sender }
+    }
+
+    pub fn create(&self, channel_name: String, pipe_name: String) -> DvcNamedPipeProxy {
+        let rdp_input_sender = self.rdp_input_sender.clone();
+
+        DvcNamedPipeProxy::new(&channel_name, &pipe_name, move |channel_id, messages| {
+            rdp_input_sender
+                .send(RdpInputEvent::SendDvcMessages { channel_id, messages })
+                .map_err(|_error| pdu_other_err!("send DVC messages to the event loop",))?;
+
+            Ok(())
+        })
+    }
+}
+
+pub type WriteDvcMessageFn = Box<dyn Fn(u32, SvcMessage) -> PduResult<()> + Send + 'static>;
+
 pub struct RdpClient {
     pub config: Config,
     pub event_loop_proxy: EventLoopProxy<RdpOutputEvent>,
     pub input_event_receiver: mpsc::UnboundedReceiver<RdpInputEvent>,
     pub cliprdr_factory: Option<Box<dyn CliprdrBackendFactory + Send>>,
+    pub dvc_pipe_proxy_factory: DvcPipeProxyFactory,
 }
 
 impl RdpClient {
     pub async fn run(mut self) {
         loop {
             let (connection_result, framed) = if let Some(rdcleanpath) = self.config.rdcleanpath.as_ref() {
-                match connect_ws(&self.config, rdcleanpath, self.cliprdr_factory.as_deref()).await {
+                match connect_ws(
+                    &self.config,
+                    rdcleanpath,
+                    self.cliprdr_factory.as_deref(),
+                    &self.dvc_pipe_proxy_factory,
+                )
+                .await
+                {
                     Ok(result) => result,
                     Err(e) => {
                         let _ = self.event_loop_proxy.send_event(RdpOutputEvent::ConnectionFailure(e));
@@ -70,7 +109,13 @@ impl RdpClient {
                     }
                 }
             } else {
-                match connect(&self.config, self.cliprdr_factory.as_deref()).await {
+                match connect(
+                    &self.config,
+                    self.cliprdr_factory.as_deref(),
+                    &self.dvc_pipe_proxy_factory,
+                )
+                .await
+                {
                     Ok(result) => result,
                     Err(e) => {
                         let _ = self.event_loop_proxy.send_event(RdpOutputEvent::ConnectionFailure(e));
@@ -118,6 +163,7 @@ type UpgradedFramed = ironrdp_tokio::TokioFramed<Box<dyn AsyncReadWrite + Unpin 
 async fn connect(
     config: &Config,
     cliprdr_factory: Option<&(dyn CliprdrBackendFactory + Send)>,
+    dvc_pipe_proxy_factory: &DvcPipeProxyFactory,
 ) -> ConnectorResult<(ConnectionResult, UpgradedFramed)> {
     let dest = format!("{}:{}", config.destination.name(), config.destination.port());
 
@@ -131,10 +177,21 @@ async fn connect(
 
     let mut framed = ironrdp_tokio::TokioFramed::new(socket);
 
+    let mut drdynvc =
+        ironrdp::dvc::DrdynvcClient::new().with_dynamic_channel(DisplayControlClient::new(|_| Ok(Vec::new())));
+
+    // Instantiate all DVC proxies
+    for proxy in config.dvc_pipe_proxies.iter() {
+        let channel_name = proxy.channel_name.clone();
+        let pipe_name = proxy.pipe_name.clone();
+
+        trace!(%channel_name, %pipe_name, "Creating DVC proxy");
+
+        drdynvc = drdynvc.with_dynamic_channel(dvc_pipe_proxy_factory.create(channel_name, pipe_name));
+    }
+
     let mut connector = connector::ClientConnector::new(config.connector.clone(), client_addr)
-        .with_static_channel(
-            ironrdp::dvc::DrdynvcClient::new().with_dynamic_channel(DisplayControlClient::new(|_| Ok(Vec::new()))),
-        )
+        .with_static_channel(drdynvc)
         .with_static_channel(rdpsnd::client::Rdpsnd::new(Box::new(cpal::RdpsndBackend::new())))
         .with_static_channel(rdpdr::Rdpdr::new(Box::new(NoopRdpdrBackend {}), "IronRDP".to_owned()).with_smartcard(0));
 
@@ -182,6 +239,7 @@ async fn connect_ws(
     config: &Config,
     rdcleanpath: &RDCleanPathConfig,
     cliprdr_factory: Option<&(dyn CliprdrBackendFactory + Send)>,
+    dvc_pipe_proxy_factory: &DvcPipeProxyFactory,
 ) -> ConnectorResult<(ConnectionResult, UpgradedFramed)> {
     let hostname = rdcleanpath
         .url
@@ -210,10 +268,21 @@ async fn connect_ws(
 
     let mut framed = ironrdp_tokio::TokioFramed::new(ws);
 
+    let mut drdynvc =
+        ironrdp::dvc::DrdynvcClient::new().with_dynamic_channel(DisplayControlClient::new(|_| Ok(Vec::new())));
+
+    // Instantiate all DVC proxies
+    for proxy in config.dvc_pipe_proxies.iter() {
+        let channel_name = proxy.channel_name.clone();
+        let pipe_name = proxy.pipe_name.clone();
+
+        trace!(%channel_name, %pipe_name, "Creating DVC proxy");
+
+        drdynvc = drdynvc.with_dynamic_channel(dvc_pipe_proxy_factory.create(channel_name, pipe_name));
+    }
+
     let mut connector = connector::ClientConnector::new(config.connector.clone(), client_addr)
-        .with_static_channel(
-            ironrdp::dvc::DrdynvcClient::new().with_dynamic_channel(DisplayControlClient::new(|_| Ok(Vec::new()))),
-        )
+        .with_static_channel(drdynvc)
         .with_static_channel(rdpsnd::client::Rdpsnd::new(Box::new(cpal::RdpsndBackend::new())))
         .with_static_channel(rdpdr::Rdpdr::new(Box::new(NoopRdpdrBackend {}), "IronRDP".to_owned()).with_smartcard(0));
 
@@ -463,6 +532,12 @@ async fn active_session(
                             warn!("Clipboard event received, but Cliprdr is not available");
                             Vec::new()
                         }
+                    }
+                    RdpInputEvent::SendDvcMessages { channel_id, messages } => {
+                        trace!(channel_id, ?messages, "Send DVC messages");
+
+                        let frame = active_stage.encode_dvc_messages(messages)?;
+                        vec![ActiveStageOutput::ResponseFrame(frame)]
                     }
                 }
             }
--- a/crates/ironrdp-dvc-pipe-proxy/CHANGELOG.md
+++ b/crates/ironrdp-dvc-pipe-proxy/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
--- a/crates/ironrdp-dvc-pipe-proxy/Cargo.toml
+++ b/crates/ironrdp-dvc-pipe-proxy/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "ironrdp-dvc-pipe-proxy"
+version = "0.1.0"
+readme = "README.md"
+description = "DVC named pipe proxy for IronRDP"
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+authors.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[lib]
+doctest = false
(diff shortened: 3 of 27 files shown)


crates/ironrdp-client/src/rdp.rs
```diff
@@ -51,26 +58,64 @@ impl RdpInputEvent {
     }
 }
 
+pub struct DvcPipeProxyFactory {
+    rdp_input_sender: mpsc::UnboundedSender<RdpInputEvent>,
```
CBenoit, line 62: note: It’s typically not ideal to use unbounded channels because there is no backpressure. You may consider a bounded channel and using async `send` or `blocking_send` instead. The sender will have to wait if the queue is full, creating (typically) necessary backpressure when the target system is overloaded.

crates/ironrdp-client/src/rdp.rs
```diff
@@ -131,10 +177,21 @@ async fn connect(
...
+    // Instantiate all DVC proxies
+    for proxy in config.dvc_pipe_proxies.iter() {
+        let channel_name = proxy.channel_name.clone();
+        let pipe_name = proxy.pipe_name.clone();
+
+        trace!(%channel_name, %pipe_name, "Creating DVC proxy");
+
+        drdynvc = drdynvc.with_dynamic_channel(dvc_pipe_proxy_factory.create(channel_name, pipe_name));
+    }
+
```
CBenoit, line 192: question: Is it best to instantiate the DVC proxies after opening the TCP stream to the server, or it does not matter? If there is a rationale, please write it down as an inline comment please

crates/ironrdp-dvc-pipe-proxy/README.md
CBenoit, line 1: question: Can you clarify just a little bit? I believe it means that it’s now possible to use many DVCs outside of our own NowProto DVC, as long as they are "named pipe clients". Do you have any other exemple? Maybe a link to some documentation?

crates/ironrdp-dvc-pipe-proxy/CHANGELOG.md
CBenoit, line 1: nitpick: You don’t have to create the CHANGELOG.md file, but it’s probably okay to leave it as-is.

crates/ironrdp-dvc-pipe-proxy/src/platform/windows/worker.rs
```diff
@@ -0,0 +1,173 @@
...
+use ironrdp_dvc::{encode_dvc_messages, DvcEncode};
+use ironrdp_pdu::PduResult;
+use ironrdp_svc::{ChannelFlags, SvcMessage};
+
+use crate::platform::windows::error::DvcPipeProxyError;
+use crate::windows::{wait_any, wait_any_with_timeout, Event, MessagePipeServer, Semaphore, WindowsError};
+
+const PIPE_CONNECT_TIMEOUT: u32 = 10_000; // 10 seconds
+const PIPE_WRITE_TIMEOUT: u32 = 3_000; // 3 seconds
+const MESSAGE_BUFFER_SIZE: usize = 64 * 1024; // 64 KiB
```
CBenoit, line 13: suggestion: For clarity, I recommend to add a suffix such as `_SECS`, OR to use the type `Duration` to avoid unit bugs.

crates/ironrdp-dvc-pipe-proxy/src/platform/windows/worker.rs
```diff
@@ -0,0 +1,173 @@
...
+                    error!(%err, %channel_name, %pipe_name, "DVC pipe proxy write callback failed");
+                }
+            }
+
+            // Queue the read operation again
+            read_ctx.overlapped_read()?;
+            continue;
+        }
+
+        // read from mpsc and write to pipe
```
CBenoit, line 115: style: https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#inline-code-comments-are-proper-sentences

Try to keep this in mind:
- It’s easier to write useful comments
- It’s easier to not write unnecessary comments

crates/ironrdp-dvc-pipe-proxy/src/windows/handle.rs
CBenoit, line 1: issue: I don’t like this abstraction, and I would rather not have it at all for new code…

You should use https://docs.rs/windows-core/latest/windows_core/struct.Owned.html or https://docs.rs/windows-core/latest/windows_core/struct.Ref.html (or `BorrowedHandle` / `OwnedHandle` from the std, but I’m pretty sure `Owned<HANDLE>` / `Ref<Handle>` are more convenient when working with the windows crate).

crates/ironrdp-dvc-pipe-proxy/src/windows/handle.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone)]
+pub(crate) struct Handle {
+    raw: HANDLE,
+}
+
+// SAFETY: A `HANDLE` is, by definition, thread safe.
+unsafe impl Send for Handle {}
+
+// SAFETY: A `HANDLE` is simply an integer, no dereferencing is done.
+unsafe impl Sync for Handle {}
```
CBenoit, line 18: issue: Yes and no. Pointers are just integers as well, but what matters is how they are used. Both objects are pointing to a ressource, and if the handle/pointer is passed to a function, if the operation on the resource is not thread-safe, then this instance of the Handle is not thread safe. My understanding here is that you can only know if its thread safe in the upper layer abstraction, where all the usages are known. It’s not okay to implement a generic handle wrapper with `Send` and `Sync` implemented on it.

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+pub(crate) use wide_string::WideString;
+use windows::Win32::Foundation::{
+    ERROR_IO_PENDING, HANDLE, WAIT_ABANDONED_0, WAIT_EVENT, WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT,
+};
+use windows::Win32::System::Threading::{WaitForMultipleObjects, INFINITE};
+
+// WaitForMultipleObjects wrapper with timeout.
+pub(crate) fn wait_any_with_timeout(handles: &[HANDLE], timeout: u32) -> Result<usize, WindowsError> {
+    // SAFETY: FFI call with no outstanding preconditions.
+    let result = unsafe { WaitForMultipleObjects(handles, false, timeout) };
```
CBenoit, line 26: issue: This API is unsound. Also there _are_ outstanding preconditions.

> An array of object handles. For a list of the object types whose handles can be specified, see the following Remarks section. The array can contain handles to objects of different types. It may not contain multiple copies of the same handle.
>
>**If one of these handles is closed while the wait is still pending, _the function's behavior is undefined._**
>
>The handles must have the SYNCHRONIZE access right. For more information, see [Standard Access Rights](https://learn.microsoft.com/en-us/windows/desktop/SecAuthZ/standard-access-rights).

https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects

As stated in the documentation, if the handle is closed, this call is UB. But for instance if you take [`BorrowedHandle`](https://doc.rust-lang.org/stable/std/os/windows/io/struct.BorrowedHandle.html) instead, you are guaranteed that the handles are not closed, so you may keep the function as safe (the safety invariants are enforced somewhere else).

Please double check that thoroughly.

crates/ironrdp-dvc-pipe-proxy/src/windows/pipe.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// RAII wrapper for WinAPI named pipe server.
+#[derive(Debug)]
+pub(crate) struct MessagePipeServer {
+    handle: Handle,
+    connected: bool,
+}
+
+/// SAFETY: It is safe to send pipe HANDLE between threads.
+unsafe impl Send for MessagePipeServer {}
```
CBenoit, line 27: Similar remarks as previously at this place: naming / typing / safety.

crates/ironrdp-dvc-pipe-proxy/src/windows/semaphore.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::windows::{Handle, WindowsError};
+
+/// RAII wrapper for WinAPI semaphore handle.
+#[derive(Debug, Clone)]
+pub(crate) struct Semaphore {
+    handle: Arc<Handle>,
+}
+
+/// SAFETY: It is safe to send semaphore HANDLE between threads.
+unsafe impl Send for Semaphore {}
```
CBenoit, line 15: https://github.com/Devolutions/sspi-rs/pull/430#pullrequestreview-2850078587

> - Explain why the code is "sound", detailing the invariants and preconditions that are maintained.
>
> - Avoid assumptions without justifications.
>	- Do not use phrases like "we assume this is safe because…" or "this should be safe".
>	- Provide concrete evidence or reasoning that justifies the safety of the operation.

crates/ironrdp-dvc-pipe-proxy/src/windows/wide_string.rs
CBenoit, line 1: suggestion: I think this is not a very good abstraction for wide strings. My recommendation would be to use widestring::U16CString as we do in Devolutions Gateway repository (we are moving away from `WideString`).

crates/ironrdp-dvc/src/client.rs
CBenoit, line 1: question: Do you know what happened to this file?

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -245,6 +249,10 @@ impl ActiveStage {
 
         None
     }
+
+    pub fn encode_dvc_messages(&mut self, message: Vec<SvcMessage>) -> SessionResult<Vec<u8>> {
+        self.process_svc_processor_messages(SvcProcessorMessages::<DrdynvcClient>::new(message))
```
CBenoit, line 254: naming: argument `messages` (plural)

note: This function seems to be for encoding **DVC** messages, but it takes `SvcMessage` values. Maybe the abstraction for SVC messages was reused as is for DVC messages, but this can be confusing.

crates/ironrdp-session/src/x224/mod.rs
```diff
@@ -186,6 +191,6 @@ impl Processor {
 /// The messages returned here are ready to be sent to the server.
 ///
 /// The caller is responsible for ensuring that the `channel_id` corresponds to the correct channel.
-fn process_svc_messages(messages: Vec<SvcMessage>, channel_id: u16, initiator_id: u16) -> SessionResult<Vec<u8>> {
+pub fn process_svc_messages(messages: Vec<SvcMessage>, channel_id: u16, initiator_id: u16) -> SessionResult<Vec<u8>> {
```
CBenoit, line 194: question: Is it necessary to increase the API surface? I think this function was meant to be an implementation detail.

CBenoit (commented):
Thank you! Very high quality PR, as always!

However, the unsafe code needs to be reviewed carefully.
The CI is also red.

crates/ironrdp-dvc-pipe-proxy/README.md
pacmancoder: This functionality is mainly required for RDM integration, but as a side effect, this could be used to implement custom DVC channels clients as long as they communicate over a named pipe.

crates/ironrdp-dvc/src/client.rs
pacmancoder: Ah. Looks like line endings were converted, I'll revert that

crates/ironrdp-client/src/rdp.rs
pacmancoder: I am re-using the already existing `RdpInputEvent` channel, which is `tokio::mpsc::Unbounded*`, I agree that we need to change that in follow-up

crates/ironrdp-client/src/rdp.rs
pacmancoder: I don't think this matters much; Historically we are already performing other DVC/SVC initialization after the socket connection, I just have added another dvc to the list

crates/ironrdp-dvc-pipe-proxy/Cargo.toml
```diff
@@ -25,6 +25,9 @@ tracing = { version = "0.1", features = ["log"] }
 
 
 [target.'cfg(windows)'.dependencies]
+
```
CBenoit, line 28: style: unintended new line?

crates/ironrdp-dvc-pipe-proxy/src/windows/semaphore.rs
```diff
@@ -44,16 +49,22 @@ impl Semaphore {
         };
 
         // SAFETY: Handle is valid and we are the owner of the handle.
-        let handle = unsafe { Handle::new_owned(handle)? };
+        let handle = unsafe { Owned::new(handle) };
 
         // CreateSemaphoreW returns a valid handle on success.
         Ok(Self {
+            // See `unsafe impl Send` comment.
+            #[allow(clippy::arc_with_non_send_sync)]
```
CBenoit, line 57: note: To myself, this is a bit suspicious. I need to verify.

pacmancoder:
@CBenoit are there any outstanding comments blocking this merge? 👀

CBenoit (commented):
I think you are implementing a safe wrapper around Windows named pipes, but I know about two libraries that are doing that:

- tokio (async): https://docs.rs/tokio/latest/tokio/net/windows/named_pipe/index.html
- named_pipe (blocking): https://blackbeam.github.io/named_pipe/named_pipe/index.html

Is there a reason for not using one of these instead? That would save us from a lot of unsafe code.

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+use smallvec::SmallVec;
+use windows::Win32::Foundation::{
+    ERROR_IO_PENDING, HANDLE, WAIT_ABANDONED_0, WAIT_EVENT, WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT,
+};
+use windows::Win32::System::Threading::{WaitForMultipleObjects, INFINITE};
+
+/// Thin wrapper around borrowed `windows` crate `HANDLE` reference.
+/// This is used to ensure handle lifetime when passing it to FFI functions
+/// (see `wait_any_with_timeout` for example).
+pub(crate) struct BorrowedHandle<'a>(&'a HANDLE);
```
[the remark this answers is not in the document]
pacmancoder: This is due to fact that `windows`'s crate `HANDLE` used in its APIs is a different type (`HANDLE(pub *mut core::ffi::c_void)`) vs `std`'s `type HANDLE = *mut c_void`, introducing new `BorrowedHandle` helps to avoid unnecessary casts

pacmancoder:
> I think you are implementing a safe wrapper around Windows named pipes, but I know about two libraries that are doing that:
> 
>     tokio (async): https://docs.rs/tokio/latest/tokio/net/windows/named_pipe/index.html
>     interprocess (blocking): https://docs.rs/interprocess/latest/x86_64-pc-windows-msvc/interprocess/os/windows/named_pipe/index.html
> 
> Did you consider them? Is there a reason for not using one of these instead? That would save us from a lot of unsafe code.
> 

I did not want to use `named_pipe` mainly because I don't want to force `tokio` dependency in a relatively lightweight crate, but I'll look closer into it again, maybe the advantages (namely its cross-platform-ness) are worth reconsidering using it here. As for `interprocess`, I haven't looked into it, I'll look it up. Let's merge as-is and then make a switch if needed.

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Safe wrapper around `WaitForMultipleObjects`.
+pub(crate) fn wait_any_with_timeout<'a, T>(handles: T, timeout: u32) -> Result<usize, WindowsError>
+where
+    T: IntoIterator<Item = BorrowedHandle<'a>>,
+{
+    let handles: SmallVec<[HANDLE; 8]> = handles.into_iter().map(|h| *h.0).collect();
+
+    // SAFETY: FFI call with no outstanding preconditions.
+    let result = unsafe { WaitForMultipleObjects(&handles, false, timeout) };
```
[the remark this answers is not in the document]
pacmancoder: done, fixed comment, added check for duplicates (in debug mode)

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Thin wrapper around borrowed `windows` crate `HANDLE` reference.
+/// This is used to ensure handle lifetime when passing it to FFI functions
+/// (see `wait_any_with_timeout` for example).
+pub(crate) struct BorrowedHandle<'a>(&'a HANDLE);
+
+/// Safe wrapper around `WaitForMultipleObjects`.
+pub(crate) fn wait_any_with_timeout<'a, T>(handles: T, timeout: u32) -> Result<usize, WindowsError>
+where
+    T: IntoIterator<Item = BorrowedHandle<'a>>,
```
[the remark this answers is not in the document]
pacmancoder: done, thanks!

crates/ironrdp-dvc-pipe-proxy/src/windows/pipe.rs
```diff
@@ -0,0 +1,288 @@
...
+        Ok(Self {
+            pipe,
+            buffer: vec![0; buffer_size],
+            overlapped,
+            event,
+        })
+    }
+
+    pub(crate) fn overlapped_read(&mut self) -> Result<(), WindowsError> {
+        // SAFETY: hfile is a valid handle to a named pipe; lpBuffer
```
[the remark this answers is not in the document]
pacmancoder: fixed

CBenoit:
> @CBenoit approve please 😅

Sure thing  – reviewing now.

crates/ironrdp-dvc-pipe-proxy/README.md
CBenoit, line 1: praise: Good README.md, thank you!

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        WAIT_EVENT(id) => Ok((id - WAIT_OBJECT_0.0) as usize),
+    }
+}
+
+/// Safe `WaitForMultipleObjects` wrapper with infinite timeout.
+pub(crate) fn wait_any(handles: &[BorrowedHandle<'_>]) -> Result<usize, WindowsError> {
+    // Standard generic syntax is used instead if `impl` because of the following lint:
+    // > warning: lifetime parameter `'a` only used once
+    //
+    // Fixing this lint (use of '_ lifetime) produces compiler error.
```
CBenoit, line 55: issue? Maybe this comment is outdated

crates/ironrdp-dvc-pipe-proxy/src/windows/mod.rs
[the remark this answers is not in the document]
CBenoit: Indeed, both are #[repr(transparent)] wrappers around `*mut c_void` so I think I would be fine with a internal cast from `std::windows::os::io::BorrowedHandle` to `windows::HANDLE`, we can write a function like the `cast_handles` function already existing, encapsulating the unsafe code, but I’m not sure it’s even necessary to have unsafe code since we can create a `windows::HANDLE` from `*mut c_void`, and we can access the `*mut c_void` from the `std::windows::os::io::BorrowedHandle` using [`as_raw_handle`](https://doc.rust-lang.org/stable/std/os/windows/io/struct.BorrowedHandle.html#method.as_raw_handle). But in fact it seems we are just using `[BorrowedHandle<'a>]`, although I may have missed a few callsites?
Now I don’t think it’s a big issue, but it may be a bit confusing. I’ll leave you decide what to do or not in a follow up!

crates/ironrdp-dvc-pipe-proxy/src/windows/semaphore.rs
```diff
@@ -0,0 +1,91 @@
...
+        let handle = unsafe {
+            CreateSemaphoreW(None, initial_count, maximum_count, None).map_err(WindowsError::CreateSemaphore)?
+        };
+
+        // SAFETY: Handle is valid and we are the owner of the handle.
+        let handle = unsafe { Owned::new(handle) };
+
+        // CreateSemaphoreW returns a valid handle on success.
+        Ok(Self {
+            // See `unsafe impl Send` comment.
```
CBenoit, line 56:
```suggestion
            // See `unsafe impl Send` comment.
            // TODO(@CBenoit): Verify this comment.
```

PR #796, opened 2025-05-28
feat(client): support HyperV VM Connect

irvingoujAtDevolution:
The main changes primarily involves changing in sequence of the connector state. With vmconnect, our connection goes from 
Preconnection Blob -> TLS -> CredSSP -> Cookie/InitRequest

where as oppose to original sequence, the cookie is sent before TLS

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -292,24 +292,29 @@ where
     {
         // RDCleanPath request
 
-        let connector::ClientConnectorState::ConnectionInitiationSendRequest { .. } = connector.state else {
-            return Err(connector::general_err!("invalid connector state (send request)"));
-        };
+        match connector.state {
+            connector::ClientConnectorState::ConnectionInitiationSendRequest { .. } => {
+                let written = connector.step_no_input(&mut buf)?;
+                let x224_pdu_len = written.size().expect("written size");
+                debug_assert_eq!(x224_pdu_len, buf.filled_len());
+                let x224_pdu = buf.filled().to_vec();
+
+                let rdcleanpath_req =
+                    ironrdp_rdcleanpath::RDCleanPathPdu::new_x224_request(x224_pdu, destination, proxy_auth_token)
+                        .map_err(|e| connector::custom_err!("new RDCleanPath request", e))?;
+                debug!(message = ?rdcleanpath_req, "Send RDCleanPath request");
+                let rdcleanpath_req = rdcleanpath_req
+                    .to_der()
+                    .map_err(|e| connector::custom_err!("RDCleanPath request encode", e))?;
+                rdcleanpath_req
+            }
+            connector::ClientConnectorState::PreconnectionBlob { .. } => {}
+            _ => {
+                return Err(connector::general_err!("invalid connector state (send request)"));
+            }
+        }
 
         debug_assert!(connector.next_pdu_hint().is_none());
-
-        let written = connector.step_no_input(&mut buf)?;
-        let x224_pdu_len = written.size().expect("written size");
-        debug_assert_eq!(x224_pdu_len, buf.filled_len());
-        let x224_pdu = buf.filled().to_vec();
-
-        let rdcleanpath_req =
-            ironrdp_rdcleanpath::RDCleanPathPdu::new_request(x224_pdu, destination, proxy_auth_token, pcb)
-                .map_err(|e| connector::custom_err!("new RDCleanPath request", e))?;
-        debug!(message = ?rdcleanpath_req, "Send RDCleanPath request");
-        let rdcleanpath_req = rdcleanpath_req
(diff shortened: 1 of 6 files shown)


crates/ironrdp-connector/src/credssp.rs
```diff
@@ -70,6 +70,7 @@ pub struct CredsspSequence {
     client: CredSspClient,
     state: CredsspState,
     selected_protocol: nego::SecurityProtocol,
+    vmconnect: bool,
```
CBenoit: question: What is the purpose of this field? I don’t see where it is used

crates/ironrdp-core/src/cursor.rs
```diff
@@ -578,6 +578,12 @@ impl<'a> WriteCursor<'a> {
         self.pos
     }
 
+    /// Returns the number of bytes written.
+    #[inline]
+    pub const fn bytes_written(&self) -> usize {
+        self.pos
+    }
```
CBenoit, line 585: issue: Remove this method. You can use `pos()` instead which is the same thing. Also, strictly speaking, it’s not necessarily the number of bytes written because it’s possible to "rewind" the cursor.

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -313,6 +313,9 @@ impl Encode for ExtendedClientOptionalInfo {
             dst.write_array(reconnect_cookie);
         }
 
+        dst.write_u16(0); // reserved1
+        dst.write_u16(0); // reserved2
```
CBenoit, line 317: question: Did you fix a bug? Could you elaborate on that? Please, open a separate PR dedicated to this with a `fix` commit type.

crates/ironrdp-testsuite-core/tests/pcb.rs
```diff
@@ -84,6 +84,15 @@ encode_decode_test! {
...
+    v2_hyperv_guid :
+        PreconnectionBlob {
+            version: PcbVersion::V2,
+            id: 0,
+            v2_payload: Some(String::from("ff995b48-a34f-404a-938f-303e4bb5bf31")),
+        },
+         hex::decode(concat!(
+            "5c0000000000000002000000000000002500660066003900390035006200340038002d0061003300340066002d0034003000340061002d0039003300380066002d003300300033006500340062006200350062006600330031000000"
+        )).expect("v2_hyperv_guid payload");
 }
```
CBenoit, line 96: issue: I think this test case does nothing more than the one just above, so let’s not add it

crates/ironrdp-web/src/session.rs
```diff
@@ -861,6 +861,8 @@ fn build_config(
         desktop_scale_factor: 0,
         hardware_id: None,
         license_cache: None,
+        // TODO: implement this
+        vmconnect: None,
```
CBenoit, line 864: issue: We already support a `pcb` extension. My understanding is that all you need to do is te place a UUID into it. No need to add extra parameters.

crates/ironrdp-connector/src/lib.rs
```diff
@@ -187,6 +187,8 @@ pub struct Config {
     pub no_server_pointer: bool,
     pub pointer_software_rendering: bool,
     pub performance_flags: PerformanceFlags,
+
+    pub vmconnect: Option<String>,
```
CBenoit, line 191: issue: I don’t think we should expose a `vmconnect` option. Also, since the PCB is something outside of the RDP protocol itself, let’s not handle it in the connector at all.

crates/ironrdp-client/src/rdp.rs
```diff
@@ -148,7 +148,7 @@ async fn connect(
 
     let should_upgrade = ironrdp_tokio::connect_begin(&mut framed, &mut connector).await?;
 
-    debug!("TLS upgrade");
+    debug!(destination = ?config.destination,"TLS upgrade");
```
CBenoit, line 151: nitpick: Add a `debug!` log before `TcpStream::connect` instead

crates/ironrdp-client/src/config.rs
```diff
@@ -238,6 +238,10 @@ struct Args {
     /// The bitmap codecs to use (remotefx:on, ...)
     #[clap(long, value_parser, num_args = 1.., value_delimiter = ',')]
     codecs: Vec<String>,
+
+    /// The Virtual Machine ID for Hyper-V, used as Pre Connection Blob
+    #[clap(long)]
+    vmconnect: Option<String>,
```
CBenoit, line 244: suggestion: I think you could use a `Uuid` type here

irvingoujAtDevolution:
You might be right, let me draft this PR, I need to make some experiment on another branch

irvingoujAtDevolution:
Closed in favor of https://github.com/Devolutions/IronRDP/pull/841

PR #804, opened 2025-06-04
feat(client): support for hardware cursor

### LINKED ISSUE #251: Support for hardware cursor on Windows
Support Windows in `ironrdp-native-cursor-icon` crate.

We’ll have to use system API such as:
- https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createcursor
- https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createiconindirect
- https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcursor

### LINKED ISSUE #252: Support for hardware cursor on Linux
Support Linux in `ironrdp-native-cursor-icon` crate.

`x11` crate is probably a good start point: https://docs.rs/x11/2.21.0/x11/xcursor/index.html

### LINKED ISSUE #253: Support for hardware cursor on Wayland
Support Wayland in `ironrdp-native-cursor-icon` crate.

`wayland_client` is a good starting point: https://docs.rs/wayland-client/latest/wayland_client/protocol/wl_pointer/struct.WlPointer.html

### LINKED ISSUE #254: Support for hardware cursor on macOS
Support macOS in `ironrdp-native-cursor-icon` crate.

### LINKED ISSUE #255: [Meta] Client-side hardware cursor
Instead of blending the cursor into the frame buffer, it’s much more efficient to set the hardware cursor.

I’m not aware of any crate which handle setting the icon for the hardware cursor directly.
`winit` only allow setting the "cursor info" (Help, VerticalText, etc).
For the native client, we should create a new crate `ironrdp-native-cursor-icon` similar to `ironrdp-native-clipboard`, providing a cross-platform API for setting the cursor. 

Issue per platform:

- Browser: https://github.com/Devolutions/IronRDP/issues/250
- Windows: https://github.com/Devolutions/IronRDP/issues/251
- X11: https://github.com/Devolutions/IronRDP/issues/252
- Wayland: https://github.com/Devolutions/IronRDP/issues/253
- macOS: https://github.com/Devolutions/IronRDP/issues/254

Note: the code for blending the cursor into the frame buffer should not be removed. It can be useful when implementing proxy-based recording and is also a good fallback for platforms where hardware cursor is not supported.

jpy794:
Render cursor icon at client side with `winit` custom cursor API. This should close #255.
Tested on Wayland.

Closes #251 
Closes #252 
Closes #253 
Closes #254 
Closes #255

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -4,6 +4,7 @@ use ironrdp::connector::{ConnectionResult, ConnectorResult};
 use ironrdp::displaycontrol::client::DisplayControlClient;
 use ironrdp::displaycontrol::pdu::MonitorLayoutEntry;
 use ironrdp::graphics::image_processing::PixelFormat;
+use ironrdp::graphics::pointer::DecodedPointer;
 use ironrdp::pdu::input::fast_path::FastPathInputEvent;
 use ironrdp::session::image::DecodedImage;
 use ironrdp::session::{fast_path, ActiveStage, ActiveStageOutput, GracefulDisconnectReason, SessionResult};
@@ -28,6 +29,7 @@ pub enum RdpOutputEvent {
     PointerDefault,
     PointerHidden,
     PointerPosition { x: u16, y: u16 },
+    PointerBitmap(DecodedPointer),
     Terminated(SessionResult<GracefulDisconnectReason>),
 }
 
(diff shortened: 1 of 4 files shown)


crates/ironrdp-client/src/rdp.rs
```diff
@@ -28,6 +29,7 @@ pub enum RdpOutputEvent {
     PointerDefault,
     PointerHidden,
     PointerPosition { x: u16, y: u16 },
+    PointerBitmap(DecodedPointer),
```
CBenoit: suggestion: I think you could avoid the deep copy by taking `Arc<DecodedPointer>` at this place, this is the type returned by the `ActiveStage` state machine.

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -38,7 +38,7 @@ pub enum PointerError {
 }
 
 /// Represents RDP pointer in decoded form (color channels stored as RGBA pre-multiplied values)
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
CBenoit: suggestion: In this case, deriving Clone is not necessary anymore. If you really want to derive this, please open a separate PR so it’s easier to track in the changelogs.

CBenoit (commented):
Thank you!!

crates/ironrdp-client/src/rdp.rs
jpy794: Good point, thanks!

PR #800, opened 2025-06-02
chore(release): prepare for publishing

devolutionsbot:
## 🤖 New release
* `iron-remote-desktop`: 0.2.0 -> 0.3.0 (⚠️ API breaking changes)

### ⚠️ `iron-remote-desktop` breaking changes

```
--- failure trait_method_added: pub trait method added ---

Description:
A non-sealed public trait added a new method without a default implementation, which breaks downstream implementations of the trait
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.38.0/src/lints/trait_method_added.ron

Failed in:
  trait method iron_remote_desktop::Session::invoke_extension in file /tmp/.tmpUoid7D/IronRDP/crates/iron-remote-desktop/src/session.rs:101

--- failure trait_method_missing: pub trait method removed or renamed ---

Description:
A trait method is no longer callable, and may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#major-any-change-to-trait-item-signatures
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.38.0/src/lints/trait_method_missing.ron

Failed in:
  method extension_call of trait Session, previously in file /tmp/.tmppwAaxU/iron-remote-desktop/src/session.rs:101
```

<details><summary><i><b>Changelog</b></i></summary><p>

## `iron-remote-desktop`
<blockquote>

### Features

- Introduce callExtension method in iron-remote-desktop API (#799) ([9f6647c341](https://github.com/Devolutions/IronRDP/commit/9f6647c34185270cd451b4e1eeacee3554a5618b)) 

### Bug Fixes

- [**breaking**] Rename callExtension to invokeExtension (#803) ([f68cd06ac3](https://github.com/Devolutions/IronRDP/commit/f68cd06ac3705608e6f2ac6bde684d9ae906ea53))
</blockquote>

</p></details>

---
This PR was generated with [release-plz](https://github.com/release-plz/release-plz/).

PR #807, opened 2025-06-05
chore: update `iron-remote-desktop` version to 0.3.1

RRRadicalEdward:
It seems that in https://github.com/Devolutions/IronRDP/pull/800, we released a version that included https://github.com/Devolutions/IronRDP/pull/799, but not https://github.com/Devolutions/IronRDP/pull/803, and it's the reason why we have https://github.com/Devolutions/devolutions-gateway/pull/1369#issuecomment-2943811482 issue.

### DIFF
--- a/crates/iron-remote-desktop/Cargo.toml
+++ b/crates/iron-remote-desktop/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iron-remote-desktop"
-version = "0.3.0"
+version = "0.3.1"
 readme = "README.md"
 description = "Helper crate for building WASM modules compatible with iron-remote-desktop WebComponent"
 edition.workspace = true
--- a/web-client/iron-remote-desktop/public/package.json
+++ b/web-client/iron-remote-desktop/public/package.json
@@ -10,7 +10,7 @@
     "Alexandr Yusuk"
   ],
   "description": "Web Component providing agnostic implementation for Iron Wasm base client",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "main": "iron-remote-desktop.js",
   "types": "index.d.ts",
   "files": [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2319,7 +2319,7 @@ dependencies = [
 
 [[package]]
 name = "iron-remote-desktop"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "console_error_panic_hook",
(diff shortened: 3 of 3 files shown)


crates/iron-remote-desktop/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iron-remote-desktop"
-version = "0.3.0"
+version = "0.3.1"
```
CBenoit, line 3: issue: I think the crate is already up to date, do not bump it

web-client/iron-remote-desktop/public/package.json
```diff
@@ -10,7 +10,7 @@
     "Alexandr Yusuk"
   ],
   "description": "Web Component providing agnostic implementation for Iron Wasm base client",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
CBenoit, line 13: issue: renaming a method is a breaking change, so this should be 0.4.0 I think

crates/iron-remote-desktop/Cargo.toml
RRRadicalEdward: I'm not sure that in https://github.com/Devolutions/IronRDP/pull/800, we actually also included https://github.com/Devolutions/IronRDP/pull/803. It would explain https://github.com/Devolutions/devolutions-gateway/pull/1369#issuecomment-2943811482 error

CBenoit: I verified, it is included (release-plz automatically catch anything related to the Rust code)

See the doc of the latest published version: https://docs.rs/iron-remote-desktop/latest/iron_remote_desktop/trait.Session.html

CBenoit:
@RRRadicalEdward I think we need to update iron-remote-desktop**-rdp**, because the latest version we released is still exposing a method `extensionCall` instead of `invokeExtension`

RRRadicalEdward:
> @RRRadicalEdward I think we need to update iron-remote-desktop-**rdp**, because the latest version we released is still exposing a method `extensionCall` instead of `invokeExtension`, and this is not compatible with iron-remote-desktop v0.3.0

But iron-remote-desktop-rdp doesn't export `extensionCall`

CBenoit:
That being said, it’s true that the error is more about the NPM package `iron-remote-desktop` itself… Though I don’t see any change that we could have missed between the release run and now?

PR #808, opened 2025-06-05
 fix!(web-client): rename callExtension to invokeExtension for UserInteraction

RRRadicalEdward:


CBenoit (approved):
Oh, that was the reason :joy:

PR #818, opened 2025-06-10
refactor!(web-client): remove RxJS depedency

RRRadicalEdward:
This PR removes RxJS dependency as discussed here - https://github.com/Devolutions/IronRDP/pull/787#issuecomment-2862401670

### DIFF
--- a/web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
+++ b/web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
@@ -3,8 +3,8 @@ import type { NewSessionInfo } from './NewSessionInfo';
 import type { SessionEvent } from './session-event';
 import { ConfigBuilder } from '../services/ConfigBuilder';
 import type { Config } from '../services/Config';
-import type { PartialObserver } from 'rxjs';
 import type { Extension } from './Extension';
+import type { Callback } from '../lib/observable';
 
 export interface UserInteraction {
     setVisibility(state: boolean): void;
@@ -25,7 +25,7 @@ export interface UserInteraction {
 
     setCursorStyleOverride(style: string | null): void;
 
-    onSessionEvent(partialObserver: PartialObserver<SessionEvent>): void;
+    onSessionEvent(partialObserver: Callback<SessionEvent>): void;
 
     resize(width: number, height: number, scale?: number): void;
 
--- a/web-client/iron-remote-desktop/src/lib/observable.ts
+++ b/web-client/iron-remote-desktop/src/lib/observable.ts
@@ -1,0 +1,19 @@
+export type Callback<T> = (_: T) => void;
+
+export class Observable<T> {
+    constructor() {
+        this.subscribers = [];
+    }
+
+    subscribers: Array<Callback<T>>;
+
+    subscribe(cb: Callback<T>) {
+        this.subscribers.push(cb);
+    }
+
+    publish(value: T) {
+        for (const cb of this.subscribers) {
+            cb(value);
+        }
+    }
+}
--- a/web-client/iron-remote-desktop/src/services/PublicAPI.ts
+++ b/web-client/iron-remote-desktop/src/services/PublicAPI.ts
@@ -21,9 +21,7 @@ export class PublicAPI {
 
     private connect(config: Config): Promise<NewSessionInfo> {
         loggingService.info('Initializing connection.');
-        const resultObservable = this.remoteDesktopService.connect(config);
-
-        return resultObservable.toPromise();
+        return this.remoteDesktopService.connect(config);
     }
 
     private ctrlAltDel() {
@@ -74,7 +72,7 @@ export class PublicAPI {
             connect: this.connect.bind(this),
             setScale: this.setScale.bind(this),
             onSessionEvent: (partialObserver) => {
-                this.remoteDesktopService.sessionObserver.subscribe(partialObserver);
+                this.remoteDesktopService.sessionEventObservable.subscribe(partialObserver);
             },
             ctrlAltDel: this.ctrlAltDel.bind(this),
             metaKey: this.metaKey.bind(this),
--- a/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
+++ b/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
@@ -1,6 +1,4 @@
-import { BehaviorSubject, from, Observable, of, Subject } from 'rxjs';
 import { loggingService } from './logging.service';
-import { catchError, filter, map } from 'rxjs/operators';
 import { scanCode } from '../lib/scancodes';
 import { OS } from '../enums/OS';
 import { ModifierKey } from '../enums/ModifierKey';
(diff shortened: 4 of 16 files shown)


web-client/iron-remote-desktop/src/lib/observable.ts
CBenoit, line 1: suggestion: Based on the rest of the codebase, maybe this file should be `Observable.ts`?

web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
```diff
@@ -25,7 +25,7 @@ export interface UserInteraction {
 
     setCursorStyleOverride(style: string | null): void;
 
-    onSessionEvent(partialObserver: PartialObserver<SessionEvent>): void;
+    onSessionEvent(partialObserver: Callback<SessionEvent>): void;
```
CBenoit: naming: not an observer anymore

web-client/iron-remote-desktop/src/services/PublicAPI.ts
```diff
@@ -74,7 +72,7 @@ export class PublicAPI {
             connect: this.connect.bind(this),
             setScale: this.setScale.bind(this),
             onSessionEvent: (partialObserver) => {
-                this.remoteDesktopService.sessionObserver.subscribe(partialObserver);
+                this.remoteDesktopService.sessionEventObservable.subscribe(partialObserver);
```
CBenoit: naming: Similar, I guess this is a callback?

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
```diff
@@ -113,14 +107,14 @@ export class RemoteDesktopService {
...
-        this.mousePosition.next(position);
+        this.mousePositionObservable.publish(position);
     }
 
     configBuilder(): ConfigBuilder {
         return new ConfigBuilder();
     }
 
-    connect(config: Config): Observable<NewSessionInfo> {
+    async connect(config: Config): Promise<NewSessionInfo> {
```
CBenoit, line 117: question: Is it necessary to return a Promise when the function is `async`? My understanding is that it’s a bit like if you had a `connect` function returning a Promise of a Promise of a NewSessionInfo.

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
```diff
@@ -153,68 +147,75 @@ export class RemoteDesktopService {
...
+                data: 'Session started',
+            });
+
+            return {
+                sessionId: 0,
+                initialDesktopSize: session.desktopSize(),
+                websocketPort: 0,
+            };
+        } catch (err: unknown) {
+            if (isIronError(err)) {
```
CBenoit: question: Is it possible to have something else than `IronError`?

web-client/iron-svelte-client/src/lib/login/login.svelte
```diff
@@ -22,33 +21,37 @@
 
     let userInteraction: UserInteraction;
 
+    const isIronError = (error: unknown): error is IronError =>
+        typeof error === 'object' &&
+        error !== null &&
+        typeof (error as IronError).backtrace === 'function' &&
+        typeof (error as IronError).kind === 'function';
```
CBenoit: issue: Do we really want to have to check all of this in the consumer code? :thinking:

Aren’t we already catching the errors as `SessionEvent`s? Could we be handling errors twice?

Maybe we could throw a `SessionEvent` instead of an `IronError`, so it’s possible to do

```ts
catch (err: unknown) {
  if (err instanceof SessionEvent) {
    // Handle the SessionEvent
  } else {
    // Generic error handling
  }
}
```

Or create a `class SessionError extends Error` that we throw instead of the `IronError`? Something that would be convenient to handle :slightly_smiling_face:

web-client/iron-svelte-client/src/lib/popup-screen/popup-screen.svelte
```diff
@@ -12,15 +12,16 @@
     userInteractionService.subscribe((val) => {
         if (val != null) {
             userInteraction = val;
```
CBenoit: note: It occurred to me that this code could be simplified

```suggestion
    userInteractionService.subscribe((userInteraction) => {
        if (userInteraction != null) {
```

web-client/iron-svelte-client/src/lib/login/login.svelte
```diff
@@ -140,38 +143,44 @@
...
+        } catch (err) {
+            if (isIronError(err)) {
+                toast.set({
+                    type: 'info',
+                    message: err.backtrace(),
+                });
+            } else {
+                toast.set({
+                    type: 'error',
+                    message: 'Failure',
```
CBenoit: suggestion: Maybe try to stringify the error at this place?

web-client/iron-svelte-client/src/lib/login/login.svelte
```diff
@@ -140,38 +143,44 @@
...
-                    toast.set({
-                        type: 'error',
-                        message: 'Failure',
-                    });
-                }
-            });
+        try {
+            const session_info = await userInteraction.connect(config);
+
+            if (session_info.initialDesktopSize !== null) {
```
CBenoit: question: Is it possible for this to be null at all?

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
RRRadicalEdward: TS requires specifying a true type. Otherwise, we will have a warning.

CBenoit: `NewSessionInfo` is not a true type?

RRRadicalEdward: No, because the real type that this function returns is `Promise<NewSessionInfo>`, we need to explicitly specify it
image

CBenoit: Oh… I didn’t know that :joy:
I thought it was like Rust. Good to know!

web-client/iron-svelte-client/src/lib/login/login.svelte
RRRadicalEdward: Yes, I've just gone through the code a little bit, and it turns out that we actually handle errors twice.
For example, an error handling in iron-remote-desktop, we do `raiseSessionEvent` [in a catch block](https://github.com/Devolutions/IronRDP/blob/remove-rxjs/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts#L185) and then [we also propagate the error](https://github.com/Devolutions/IronRDP/blob/remove-rxjs/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts#L188). 
 
Then, in the consumer code, we handle both the `SessionEvent` and the error that we raised.
 
So, my solution would be to remove the propagation of the `IronError` error and instead just propagate `new Error("...")` if we have already sent it as an event to the consumer. In this case, we won't need the `IronError` type guard in the consumer anymore, since there will be no such errors in the consumer's catch block - we can just write to the console.
What do you think about it?

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
RRRadicalEdward: I believe that the error can only be of type `IronError`, but we need this type guard so that TS can safely assume that the `err` is of type `IronError`.

web-client/iron-svelte-client/src/lib/login/login.svelte
RRRadicalEdward: No. I removed the check

web-client/iron-svelte-client/src/lib/popup-screen/popup-screen.svelte
RRRadicalEdward: Done

web-client/iron-remote-desktop/src/services/PublicAPI.ts
RRRadicalEdward: Fixed

web-client/iron-remote-desktop/src/interfaces/UserInteraction.ts
RRRadicalEdward: Fixed

web-client/iron-remote-desktop/src/lib/observable.ts
RRRadicalEdward: Done

web-client/iron-svelte-client/src/lib/login/login.svelte
RRRadicalEdward: No longer actual since I refactored it and related code

web-client/iron-svelte-client/src/lib/login/login.svelte
CBenoit: Sounds good to me! :+1:

RRRadicalEdward: @CBenoit It's done in https://github.com/Devolutions/IronRDP/pull/818/commits/42826f22615b0ea39817c5dd56b622bf7cd34f3e

CBenoit: My bad!

PR #820, opened 2025-06-16
fix(web): fix session dangling after calling `run` method

RRRadicalEdward:


CBenoit (approved):
Thank you! What is the problem exactly though? I’m not sure to understand with my JavaScript knowledge...

PR #823, opened 2025-06-19
fix(web): remove window resize listener on component destroy

RRRadicalEdward:
QA hasn't created an issue about it yet, but it was causing "Uncaught TypeError displayed in dev console after enabling Full Screen mode".


CBenoit (approved):
LGTM! Thanks!

By the way, I’m using the scope `web` for all the web client related stuff: `ironrdp-web`, `iron-remote-desktop`, `iron-remote-desktop-rdp` :slightly_smiling_face:

PR #824, opened 2025-06-19
feat(web): add dynamic_resizing_supported callback

RRRadicalEdward:


web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -517,19 +517,18 @@
...
-        const containerWidth = windowSize.x - wrapperBoundingBox.x;
-        const containerHeight = windowSize.y - wrapperBoundingBox.y;
+        const containerWidth = windowSize.x;
+        const containerHeight = windowSize.y;
 
         let width = canvas.width;
         let height = canvas.height;
 
-        const ratio = Math.max(containerWidth / canvas.width, containerHeight / canvas.height);
+        const ratio = Math.min(containerWidth / canvas.width, containerHeight / canvas.height);
```
RRRadicalEdward: Also, I fixed the scale calculation in full-screen mode

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -517,19 +517,18 @@
...
         width = width * ratio;
         height = height * ratio;
 
-        setWrapperStyle(`${containerHeight}px`, `${containerWidth}px`, 'auto');
+        setWrapperStyle(`${containerHeight}px`, `${containerWidth}px`, 'hidden');
```
RRRadicalEdward: This option controls the sliders. If it's set to auto, for some reason slider is shown. So I decided to hide it.

CBenoit: Thank you! As a rule of thumb, it’s best to open a dedicated PR even if the change is small (so we can properly track the patches)

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
CBenoit: Thank you! As a rule of thumb, it’s best to open a dedicated PR even if the change is small (so we can properly track the patches)

web-client/iron-remote-desktop/public/package.json
```diff
@@ -10,7 +10,7 @@
     "Alexandr Yusuk"
   ],
   "description": "Backend-agnostic Web Component for remote desktop protocols",
-  "version": "0.5.1",
+  "version": "0.5.2",
```
CBenoit: note: I prefer having a separate PR `chore(release): prepare iron-remote-desktop vX.Y.Z` so it’s easier to discover the bumps

CBenoit (commented):
It seems to me that we do not need to add this to the common API surface. `ironrdp-web` seems to be essentially reporting the value of the extension `use_display_control`, which is already controlled by the consumer of `iron-remote-desktop-rdp`/`ironrdp-web`.
This sounds like this should be an extension in `iron-remote-desktop-vnc`/`ironvnc-web`. What do you think?

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: My bad. I will open a seperate PR with the fix 👍

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: My bad. I will open a seperate PR with the fix 👍

RRRadicalEdward: https://github.com/Devolutions/IronRDP/pull/827

RRRadicalEdward:
> It seems to me that we do not need to add this to the common API surface. `ironrdp-web` seems to be essentially reporting the value of the extension `use_display_control`, which is already controlled by the consumer of `iron-remote-desktop-rdp`/`ironrdp-web`. This sounds like this callback should be provided as an extension in `iron-remote-desktop-vnc`/`ironvnc-web`. What do you think?

Make sense. Then, there is no need for this PR at all, as I moved out every small fix into separate PRs anyway, and this PR doesn't add anything.

web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
RRRadicalEdward: https://github.com/Devolutions/IronRDP/pull/828

PR #833, opened 2025-06-23
fix(web): keyboard input not working after launching a session via Firefox

RRRadicalEdward:
Closes https://github.com/Devolutions/IronVNC/issues/863.


CBenoit (approved):
Thank you! Small change but very consequent fix :joy:

PR #839, opened 2025-06-24
feat!: update sspi dependency

TheBestTvarynka:
Newer version of sspi adds support for server-side Kerberos.
This is relevant for the ironrdp-acceptor crate.

### DIFF
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -159,9 +159,12 @@ impl<'a> CredsspSequence<'a> {
         output: &mut WriteBuf,
     ) -> ConnectorResult<Written> {
         let (ts_request, next_state) = match result {
-            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(Box::new(ts_request)), CredsspState::Ongoing),
+            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(ts_request), CredsspState::Ongoing),
             Ok(ServerState::Finished(_id)) => (None, CredsspState::Finished),
-            Err(err) => (err.ts_request, CredsspState::ServerError(err.error)),
+            Err(err) => (
+                err.ts_request.map(|ts_request| *ts_request),
+                CredsspState::ServerError(err.error),
+            ),
         };
 
         self.state = next_state;
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -34,7 +34,7 @@ impl PduHint for CredsspTsRequestHint {
 }
 
 pub type CredsspProcessGenerator<'a> =
-    Generator<'a, NetworkRequest, sspi::Result<Vec<u8>>, Result<ServerState, Box<ServerError>>>;
+    Generator<'a, NetworkRequest, sspi::Result<Vec<u8>>, Result<ServerState, ServerError>>;
 
 #[derive(Debug)]
 pub struct CredsspSequence<'a> {
@@ -71,7 +71,7 @@ impl CredentialsProxy for CredentialsProxyImpl<'_> {
 pub(crate) async fn resolve_generator(
     generator: &mut CredsspProcessGenerator<'_>,
     network_client: &mut dyn AsyncNetworkClient,
-) -> Result<ServerState, Box<ServerError>> {
+) -> Result<ServerState, ServerError> {
     let mut state = generator.start();
 
     loop {
@@ -81,11 +81,9 @@ pub(crate) async fn resolve_generator(
                     .send(&request)
                     .await
                     .inspect_err(|err| error!(?err, "Failed to send a Kerberos message"))
-                    .map_err(|err| {
-                        Box::new(ServerError {
-                            ts_request: None,
-                            error: sspi::Error::new(sspi::ErrorKind::InternalError, err),
-                        })
+                    .map_err(|err| ServerError {
+                        ts_request: None,
+                        error: sspi::Error::new(sspi::ErrorKind::InternalError, err),
                     })?;
                 state = generator.resume(Ok(response));
             }
@@ -157,11 +155,11 @@ impl<'a> CredsspSequence<'a> {
 
     pub fn handle_process_result(
         &mut self,
-        result: Result<ServerState, Box<ServerError>>,
+        result: Result<ServerState, ServerError>,
         output: &mut WriteBuf,
     ) -> ConnectorResult<Written> {
         let (ts_request, next_state) = match result {
-            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(ts_request), CredsspState::Ongoing),
+            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(Box::new(ts_request)), CredsspState::Ongoing),
             Ok(ServerState::Finished(_id)) => (None, CredsspState::Finished),
             Err(err) => (err.ts_request, CredsspState::ServerError(err.error)),
         };
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -77,13 +77,16 @@ pub(crate) async fn resolve_generator(
     loop {
         match state {
             GeneratorState::Suspended(request) => {
-                let response = network_client.send(&request).await.map_err(|err| {
-                    error!(?err, "Failed to send a Kerberos message");
-                    Box::new(ServerError {
-                        ts_request: None,
-                        error: sspi::Error::new(sspi::ErrorKind::InternalError, err),
-                    })
-                })?;
+                let response = network_client
+                    .send(&request)
+                    .await
+                    .inspect_err(|err| error!(?err, "Failed to send a Kerberos message"))
+                    .map_err(|err| {
+                        Box::new(ServerError {
+                            ts_request: None,
+                            error: sspi::Error::new(sspi::ErrorKind::InternalError, err),
+                        })
+                    })?;
                 state = generator.resume(Ok(response));
             }
             GeneratorState::Completed(client_state) => break client_state,
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -79,7 +79,10 @@ pub(crate) async fn resolve_generator(
             GeneratorState::Suspended(request) => {
                 let response = network_client.send(&request).await.map_err(|err| {
                     error!(?err, "Failed to send a Kerberos message");
-                    Box::new(ServerError { ts_request: Default::default(), error: sspi::Error::new(sspi::ErrorKind::InternalError, err) })
+                    Box::new(ServerError {
+                        ts_request: None,
+                        error: sspi::Error::new(sspi::ErrorKind::InternalError, err),
+                    })
                 })?;
                 state = generator.resume(Ok(response));
             }
@@ -157,7 +160,7 @@ impl<'a> CredsspSequence<'a> {
         let (ts_request, next_state) = match result {
             Ok(ServerState::ReplyNeeded(ts_request)) => (Some(ts_request), CredsspState::Ongoing),
             Ok(ServerState::Finished(_id)) => (None, CredsspState::Finished),
-            Err(err) => (Some(err.ts_request), CredsspState::ServerError(err.error)),
+            Err(err) => (err.ts_request, CredsspState::ServerError(err.error)),
         };
 
         self.state = next_state;
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -77,14 +77,10 @@ pub(crate) async fn resolve_generator(
     loop {
         match state {
             GeneratorState::Suspended(request) => {
-                let response = network_client
-                    .send(&request)
-                    .await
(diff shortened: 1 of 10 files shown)


crates/ironrdp-connector/Cargo.toml
```diff
@@ -24,13 +24,13 @@ ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
 ironrdp-error = { path = "../ironrdp-error", version = "0.1" } # public
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.5", features = ["std"] } # public
 arbitrary = { version = "1", features = ["derive"], optional = true } # public
-sspi = "0.15" # public
+sspi = { path = "../../../sspi-rs" } # public
```
TheBestTvarynka: https://github.com/Devolutions/sspi-rs/pull/457

crates/ironrdp-connector/Cargo.toml
```diff
@@ -24,13 +24,13 @@ ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
...
 url = "2.5" # public
 rand_core = { version = "0.6", features = ["std"] } # TODO: dependency injection?
 tracing = { version = "0.1", features = ["log"] }
-picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky = "7.0.0-rc.12"
+picky-asn1-der = { path = "../../../picky-rs/picky-asn1-der" }
+picky-asn1-x509 = { path = "../../../picky-rs/picky-asn1-x509" }
+picky = { path = "../../../picky-rs/picky" }
```
TheBestTvarynka: https://github.com/Devolutions/picky-rs/pull/380

crates/ironrdp-tokio/src/lib.rs
```diff
@@ -13,6 +13,14 @@ use std::io;
 use bytes::BytesMut;
 use tokio::io::{AsyncRead, AsyncWrite, ReadHalf, WriteHalf};
 
+#[cfg(feature = "reqwest")]
+pub trait AsyncSendableNetworkClient: Send + Sync {
+    fn send<'a>(
+        &'a mut self,
+        network_request: &'a sspi::NetworkRequest,
+    ) -> Pin<Box<dyn core::future::Future<Output = ironrdp_connector::ConnectorResult<Vec<u8>>> + Send + 'a>>;
+}
```
TheBestTvarynka: ### What?

It may look weird, but let me explain myself.

The existing `AsyncNetworkClient` trait is `!Send` and returns `!Send` future. This is a problem for the Devolutions-Gateway because it requires the network client to be `Send`, and, even more, the returned future should also be `Send`.
I cannot _just_ change the existing trait, because the implemented network client in `ironrdp-web` returns `!Send` future, and there is no way of making it `Send`. Moreover, `ironrdp-async` and all other crates work pretty well with the `!Send` network client, and I do not want to restrict it.
Thus, I added a new trait which is almost identical but is `Send` but returns a `Send`able future. It will be used (it is already used, but only locally on my laptop) by Devolutions-Gateway (and other future RDP clients which require `Send`able futures).
This approach is also great because it doesn't force me to refactor other `ironrdp-*` crates.

### Why do not define this trait in the DG?

Because we need the `ReqwestNetworkClient` in the DG, an alternative approach is to make all `ReqwestNetworkClient` methods public (like `send_tcp`, `send_http`, etc.) and implement everything in the DG. But I do not like this idea and do not want to expose these methods.

crates/ironrdp-connector/Cargo.toml
CBenoit: The PR was merged and crates were published!

TheBestTvarynka: done!

crates/ironrdp-connector/Cargo.toml
CBenoit: Note that we can’t publish until the dev branch is merged onto master.

TheBestTvarynka: Yes, I remember. I am working on one more fix for sspi-rs (I found a problem when connecting using `mstsc`)

crates/ironrdp-tokio/src/lib.rs
CBenoit: I've thought this through, and I don't believe this is a good direction for a few reasons:

- Unused trait: This introduces a new trait that isn't used elsewhere in IronRDP. That adds cognitive overhead without clear benefit.
- Unnecessary API surface: The trait wraps functionality that could be exposed directly on `ReqwestNetworkClient`. It doesn't meaningfully improve abstraction, and adds public API surface we’ll need to maintain.
- Cross-cutting concerns: This mixes reqwest-specific logic (behind a feature flag) with core code (framed trait implementation). The `reqwest` feature in `ironrdp-tokio` is already something I’m not really happy about. I’d rather keep its logic self-contained. Granted this can be fixed by moving the trait definition in the appropriate module.

### Alternatives

Here are two options I believe are clearer. Feel free to pick the one that fits best:

#### Option 1: Add a `send` method directly to `ReqwestNetworkClient`

Add an async `send()` method on `impl ReqwestNetworkClient`, instead of exposing `send_tcp` etc. directly. This keeps the abstraction clear: it's a helper specific to `ReqwestNetworkClient`, not meant for generic use.  

Given that the sister PR doesn’t require trait-level genericity, this helper approach would be sufficient and avoids trait proliferation.

We also save an unnecessary heap-allocation, but it’s not really critical.

#### Option 2: Make `AsyncNetworkClient` non-object-safe

Define the trait as:

```rust
pub trait AsyncNetworkClient {
    async fn send<'a>(&'a mut self, network_request: &'a NetworkRequest) -> ConnectorResult<Vec<u8>>;
}
```

And monomorphize the code in `ironrdp-async::connector`:
```rust
#[instrument(skip_all)]
pub async fn connect_finalize<S, R>(
    _: Upgraded,
    framed: &mut Framed<S>,
    mut connector: ClientConnector,
    server_name: ServerName,
    server_public_key: Vec<u8>,
    network_client: Option<&mut R>,
    kerberos_config: Option<KerberosConfig>,
) -> ConnectorResult<ConnectionResult>
where
    S: FramedRead + FramedWrite,
    R: AsyncNetworkClient,
{
```

This is fine because the function is already monomorphized for similar reasons for `FramedRead` and `FramedWrite` traits.

crates/ironrdp-acceptor/src/credssp.rs
```diff
@@ -122,19 +149,19 @@ impl<'a> CredsspSequence<'a> {
...
-        result: Result<ServerState, Box<ServerError>>,
+        result: Result<ServerState, ServerError>,
         output: &mut WriteBuf,
     ) -> ConnectorResult<Written> {
         let (ts_request, next_state) = match result {
-            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(ts_request), CredsspState::Ongoing),
+            Ok(ServerState::ReplyNeeded(ts_request)) => (Some(Box::new(ts_request)), CredsspState::Ongoing),
             Ok(ServerState::Finished(_id)) => (None, CredsspState::Finished),
-            Err(err) => (Some(err.ts_request), CredsspState::ServerError(err.error)),
+            Err(err) => (err.ts_request, CredsspState::ServerError(err.error)),
```
CBenoit: suggestion: Instead of boxing the TS request in the success path, what about unboxing the TS request in the error path? The Rust compiler has a specialization when dereferencing `Box` types for that: `*err.ts_request`.

crates/ironrdp-acceptor/src/credssp.rs
```diff
@@ -64,6 +68,30 @@ impl CredentialsProxy for CredentialsProxyImpl<'_> {
...
+) -> Result<ServerState, ServerError> {
+    let mut state = generator.start();
+
+    loop {
+        match state {
+            GeneratorState::Suspended(request) => {
+                let response = network_client
+                    .send(&request)
+                    .await
+                    .inspect_err(|err| error!(?err, "Failed to send a Kerberos message"))
```
CBenoit: question: Do we really need `inspect_err` here? I think we are propagating the error using `map_err`.

CBenoit (commented):
Looking mostly good to me. However, I would like to have changes in `ironrdp-tokio` (https://github.com/Devolutions/IronRDP/pull/839#discussion_r2177235421) in a separate PR. We can merge this now, and this should be a dedicated changelog item.

TheBestTvarynka:
Thank you for the review. I will address these comments when I finish `sspi-rs` fix

crates/ironrdp-tokio/src/reqwest.rs
```diff
@@ -45,6 +37,21 @@ impl Default for ReqwestNetworkClient {
...
+        Box::pin(async move {
+            match &request.protocol {
+                sspi::network_client::NetworkProtocol::Tcp => self.send_tcp(&request.url, &request.data).await,
+                sspi::network_client::NetworkProtocol::Udp => self.send_udp(&request.url, &request.data).await,
+                sspi::network_client::NetworkProtocol::Http | sspi::network_client::NetworkProtocol::Https => {
+                    self.send_http(&request.url, &request.data).await
+                }
+            }
+        })
+    }
```
CBenoit: question: Is it impossible to write it as an `async fn` without boxing?
```suggestion
    pub async fn send(
        &mut self,
        request: &sspi::generator::NetworkRequest,
    ) -> ConnectorResult<Vec<u8>>> {
        match &request.protocol {
            sspi::network_client::NetworkProtocol::Tcp => self.send_tcp(&request.url, &request.data).await,
            sspi::network_client::NetworkProtocol::Udp => self.send_udp(&request.url, &request.data).await,
            sspi::network_client::NetworkProtocol::Http | sspi::network_client::NetworkProtocol::Https => {
                self.send_http(&request.url, &request.data).await
            }
        }
    }
```

crates/ironrdp-tokio/src/reqwest.rs
```diff
@@ -20,15 +20,7 @@ impl AsyncNetworkClient for ReqwestNetworkClient {
...
-        Box::pin(async move {
-            match &request.protocol {
-                sspi::network_client::NetworkProtocol::Tcp => self.send_tcp(&request.url, &request.data).await,
-                sspi::network_client::NetworkProtocol::Udp => self.send_udp(&request.url, &request.data).await,
-                sspi::network_client::NetworkProtocol::Http | sspi::network_client::NetworkProtocol::Https => {
-                    self.send_http(&request.url, &request.data).await
-                }
-            }
-        })
+        ReqwestNetworkClient::send(self, request)
```
CBenoit: suggestion: If `send` was an `async fn`, you could `Box::pin` at this place only instead

crates/ironrdp-connector/Cargo.toml
CBenoit: Version 0.16 is published!

crates/ironrdp-tokio/src/reqwest.rs
TheBestTvarynka: done in https://github.com/Devolutions/IronRDP/pull/859

crates/ironrdp-tokio/src/reqwest.rs
TheBestTvarynka: done in https://github.com/Devolutions/IronRDP/pull/859

crates/ironrdp-acceptor/src/credssp.rs
TheBestTvarynka: > I think we are propagating the error using map_err.

Yes, but `map_err` uses the `Display` error implementation, but in the `error!` macro uses error's `Debug` implementation. It can give us more details about the error.

CBenoit: If that’s the case, we have a problem in the way we display the error, and this should be fixed separately, without using `inspect_err` and causing a duplication
In IronRDP, we typically use `anyhow::Error` or the [`ironrdp_error::ErrorReport`](https://github.com/Devolutions/IronRDP/blob/4dc5945019bd30e0bbb2ee28f12a206b3f78ff4f/crates/ironrdp-error/src/lib.rs#L125-L146), which are capable of crawling and displaying all the source errors.
Unless the Display implementation of some underlying error type is lacking information, this should be sufficient

TheBestTvarynka: got it. removed `inspect_err`

crates/ironrdp-connector/Cargo.toml
```diff
@@ -24,7 +24,7 @@ ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
 ironrdp-error = { path = "../ironrdp-error", version = "0.1" } # public
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.5", features = ["std"] } # public
 arbitrary = { version = "1", features = ["derive"], optional = true } # public
-sspi = "0.15" # public
+sspi = "0.16" # public
```
CBenoit, line 27: issue: Oh right, since this is a public dependency, this makes sspi’s major bumps a breaking change for the whole ironrdp-connector crate, I’ll open a separate PR for that

CBenoit: Oh, well, this is basically what your PR is doing. I guess this is a limit of the "generate changelog from commits" approach where we can’t really give a different description for each affected crate

CBenoit: I updated the PR title and description to reflect that better, and I’ll update the changelog later

CBenoit (approved):
LGTM!

PR #842, opened 2025-06-27
feat(web): add `canvasResizedCallback` method to the `SessionBuilder`

RRRadicalEdward:


crates/ironrdp-web/src/session.rs
```diff
@@ -205,6 +205,11 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
         self.clone()
     }
 
+    /// Not used in IronRDP.
```
CBenoit: suggestion: You could say that the server cannot resize the framebuffer in the RDP protocol

CBenoit:
```suggestion
    /// The server never resizes the framebuffer in the RDP protocol, so this is not used by IronRDP.
```

crates/ironrdp-web/src/session.rs
```diff
@@ -205,6 +205,11 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
         self.clone()
     }
 
+    /// The server never resizes the framebuffer in the RDP protocol, so this is not used by IronRDP.
```
CBenoit:
```suggestion
    /// Because the server does not resize the framebuffer in the RDP protocol, this feature is unused in IronRDP.
```

PR #843, opened 2025-06-27
chore(release): prepare web packages for publishing

CBenoit:
- iron-remote-desktop v0.6.0
- iron-remote-desktop-rdp v0.3.0


CBenoit:
cc @RRRadicalEdward 
Major bump because the web component requires a new function, and will fail when it’s not present.

PR #850, opened 2025-07-01
build(deps): bump picky to v7.0.0-rc.15

CBenoit:


TheBestTvarynka (approved):

CBenoit:
@TheBestTvarynka This is also something that is not directly about the feature "server-side Kerberos".

For:
- Simpler review
- Cleaner history
- Better changelog generation

Let’s merge this change separately.

PR #855, opened 2025-07-03
feat(ffi): expose API for RDCleanPath

irvingoujAtDevolution:
Expose RDCleanPath via FFI, enable IronRDP .NET connections via RDCleanPath.

Issue: ARC-310


ffi/dotnet/Devolutions.IronRdp.ConnectExample/Properties/launchSettings.json
irvingoujAtDevolution, line 1: I suggest we make this file ignored

crates/ironrdp/src/lib.rs
```diff
@@ -63,3 +63,7 @@ pub use ironrdp_session as session;
 #[cfg(feature = "svc")]
 #[doc(inline)]
 pub use ironrdp_svc as svc;
+
+#[cfg(feature = "rdcleanpath")]
+#[doc(inline)]  
+pub use ironrdp_rdcleanpath as rdclean_path;
```
CBenoit, line 69: issue: Since RDCleanPath is a custom optional extension, let’s not expose it from `ironrdp` crate.

ffi/dotnet/Devolutions.IronRdp.ConnectExample/Properties/launchSettings.json
CBenoit: Is it a VS Code file? Sure, we can ignore that. Note that you can locally decide what to ignore using the `$GIT_DIR/info/exclude` file. It’s the same format as gitignore.

CBenoit:
Please, fix the errors reported by the CI

ffi/dotnet/IronRdp.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
CBenoit: issue: Unwanted file?

ffi/src/rdcleanpath.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    #[diplomat::opaque]
+    pub struct RdCleanPathPdu(pub Option<ironrdp::rdclean_path::RDCleanPathPdu>);
+
+    impl RdCleanPathPdu {
+        pub fn new_request(
+            x224_pdu: &VecU8,
+            destination: &str,
+            proxy_auth: &str,
+            pcb: &OptionalString,
```
CBenoit: suggestion: Instead of creating this new "OptionalString" type, remove the parameter all together, and add setters and getters for all fields.

ffi/src/rdcleanpath.rs
```diff
@@ -0,0 +1,160 @@
...
+    use diplomat_runtime::DiplomatWriteable;
+    use ironrdp::rdclean_path::der::asn1::OctetString;
+    use std::fmt::Write;
+
+    use crate::error::ffi::{IronRdpError, IronRdpErrorKind};
+    use crate::error::ValueConsumedError;
+    use crate::utils::ffi::{OptionalString, VecU8, VecVecU8};
+
+    #[diplomat::opaque]
+    pub struct RdCleanPathPdu(pub Option<ironrdp::rdclean_path::RDCleanPathPdu>);
```
CBenoit: issue: I think, as a whole, that it’s not a great approach for the FFI API. Since most of the other languages, including C#, do not have first class support for algebraic datatypes such as Rust enums, I think it’s not worth to attempt to mimic this. Instead, focus on exposing `RDCleanPathPdu` itself. In this case, I think we could define getters and setters for all the fields, without the enum helper.

suggestion: https://github.com/Devolutions/IronRDP/pull/855/files#r2185470683

ffi/src/utils/mod.rs
```diff
@@ -3,6 +3,29 @@ pub mod ffi {
...
+            Ok(Box::new(VecU8(self.0[index].clone())))
+        }
+
+        pub fn get_slice<'a>(&'a self, index: usize) -> Result<Box<BytesSlice<'a>>, Box<IronRdpError>> {
+            if index >= self.0.len() {
+                return Err("index out of bounds".into());
+            }
+            Ok(Box::new(BytesSlice(&self.0[index])))
+        }
+    }
```
CBenoit: suggestion: Instead of defining a `VecVecU8` type, I would prefer something like a `CertificationChain` type.

ffi/dotnet/Devolutions.IronRdp/src/RdcleanPath.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+public class RdcleanPathConfig
+{
+    public Uri Uri { get; private set; }
+
+    public string  AuthToken { get; private set; }
+
+    public RdcleanPathConfig(Uri url, string authToken)
```
CBenoit:
```suggestion
    public RdCleanPathConfig(Uri url, string authToken)
```

ffi/dotnet/Devolutions.IronRdp/src/RdcleanPath.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+public class RdcleanPathConfig
+{
+    public Uri Uri { get; private set; }
```
CBenoit: question: What is `Uri` referring to in this context? The proxy URL?

suggestion: name it `ProxyUrl`

ffi/dotnet/Devolutions.IronRdp/src/WebsocketStream.cs
```diff
@@ -0,0 +1,64 @@
...
+            return result.Count;
+        }
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (count == 0)
+            {
+                // Note: this is particularly important, if we send a zero-length frame,
+                // somehow Gateway will raise TLS issue during the proxy.
+                return; // Nothing to write
```
CBenoit, line 44: note: I’m not sure why you get a TLS issue at this point, since you are supposed to have an already established TLS connection at this point, but `SendAsync` with an empty ArraySegment will result in a WebSocket frame to be send. A "Binary" frame that contains an empty byte array.

ffi/dotnet/Devolutions.IronRdp/src/WebsocketStream.cs
```diff
@@ -0,0 +1,64 @@
...
+            await _webSocket.SendAsync(
+                new ArraySegment<byte>(buffer, offset, count),
+                WebSocketMessageType.Binary,
+                true,
+                cancellationToken);
+        }
+
+        public override void Flush()
+        {
+        }
```
CBenoit: question: Nothing to do here? Could you add a comment explaining why?

CBenoit (commented):
I went through it!

CBenoit:
Also, ideally I would like two PRs:
- One dedicated to exposing the RDCleanPath API, and
- One dedicated to adding the WebSocket helper in the .NET library.

crates/ironrdp/src/lib.rs
irvingoujAtDevolution: So the other work around would be having ffi import from crate/ironrdp-vmconnect directly, does that work with you?

ffi/dotnet/IronRdp.sln.DotSettings
irvingoujAtDevolution: updated

ffi/src/rdcleanpath.rs
irvingoujAtDevolution: updated

ffi/src/rdcleanpath.rs
irvingoujAtDevolution: updated

ffi/src/utils/mod.rs
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp/src/RdcleanPath.cs
irvingoujAtDevolution: updated

ffi/dotnet/Devolutions.IronRdp/src/WebsocketStream.cs
irvingoujAtDevolution: The flush is never called as well, I'm not an expert in C# but I think there's some related to the "can seek" property. However, since this does not block us in any meaningful way, I;ll just leave a comment.

crates/ironrdp/src/lib.rs
CBenoit: I’m not sure to understand what you mean here?
It seems to me that you can add a dependency on `ironrdp-rdcleanpath` in `ffi`, what is the connection with `ironrdp-vmconnect`?

ffi/dotnet/Devolutions.IronRdp/src/WebsocketStream.cs
CBenoit: Flush is something that the consumer may call himself, in this case, us. It’s supposed to flush the write buffer, and immediately send the contents on the wire. Maybe it’s not required for ClientWebSocket because it already flushes before completing future returned by `SendAsync`, etc functions. The `tungstenite` library for Rust does expose a `flush` function, but it’s not the case for the C# class found in the standard library.

irvingoujAtDevolution:
> Also, ideally I would like two PRs:
> 
>     * One dedicated to exposing the RDCleanPath API, and
> 
>     * One dedicated to adding the WebSocket helper in the .NET library.

I created this one dedicated for FFI changes, after it's merged I can rebase current PR to master, because the  changes in FFI source would not compile without generated C# code.
https://github.com/Devolutions/IronRDP/pull/864

crates/ironrdp/src/lib.rs
irvingoujAtDevolution: Oh, I mixed up with the other PR

irvingoujAtDevolution: fixed in separate PR as suggested

CBenoit:
> > Also, ideally I would like two PRs:
> 
> > 
> 
> >     * One dedicated to exposing the RDCleanPath API, and
> 
> > 
> 
> >     * One dedicated to adding the WebSocket helper in the .NET library.
> 
> 
> 
> I created this one dedicated for FFI changes, after it's merged I can rebase current PR to master, because the  changes in FFI source would not compile without generated C# code.
> 
> https://github.com/Devolutions/IronRDP/pull/864

Great, thank you!

irvingoujAtDevolution:
Superseded by #1014.

PR #857, opened 2025-07-04
fix(web): fix issue when clipboard monitoring treats clipboard update  from server as the local update

RRRadicalEdward:
When the server sends the clipboard update, we write it to our clipboard. But this new clipboard data was then processed as a new one, so we sent it back to the server. This commit fixes this behavior by tracking the data that we received from the server.

### DIFF
--- a/web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
+++ b/web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
@@ -66,8 +66,9 @@
     const CLIPBOARD_MONITORING_INTERVAL = 100; // ms
 
     let isClipboardApiSupported = false;
-    let lastClientClipboardItems = new Map<string, string | Uint8Array>();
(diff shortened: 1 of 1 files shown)


web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
```diff
@@ -252,8 +274,9 @@
                 });
 
                 if (!clipboardData.isEmpty()) {
-                    lastClientClipboardData = clipboardData;
-                    remoteDesktopService.onClipboardChanged(clipboardData);
+                    lastSentClipboardData = clipboardData;
+                    // TODO(Fix): onClipboardChanged takes an ownership over clipboardData, so lastSentClipboardData will be nullptr.
```
CBenoit, line 278: question: Should this be a FIXME?

CBenoit (approved):
LGTM! Thank you!

PR #858, opened 2025-07-04
fix(web): run navigator.clipboard.write only when window has focus

RRRadicalEdward:
When we receive clipboard update from the server and the browser window is not in focus (for example, when the user copies some text directly on the machine, not via the browser's VNC viewer), we got an error that `navigator.clipboard.write` is not allowed when window is not in focus.
This PR adds a window check that the window has focus, and now `clipboard.write` runs only when the window is in focus.

https://github.com/Devolutions/IronRDP/pull/857 has to be merged first.


CBenoit (approved):
LGTM!

PR #866, opened 2025-07-11
fix(web): apply correct key codes to scan codes mapping

RRRadicalEdward:
The root cause of the bug is the incorrect keycodes-to-scancodes mappings. These mappings translated the browser's key codes (like "KeyA") to OS-specific scancode (OS where the browser runs). However, that's not exactly what we need to do. We need to map the browser's key codes to _Windows_ scancodes - the only format that is accepted by _VNC_ and _RDP_) modules. 
Let's look at an example for a better understanding. _Safari_ browser used _Linux_'s _Gecko) mappings as a fallback (because there were no MacOS-specific mappings). For a given key, _iron-remote-desktop_ was providing a scancode that did not correspond to the _Windows_ scancode for that same key. As a result, the `IronVNC` module incorrectly mapped this scancode to a `KeySym`, which resulted in an incorrect `KeySym` or `NO_SYMBOL`.

Fixes https://github.com/Devolutions/IronVNC/issues/882.


CBenoit (approved):
LGTM!

PR #869, opened 2025-07-17
chore: run cargo +nightly fmt --all

RRRadicalEdward:


crates/ironrdp-client/src/config.rs
```diff
@@ -1,14 +1,14 @@
 #![allow(clippy::print_stdout)]
 use core::num::ParseIntError;
```
CBenoit, line 3:
```suggestion

use core::num::ParseIntError;
```

CBenoit (approved):
LGTM!

PR #876, opened 2025-07-22
refactor: add more clippy lints to "Let’s not merge unintended" section

RRRadicalEdward:


Cargo.toml
```diff
@@ -126,6 +126,9 @@ wildcard_dependencies = "warn"
 print_stderr = "warn"
 print_stdout = "warn"
 dbg_macro = "warn"
+todo = "warn"
+unreachable = "warn"
+unimplemented = "warn"
```
CBenoit: I understand for `todo!`, but I’m not convinced that we need to remove `unimplemented!` and `unreachable!`. Especially, `unreachable!` seems perfectly reasonable to me 🤔 
I think this lint is useful for people who want to prove the code is panic-free in order to, e.g.: reduce the binary size

RRRadicalEdward: Make sense. Let me add lint for `todo!` only then

RRRadicalEdward: Done

CBenoit (approved):
LGTM!

PR #877, opened 2025-07-22
feat(web): add `ConfigParser` structure to parse RDP config with convenient getters

RRRadicalEdward:
This PR adds `ConfigParser` interface and structure that Devolutions Gateway Standalone can use to parse config from URL and get the required properties. For example, Devolutions Gateway Standalone requires the `hostname` to be known before calling `connect`due to the _authToken_ generation.


RRRadicalEdward:
> That being said, I don’t think this should be part of iron-remote-desktop interface.
Instead, you could re-export that from iron-remote-desktop-rdp specifically.

Done.

crates/ironrdp-web/src/config_parser.rs
```diff
@@ -0,0 +1,30 @@
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[wasm_bindgen]
+pub struct RdpConfigParser(ironrdp_propertyset::PropertySet);
+
+#[wasm_bindgen]
+impl RdpConfigParser {
+    #[wasm_bindgen(constructor)]
+    pub fn create(config: &str) -> Self {
```
CBenoit: suggestion: I feel this should be called `load` as well, since this is wrapping `ironrdp_rdpfile::load` (or `parse`)

suggestion: I think we could go a bit further by adding a constructor creating an empty `Config`, and adding `insert_str` and `insert_int` methods.

rationale: Ultimately, we could add an advanced configuration menu giving access to [all the properties](https://learn.microsoft.com/en-us/azure/virtual-desktop/rdp-properties) using checkboxes, etc. In this case, we could start a "RdpFile" from scratch, or modify an existing one before launching the session.

crates/ironrdp-web/src/config_parser.rs
```diff
@@ -0,0 +1,30 @@
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[wasm_bindgen]
+pub struct RdpConfigParser(ironrdp_propertyset::PropertySet);
```
CBenoit: suggestion: I think, I would prefer we name it `RdpFile`, so it’s clear we are dealing with the .RDP file format.

```rust
let rdp_file = RdpFile::load(rdp_file_config);
let full_address = rdp_file.get_str("full address");
```

Alternatively, `PropertySet` is also exactly what it is. Could be:

```rust
let properties = PropertySet::load_rdp_file(rdp_file_config);
// or
let properties = PropertySet::from_rdp_file(rdp_file_config);
```

web-client/iron-remote-desktop-rdp/src/main.ts
```diff
@@ -6,13 +6,16 @@ import wasm_init, {
...
     Extension,
+    RdpConfigParser,
 } from '../../../crates/ironrdp-web/pkg/ironrdp_web';
 
 export async function init(log_level: string) {
     await wasm_init();
     setup(log_level);
 }
 
+export { RdpConfigParser };
```
CBenoit: Exactly what I want to see 👍 

This gives us a new "special feature" only exposed from IronRDP.

crates/ironrdp-web/src/config_parser.rs
CBenoit: note: It could also be useful to wrap `ironrdp_rdpfile::write`.

-> Load RDP file using upload form -> Get a `RdpFile` object -> Modify it used advanced menu -> Turn it back into a stringified form -> Feed that string to IronRDP WASM module as an extension

In the future, I also want to support "RDP files" for VNC, ARD, etc, so it would help

crates/ironrdp-web/src/config_parser.rs
RRRadicalEdward: Done

crates/ironrdp-web/src/config_parser.rs
RRRadicalEdward: Done

CBenoit (approved):
LGTM!

PR #878, opened 2025-07-22
feat(web): add enableCredssp extension to support auth to AzureAd/Entra

gabrielbauman:
This change adds a new extension allowing users to turn off Network Level Authentication (CredSSP) for Azure AD/Entra scenarios.

Hi from Cloudflare!

### DIFF
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -68,6 +68,7 @@ struct SessionBuilderInner {
     force_clipboard_update_callback: Option<js_sys::Function>,
 
     use_display_control: bool,
+    enable_credssp: bool,
 }
 
 impl Default for SessionBuilderInner {
@@ -95,6 +96,7 @@ impl Default for SessionBuilderInner {
             force_clipboard_update_callback: None,
 
             use_display_control: false,
+            enable_credssp: true,
         }
     }
 }
@@ -216,6 +218,7 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             |pcb: String| { self.0.borrow_mut().pcb = Some(pcb) };
             |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
             |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
+            |enable_credssp: bool| { self.0.borrow_mut().enable_credssp = enable_credssp };
         }
 
         self.clone()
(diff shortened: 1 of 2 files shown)


crates/iron-remote-desktop/src/session.rs
```diff
@@ -54,6 +54,9 @@ pub trait SessionBuilder {
     #[must_use]
     fn canvas_resized_callback(&self, callback: js_sys::Function) -> Self;
 
+    #[must_use]
+    fn disable_nla(&self, disable: bool) -> Self;
```
CBenoit: issue: You probably forgot to remove your modifications to the common interface here

CBenoit:
Hello! Thank you for the contribution!
This is looking mostly good, but do you think you could remove the modifications made to the "common interface" (the trait), and only keep the extension, please? :pray:

crates/ironrdp-web/src/session.rs
```diff
@@ -210,12 +212,22 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
...
+        self.clone()
+    }
+
     fn extension(&self, ext: Extension) -> Self {
         iron_remote_desktop::extension_match! {
             match ext;
             |pcb: String| { self.0.borrow_mut().pcb = Some(pcb) };
             |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
             |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
+            |disable_nla: bool| { self.0.borrow_mut().disable_nla = disable_nla };
```
CBenoit: suggestion: Another suggestion. I have a strong preference for using positive flags, because it leads to code that is overall easier to understand. For instance: `!disable_nla` "do not not enable NLA" means "enable NLA". Also, we are using `enable_credssp` in `connector::Config`, I think we should stick to that.

```suggestion
            |enable_credssp: bool| { self.0.borrow_mut().enable_credssp = enable_credssp };
```

for the default value:
```rust
enable_credssp: true,
```

and finally
Image

gabrielbauman: Sure, I'll get this tidied up tomorrow!

gabrielbauman: Actually... should we rename the extension to enableCredSsp() as well?

CBenoit: Yes please! Also, could you name it `enableCredssp`? We’ve been spelling CredSSP as a single word across the codebase, so it would be more consistent :slightly_smiling_face: 
Thanks!

CBenoit (approved):
LGTM! Thanks!

PR #886, opened 2025-07-25
chore: add rustc "Correctness" lints

RRRadicalEdward:


crates/ironrdp-cliprdr-native/src/windows/clipboard_impl.rs
```diff
@@ -320,7 +320,7 @@ pub(crate) unsafe extern "system" fn clipboard_subproc(
 
         // SAFETY: `data` is a valid pointer, returned by `Box::into_raw`, transferred to OS earlier
         // via `SetWindowSubclass` call.
-        let _ = unsafe { Box::from_raw(data as *mut WinClipboardImpl) };
+        let _unused = unsafe { Box::from_raw(data as *mut WinClipboardImpl) };
```
CBenoit: I'm not sure about that, in all the modified lines in seems to me that immediately dropping the value is intended 🤔

RRRadicalEdward: What I saw was that in all the cases, we ignored the `Result` from a method with `let _ =`. In this case, it's not required to drop it immediately.

CBenoit: That’s true. However, every let_underscore_drop hit in this diff is a false positive, so I propose we keep the lint at allow :thinking: 

- `Box::from_raw(ptr)`: we _want_ to deallocate; but I admit that `unsafe { drop(Box::from_raw(ptr)) }` is the clearest way to reflect that. We can rewrite this accordingly.
- `UnboundedSender::send(msg)`: the returned `Result` has no resources to guard; a lint hit suggests a bug in the linter.
- `let _ = …` is the idiomatic way of suppressing `#[must_use]` without extending the lifetime. For guards we intentionally keep alive, we already use `let _guard = …`. Renaming to `_unused` would extend lifetimes, clutter diffs, and encourage unnecessary shadowing.

Also, the stricter `let_underscore_lock` stays at `deny` and continues catching real issues. Overall I think this one is not worth all the noise.

What do you think?

RRRadicalEdward: Make sense. Let me revert

Cargo.toml
```diff
@@ -52,6 +52,13 @@ unused_unsafe = "warn"
 missing_unsafe_on_extern = "warn"
 unsafe_attr_outside_unsafe = "warn"
 
+# == Correctness == #
+ambiguous_negative_literals  = "warn"
+arithmetic_overflow = "warn"
+future_incompatible = "warn"
```
CBenoit: question: Is it really a correctness lint? I understand this lint is more about warning about code that will stop compiling in the future. It’s also my understanding that all these lints are warn-by-default (or deny-by-default), so we don’t have to enable it again here. Did I miss something?

Cargo.toml
```diff
@@ -52,6 +52,13 @@ unused_unsafe = "warn"
...
+keyword_idents_2024 = "warn"
```
CBenoit: suggestion: Add a `FIXME: remove when switched to 2024 edition`

Cargo.toml
```diff
@@ -52,6 +52,13 @@ unused_unsafe = "warn"
...
+refining_impl_trait = "warn"
```
CBenoit: This is warn-by-default, so we can remove

Cargo.toml
```diff
@@ -52,6 +52,13 @@ unused_unsafe = "warn"
 missing_unsafe_on_extern = "warn"
 unsafe_attr_outside_unsafe = "warn"
 
+# == Correctness == #
+ambiguous_negative_literals  = "warn"
+arithmetic_overflow = "warn"
```
CBenoit: This is warn-by-default, so we can remove

Cargo.toml
RRRadicalEdward: > It’s also my understanding that all these lints are warn-by-default (or deny-by-default), so we don’t have to enable it again here. Did I miss something?

Ooh, no...  I thought they're disabled by default.

RRRadicalEdward: My bad. I removed not warn-by-default lints.

Cargo.toml
RRRadicalEdward: Done

Cargo.toml
RRRadicalEdward: Removed

Cargo.toml
RRRadicalEdward: Removed

PR #887, opened 2025-07-25
refactor: add field_scoped_visibility_modifiers clippy extra-pedantic lint

RRRadicalEdward:


crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -159,10 +159,14 @@ impl Acceptor {
             }
         }
     }
+
+    pub fn creds(&self) -> Option<&Credentials> {
```
CBenoit, line 163: issue: The field used to be pub(crate)

crates/ironrdp-acceptor/src/lib.rs
```diff
@@ -113,10 +113,12 @@ where
     S: FramedRead + FramedWrite,
 {
     assert!(acceptor.should_perform_credssp());
-    let AcceptorState::Credssp { protocol, .. } = acceptor.state else {
+    let Some(AcceptorState::Credssp { protocol, .. }) = acceptor.state().as_any().downcast_ref() else {
         unreachable!()
     };
 
+    let protocol = *protocol;
```
CBenoit, line 120: issue: I think we should not follow blindly the lint in this case, as it's leading to worse code. Allowing a field to be moved can be a good reason to use a scoped visibility modifier on the field. Here, downcasting is also arguably equivalent to making the field `pub(crate)` in the first place, but feels like a hack.

crates/ironrdp-graphics/src/zgfx/control_messages.rs
```diff
@@ -46,8 +46,8 @@ impl<'a> SegmentedDataPdu<'a> {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct BulkEncodedData<'a> {
-    pub(crate) compression_flags: CompressionFlags,
-    pub(crate) data: &'a [u8],
+    compression_flags: CompressionFlags,
+    data: &'a [u8],
 }
```
CBenoit, line 51: issue: For POD structs, or context structs, I think it’s perfectly fine to keep the `pub(crate)`. In this case, it’s exactly the same as if everything was `pub`, but the only difference is that it’s an implementation detail.

crates/ironrdp-web/src/clipboard.rs
```diff
@@ -122,9 +122,23 @@ pub(crate) struct WasmClipboard {
 
 /// Callbacks, required to interact with JS code from within the backend.
 pub(crate) struct JsClipboardCallbacks {
-    pub(crate) on_remote_clipboard_changed: js_sys::Function,
-    pub(crate) on_remote_received_format_list: Option<js_sys::Function>,
-    pub(crate) on_force_clipboard_update: Option<js_sys::Function>,
+    on_remote_clipboard_changed: js_sys::Function,
+    on_remote_received_format_list: Option<js_sys::Function>,
+    on_force_clipboard_update: Option<js_sys::Function>,
+}
```
CBenoit, line 128: issue: Likewise, it was actually better the way it was before because the parameters where named whereas in the new version there is a risk of mixing up the parameter order, creating subtle bugs. It’s also not worth to use a fully-fledged builder pattern.

ffi/src/error.rs
```diff
@@ -143,9 +143,13 @@ pub mod ffi {
 
     /// Stringified Picky error along with an error kind.
     #[diplomat::opaque]
-    pub struct IronRdpError(pub(super) super::IronRdpErrorInner);
+    pub struct IronRdpError(super::IronRdpErrorInner);
```
CBenoit, line 146: issue: The original code was also perfectly fine at this place

CBenoit (commented):
I’m not sure about this lint, based on the issues I mention in my comments. I would recommend that we keep it "allow", but feel free to keep the code improvements that are indeed improvements. I consider improvements:
- Removal of unneeded pub(crate) (because the field is never used outside of the module)
- ExtendedClientOptionalInfoBuilder is also actually better that way

RRRadicalEdward:
> I’m not sure about this lint, based on the issues I mention in my comments. I would recommend that we keep it "allow", but feel free to keep the code improvements that are indeed improvements. Overall it seems to forbid actually useful patterns, and we already tend to not write too much code leaking invariants in outside modules already.
> 
> I consider improvements:
> 
>     * Removal of unneeded pub(crate) (because the field is never used outside of the module)
> 
>     * ExtendedClientOptionalInfoBuilder is also actually better that way

Thank you for the insights. I will close this PR and open separate PRs for these improvements.

PR #889, opened 2025-07-25
feat(web): add outbound WebSocket message size limit extension

gabrielbauman:
Add support for chunking outbound WebSocket messages when they exceed a configurable size limit. This helps avoid browser- or proxy-specific WebSocket message size restrictions while maintaining wire compatibility.

Changes:
- Add outbound_message_size_limit field to SessionBuilderInner
- Implement extension handler with safe f64->u32 casting and validation
- Update writer_task to chunk large messages when limit is set
- Add outboundMessageSizeLimit() helper function to JavaScript API

Hope this is the right-ish way to do this.

### DIFF
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -307,7 +307,7 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             )
         });
 
-        let ws = WebSocket::open(&proxy_address).context("Couldn't open WebSocket")?;
+        let ws = WebSocket::open(&proxy_address).context("couldn't open WebSocket")?;
 
         // NOTE: ideally, when the WebSocket can't be opened, the above call should fail with details on why is that
         // (e.g., the proxy hostname could not be resolved, proxy service is not running), but errors are neved


crates/ironrdp-web/src/session.rs
```diff
@@ -212,13 +214,22 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
         self.clone()
     }
 
+    #[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
CBenoit: suggestion: There is a function `f64_to_u16_saturating_cast` at the bottom of the file, I suggest you implement something similar to isolate the suppression further. In your case, that’s not `saturating`, so it would be `f64_to_u16_cast(value: f64) -> Option<u16>`. Alternatively, it’s possible to create a local scope, like this:

```rust
#[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
let limit = { outbound_message_size_limit as u32 };
```

crates/ironrdp-web/src/session.rs
```diff
@@ -212,13 +214,22 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
...
             match ext;
             |pcb: String| { self.0.borrow_mut().pcb = Some(pcb) };
             |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
             |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
             |enable_credssp: bool| { self.0.borrow_mut().enable_credssp = enable_credssp };
+            |outbound_message_size_limit: f64| {
+                let limit = if outbound_message_size_limit >= 0.0 && outbound_message_size_limit <= f64::from(u32::MAX) {
+                    outbound_message_size_limit as u32
+                } else {
+                    0 // Fallback to no limit for invalid values
```
CBenoit: suggestion: While at it, let’s log a warning
```suggestion
                    warn!(outbound_message_size_limit, "Invalid outbound message size limit; fallback to not limit");
                    0
```

crates/ironrdp-web/src/session.rs
```diff
@@ -885,22 +898,38 @@ fn build_config(
...
+                    for chunk in frame.chunks(max_size as usize) {
+                        rdp_writer.write_all(chunk).await.context("Couldn't write chunk")?;
+                        rdp_writer.flush().await.context("Couldn't flush chunk")?;
+                    }
+                    continue;
+                }
+            }
+            // Send complete frame
             rdp_writer.write_all(&frame).await.context("Couldn’t write frame")?;
             rdp_writer.flush().await.context("Couldn’t flush")?;
```
CBenoit: suggestion: To make the code easier to scan, I suggest to avoid the `continue`, so we maintain a more "straightforward" control flow. It’s not necessarily bad, but in this case, I think it’s a bit too buried, and we are not saving indentation so much as the else case is even smaller. Just based on my own experience of reviewing the code, overlooking the `continue` and being a bit confused.

```suggestion
            match outbound_limit {
                Some(max_size) if frame.len() > max_size as usize {
                    // Send in chunks.
                    for chunk in frame.chunks(max_size as usize) {
                        rdp_writer.write_all(chunk).await.context("couldn't write chunk")?;
                        rdp_writer.flush().await.context("couldn't flush chunk")?;
                    }
                }
                _ => {
                     // Send complete frame.
                    rdp_writer.write_all(&frame).await.context("couldn’t write frame")?;
                    rdp_writer.flush().await.context("couldn’t flush")?;
                }
            }
```

style: Also not that error messages should not be capitalized (general Rust convention)
https://github.com/Devolutions/IronRDP/blob/ae052ed83598ad1f4ad7038b153e3c5398d2a738/STYLE.md#formatting-of-error-messages

("Couldn’t write frame" is likely on me though 😂)

style: And that in IronRDP codebase, we try to keep comments as "proper sentences" (i.e.: first word capitalized, properly punctuated…).
https://github.com/Devolutions/IronRDP/blob/ae052ed83598ad1f4ad7038b153e3c5398d2a738/STYLE.md#inline-code-comments-are-proper-sentences

CBenoit (commented):
> Add support for chunking outbound WebSocket messages when they exceed a configurable size limit. This helps avoid browser- or proxy-specific WebSocket message size restrictions while maintaining wire compatibility.

Sounds useful! Out of curiosity, can you give me a few cases you discovered where this was a problem?

> Changes:
> […]
> Hope this is the right-ish way to do this.

Looks like the right way to do it to me! Thanks!
I left a few suggestions, but it’s more about code quality and style than anything.
Let me know what you think!

gabrielbauman:
Cleaned up based on feedback.

Replaced the if-let + continue pattern in writer_task with the suggested match statement. Makes it easier to scan since chunking is clearly the special case.

Also fixed smart quotes that had snuck in somehow, made comments proper sentences, and uncapitalized error messages per style comments.

Moved the `#[expect(clippy::cast_*)]` from function-level down to a local scope around just the f64 as u32 cast when reading the config. Keeps the suppression tight and isolated.

Added warning when someone passes a negative number or something > u32::MAX for chunk size.

Hopefully this looks ok!

crates/ironrdp-web/src/session.rs
```diff
@@ -219,6 +221,17 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
             |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
             |enable_credssp: bool| { self.0.borrow_mut().enable_credssp = enable_credssp };
+            |outbound_message_size_limit: f64| {
+                let limit = if outbound_message_size_limit >= 0.0 && outbound_message_size_limit <= f64::from(u32::MAX) {
+                    #[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+                    let limit = { outbound_message_size_limit as u32 };
+                    limit
```
CBenoit:
```suggestion
                    { outbound_message_size_limit as u32 }
```

crates/ironrdp-web/src/session.rs
```diff
@@ -219,6 +221,16 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             |kdc_proxy_url: String| { self.0.borrow_mut().kdc_proxy_url = Some(kdc_proxy_url) };
             |display_control: bool| { self.0.borrow_mut().use_display_control = display_control };
             |enable_credssp: bool| { self.0.borrow_mut().enable_credssp = enable_credssp };
+            |outbound_message_size_limit: f64| {
+                let limit = if outbound_message_size_limit >= 0.0 && outbound_message_size_limit <= f64::from(u32::MAX) {
+                    #[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+                    { outbound_message_size_limit as u32 }
+                } else {
+                    warn!("Invalid outbound message size limit; fallback to unlimited");
```
CBenoit:
```suggestion
                    warn!(outbound_message_size_limit, "Invalid outbound message size limit; fallback to unlimited");
```

crates/ironrdp-web/src/session.rs
```diff
@@ -293,9 +307,9 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             )
         });
 
-        let ws = WebSocket::open(&proxy_address).context("Couldn’t open WebSocket")?;
+        let ws = WebSocket::open(&proxy_address).context("Couldn't open WebSocket")?;
```
CBenoit:
```suggestion
        let ws = WebSocket::open(&proxy_address).context("couldn't open WebSocket")?;
```

CBenoit (approved):
Perfect! LGTM!

PR #895, opened 2025-07-29
refactor: add non_zero_suggestions clippy extra-pedantic lint

RRRadicalEdward:


crates/ironrdp/examples/server.rs
```diff
@@ -177,14 +178,15 @@ impl RdpServerDisplayUpdates for DisplayUpdates {
         }
 
         info!("get_update +{x}+{y} {width}x{height}");
+        let stride = NonZeroUsize::from(width).checked_mul(NonZero::new(4).unwrap()).unwrap();
         let bitmap = BitmapUpdate {
```
CBenoit, line 183: suggestion: BitmapUpdate could be updated so that `stride` is NonZeroUsize

crates/ironrdp-server/src/display.rs
```diff
@@ -84,17 +84,17 @@ impl TryInto<Framebuffer> for BitmapUpdate {
 impl Framebuffer {
     pub fn new(width: NonZeroU16, height: NonZeroU16, format: PixelFormat) -> Self {
         let mut data = BytesMut::new();
-        let w = usize::from(width.get());
-        let h = usize::from(height.get());
+        let w = NonZeroUsize::from(width);
+        let h = NonZeroUsize::from(height);
         let bpp = usize::from(format.bytes_per_pixel());
-        data.resize(bpp * w * h, 0);
+        data.resize(bpp * w.get() * h.get(), 0);
```
CBenoit: suggestion: I think it’s best to use the `.get()` only once, instead of everywhere, just to remove noise

```rust
        let w = NonZeroUsize::from(width).get();
        let h = NonZeroUsize::from(height).get();
        let bpp = usize::from(format.bytes_per_pixel());
        data.resize(bpp * w * h, 0);
        data.resize(bpp * w * h, 0);
```

In this case, I don’t think it changes anything, but it’s fine.

crates/ironrdp-server/src/display.rs
```diff
@@ -106,21 +106,21 @@ impl Framebuffer {
         let bpp = usize::from(self.format.bytes_per_pixel());
         let x = usize::from(bitmap.x);
         let y = usize::from(bitmap.y);
-        let width = usize::from(bitmap.width.get());
-        let height = usize::from(bitmap.height.get());
+        let width = NonZeroUsize::from(bitmap.width);
+        let height = NonZeroUsize::from(bitmap.height);
```
CBenoit: suggestion: Same here

crates/ironrdp/examples/server.rs
```diff
@@ -163,12 +163,13 @@ impl RdpServerDisplayUpdates for DisplayUpdates {
...
         let x: u16 = rng.random_range(0..WIDTH);
         let width = NonZeroU16::new(rng.random_range(1..=WIDTH.checked_sub(x).unwrap())).unwrap();
-        let capacity = usize::from(width.get())
-            .checked_mul(usize::from(height.get()))
+        let capacity = NonZeroUsize::from(width)
+            .checked_mul(NonZeroUsize::from(height))
             .unwrap()
-            .checked_mul(4)
+            .checked_mul(NonZeroUsize::new(4).unwrap())
             .unwrap();
```
CBenoit: suggestion: Also here,

```rust
        let capacity = NonZeroUsize::from(width)
            .checked_mul(NonZeroUsize::from(height))
            .unwrap()
            .get()
            .checked_mul(4)
            .unwrap();
```

CBenoit:
> I don’t get why it breaks now, even though nothing changed in the dependencies, but what about upgrading the nightly toolchain?

Oh! We didn’t pin the nightly version in IronRDP…! So actually we should downgrade to yesterday’s version, as I understand it!

crates/ironrdp/examples/server.rs
RRRadicalEdward: Done

crates/ironrdp-server/src/display.rs
RRRadicalEdward: Done

crates/ironrdp-server/src/display.rs
RRRadicalEdward: Done

crates/ironrdp/examples/server.rs
RRRadicalEdward: Done

CBenoit:
> Do you want me to do it?

If you have some time for it today, yes, please!
Otherwise, I’ll handle that a little bit later today :slightly_smiling_face:

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -203,7 +203,7 @@ impl UpdateEncoder {
                 width.get().into(),
                 height.get().into(),
                 &bitmap.data,
-                bitmap.stride,
+                bitmap.stride.into(),
```
CBenoit: suggestion: I largely prefer `get` over `into` (and I’m generally averse to `into` when `from` can be used, because it’s harder to read/review)

RRRadicalEdward:
> > Do you want me to do it?
> 
> If you have some time for it now, yes, please! Otherwise, I’ll handle that a little bit later today 🙂

I will do it

crates/ironrdp-server/src/encoder/mod.rs
RRRadicalEdward: Done

crates/ironrdp-server/src/display.rs
```diff
@@ -177,6 +178,7 @@ impl BitmapUpdate {
     ///
     /// ```
     /// # use core::num::NonZeroU16;
+    /// use std::num::NonZeroUsize;
```
CBenoit:
```suggestion
    /// # use std::num::NonZeroUsize;
```

CBenoit (approved):
Pre-approving! LGTM!

crates/ironrdp-server/src/encoder/mod.rs
```diff
@@ -356,7 +356,7 @@ impl BitmapUpdateHandler for NoneHandler {
     fn handle(&mut self, bitmap: &BitmapUpdate) -> Result<UpdateFragmenter> {
         let stride = usize::from(bitmap.format.bytes_per_pixel()) * usize::from(bitmap.width.get());
         let mut data = Vec::with_capacity(stride * usize::from(bitmap.height.get()));
-        for row in bitmap.data.chunks(bitmap.stride).rev() {
+        for row in bitmap.data.chunks(bitmap.stride.into()).rev() {
```
CBenoit: Ideally, also apply my previous comment at the other places, but it’s a nitpick

RRRadicalEdward: Done here and in other places as well

PR #896, opened 2025-07-29
[DRAFT] feat(dvc): make dvc named pipe proxy cross-platform

pacmancoder:


crates/ironrdp-dvc-pipe-proxy/src/worker.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+async fn process_client<P: OsPipe>(ctx: &mut WorkerCtx) -> Result<NextWorkerState, DvcPipeProxyError> {
+    let pipe_name = &ctx.pipe_name;
+    let channel_name = &ctx.channel_name;
+
+    /*
+     match fs::metadata(&ctx.pipe_name).await
+     {
+         Ok(metadata) => {
```
CBenoit: question: Leftover?

crates/ironrdp-dvc-pipe-proxy/Cargo.toml
```diff
@@ -22,17 +22,8 @@ ironrdp-dvc = { path = "../ironrdp-dvc", version = "0.3" }
...
-windows = { version = "0.61", features = [
-    "Win32_Foundation",
-    "Win32_Security",
-    "Win32_System_Threading",
-    "Win32_Storage_FileSystem",
-    "Win32_System_Pipes",
-    "Win32_System_IO",
-] }
+tokio = { version = "1", features = ["net", "rt", "sync", "macros", "io-util"]}
+async-trait = "0.1"
```
CBenoit, line 26: note: We loose the sync-friendly non-tokio-specific interface you made efforts to keep, but at the same time the tradeoff seems worth it.

question: Just to be on the safe side, can you double check you really enabled only the minimal amount of features?

CBenoit (approved):
LGTM!

I appreciate that the API surface is very small. I see you were careful about making tokio an internal, implementation detail.

However, the downside is that we end up using two runtimes, and it’s overall a bit heavier. Maybe you could expose an async function to "ignite" the worker manually into the tokio thread pool of the consumer, and use some opaque handle that can be re-injected into DvcNamedPipeProxy, and re-used in `start`.

Feel free to modify and merge yourself!

crates/ironrdp-dvc-pipe-proxy/Cargo.toml
pacmancoder: Yep, feature set is minimal

PR #897, opened 2025-07-30
ci(xtask): pin nightly version

RRRadicalEdward:


xtask/src/bin_version.rs
```diff
@@ -9,5 +9,6 @@ pub const GRCOV: CargoPackage = CargoPackage::new("grcov", "0.8.20");
 pub const WASM_PACK: CargoPackage = CargoPackage::new("wasm-pack", "0.13.1");
 pub const TYPOS_CLI: CargoPackage = CargoPackage::new("typos-cli", "1.29.5").with_binary_name("typos");
 pub const DIPLOMAT_TOOL: CargoPackage = CargoPackage::new("diplomat-tool", "0.7.1");
+pub const NIGHTLY_TOOLCHAIN: &str = "nightly-2025-07-28";
 
 pub const WABT_VERSION: &str = "1.0.36";
```
CBenoit:
```suggestion

pub const WABT_VERSION: &str = "1.0.36";
pub const NIGHTLY_TOOLCHAIN: &str = "nightly-2025-07-28";
```

xtask/src/fuzz.rs
```diff
@@ -86,9 +84,8 @@ pub fn run(sh: &Shell, duration: Option<u32>, target: Option<String>) -> anyhow:
     for target in targets {
         cmd!(
             sh,
-            "rustup run nightly cargo fuzz run {target} -- -max_total_time={duration}"
+            "rustup run {NIGHTLY_TOOLCHAIN} cargo fuzz run {target} -- -max_total_time={duration}"
         )
-        .env("RUSTUP_TOOLCHAIN", "nightly")
```
CBenoit, line 91: question: Is it working as expected? I remember there was an issue when the environment was not set, I guess it’s fixed?

xtask/src/bin_version.rs
RRRadicalEdward: I will update it. First, I want to make sure CI completes successfully with the fix

xtask/src/fuzz.rs
RRRadicalEdward: I didn't notice any issue. In fact, we have similar code in IronVNC, and there we don't set the environment - https://github.com/Devolutions/IronVNC/blob/master/xtask/src/tasks.rs#L190-L196. So I guess it should be fine without it.

xtask/src/bin_version.rs
RRRadicalEdward: Fuzzing passed successfully 😎
https://github.com/Devolutions/IronRDP/actions/runs/16616195441/job/47009327158?pr=897

RRRadicalEdward: Done

PR #898, opened 2025-07-30
refactor: add renamed_function_params clippy extra-pedantic lint

RRRadicalEdward:


crates/ironrdp-cliprdr/src/backend.rs
```diff
@@ -111,7 +111,7 @@ pub trait CliprdrBackend: AsAny + core::fmt::Debug + Send {
     /// This method only signals the client that server requests data in the given format.
     /// Implementors should respond by compiling a [`FormatDataResponse`] and calling
     /// [`crate::Cliprdr::submit_format_data`]
-    fn on_format_data_request(&mut self, format: FormatDataRequest);
+    fn on_format_data_request(&mut self, request: FormatDataRequest);
```
RRRadicalEdward, line 114: Instead of changing the parameter name on the implementation side, I changed the name directly in the trait. This way it aligns with the naming style of other methods of this trait.

CBenoit (approved):
LGTM!

PR #902, opened 2025-07-31
refactor: add wildcard_imports clippy extra-pedantic lint

RRRadicalEdward:


crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -714,7 +714,7 @@ fn string_len(value: &str, character_set: CharacterSet) -> u16 {
 pub mod builder {
     use core::marker::PhantomData;
 
-    use super::*;
+    use super::{ExtendedClientOptionalInfo, PerformanceFlags, TimezoneInfo, RECONNECT_COOKIE_LEN};
```
CBenoit, line 717: note: I feel this was a valid glob, but I’m not sure we can tune the lint for that…? Not too bad like that either

PR #904, opened 2025-07-31
WIP: feat(web): Web Audio support for RDP web clients

gabrielbauman:
Enables RDP web clients to receive and play PCM audio from remote sessions through the browser's native Web Audio API, supporting various sample rates.

Highlights:

- Web Audio API backend with AudioContext management and sample buffering
- PCM sample rate conversion
- Extension helpers for web client session integration
- Error handling for audio context creation and playback failures

### DIFF
--- a/crates/ironrdp-web/src/audio.rs
+++ b/crates/ironrdp-web/src/audio.rs
@@ -1,0 +1,773 @@
+use core::cell::RefCell;
+use std::borrow::Cow;
+use std::collections::VecDeque;
+use std::rc::Rc;
+
+use ironrdp::rdpsnd::client::RdpsndClientHandler;
+use ironrdp::rdpsnd::pdu::{AudioFormat, AudioFormatFlags, PitchPdu, VolumePdu, WaveFormat};
+use wasm_bindgen::{closure::Closure, JsCast as _};
+use web_sys::{AudioBuffer, AudioContext, Event, GainNode};
+
+use crate::error::IronError;
+
+/// Web Audio API backend for RDP audio playback
+///
+/// Features:
+/// - PCM 16-bit audio format support
+/// - Sample rate conversion for compatibility with any RDP server
+/// - Channel count conversion (mono/stereo)
+/// - User gesture activation handling for Web Audio policy
+#[derive(Debug)]
+pub(crate) struct WebAudioBackend {
+    audio_context: AudioContext,
+    gain_node: GainNode,
+    context_sample_rate: f32,
+    supported_formats: Vec<AudioFormat>,
+    #[expect(dead_code)] // Will be used for future audio optimization
+    browser_capabilities: BrowserAudioCapabilities,
+    volume: f32,
+    pitch: f32,
+    is_active: bool,
+    context_ready: Rc<RefCell<bool>>,
+    pending_audio_data: VecDeque<(Vec<u8>, AudioFormat)>,
+    audio_queue: AudioQueue,
+}
+
+#[derive(Debug)]
+struct AudioQueue {
+    context: AudioContext,
+    current_time: f64,
+}
+
+#[derive(Debug, Clone)]
+struct BrowserAudioCapabilities {
+    supported_sample_rates: Vec<u32>,
+    #[expect(dead_code)] // Reserved for future bandwidth optimization
+    max_sample_rate: f32,
+    #[expect(dead_code)] // Reserved for future quality fallback
+    min_sample_rate: f32,
+}
+
+impl WebAudioBackend {
+    /// Create a new WebAudioBackend with the specified sample rate
+    pub(crate) fn new(sample_rate: Option<f32>) -> Result<Self, IronError> {
+        let audio_context = AudioContext::new().map_err(|e| {
+            anyhow::Error::msg(format!(
+                "failed to create Web Audio API context (check browser support and user gesture requirement): {e:?}"
+            ))
+        })?;
+
+        let gain_node = audio_context
+            .create_gain()
+            .map_err(|e| anyhow::Error::msg(format!("failed to create Web Audio gain node: {e:?}")))?;
+
+        // Connect gain node to destination
+        gain_node
+            .connect_with_audio_node(&audio_context.destination())
+            .map_err(|e| anyhow::Error::msg(format!("failed to connect gain node to audio destination: {e:?}")))?;
+
+        let context_sample_rate = audio_context.sample_rate();
+        let requested_rate = sample_rate.unwrap_or(context_sample_rate);
+
+        let browser_capabilities = Self::detect_browser_capabilities(&audio_context);
+        let supported_formats = Self::create_supported_formats(requested_rate, &browser_capabilities);
+
+        let audio_queue = AudioQueue {
+            context: audio_context.clone(),
+            current_time: audio_context.current_time(),
+        };
+
+        let context_ready = Rc::new(RefCell::new(false));
+
+        // Set up user gesture activation
+        let backend = Self {
+            audio_context: audio_context.clone(),
+            gain_node,
+            context_sample_rate,
+            supported_formats,
+            browser_capabilities,
+            volume: 1.0,
+            pitch: 1.0,
+            is_active: true, // Mark as active by default
+            context_ready: Rc::clone(&context_ready),
+            pending_audio_data: VecDeque::new(),
+            audio_queue,
+        };
+
+        info!(
+            "WebAudioBackend initialized: {} supported formats, context sample rate: {}Hz",
+            backend.supported_formats.len(),
+            backend.context_sample_rate
+        );
+        for (i, format) in backend.supported_formats.iter().enumerate() {
+            info!(
+                "Audio format {}: {:?} {}Hz {}ch ({}bps)",
+                i, format.format, format.n_samples_per_sec, format.n_channels, format.bits_per_sample
+            );
+        }
+
+        // Set up one-time user gesture listener on document
+        Self::setup_user_gesture_listener(audio_context, context_ready);
+
+        Ok(backend)
+    }
+
+    /// Detect browser audio capabilities
+    /// Simplified to focus on sample rate support since we only use PCM
+    fn detect_browser_capabilities(audio_context: &AudioContext) -> BrowserAudioCapabilities {
+        let context_sample_rate = audio_context.sample_rate();
+
+        // Common sample rates that browsers and RDP servers typically support
+        let mut supported_sample_rates = vec![22050, 44100, 48000];
+
+        // Add the context's native sample rate if not already included
+        #[expect(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+        let native_rate = context_sample_rate.round() as u32;
+        if !supported_sample_rates.contains(&native_rate) {
+            supported_sample_rates.push(native_rate);
+            supported_sample_rates.sort_unstable();
+        }
+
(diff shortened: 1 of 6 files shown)


crates/ironrdp-web/src/session.rs
```diff
@@ -231,6 +241,12 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
                 };
                 self.0.borrow_mut().outbound_message_size_limit = if limit > 0 { Some(limit) } else { None };
             };
+            |enable_audio: bool| { self.0.borrow_mut().enable_audio = enable_audio };
+            |audio_sample_rate: f64| {
+                #[expect(clippy::cast_possible_truncation)] // JavaScript numbers are f64, audio uses f32
```
CBenoit, line 246: praise: Thank you for also including the reason of the suppression.

crates/ironrdp-web/src/session.rs
```diff
@@ -977,6 +1013,13 @@ async fn connect(
         );
     }
 
+    if enable_audio {
+        debug!("Enabling audio with sample rate: {:?}", audio_sample_rate);
```
CBenoit, line 1017: style: It’s not bad per se, but the general preference goes for structured logging, like so
```suggestion
        debug!(audio_sample_rate, "Enabling audio");
```
It’s generally more concise while maintaining the same density of information.
Also, in theory, we could also parse the logs since it’s then structured.

crates/ironrdp-web/src/session.rs
```diff
@@ -977,6 +1013,13 @@ async fn connect(
...
+        let audio_backend = WebAudioBackend::new(audio_sample_rate)
+            .map_err(|e| anyhow::Error::msg(format!("failed to initialize Web Audio backend: {e:?}")))?;
```
CBenoit, line 1019: suggestion: You can use the `context` method provided by the `anyhow::Context` trait.

```suggestion
            .context("failed to initialize Web Audio backend)?;
```

I’ve spotted similar patterns in the audio module as well, you may want to double check.

crates/ironrdp-web/src/session.rs
```diff
@@ -889,7 +921,7 @@ fn build_config(
         platform: ironrdp::pdu::rdp::capability_sets::MajorPlatformType::UNSPECIFIED,
         no_server_pointer: false,
         autologon: false,
-        no_audio_playback: true,
+        no_audio_playback: !enable_audio,
```
CBenoit, line 924: thought: For a follow up PR, `no_audio_playback` should be changed to `enable_audio` or `enable_audio_playback`, as I believe its a better naming.

web-client/iron-remote-desktop-rdp/src/main.ts
```diff
@@ -43,3 +43,42 @@ export function outboundMessageSizeLimit(limit: number): Extension {
...
+ * - 44100 Hz - CD quality
+ * - 48000 Hz - Professional audio, often browser native
+ * 
+ * If not specified, the browser's native sample rate is used as the preference.
+ * 
+ * @param rate - Preferred sample rate in Hz (e.g., 48000 for 48kHz)
+ * @returns Extension for sample rate preference
+ */
+export function audioSampleRate(rate: number): Extension {
+    return new Extension('audio_sample_rate', rate);
```
CBenoit, line 83: praise: Many thanks for all the documentation!

crates/ironrdp-web/src/error.rs
```diff
@@ -6,6 +6,12 @@ pub(crate) struct IronError {
     source: anyhow::Error,
 }
 
+impl core::fmt::Debug for IronError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("IronError").field("source", &self.source).finish()
+    }
+}
```
CBenoit, line 13: question: Why was it necessary to implement Debug on IronError? The main purpose of IronError is WASM/JavaScript interop, internally we don’t really use it like the usual Rust idiomatic errors.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+            context_ready: Rc::clone(&context_ready),
+            pending_audio_data: VecDeque::new(),
+            audio_queue,
+        };
+
+        info!(
+            "WebAudioBackend initialized: {} supported formats, context sample rate: {}Hz",
+            backend.supported_formats.len(),
+            backend.context_sample_rate
+        );
```
CBenoit, line 101: style: I would also suggest the tracing structured logging approach here, but it’s fine like this as well. I’ll let you decide for the rest of the file.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+            anyhow::Error::msg(format!(
+                "failed to create Web Audio API context (check browser support and user gesture requirement): {e:?}"
+            ))
+        })?;
+
+        let gain_node = audio_context
+            .create_gain()
+            .map_err(|e| anyhow::Error::msg(format!("failed to create Web Audio gain node: {e:?}")))?;
+
+        // Connect gain node to destination
```
CBenoit, line 64: style: https://github.com/Devolutions/IronRDP/blob/82d7977672f4f691722b30708f35c6aee9f47e14/STYLE.md#inline-code-comments-are-proper-sentences

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+        }
+
+        // Set up one-time user gesture listener on document
+        Self::setup_user_gesture_listener(audio_context, context_ready);
+
+        Ok(backend)
+    }
+
+    /// Detect browser audio capabilities
+    /// Simplified to focus on sample rate support since we only use PCM
```
CBenoit, line 116: suggestion: Proper docstring should be more like this:

```suggestion
    /// Detect browser audio capabilities
    ///
    /// Simplified to focus on sample rate support since we only use PCM.
```

The first paragraph is used as the summary in the overview, and since it’s markdown, the lines a merged together, giving something like:

> Detect browser audio capabilities Simplified to focus on sample rate support since we only use PCM

Whereas my suggestion will result into:

> Detect browser audio capabilities
>
> Simplified to focus on sample rate support since we only use PCM.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+        const MAX_REASONABLE_AUDIO_BUFFER_SIZE: usize = 48000 * 2 * 2 * 10; // ~1.9MB
+
+        if pcm_data.len() > MAX_REASONABLE_AUDIO_BUFFER_SIZE {
+            return Err(anyhow::Error::msg(format!(
+                "audio buffer too large ({} bytes), possible malformed data (max: {} bytes)",
+                pcm_data.len(),
+                MAX_REASONABLE_AUDIO_BUFFER_SIZE
+            ))
+            .into());
+        }
```
CBenoit, line 255: praise: Good defensive pattern.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+        let mut mono = Vec::with_capacity(samples.len() / 2);
+        for chunk in samples.chunks_exact(2) {
+            let avg = (chunk[0] + chunk[1]) / 2.0;
+            mono.push(avg);
+        }
+        mono
+    }
+
+    /// Convert PCM 16-bit signed integer data to 32-bit float samples
+    fn convert_pcm_to_float(pcm_data: &[u8], format: &AudioFormat) -> Result<Vec<f32>, IronError> {
```
CBenoit, line 244: suggestion: It would be more idiomatic to return `anyhow::Error` (you can use `anyhow::Result<Vec<f32>>`) at this point, and only convert to `IronError` in the top-level function, unless you really need to specify an error kind, but I don’t think you do so so far.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+            self.current_time,
+            context_time
+        );
+
+        Ok(())
+    }
+}
+
+// SAFETY: In WebAssembly single-threaded environment, Send is safe for WebAudioBackend
+unsafe impl Send for WebAudioBackend {}
```
CBenoit, line 521: nitpick: Strictly speaking, Send is really to communicate that the type is thread-safe, that it is safe to send it to another thread, and it’s obviously **not the case** for WebAudioBackend (using `Rc` instead of `Arc`, etc).
So we should rather say this is a workaround that is not compromising the memory safety of the program, because as far as we are concerned for _this_ crate which is compiled into a WASM module and run in a single-threaded environment, `WebAudioBackend` is never actually used in a multi-threaded context.
The real fix would be to relax the Send bound somewhere, but I’m okay with the workaround for the purpose of getting stuff done.

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+            }
+        };
+
+        // Try to resume context and process audio
+        if let Ok(()) = self.try_resume_and_process_pending() {
+            // Context is ready, play audio immediately
+            match self.create_audio_buffer(&data, &format) {
+                Ok(buffer) => {
+                    if let Err(e) = self.audio_queue.enqueue_audio(buffer, &self.gain_node) {
+                        error!("Failed to enqueue audio: {:?}", e);
```
CBenoit, line 560: https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#logging

crates/ironrdp-web/src/audio.rs
```diff
@@ -0,0 +1,773 @@
...
+        debug!("Audio backend closed, context suspended");
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_convert_pcm_to_float_valid_16bit() {
```
CBenoit, line 635: note: Unit tests are not run by default (see the Cargo.toml), because:
- https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#testing
- https://github.com/Devolutions/IronRDP/blob/39566bff58b07637bc89d78570e043befb5123a7/crates/ironrdp-testsuite-core/tests/main.rs#L3-L12

But you don’t have to care now, I can look into moving that on my side, in a follow up PR.

crates/ironrdp-web/src/audio.rs
gabrielbauman: I just knew you'd call this out ;) Let me see what I can do.

crates/ironrdp-web/src/error.rs
gabrielbauman: I was tired and it's probably not needed. Want me to pull it?

crates/ironrdp-web/src/audio.rs
gabrielbauman: I'll adapt it to your preferred style.

crates/ironrdp-web/src/error.rs
gabrielbauman: I take your point.

crates/ironrdp-web/src/session.rs
CBenoit: Done in https://github.com/Devolutions/IronRDP/pull/907

gabrielbauman:
Note: I haven't had time to pick this up again, will probably be a couple of weeks

PR #905, opened 2025-08-01
refactor(web): add IronErrorKind variants for connector FailureCode types

gabrielbauman:
Allows Javascript embedders to handle negotiation errors that occur before the RDP session begins.

Changes:
- Add specific IronErrorKind variants for negotiation failures:
  - SslRequiredByServer: Server requires TLS/CredSSP
  - SslNotAllowedByServer: Server only supports Standard RDP Security
  - SslCertNotOnServer: Server lacks valid authentication certificate
  - InconsistentFlags: Inconsistent security protocol flags
  - HybridRequiredByServer: Server requires CredSSP
  - SslWithUserAuthRequiredByServer: Server requires TLS + client cert
- Update ironrdp-web to map FailureCode constants to specific error kinds

In JavaScript:
  if (error.kind() === IronErrorKind.SslRequiredByServer) { ... }

### DIFF
--- a/crates/iron-remote-desktop/src/error.rs
+++ b/crates/iron-remote-desktop/src/error.rs
@@ -21,4 +21,16 @@ pub enum IronErrorKind {
     RDCleanPath,
     /// Couldn’t connect to proxy
     ProxyConnect,
+    /// Server requires Enhanced RDP Security with TLS or CredSSP
+    SslRequiredByServer,
+    /// Server only supports Standard RDP Security
+    SslNotAllowedByServer,
+    /// Server lacks valid authentication certificate
+    SslCertNotOnServer,
+    /// Inconsistent security protocol flags
+    InconsistentFlags,
+    /// Server requires Enhanced RDP Security with CredSSP
+    HybridRequiredByServer,
+    /// Server requires Enhanced RDP Security with TLS and client certificate
+    SslWithUserAuthRequiredByServer,
 }
--- a/crates/ironrdp-connector/src/lib.rs
+++ b/crates/ironrdp-connector/src/lib.rs
@@ -276,6 +276,7 @@ pub enum ConnectorErrorKind {
     AccessDenied,
     General,
     Custom,
+    NegotiationFailure(ironrdp_pdu::nego::FailureCode),
 }
 
 impl fmt::Display for ConnectorErrorKind {
@@ -288,6 +289,7 @@ impl fmt::Display for ConnectorErrorKind {
             ConnectorErrorKind::AccessDenied => write!(f, "access denied"),
             ConnectorErrorKind::General => write!(f, "general error"),
             ConnectorErrorKind::Custom => write!(f, "custom error"),
+            ConnectorErrorKind::NegotiationFailure(code) => write!(f, "negotiation failure: {code}"),
         }
     }
 }
@@ -302,6 +304,7 @@ impl core::error::Error for ConnectorErrorKind {
             ConnectorErrorKind::AccessDenied => None,
             ConnectorErrorKind::Custom => None,
             ConnectorErrorKind::General => None,
+            ConnectorErrorKind::NegotiationFailure(_) => None,
         }
     }
 }
--- a/crates/ironrdp-web/src/error.rs
+++ b/crates/ironrdp-web/src/error.rs
@@ -1,5 +1,6 @@
 use iron_remote_desktop::IronErrorKind;
 use ironrdp::connector::{self, sspi, ConnectorErrorKind};
+use tracing::info;
 
 pub(crate) struct IronError {
     kind: IronErrorKind,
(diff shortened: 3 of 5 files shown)


crates/ironrdp-connector/src/lib.rs
```diff
@@ -36,6 +36,22 @@ pub use self::license_exchange::{LicenseExchangeSequence, LicenseExchangeState};
...
+        FailureCode::SSL_NOT_ALLOWED_BY_SERVER => "server only supports Standard RDP Security".to_owned(),
+        FailureCode::SSL_CERT_NOT_ON_SERVER => "server lacks valid authentication certificate".to_owned(),
+        FailureCode::INCONSISTENT_FLAGS => "inconsistent security protocol flags".to_owned(),
+        FailureCode::HYBRID_REQUIRED_BY_SERVER => "server requires Enhanced RDP Security with CredSSP".to_owned(),
+        FailureCode::SSL_WITH_USER_AUTH_REQUIRED_BY_SERVER => {
+            "server requires Enhanced RDP Security with TLS and client certificate".to_owned()
+        }
+        _ => format!("unknown negotiation failure (code: 0x{:08x})", u32::from(code)),
+    }
+}
```
CBenoit: praise: I like the idea!

suggestion: Use a NewType wrapper to create a full-fledged error. Also make it public.

```rust
/// Provides user-friendly error messages for RDP negotiation failures
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub struct NegotiationFailure(ironrdp_pdu::nego::FailureCode);

impl NegotiationFailure {
    pub fn code(self) -> ironrdp_pdu::nego::FailureCode {
        self.0
    }
}

impl core::error::Error from NegotiationFailure {}

impl From<ironrdp_pdu::nego::FailureCode> for NegotiationFailure {
    fn from(code: ironrdp_pdu::nego::FailureCode) -> Self {
        Self(code)
    }
}

impl std::fmt::Display for NegotiationFailure {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        use ironrdp_pdu::nego::FailureCode;
        
        match self.0 {
            FailureCode::SSL_REQUIRED_BY_SERVER => {
                write!(f, "server requires Enhanced RDP Security with TLS or CredSSP")
            }
            FailureCode::SSL_NOT_ALLOWED_BY_SERVER => {
                write!(f, "server only supports Standard RDP Security")
            }
            FailureCode::SSL_CERT_NOT_ON_SERVER => {
                write!(f, "server lacks valid authentication certificate")
            }
            FailureCode::INCONSISTENT_FLAGS => {
                write!(f, "inconsistent security protocol flags")
            }
            FailureCode::HYBRID_REQUIRED_BY_SERVER => {
                write!(f, "server requires Enhanced RDP Security with CredSSP")
            }
            FailureCode::SSL_WITH_USER_AUTH_REQUIRED_BY_SERVER => {
                write!(f, "server requires Enhanced RDP Security with TLS and client certificate")
            }
            _ => write!(f, "unknown negotiation failure (code: 0x{:08x})", u32::from(self.0)),
        }
    }
}
```

rationale: No string allocation when you just need to format the message. The message is generated on-demand during formatting. More efficient for cases where you're just passing the message to `println!`, `format!`, or other formatting macros. Still allows creating an owned String when needed via `failure_message.to_string()`. It’s possible to retrieve the original error code using `.code()`.

suggestion: Rename `NegotiationFailure` variant in `ConnectorErrorKind` into `Negotiation`. (We typically don’t duplicate the word "Error" or "Failure" in the variant names.)

suggestion: Use the `NegotiationFailure` type in the `Negotiation` variant of `ConnectorErrorKind`.

You can construct the `ConnectorError` like so:
```rust
ConnectorError::new("connection initiation", ConnectorErrorKind::Negotiation(NegotiationFailure::from(code)))
```

(The failure happens during the connection initiation step of the connection sequence.)

crates/iron-remote-desktop/src/error.rs
```diff
@@ -21,4 +21,6 @@ pub enum IronErrorKind {
     RDCleanPath,
     /// Couldn’t connect to proxy
     ProxyConnect,
+    /// Protocol negotiation failed
+    NegotiationFailure,
```
CBenoit, line 25: praise: I had a peek at the code before you update it, and I think I much more prefer the current choice 👍

crates/ironrdp-web/src/error.rs
```diff
@@ -27,23 +28,24 @@ impl From<connector::ConnectorError> for IronError {
...
             ConnectorErrorKind::Credssp(sspi::Error {
                 nstatus: Some(NStatusCode::LOGON_FAILURE),
                 ..
-            }) => IronErrorKind::LogonFailure,
-            ConnectorErrorKind::AccessDenied => IronErrorKind::AccessDenied,
-            _ => IronErrorKind::General,
+            }) => (IronErrorKind::LogonFailure, anyhow::Error::new(e)),
+            ConnectorErrorKind::AccessDenied => (IronErrorKind::AccessDenied, anyhow::Error::new(e)),
+            ConnectorErrorKind::NegotiationFailure(code) => {
+                info!("RDP negotiation failure: {} (code: 0x{:08x})", code, u32::from(code));
```
CBenoit: issue: I think putting this kind of logging inside conversion functions such as `From::from` is a bit confusing. Besides, I expect the error to be already logged or surfaced elsewhere, is it not the case?

crates/ironrdp-web/src/error.rs
```diff
@@ -27,23 +28,24 @@ impl From<connector::ConnectorError> for IronError {
     fn from(e: connector::ConnectorError) -> Self {
         use sspi::credssp::NStatusCode;
 
-        let kind = match e.kind {
+        let (kind, source) = match e.kind {
```
CBenoit: question: I’m not sure to understand why you changed the code to return the source from the match, seems equivalent to the previous code but a bit noisier.

crates/ironrdp-web/src/error.rs
```diff
@@ -1,5 +1,6 @@
 use iron_remote_desktop::IronErrorKind;
 use ironrdp::connector::{self, sspi, ConnectorErrorKind};
+use tracing::info;
```
CBenoit: nitpick: You don’t have to bring tracing macros in scope, we are using `#[macro_use] extern crate tracing` in the lib.rs file, because tracing macros are used pervasively.

crates/ironrdp-web/src/error.rs
gabrielbauman: I'll remove it, I was just debugging a thing. Good catch

crates/ironrdp-web/src/error.rs
gabrielbauman: You're right, no need to create a tuple. Blah, sorry.

CBenoit (approved):
LGTM! 🎉

PR #911, opened 2025-08-05
refactor(server): remove needless sync primitives

CBenoit:
Instead use a `Option` as a slot from which the value can be moved out temporarily when necessary.


CBenoit:
cc @elmarco

pacmancoder (approved):

PR #913, opened 2025-08-05
Add MS-TSGU (Microsoft RD Gateway) support

steffengy:
This adds a working state to connect with the ironrdp-client CLI against a server behind a microsoft remote desktop gateway. 
During my testing this was robust enough to work with sessions for more than 30 minutes.

CLI Flags and prompts are implemented and can be mixed, so the following would prompt only for 2 passwords:
> ironrdp-client --gw-user username@domain --gw-endpoint rdp.gw.host:443 -u username@domain rdp.internal.host

[MS-TSGU] https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tsgu/0007d661-a86d-4e8f-89f7-7f77f8824188
* This implements a MVP (in terms of recentness) state needed to connect through microsoft rdp gateway.
* This only supports the HTTPS protocol with Websocket (and not the legacy HTTP, HTTP-RPC or UDP protocols).
* This does not implement reconnection/reauthentication.
* This only supports basic auth.

Mostly looking for rough initial feedback (e.g. in terms of if there are parts that dont align with projects architecture or other areas needing major rework) as well as if the implemented scope would be considered complete enough to land this in the first place.

### DIFF
--- a/crates/ironrdp-client/src/config.rs
+++ b/crates/ironrdp-client/src/config.rs
@@ -363,6 +363,7 @@ impl Config {
                 .context("Address prompt")?
                 .pipe(Destination::new)?
         };
+
         if let Some(ref mut gw) = gw {
             gw.server = destination.name.clone(); // TODO
         }


crates/ironrdp-client/src/config.rs
```diff
@@ -310,6 +319,44 @@ impl Config {
...
 
+        let mut gw: Option<GwConnectTarget> = None;
+
+        let gw_addr = if let Some(gw_addr) = args.gw_endpoint {
+            gw_addr
+        } else {
+            inquire::Text::new("Gateway address:")
+                .prompt()
+                .context("Address prompt")?
+        };
```
CBenoit: issue: According to this logic, we’ll always be prompted for the gateway address unless we specify it in the CLI. I would prefer we preserve the current behavior when none of these arguments are specified, as I think not using an RD Gateway is the most frequent use case.

suggestion:
```suggestion
        let mut gw: Option<GwConnectTarget> = None;

        let gw_addr = args.gw_endpoint;
```

The username and password will still be prompted for when the --gw-endpoint argument is specified.

crates/ironrdp-client/src/config.rs
```diff
@@ -310,6 +319,44 @@ impl Config {
...
+            } else {
+                inquire::Text::new("Gateway Username:")
+                    .prompt()
+                    .context("Username prompt")?
+            };
+
+            gw.gw_pass = if let Some(gw_pass) = args.gw_pass {
+                gw_pass
+            } else {
+                inquire::Password::new("Gateway Password:")
```
CBenoit: suggestion: For consistency with the other prompts.
```suggestion
                inquire::Text::new("Gateway username:")
                    .prompt()
                    .context("Username prompt")?
            };

            gw.gw_pass = if let Some(gw_pass) = args.gw_pass {
                gw_pass
            } else {
                inquire::Password::new("Gateway password:")
```

crates/ironrdp-client/src/config.rs
```diff
@@ -325,6 +372,9 @@ impl Config {
                 .context("Address prompt")?
                 .pipe(Destination::new)?
         };
+        if let Some(ref mut gw) = gw {
```
CBenoit: style: Spacing
```suggestion
};

        if let Some(ref mut gw) = gw {
```

steffengy: You mean just the separation spacing to the upper line/block?
    Because this also seems to move the `};` to a non-indented position. 
I also meanwhile ran cargo fmt.

CBenoit: Yes, not sure what happened with the `};`! Thanks!

steffengy: Added.

PR #916, opened 2025-08-12
feat: convert X.224 negotiation failures in RDCleanPath error responses to NegotiationFailure when present

gabrielbauman:
This is a follow-up to the previous PR that added IronErrorKind.NegotiationFailure etc. The idea is that both web and desktop clients check RDCleanPath error responses for X.224 connection errors from the server. If present we throw a Negotiation error.

If the proxy doesn't include the server's X.224 connection response in the RDCleanPath error response, behaviour is unchanged.

* When X.224 Connection Confirm failure is found, convert to specific NegotiationFailure error type instead of generic RDCleanPath error
* Enable clients to show meaningful error messages like "CredSSP authentication required" instead of generic connection failures
* Maintain backward compatibility - existing proxies sending empty x224_connection_pdu continue working as before
* Helper for proxies creating an RDCleanPath error with server response

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1510,7 +1510,7 @@ dependencies = [
  "embed-resource",
  "ironrdp",
  "ironrdp-cliprdr-native",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "sspi",
  "thiserror 2.0.14",
  "tracing",
@@ -2332,7 +2332,7 @@ dependencies = [
  "ironrdp-cliprdr",
  "ironrdp-cliprdr-native",
  "ironrdp-connector",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-displaycontrol",
  "ironrdp-dvc",
  "ironrdp-graphics",
@@ -2359,7 +2359,7 @@ version = "0.6.0"
 dependencies = [
  "ironrdp-async",
  "ironrdp-connector",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-pdu",
  "ironrdp-svc",
  "tracing",
@@ -2370,7 +2370,7 @@ name = "ironrdp-ainput"
 version = "0.3.0"
 dependencies = [
  "bitflags 2.9.1",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-dvc",
  "num-derive",
  "num-traits",
@@ -2382,7 +2382,7 @@ version = "0.6.0"
 dependencies = [
  "bytes",
  "ironrdp-connector",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-pdu",
  "tracing",
 ]
@@ -2403,7 +2403,7 @@ version = "0.6.0"
 dependencies = [
  "bytes",
  "ironrdp-connector",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-pdu",
  "tracing",
 ]
@@ -2426,13 +2426,14 @@ dependencies = [
  "ironrdp",
  "ironrdp-cfg",
  "ironrdp-cliprdr-native",
- "ironrdp-core",
+ "ironrdp-core 0.1.5",
  "ironrdp-dvc-pipe-proxy",
+ "ironrdp-mstsgu",
  "ironrdp-propertyset",
  "ironrdp-rdcleanpath",
  "ironrdp-rdpfile",
  "ironrdp-rdpsnd-native",
- "ironrdp-tls",
+ "ironrdp-tls 0.1.3",
  "ironrdp-tokio",
  "proc-exit",
  "raw-window-handle",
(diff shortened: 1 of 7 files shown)


gabrielbauman:
Yes, that's right. We wanted the client to be able to show detailed errors if the X224 setup failed on the target server.

crates/ironrdp-rdcleanpath/src/lib.rs
```diff
@@ -258,6 +258,34 @@ impl RDCleanPathPdu {
...
+    /// // Server rejected connection with "CredSSP required" - preserve this info
+    /// let server_response = vec![/* X.224 Connection Confirm with failure code */];
+    /// let error_pdu = RDCleanPathPdu::new_general_error_with_server_response(server_response)?;
+    /// # Ok::<(), der::Error>(())
+    /// ```
+    pub fn new_general_error_with_server_response(server_x224_response: Vec<u8>) -> der::Result<Self> {
+        Ok(Self {
+            version: VERSION_1,
+            error: Some(RDCleanPathErr {
+                error_code: GENERAL_ERROR_CODE,
```
CBenoit: thought: What about adding a new error code "NEGOTATION_ERROR_CODE" or alike?

suggestion: It’s then possible to simplify the function name: `new_negotation_error(server_x224_response: Vec<u8>) -> der::Result<Self>`

crates/ironrdp-web/src/session.rs
```diff
@@ -1077,6 +1077,26 @@ where
...
+                            anyhow::Error::new(negotiation_failure).context("RDP negotiation failed"),
+                        )
+                        .with_kind(IronErrorKind::NegotiationFailure));
+                    }
+                }
+            }
+        }
+
         let (x224_connection_response, server_cert_chain) =
             match rdcleanpath_res.into_enum().context("invalid RDCleanPath PDU")? {
```
CBenoit: thought: I think we may improve the helper enum as well. This is a breaking change in the Rust API, but not a breaking change in the on-wire representation:
- Rename `Err` into `GeneralErr`,
- Add a new `NegotationErr { x224_connection_response: Vec<u8> }` variant that is produced when the error code is NEGOTATION_ERROR_CODE and the x224_connection_response field is `Some`

What do you think?

gabrielbauman: I like this. I wasn't sure how you felt about API changes. I'll play with this a bit.

crates/ironrdp-rdcleanpath/src/lib.rs
gabrielbauman: Adding a specific NEGOTIATION_ERROR_CODE would make the error semantics much clearer, and having a dedicated new_negotiation_error() function would also make the API more explicit about when this error type should be used. I like it.

crates/ironrdp-rdcleanpath/src/lib.rs
```diff
@@ -323,7 +355,18 @@ impl TryFrom<RDCleanPathPdu> for RDCleanPath {
...
+                if let Some(x224_pdu) = pdu.x224_connection_pdu {
+                    Self::NegotiationErr {
+                        x224_connection_response: x224_pdu.as_bytes().to_vec(),
+                    }
+                } else {
+                    Self::GeneralErr(error)
+                }
+            } else {
+                Self::GeneralErr(error)
+            }
```
CBenoit: nitpick: I think this concise version is more readable

```suggestion
            match (error.error_code, pdu.x224_connection_pdu) {
                (NEGOTIATION_ERROR_CODE, Some(x224_pdu)) => Self::NegotiationErr {
                    x224_connection_response: x224_pdu.as_bytes().to_vec(),
                },
                _ => Self::GeneralErr(error),
            }
```

note: In the 2024 edition, it’s also possible to use if-let-chains, but we didn’t switch yet.

crates/ironrdp-web/src/session.rs
```diff
@@ -1087,12 +1107,36 @@ where
                     server_cert_chain,
                     server_addr: _,
                 } => (x224_connection_response, server_cert_chain),
-                ironrdp_rdcleanpath::RDCleanPath::Err(error) => {
+                ironrdp_rdcleanpath::RDCleanPath::GeneralErr(error) => {
+                    // Fallback to generic RDCleanPath error if no negotiation failure data found
```
CBenoit: nitpick: This branch is not a fallback anymore, only the one below is

crates/ironrdp-web/src/session.rs
```diff
@@ -1077,6 +1077,26 @@ where
...
+                        // Convert to negotiation failure instead of generic RDCleanPath error
+                        let negotiation_failure = connector::NegotiationFailure::from(code);
+                        return Err(IronError::from(
+                            anyhow::Error::new(negotiation_failure).context("RDP negotiation failed"),
+                        )
+                        .with_kind(IronErrorKind::NegotiationFailure));
+                    }
+                }
+            }
+        }
```
CBenoit: issue: We don’t need this code anymore, now that we extended the enum helper

CBenoit:
@gabrielbauman looks like you removed the only commit when you push-forced, so it’s not possible to re-open the PR if there is no commit… Did you lose the changes? Try to look into `git reflog` if you want to restore a previous working state!

PR #917, opened 2025-08-15
feat: allow users to specify a timezone

danielashare:
This PR allows you to pass a timezone to the remote desktop.

At some point, the Microsoft spec for the timezone bias types should be unsigned based on implementations of this spec in other clients. Now the docs have been updated to have the timezone bias type as a signed integer, so I've updated the `TimezoneInfo` struct to reflect that.

### DIFF
--- a/crates/ironrdp-connector/src/lib.rs
+++ b/crates/ironrdp-connector/src/lib.rs
@@ -20,7 +20,7 @@ use std::sync::Arc;
 use ironrdp_core::{encode_buf, encode_vec, Encode, WriteBuf};
 use ironrdp_pdu::nego::NegoRequestData;
 use ironrdp_pdu::rdp::capability_sets::{self, BitmapCodecs};
-use ironrdp_pdu::rdp::client_info::PerformanceFlags;
+use ironrdp_pdu::rdp::client_info::{PerformanceFlags, TimezoneInfo};
 use ironrdp_pdu::x224::X224;
 use ironrdp_pdu::{gcc, x224, PduHint};
 pub use sspi;
@@ -228,6 +228,9 @@ pub struct Config {
     pub performance_flags: PerformanceFlags,
 
     pub license_cache: Option<Arc<dyn LicenseCache>>,
+
+    // For Timezone Redirection to sync the server's timezone with the client's.
+    pub timezone_info: TimezoneInfo,
 
     // FIXME(@CBenoit): these are client-only options, not part of the connector.
     pub enable_server_pointer: bool,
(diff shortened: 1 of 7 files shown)


crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -385,13 +385,13 @@ impl<'de> Decode<'de> for ExtendedClientOptionalInfo {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TimezoneInfo {
-    pub bias: u32,
+    pub bias: i32,
```
CBenoit: suggestion: Thank you for updating the definition. While we are at it, you could also add a link to the spec, so it’s easier to review? Thanks!

https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#doc-comments-should-link-to-reference-documents

crates/ironrdp-connector/src/lib.rs
```diff
@@ -187,6 +187,9 @@ pub struct Config {
     pub no_server_pointer: bool,
     pub pointer_software_rendering: bool,
     pub performance_flags: PerformanceFlags,
+
+    // For Timezone Redirection to sync the server's timezone with the client's.
+    pub timezone_info: TimezoneInfo,
```
CBenoit, line 233: suggestion: Move this above the FIXME

CBenoit (commented):
Thank you for this! This is looking good.

Before we merge, do you think you could separate the change in two, please?
- One commit / PR for updating the PDU definition (breaking change)
- One commit / PR for updating adding the new option (new feature)

We are relying on the commits following the conventional commit specification for auto generating the changelogs and bumping the versions 🙂 

Thanks!

CBenoit:
> Thank you for your review @CBenoit.
> 
> I've split out the PDU change in to this [PR](https://github.com/Devolutions/IronRDP/pull/921).
> 
> I hope I've understood your request and the commit specification properly, if not, please do let me know how I can improve them.

This is perfect, thank you!

CBenoit (approved):
LGTM!

danielashare:
Hey @CBenoit, it looks like some FFI changes are required, but I no longer have permission to push to this branch

ffi/src/connector/config.rs
```diff
@@ -124,6 +125,10 @@ pub mod ffi {
             self.performance_flags = Some(performance_flags.0);
         }
 
+        pub fn set_timezone_info(&mut self, timezone_info: Option<ironrdp::pdu::rdp::client_info::TimezoneInfo>) {
+            self.timezone_info = timezone_info;
+        }
```
CBenoit:
```suggestion
        // pub fn set_timezone_info(&mut self, timezone_info: Option<ironrdp::pdu::rdp::client_info::TimezoneInfo>) {
        //     self.timezone_info = timezone_info;
        // }
```

PR #919, opened 2025-08-19
refactor: add `macro_use_extern_crate` rustc style/readability lint

RRRadicalEdward:
This lint warns on each `macro_use` attribute on an extern crate. Such a way of importing macros from a crate is being phased out. Instead of it, we should directly import the needed macro using `use` import.


RRRadicalEdward:
https://github.com/Devolutions/IronRDP/pull/918 should be merged first.

xtask/src/macros.rs
```diff
@@ -1,3 +1,4 @@
+#[macro_export]
```
CBenoit: issue: Do not use macro_export for internal macros, as it means for exporting the macro outside the crate. Instead, you can `pub(crate) use` it to export it from the module: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=93ee277d14ed04bb40eda0b502961233

xtask/src/main.rs
```diff
@@ -3,8 +3,7 @@
 #![allow(clippy::unwrap_used)]
 #![allow(unreachable_pub)]
 
-#[macro_use]
-mod macros;
+pub mod macros;
```
CBenoit: issue: For the same reason, this should not be `pub`, since we are not exporting anything as a library. Or am I missing something?

xtask/src/fuzz.rs
```diff
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-
+use crate::windows_skip;
```
CBenoit: suggestion: Re-export the xtask macros in the prelude, I think this aligns better with the initial intention.

crates/ironrdp-session/src/rfx.rs
```diff
@@ -5,11 +5,12 @@ use ironrdp_graphics::image_processing::PixelFormat;
 use ironrdp_graphics::rectangle_processing::Region;
 use ironrdp_graphics::{dwt, quantization, rlgr, subband_reconstruction};
 use ironrdp_pdu::codecs::rfx::{self, EntropyAlgorithm, Quant, RfxRectangle, Tile};
-use ironrdp_pdu::geometry::{InclusiveRectangle, Rectangle as _};
-use ironrdp_pdu::{decode_cursor, Decode as _, ReadCursor};
+use ironrdp_pdu::geometry::{InclusiveRectangle, Rectangle};
+use ironrdp_pdu::{decode_cursor, Decode, ReadCursor};
```
CBenoit: issue: The style I recommend for unnamed traits is to suppress them from the namespace using `as _`. This seems unrelated to the current change as well. There is a lint for that (clippy::unused_trait_names), that I would like to see enabled in the future. EDIT: Apparently already enabled. This is causing the CI to be failed.

crates/ironrdp-rdpdr-native/src/lib.rs
```diff
@@ -10,7 +10,6 @@
 #![warn(clippy::ptr_cast_constness)]
 
 #[cfg(any(target_os = "macos", target_os = "linux"))]
-#[macro_use]
 extern crate tracing;
```
CBenoit: suggestion: You can remove the whole thing. We don’t need the extern crate alone.

xtask/src/macros.rs
RRRadicalEdward: I missed this point. Let me address it

crates/ironrdp-rdpdr-native/src/lib.rs
RRRadicalEdward: Done

xtask/src/fuzz.rs
RRRadicalEdward: Done, but I had to reexport macros via `pub(crate)` as otherwise, It fails with:
```bash
error[E0364]: `trace` is only public within the crate, and cannot be re-exported outside
 --> xtask/src/prelude.rs:8:25
  |
8 | pub use crate::macros::{trace, run_cmd_in, windows_skip};
  |                         ^^^^^
  |
note: consider marking `trace` as `pub` in the imported module
 --> xtask/src/prelude.rs:8:25
  |
8 | pub use crate::macros::{trace, run_cmd_in, windows_skip};
```

xtask/src/main.rs
RRRadicalEdward: Fixed

xtask/src/macros.rs
RRRadicalEdward: Done

xtask/src/fuzz.rs
CBenoit: Makes sense!

crates/ironrdp-connector/src/license_exchange.rs
```diff
@@ -7,9 +7,10 @@ use std::sync::Arc;
 use ironrdp_core::WriteBuf;
 use ironrdp_pdu::rdp::server_license::{self, LicenseInformation, LicensePdu, ServerLicenseError};
 use ironrdp_pdu::PduHint;
-use rand::prelude::*;
+use rand::prelude::RngCore as _;
```
CBenoit:
```suggestion
use rand::RngCore as _;
```

PR #920, opened 2025-08-19
refactor(mstsgu): follow up to PR 913

CBenoit:
- Update tokio-tungstenite to latest
- Fix the dependencies
- Move the top-level documentation into a README.md that we re-include in the source code
- Re-order the imports using the nightly formater
- Audit the unwraps and remove them
- Fix the UTF-16 string length computation (related to https://github.com/Devolutions/IronRDP/issues/130)

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -256,6 +256,10 @@ State machines to drive an RDP connection acceptance sequence
 
 Extendable skeleton for implementing custom RDP servers.
 
+#### [`crates/ironrdp-mstsgu`](./crates/ironrdp-mstsgu) (@steffengy)
+
+Terminal Services Gateway Server Protocol implementation.
+
 #### [`crates/ironrdp-glutin-renderer`](./crates/ironrdp-glutin-renderer) (no maintainer)
 
 `glutin` primitives for OpenGL rendering.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1510,7 +1510,7 @@ dependencies = [
  "embed-resource",
  "ironrdp",
  "ironrdp-cliprdr-native",
- "ironrdp-core 0.1.5",
+ "ironrdp-core",
  "sspi",
  "thiserror 2.0.14",
  "tracing",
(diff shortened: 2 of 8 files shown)


CBenoit:
## Extra feedback on https://github.com/Devolutions/IronRDP/pull/913

issue: The correct way of computing the number of bytes for a UTF-16 string using the UTF-8 version is more complex than it may seem. See this issue: https://github.com/Devolutions/IronRDP/issues/130

suggestion: Implement both Encode and Decode. It’s nice to be able to refer to the `FIXED_SIZE_PART` constant, and it’s possible to write powerful property tests.

suggestion: Add various tests (encoding/decoding roudtrips, fuzzing, etc).

suggestion: Add spec references as docstring attached to the types. See [STYLE.md](https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#doc-comments-should-link-to-reference-documents).

suggestion: I recommend to split the proto/pdu part in a no-IO crate, and leave the remaining transport-bits in ironrdp-mstsgu. It would make it much easier to fuzz and test the non-IO part without bringing unneeded dependencies.

thought: I think we need to change the semantic of Encode and Decode so it’s a bit more flexible. The definition is "produces a complete PDU ready to be sent on the network", but in practice we implemented in other situations out of convenience. I think we should define sub-traits to build upon the more generic semantic or "encoding / decoding". I may need to elaborate further.
  
thought: I’ve been thinking about this in the past, but we may need a better type than `String` for passwords.

steffengy:
@CBenoit 
Thanks! Some quick impressions:

> suggestion: I recommend to split the proto/pdu part in a no-IO crate, and leave the remaining transport-bits in ironrdp-mstsgu. It would make it much easier to fuzz and test the non-IO part without bringing unneeded dependencies.

I even thought about abstracting away tungtenite at some point, but in this case this being used e.g. from wasm due to the nature of the protocol/RDP gateways is fairly unlikely and even if it was hard to get a general abstraction that fits multiple implementations right.

> thought: I’ve been thinking about this in the past, but we may need a better type than String for passwords.

Functionality wise something like secrecy crate? (so reasons being wanting zero/wipe on drop, not displayed in logs etc. or more?)

PR #922, opened 2025-08-19
fix(web): remove `contenteditable=true` from top level element

RRRadicalEdward:
Setting `contenteditable=true` causes spurious edits of that element (when pressing dead keys using the BEPO or AZERTY keyboard).

`contenteditable=true`  was added to workaround the clipboard on Firefox: https://github.com/Devolutions/IronRDP/blob/35c19ce444bcbb1b4250d00977d60b30fab42eca/web-client/iron-remote-desktop/src/iron-remote-desktop.svelte#L59-L63

Its removal in this PR will likely break clibpoard on Firefox, but it shouldn't be an issue since we're going to rework clibpoard logic. I might good idea to merge this PR with the clipboard rework PRs to have the least possible impact.

### DIFF
--- a/web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
+++ b/web-client/iron-remote-desktop/src/iron-remote-desktop.svelte
@@ -759,12 +759,7 @@
         class:capturing-inputs={capturingInputs}
         style={wrapperStyle}
     >
-        <div
-            bind:this={screenViewer}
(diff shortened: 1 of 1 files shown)


CBenoit:
Awesome! I’ll give a try!

CBenoit (approved):
I did not observe any issue with this PR 👍 
Thank you!

PR #928, opened 2025-08-21
chore: add `mem_forget` clippy correctness lint

RRRadicalEdward:
This lint warns on usage of `std::mem::forget`. If it's needed, we should silence the lint and justify that it's used mindfully.


CBenoit (approved):
Sounds reasonable! Thank you!

PR #929, opened 2025-08-22
chore: add `mixed_read_write_in_expression` clippy correctness lint

RRRadicalEdward:
`mixed_read_write_in_expression`  checks for a read and a write to the same variable, where whether the read occurs before or after the write depends on the evaluation order of sub-expressions.


CBenoit (approved):
LGTM!

PR #932, opened 2025-08-25
refactor: add `non_ascii_literal` clippy correctness lint

RRRadicalEdward:
This lint checks for non-ASCII characters in string and char literals.


CBenoit (approved):
I feel this is a bit too strict, but it does not seem to cause any harm in our case. LGTM!

PR #934, opened 2025-08-26
refactor: add `panic` clippy correctness lint

RRRadicalEdward:
This lint checks for `panic!` usage. I allowed `panic!` usage in tests, because it's useful and doesn't hurt the production code.


crates/ironrdp-testsuite-core/tests/pdu/pointer.rs
```diff
@@ -5,6 +5,7 @@ use ironrdp_pdu::pointer::{
     PointerPositionAttribute,
 };
 
+#[expect(clippy::panic, reason = "panic is acceptable in tests")]
```
CBenoit: suggestion: Move the #![allow(clippy::panic)] at the top of the main entry point instead (in this case, use allow, instead of expect, because we want to generally allow it in the whole test crate)

crates/ironrdp-session/src/legacy.rs
```diff
@@ -5,10 +5,9 @@ use crate::SessionError;
 impl From<ironrdp_connector::ConnectorErrorKind> for crate::SessionErrorKind {
     fn from(value: ironrdp_connector::ConnectorErrorKind) -> Self {
         match value {
-            ironrdp_connector::ConnectorErrorKind::Credssp(_) => panic!("unexpected"),
-            ironrdp_connector::ConnectorErrorKind::AccessDenied => panic!("unexpected"),
-            ironrdp_connector::ConnectorErrorKind::General => crate::SessionErrorKind::General,
-            ironrdp_connector::ConnectorErrorKind::Custom => crate::SessionErrorKind::Custom,
+            ironrdp_connector::ConnectorErrorKind::Custom | ironrdp_connector::ConnectorErrorKind::Credssp(_) => {
+                crate::SessionErrorKind::Custom
+            }
```
CBenoit, line 10: question: Maybe use `unreachable!` in this case? Although I’m not sure if that’s correct… What do you think?

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -392,7 +400,7 @@ impl ColorStrideReader {
                         let color_32bit = cursor.read_array::<4>();
                         [color_32bit[2], color_32bit[1], color_32bit[0], color_32bit[3]]
                     }
-                    _ => panic!("BUG: should be validated in the calling code"),
+                    _ => unreachable!("Per the invariant on self.bpp, this path is unreachable"),
```
CBenoit:
```suggestion
                    _ => unreachable!("per the invariant on self.bpp, this path is unreachable"),
```

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -349,16 +350,23 @@ enum ColorStrideReader {
...
+    fn new(bpp: u16, stride: Stride) -> Result<Self, PointerError> {
+        Ok(match bpp {
             1 => Self::Bitmask(BitmaskStrideReader::new(stride)),
             bpp => Self::Color {
-                bpp,
+                bpp: {
+                    // INVARIANT: bpp == 16 || bpp == 24 || bpp == 32
+                    if !SUPPORTED_COLOR_BPP[1..].contains(&bpp) {
+                        return Err(PointerError::NotSupportedBpp { bpp });
+                    }
```
CBenoit: style: https://github.com/Devolutions/IronRDP/blob/8fe288400d617ff24c67ec2cc47df020c9fa7229/STYLE.md#invariants

Only use `INVARIANT:` for declaring invariants (in this case, on the field).

suggestion: You can write something like `Enforce the bpp == 16 || bpp == 24 || bpp == 32 invariant.`

crates/ironrdp-session/src/legacy.rs
RRRadicalEdward: Not sure either. As for me, both versions are equal. I would prefer to leave as is

crates/ironrdp-graphics/src/pointer.rs
```diff
@@ -340,6 +340,7 @@ impl BitmaskStrideReader {
 
 enum ColorStrideReader {
     Color {
+        /// INVARIANT: `bpp == 16 || bpp == 24 || bpp == 32`
```
CBenoit, line 343: praise: Good approach

crates/ironrdp-session/src/legacy.rs
CBenoit: Let’s go with that

crates/ironrdp-graphics/src/pointer.rs
RRRadicalEdward: Done

crates/ironrdp-graphics/src/pointer.rs
RRRadicalEdward: Done

crates/ironrdp-testsuite-core/tests/pdu/pointer.rs
RRRadicalEdward: Done

PR #935, opened 2025-08-26
fix(web)!: replace current clipboard logic with auto and manual clipboard modes

RRRadicalEdward:
This PR adds:
* auto clipboard mode. 
When it's enabled, the clipboard will be automatically monitored for changes, and updates from the server will be automatically saved to the local clipboard (this is the old logic; nothing has changed). 
* manual clipboard mode. 
One calls dedicated functions to interact with the clipboard.

### DIFF
--- a/web-client/iron-remote-desktop/src/services/clipboard.service.ts
+++ b/web-client/iron-remote-desktop/src/services/clipboard.service.ts
@@ -1,0 +1,312 @@
+import type { RemoteDesktopService } from './remote-desktop.service';
+import { isComponentDestroyed } from '../lib/stores/componentLifecycleStore';
+import { get } from 'svelte/store';
+import type { ClipboardData } from '../interfaces/ClipboardData';
+import type { RemoteDesktopModule } from '../interfaces/RemoteDesktopModule';
+import { runWhenFocusedQueue } from '../lib/stores/runWhenFocusedStore';
+import { SessionEventType } from '../enums/SessionEventType';
+
+const CLIPBOARD_MONITORING_INTERVAL = 100; // ms
+
+export class ClipboardService {
+    private remoteDesktopService: RemoteDesktopService;
+    private module: RemoteDesktopModule;
+
+    private isClipboardApiSupported: boolean = false;
+
+    private lastClientClipboardItems: Record<string, string | Uint8Array> = {};
+    private lastReceivedClipboardData: Record<string, string | Uint8Array> = {};
+    private lastSentClipboardData: ClipboardData | null = null;
+    private clipboardDataToSave: ClipboardData | null = null;
+    private lastClipboardMonitorLoopError: Error | null = null;
+
+    constructor(remoteDesktopService: RemoteDesktopService, module: RemoteDesktopModule) {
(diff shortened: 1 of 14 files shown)


RRRadicalEdward:
Hi @CBenoit! Could you review when you get a chance?

web-client/iron-remote-desktop/src/services/clipboard.service.ts
```diff
@@ -0,0 +1,312 @@
+import type { RemoteDesktopService } from './remote-desktop.service';
+import { isComponentDestroyed } from '../lib/stores/componentLifecycleStore';
+import { get } from 'svelte/store';
+import type { ClipboardData } from '../interfaces/ClipboardData';
+import type { RemoteDesktopModule } from '../interfaces/RemoteDesktopModule';
+import { runWhenFocusedQueue } from '../lib/stores/runWhenFocusedStore';
+import { SessionEventType } from '../enums/SessionEventType';
+
+const CLIPBOARD_MONITORING_INTERVAL = 100; // ms
```
CBenoit, line 9: suggestion: If using a stronger type (such as Duration) is not an option, I would go for a longer naming which includes the unit
```suggestion
const CLIPBOARD_MONITORING_INTERVAL_MS = 100;
```

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
```diff
@@ -36,6 +36,7 @@ export class RemoteDesktopService {
     private cursorHasOverride: boolean = false;
     private lastCursorStyle: string = 'default';
     private enableClipboard: boolean = true;
+    private _autoClipboard: boolean = true;
```
CBenoit, line 39: question: Why the `_` prefix? Does not seem to be the style we used so far (I also think the `private` keyword is enough)

web-client/iron-remote-desktop/src/services/clipboard.service.ts
CBenoit, line 1: praise: Good thing this is now isolated in its own file.

CBenoit (approved):
LGTM! Nothing blocking, but feel free to open a follow up PR! Thanks!

web-client/iron-remote-desktop/src/services/remote-desktop.service.ts
RRRadicalEdward: I've added a getter with the `autoClipboard` name, see - https://github.com/Devolutions/IronRDP/pull/935/files/0446b8145c61fcdeda109b178b4cb0e5cf2b75c8#diff-f10886062174a17029195ac21b4ce32308bcbe335892272adc5e5198fee74714R58

CBenoit: Oh? And this prevents you from naming the field the same way I guess? A bit unfortunate, but I understand now

web-client/iron-remote-desktop/src/services/clipboard.service.ts
RRRadicalEdward: Done in https://github.com/Devolutions/IronRDP/pull/955

PR #936, opened 2025-08-26
fix(clipbrdr-native): map `E_ACCESSDENIED` WinAPI error code to `ClipboarAccessDenied` error

RRRadicalEdward:
When the system clipboard updates, we receive an `Updated` event. Then we try to open it, but we can get `AccessDenied` error because the clipboard may still be locked for another window (like _Notepad_). To handle this, we have special logic that attempts to open the clipboard in the event of such errors. 
The problem is that nothing in the code actually sets the `ClipboardAccessDenied` error to be able to run the retrieval logic. This PR fixes it.


CBenoit (approved):
LGTM!

PR #937, opened 2025-08-26
refactor(clipbrdr-native): provide information about source WinAPI error when `Display` formatting

RRRadicalEdward:
This way it's easier to troubleshoot.


CBenoit (commented):
Unfortunately, I will give you the exact same answer I gave here: https://github.com/Devolutions/IronRDP/pull/525#pullrequestreview-2231978013

But I should document that in our STYLE.md document so it’s easier to discover

TL;DR: when formatting the error, use `ironrdp_error::ErrorReport` or `anyhow::Error`, and you’ll get all the sources formatted properly. Whatever the approach chosen locally, you need to do that because you are never guaranteed all the nested errors will always write their source error in their Display implementation (and we can’t blame them, that’s how its supposed to be done in general).

PR #938, opened 2025-08-26
feat: add support for DVC pipe proxy in FFI

pacmancoder:


ffi/src/connector/config.rs
```diff
@@ -7,15 +7,23 @@ pub mod ffi {
     use ironrdp::connector::Credentials;
     use ironrdp::pdu::rdp::capability_sets::MajorPlatformType;
 
+    use crate::dvc::ffi::DvcPipeProxyConfig;
     use crate::error::ffi::IronRdpError;
 
     #[diplomat::opaque]
-    pub struct Config(pub ironrdp::connector::Config);
+    pub struct Config {
```
pacmancoder, line 14: Although adding non-pure (contains mpsc sender inside) config-related `DvcPipeProxyConfig` to `Config` may not be 100% correct, adding additional argument to Connection constructor (`Connection.cs`) feels even more wrong.

ffi/src/dvc/dvc_pipe_proxy_message_queue.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        pub fn next_message(&self) -> Result<Option<Box<DvcPipeProxyMessage>>, Box<IronRdpError>> {
+            Ok(self.0.next_message().map(DvcPipeProxyMessage).map(Box::new))
+        }
+
+        pub fn next_message_blocking(&self) -> Result<Box<DvcPipeProxyMessage>, Box<IronRdpError>> {
+            let message = self.0.next_message_blocking().map(DvcPipeProxyMessage).map(Box::new)?;
+
+            Ok(message)
+        }
```
CBenoit, line 42: suggestion: I suppose next_message will return None when no message is ready to be received? The conventional naming in Rust is to use `try_`: `try_next_message` for the non-blocking one, and `next_message` for the blocking one. May also be a good idea to add documentation.

ffi/src/dvc/dvc_pipe_proxy_message_queue.rs
```diff
@@ -0,0 +1,70 @@
...
+        pub fn get_sink(&self) -> Box<DvcPipeProxyMessageSink> {
+            Box::new(DvcPipeProxyMessageSink(self.0.tx.clone()))
+        }
+    }
+}
+
+struct DvcPipeProxyMessageQueueInner {
+    pub tx: mpsc::SyncSender<DvcPipeProxyMessageInner>,
+    pub rx: mpsc::Receiver<DvcPipeProxyMessageInner>,
+}
```
CBenoit: nitpick: pub fields on non-pub struct. Maybe everything should be private?

ffi/src/dvc/dvc_pipe_proxy_message_queue.rs
pacmancoder: The naming/signature was made consistent with other similar type in FFI: https://github.com/Devolutions/IronRDP/blob/feat/ffi-dvc-proxy/ffi/src/clipboard/windows.rs#L37-L46

CBenoit: I see :thinking: 

I would be in favor of changing that, but can be a follow up!

ffi/src/connector/config.rs
CBenoit: This is because it’s hard to pass new arguments around in the RDM codebase? I guess it’s fine as-is!

crates/ironrdp-connector/src/connection.rs
```diff
@@ -154,6 +154,24 @@ impl ClientConnector {
...
+    }
+
+    pub fn get_static_channel_processor_mut<T>(&mut self) -> Option<&mut T>
+    where
+        T: SvcClientProcessor + 'static,
+    {
+        self.static_channels
+            .get_by_type_mut::<T>()
+            .and_then(|channel| channel.channel_processor_downcast_mut())
+    }
```
CBenoit, line 173: issue: I can workaround that myself, but you you could ease my job with releasing the crates by creating one commit or PR per CHANGELOG item. Adding new methods to the public API of a crate requires a CHANGELOG item, and a new version bump, which will be handled automatically by release-plz given the correct commits are merged into master.

ffi/src/dvc/dvc_pipe_proxy_message_queue.rs
pacmancoder: fixed

crates/ironrdp-connector/src/connection.rs
pacmancoder: done

CBenoit (approved):
LGTM!

PR #944, opened 2025-08-27
refactor(web): raise `TERMINATED` event on connection error

RRRadicalEdward:
This PR changes the session to not return an error on the connection stage. The `iron-remote-desktop` raises `TERMINATED` instead of `ERROR`, because `ERROR` is now handled as non-critical and does not close the session.


RRRadicalEdward:
@CBenoit 

I think is a thing we discussed a while ago in DM.  I will open a PR with the error handling improvement in Devolutions Gateway Standalone soon. I partially need it for the clipboard refactoring.

CBenoit (commented):
For this one, review left on Slack

PR #946, opened 2025-08-27
refactor: add `let_underscore_must_use` clippy correctness lint

RRRadicalEdward:
This lint checks for `let _ = <expr>` where expr is `#[must_use]`.

### DIFF
--- a/crates/ironrdp-client/src/app.rs
+++ b/crates/ironrdp-client/src/app.rs
@@ -5,6 +5,7 @@ use core::time::Duration;
 use std::sync::Arc;
 use std::time::Instant;
 
+use anyhow::bail;
 use raw_window_handle::{DisplayHandle, HasDisplayHandle as _};
 use tokio::sync::mpsc;
 use tracing::{debug, error, trace};
@@ -56,25 +57,33 @@ impl App {
         })
     }
 
-    fn send_resize_event(&mut self) {
+    fn send_resize_event(&mut self) -> anyhow::Result<()> {
         let Some(size) = self.last_size.take() else {
-            return;
+            return Ok(());
         };
         let Some((window, _)) = self.window.as_mut() else {
-            return;
+            return Ok(());
         };
         let scale_factor = (window.scale_factor() * 100.0) as u32;
 
-        let _ = self.input_event_sender.send(RdpInputEvent::Resize {
-            width: u16::try_from(size.width).unwrap(),
-            height: u16::try_from(size.height).unwrap(),
-            scale_factor,
-            // TODO: it should be possible to get the physical size here, however winit doesn't make it straightforward.
-            // FreeRDP does it based on DPI reading grabbed via [`SDL_GetDisplayDPI`](https://wiki.libsdl.org/SDL2/SDL_GetDisplayDPI):
-            // https://github.com/FreeRDP/FreeRDP/blob/ba8cf8cf2158018fb7abbedb51ab245f369be813/client/SDL/sdl_monitor.cpp#L250-L262
-            // See also: https://github.com/rust-windowing/winit/issues/826
-            physical_size: None,
-        });
+        if self
+            .input_event_sender
+            .send(RdpInputEvent::Resize {
+                width: u16::try_from(size.width).unwrap(),
+                height: u16::try_from(size.height).unwrap(),
+                scale_factor,
+                // TODO: it should be possible to get the physical size here, however winit doesn't make it straightforward.
+                // FreeRDP does it based on DPI reading grabbed via [`SDL_GetDisplayDPI`](https://wiki.libsdl.org/SDL2/SDL_GetDisplayDPI):
+                // https://github.com/FreeRDP/FreeRDP/blob/ba8cf8cf2158018fb7abbedb51ab245f369be813/client/SDL/sdl_monitor.cpp#L250-L262
+                // See also: https://github.com/rust-windowing/winit/issues/826
+                physical_size: None,
+            })
+            .is_err()
+        {
+            bail!("failed to send resize event");
+        };
+
+        Ok(())
     }
 
     fn draw(&mut self) {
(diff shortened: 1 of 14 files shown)


crates/ironrdp-client/src/app.rs
```diff
@@ -56,25 +57,33 @@ impl App {
...
+                // TODO: it should be possible to get the physical size here, however winit doesn't make it straightforward.
+                // FreeRDP does it based on DPI reading grabbed via [`SDL_GetDisplayDPI`](https://wiki.libsdl.org/SDL2/SDL_GetDisplayDPI):
+                // https://github.com/FreeRDP/FreeRDP/blob/ba8cf8cf2158018fb7abbedb51ab245f369be813/client/SDL/sdl_monitor.cpp#L250-L262
+                // See also: https://github.com/rust-windowing/winit/issues/826
+                physical_size: None,
+            })
+            .is_err()
+        {
+            bail!("failed to send resize event");
+        };
```
CBenoit, line 84: Ignoring the return value of a channel send -> Even if the channel was closed, we decided that we didn’t care about that explicitly by using `let _ =` (not accidental!): there is nothing relevant we can do about that, and we don’t need to retrieve the item we tried to send either. Could happen when the program is being stopped. Typically, everything will be shutdown one after another, and it’s just a harmless race condition. I would even argue that we don’t even want to log anything in that case, or even if we did, it should be a trace.

crates/ironrdp-client/src/app.rs
```diff
@@ -391,7 +402,7 @@ fn send_fast_path_events(
     input_event_sender: &mpsc::UnboundedSender<RdpInputEvent>,
     input_events: smallvec::SmallVec<[ironrdp::pdu::input::fast_path::FastPathInputEvent; 2]>,
 ) {
-    if !input_events.is_empty() {
-        let _ = input_event_sender.send(RdpInputEvent::FastPath(input_events));
-    }
+    if !input_events.is_empty() && input_event_sender.send(RdpInputEvent::FastPath(input_events)).is_err() {
+        error!("Failed to send FastPath event");
+    };
```
CBenoit, line 407: This new logic is erroneous. If `input_events` is empty, it’s not an error, we should not log "Failed to send FastPath event, since there was no event to send in the first place.

crates/ironrdp-cliprdr-native/src/windows/utils.rs
```diff
@@ -65,7 +65,7 @@ pub(crate) unsafe fn render_format(format: ClipboardFormatId, data: &[u8]) -> Wi
 
     // SAFETY: If described above safety requirements of `render_format` call are met, then
     // `SetClipboardData` is safe to call.
-    let _ = unsafe { SetClipboardData(format.value(), Some(handle)) };
+    let _handle = unsafe { SetClipboardData(format.value(), Some(handle)) };
```
CBenoit, line 68: This is a non-idiomatic way of ignoring a return value. It also changes the drop order. It should not be necessary to think about a name for a binding we don’t care about. I would expect this to be necessary in order to keep some handle alive, but here it’s not necessary, so it’s a bit misleading.

crates/ironrdp-rdpsnd-native/examples/cpal.rs
```diff
@@ -56,7 +56,7 @@ fn main() -> anyhow::Result<()> {
 
     stream.stream.play()?;
     thread::sleep(Duration::from_secs(3));
-    let _ = producer.join();
+    producer.join().expect("thread failed");
```
CBenoit, line 59: Is it really the original intention? Do we care about this error? And if we did, shouldn’t it be a normal error instead of a panic?

crates/ironrdp-server/src/server.rs
```diff
@@ -385,7 +385,9 @@ impl RdpServer {
                             break;
                         }
                         ServerEvent::GetLocalAddr(tx) => {
-                            let _ = tx.send(self.local_addr);
+                            if tx.send(self.local_addr).is_err() {
+                                bail!("failed to send the local address");
+                            }
```
CBenoit, line 390: Greatly changing the behavior: we are short-circuiting the code when an error happens, but that’s not how the code is supposed to flow. Maybe it’s fine, maybe its not; there are several similar places where we should be careful.

xtask/src/cov.rs
```diff
@@ -22,7 +22,7 @@ pub fn update(sh: &Shell) -> anyhow::Result<()> {
     let initial_branch = cmd!(sh, "git rev-parse --abbrev-ref HEAD").read()?;
 
     println!("Switch branch");
-    let _ = cmd!(sh, "git branch -D cov-data").run();
+    cmd!(sh, "git branch -D cov-data").run()?;
```
CBenoit, line 25: This is not the intended behavior. On first run, when the branch does not exist yet, this will return an error, preventing us from proceeding with the rest of the task. We also really don’t care about whether this fails or not. Alternatively, we could check for the existence of the branch before running this, but it’s probably not very interesting.

xtask/src/section.rs
```diff
@@ -25,6 +25,10 @@ impl Drop for Section {
...
     use std::io::{self, Write as _};
 
-    let _ = io::stdout().flush();
-    let _ = io::stderr().flush();
+    if let Err(err) = io::stdout().flush() {
+        eprintln!("stdout flush error: {err}");
+    }
+    if let Err(err) = io::stderr().flush() {
+        eprintln!("stderr flush error: {err}");
+    }
```
CBenoit, line 33: We don’t really care about the error, and I don’t think writing further to stderr is the right thing to do even when this fails

CBenoit (commented):
I’m sorry, I’ll put a veto on this lint… I understand that some projects may want to forbid ignoring #[must_use], even with a `let _`, but this seems to be pretty niche restriction to me, and I don’t think it makes sense to enforce that in IronRDP. I think the default behavior: #[must_use] throws a warning when the result is not handled NOR ignored with `let _`, fits us very well: it prevents careless misses, but gets out of the way when unnecessary.

I’ll add some comments, to justify my point.

RRRadicalEdward:
> I’m sorry, I’ll put a veto on this lint… I understand that some projects may want to forbid ignoring #[must_use], even with a `let _`, but this seems to be pretty niche restriction to me, and I don’t think it makes sense to enforce that in IronRDP. I think the default behavior: #[must_use] throws a warning when the result is not handled NOR ignored with `let _`, fits us very well: it prevents careless misses, but gets out of the way when unnecessary.
> 
> I’ll add some comments, to justify my point.

I see your points, and they make sense to me! I'm not strongly for the lint, and it seems to bring more inconvenience than usefulness. Let's not add it

PR #948, opened 2025-08-28
refactor: add `same_name_method` clippy correctness lint

RRRadicalEdward:
From the lint [docs](https://rust-lang.github.io/rust-clippy/master/index.html#/same_name_method):
> It lints if a struct has two methods with the same name: one from a trait, another not from a trait.

Because it can be confusing.

### DIFF
--- a/crates/ironrdp-rdpdr/src/pdu/esc/rpce.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/esc/rpce.rs
@@ -99,7 +99,7 @@ impl<T: HeaderlessDecode> Pdu<T> {
         // omit that check here.
         let _stream_header = StreamHeader::decode(src)?;
         let _type_header = TypeHeader::decode(src)?;
-        let pdu = T::decode(src, charset)?;
+        let pdu = T::decode_without_headers(src, charset)?;
         Ok(Self(pdu))
     }
 }
@@ -163,7 +163,7 @@ pub trait HeaderlessDecode: Sized {
     /// `charset` is an optional parameter that can be used to specify the character set
     /// when relevant. This is useful for accounting for the "A" vs "W" variants of certain
     /// opcodes e.g. [`ListReadersA`][`super::ScardIoCtlCode::ListReadersA`] vs [`ListReadersW`][`super::ScardIoCtlCode::ListReadersW`].
-    fn decode(src: &mut ReadCursor<'_>, charset: Option<CharacterSet>) -> DecodeResult<Self>;
+    fn decode_without_headers(src: &mut ReadCursor<'_>, charset: Option<CharacterSet>) -> DecodeResult<Self>;
 }
 
 /// [2.2.6.1] Common Type Header for the Serialization Stream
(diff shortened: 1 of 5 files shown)


crates/ironrdp-web/src/clipboard.rs
```diff
@@ -601,14 +602,6 @@ impl ClipboardData {
         self.items.push(item);
     }
 
-    pub(crate) fn items(&self) -> &[ClipboardItem] {
```
RRRadicalEdward, line 604: Removed these methods:
* `pub(crate) fn is_empty(&self) -> bool`
Not used.
*  `pub(crate) fn items(&self) -> &[ClipboardItem]` 
It's not used, but the same method from `iron_remote_desktop::ClipboardData` trait, which is implemented for the struct.

crates/ironrdp-rdpdr/src/pdu/esc/rpce.rs
```diff
@@ -163,7 +163,7 @@ pub trait HeaderlessDecode: Sized {
     /// `charset` is an optional parameter that can be used to specify the character set
     /// when relevant. This is useful for accounting for the "A" vs "W" variants of certain
     /// opcodes e.g. [`ListReadersA`][`super::ScardIoCtlCode::ListReadersA`] vs [`ListReadersW`][`super::ScardIoCtlCode::ListReadersW`].
-    fn decode(src: &mut ReadCursor<'_>, charset: Option<CharacterSet>) -> DecodeResult<Self>;
+    fn decode_without_headers(src: &mut ReadCursor<'_>, charset: Option<CharacterSet>) -> DecodeResult<Self>;
```
RRRadicalEdward: The trait is called `HeaderlessDecode`. It had its own `decode` as well as `decode` via `impl<T: HeaderlessDecode> Pdu<T>`.  Thus, it was triggering the lint.  I renamed `HeaderlessDecode`'s `decode` to `decode_without_headers`.

CBenoit: suggestion: Maybe name it `headerless_decode`? It’s a bit shorter

RRRadicalEdward: Yeah. Make sense

crates/ironrdp-tokio/src/reqwest.rs
```diff
@@ -37,7 +37,7 @@ impl Default for ReqwestNetworkClient {
 }
 
 impl ReqwestNetworkClient {
-    pub async fn send<'a>(&'a mut self, request: &'a sspi::generator::NetworkRequest) -> ConnectorResult<Vec<u8>> {
+    async fn send_impl<'a>(&'a mut self, request: &'a sspi::generator::NetworkRequest) -> ConnectorResult<Vec<u8>> {
```
CBenoit: issue: This is a breaking change, and we need the non-trait-method one in Devolutions Gateway: https://github.com/Devolutions/IronRDP/pull/859

Not sure of what would be a better name here though… Maybe `send_request` or `execute`?

CBenoit (commented):
Sounds reasonable!

crates/ironrdp-tokio/src/reqwest.rs
RRRadicalEdward: `send_request` looks good to me

RRRadicalEdward: My bad. I missed that it's a `pub`. I will revert it

crates/ironrdp-rdpdr/src/pdu/esc/rpce.rs
RRRadicalEdward: Done

crates/ironrdp-tokio/src/reqwest.rs
RRRadicalEdward: Done

PR #951, opened 2025-08-29
fix(web): implement text-only clipboard for outdated versions of Firefox browser

RRRadicalEdward:
Firefox versions before v127 do not support the write() method - only writeText(). Therefore, the Extended Clipboard checkbox must be automatically disabled.


CBenoit (approved):
Sounds reasonable! Thank you for looking into that 💯

PR #952, opened 2025-08-29
refactor(web): extend `DeviceEvent.wheelRotations` event to support passing rotation units other than pixels

RRRadicalEdward:
Closes https://github.com/Devolutions/IronVNC/issues/905.


CBenoit (approved):
LGTM! Thanks!!

CBenoit:
By the way, this is a breaking change, not a trivial internal refactoring, so I updated the commit accordingly. Could be a `fix` too, but let’s consider this a `feat` :thinking:

PR #959, opened 2025-09-02
fix(web): disable clipboard in non-secure context

RRRadicalEdward:


CBenoit (commented):
question: The PR title states "disable", but all I see is a warning. I understand the clipboard is already disabled in non-secure context? Maybe word it more accurately

PR #963, opened 2025-09-03
refactor: add `unwrap_used` clippy correctness lint

RRRadicalEdward:

PR #964, opened 2025-09-03
refactor(pdu)!: fix unwrap_used clippy lint warnings

RRRadicalEdward:
Extract code changes in `ironrdp-pdu` from the previous PR - https://github.com/Devolutions/IronRDP/pull/963. 
Github shows that we have 1K of changes, but it should be easy to review since the changes are really simple: replace `unwrap` with `expect`,  propagate the error instead of unwrapping, add `cast_length!` usage, etc.

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -350,7 +350,7 @@ impl Sequence for ClientConnector {
 
                 let client_gcc_blocks = connect_initial.conference_create_request.gcc_blocks();
 
-                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks();
+                let server_gcc_blocks = connect_response.conference_create_response.into_gcc_blocks();
 
                 if client_gcc_blocks.security == gcc::ClientSecurityData::no_security()
                     && server_gcc_blocks.security != gcc::ServerSecurityData::no_security()
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -348,7 +348,7 @@ impl Sequence for ClientConnector {
 
                 debug!(message = ?connect_response, "Received");
 
-                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks();
+                let client_gcc_blocks = connect_initial.conference_create_request.gcc_blocks();
 
                 let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks();
 
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -366,7 +366,7 @@ impl Sequence for ClientConnector {
                     warn!("Unexpected MultiTransportChannelData GCC block (not supported)");
                 }
 
-                let static_channel_ids = server_gcc_blocks.network.channel_ids.clone();
+                let static_channel_ids = server_gcc_blocks.network.channel_ids;
                 let io_channel_id = server_gcc_blocks.network.io_channel;
 
                 debug!(?static_channel_ids, io_channel_id);
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -327,7 +327,8 @@ impl Sequence for ClientConnector {
                 let client_gcc_blocks =
                     create_gcc_blocks(&self.config, selected_protocol, self.static_channels.values())?;
 
-                let connect_initial = mcs::ConnectInitial::with_gcc_blocks(client_gcc_blocks);
+                let connect_initial =
+                    mcs::ConnectInitial::with_gcc_blocks(client_gcc_blocks).map_err(ConnectorError::decode)?;
 
                 debug!(message = ?connect_initial, "Send");
 
@@ -347,9 +348,9 @@ impl Sequence for ClientConnector {
 
                 debug!(message = ?connect_response, "Received");
 
-                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks;
+                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks();
 
-                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks;
+                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks();
 
                 if client_gcc_blocks.security == gcc::ClientSecurityData::no_security()
                     && server_gcc_blocks.security != gcc::ServerSecurityData::no_security()
@@ -365,7 +366,7 @@ impl Sequence for ClientConnector {
                     warn!("Unexpected MultiTransportChannelData GCC block (not supported)");
                 }
 
-                let static_channel_ids = server_gcc_blocks.network.channel_ids;
+                let static_channel_ids = server_gcc_blocks.network.channel_ids.clone();
                 let io_channel_id = server_gcc_blocks.network.io_channel;
 
                 debug!(?static_channel_ids, io_channel_id);
--- a/crates/ironrdp-pdu/src/basic_output/fast_path.rs
+++ b/crates/ironrdp-pdu/src/basic_output/fast_path.rs
@@ -312,6 +312,7 @@ impl Encode for FastPathUpdate<'_> {
     }
 }
 
+#[repr(u8)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, FromPrimitive)]
 pub enum UpdateCode {
     Orders = 0x0,
@@ -329,21 +330,12 @@ pub enum UpdateCode {
 }
 
 impl UpdateCode {
+    #[expect(
+        clippy::as_conversions,
+        reason = "guarantees discriminant layout, and as is the only way to cast enum -> primitive"
+    )]
     pub fn as_u8(&self) -> u8 {
-        match self {
-            Self::Orders => 0x0,
-            Self::Bitmap => 0x1,
-            Self::Palette => 0x2,
-            Self::Synchronize => 0x3,
-            Self::SurfaceCommands => 0x4,
-            Self::HiddenPointer => 0x5,
-            Self::DefaultPointer => 0x6,
-            Self::PositionPointer => 0x8,
-            Self::ColorPointer => 0x9,
-            Self::CachedPointer => 0xa,
-            Self::NewPointer => 0xb,
-            Self::LargePointer => 0xc,
-        }
+        *self as u8
     }
 }
 
@@ -365,6 +357,7 @@ impl From<&FastPathUpdate<'_>> for UpdateCode {
     }
 }
 
+#[repr(u8)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, FromPrimitive)]
 pub enum Fragmentation {
     Single = 0x0,
@@ -374,13 +367,12 @@ pub enum Fragmentation {
 }
 
 impl Fragmentation {
+    #[expect(
+        clippy::as_conversions,
+        reason = "guarantees discriminant layout, and as is the only way to cast enum -> primitive"
+    )]
     pub fn as_u8(&self) -> u8 {
-        match self {
-            Self::Single => 0x0,
-            Self::Last => 0x1,
-            Self::First => 0x2,
-            Self::Next => 0x3,
-        }
+        *self as u8
     }
 }
 
--- a/crates/ironrdp-pdu/src/basic_output/fast_path.rs
+++ b/crates/ironrdp-pdu/src/basic_output/fast_path.rs
@@ -334,8 +334,8 @@ impl UpdateCode {
         clippy::as_conversions,
         reason = "guarantees discriminant layout, and as is the only way to cast enum -> primitive"
     )]
-    pub fn as_u8(&self) -> u8 {
-        *self as u8
+    pub fn as_u8(self) -> u8 {
+        self as u8
     }
 }
 
@@ -371,8 +371,8 @@ impl Fragmentation {
         clippy::as_conversions,
         reason = "guarantees discriminant layout, and as is the only way to cast enum -> primitive"
     )]
-    pub fn as_u8(&self) -> u8 {
-        *self as u8
+    pub fn as_u8(self) -> u8 {
+        self as u8
     }
 }
 
--- a/crates/ironrdp-pdu/src/basic_output/fast_path.rs
+++ b/crates/ironrdp-pdu/src/basic_output/fast_path.rs
@@ -7,8 +7,8 @@ use ironrdp_core::{
     decode_cursor, ensure_fixed_part_size, ensure_size, invalid_field_err, Decode, DecodeError, DecodeResult, Encode,
     EncodeResult, InvalidFieldErr as _, ReadCursor, WriteCursor,
 };
-use num_derive::{FromPrimitive, ToPrimitive};
-use num_traits::{FromPrimitive as _, ToPrimitive as _};
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive as _;
 
 use super::bitmap::BitmapUpdateData;
 use super::pointer::PointerUpdateData;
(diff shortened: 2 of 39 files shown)


crates/ironrdp-pdu/src/basic_output/fast_path.rs
```diff
@@ -346,14 +365,25 @@ impl From<&FastPathUpdate<'_>> for UpdateCode {
...
 
+impl Fragmentation {
+    pub fn as_u8(&self) -> u8 {
+        match self {
+            Self::Single => 0x0,
+            Self::Last => 0x1,
+            Self::First => 0x2,
+            Self::Next => 0x3,
+        }
+    }
```
CBenoit: suggestion: Since we define the value of each variant in the enum itself, this should do the trick!

```suggestion
    pub fn as_u8(self) -> u8 {
        self as u8
    }
```

I recommend this pattern whenever possible to avoid repetition 🙂

crates/ironrdp-pdu/src/gcc/core_data/client.rs
```diff
@@ -3,8 +3,8 @@ use ironrdp_core::{
     ensure_fixed_part_size, ensure_size, invalid_field_err, write_padding, Decode, DecodeResult, Encode, EncodeResult,
     ReadCursor, WriteCursor,
 };
-use num_derive::{FromPrimitive, ToPrimitive};
-use num_traits::{FromPrimitive as _, ToPrimitive as _};
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive as _;
```
CBenoit, line 7: praise: Thank you for moving a little bit away from num_traits. I think we don’t really benefit much from it.

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -26,11 +26,26 @@ const CONFERENCE_NAME: &[u8] = b"1";
...
 
 impl ConferenceCreateRequest {
     const NAME: &'static str = "ConferenceCreateRequest";
+
+    pub fn new(gcc_blocks: ClientGccBlocks) -> DecodeResult<Self> {
+        // INVARIANT: gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= u16::MAX
+        let _: usize = cast_length!(
+            "gcc blocks length",
+            gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE
+        )?;
```
CBenoit: issue? Did you mean this?

```suggestion
        let _: u16 = cast_length!(
            "gcc blocks length",
            gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE
        )?;
```

Style-wise, see https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#invariants

I would go for something like this:

```suggestion
        // Ensure the invariant on gcc_blocks.size() is respected.
        if !(gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= usize::from(u16::MAX)) {
            return Err(…);
        }
```

Only use `INVARIANT:` when defining the invariant (on the field), and reuse the same expression as-is, so it’s easy to verify.

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -84,16 +99,16 @@ impl Encode for ConferenceCreateRequest {
...
         let gcc_blocks_buffer_length = self.gcc_blocks.size();
-        let req_length: DecodeResult<u16> = cast_length!(
-            "gccBlocksLen",
-            CONFERENCE_REQUEST_CONNECT_PDU_SIZE + gcc_blocks_buffer_length
-        );
-        let length: DecodeResult<u16> = cast_length!("gccBlocksLen", gcc_blocks_buffer_length);
+        let req_length = u16::try_from(CONFERENCE_REQUEST_CONNECT_PDU_SIZE + gcc_blocks_buffer_length)
+            .expect("Per the invariant on self.gcc_blocks, this cast is infallible");
+        let length = u16::try_from(gcc_blocks_buffer_length)
+            .expect("Per the invariant on self.gcc_blocks, this cast is infallible");
```
CBenoit: style:
```suggestion
            .expect("per the invariant on self.gcc_blocks, this cast is infallible");
        let length = u16::try_from(gcc_blocks_buffer_length)
            .expect("per the invariant on self.gcc_blocks, this cast is infallible");
```

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -26,11 +26,26 @@ const CONFERENCE_NAME: &[u8] = b"1";
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ConferenceCreateRequest {
-    pub gcc_blocks: ClientGccBlocks,
+    /// INVARIANT: `gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= u16::MAX`
+    gcc_blocks: ClientGccBlocks,
```
CBenoit: praise: Good use of the invariant tactic.

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -197,6 +227,8 @@ impl Encode for ConferenceCreateResponse {
             dst,
             cast_length!(
                 "gccBlocksLen",
+                // QUESTION: It seems that the addition of 1 here is a bug.
+                // The fuzzing is not failing because this length is ignored.
                 gcc_blocks_buffer_length + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE + 1
```
CBenoit: Oh? Maybe add a `FIXME` too.

crates/ironrdp-pdu/src/basic_output/fast_path.rs
RRRadicalEdward: On the other hand, when we add a lint for `as` conversion, we will need to mark all these functions as `#[expect(clippy::as_conversions)]`. Wouldn't it be better to use `match` as the middle ground? I understand that it's tedious, but it won't require a lint silence.

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -723,10 +797,10 @@ impl From<PduError> for ClientInfoError {
...
 
-fn string_len(value: &str, character_set: CharacterSet) -> u16 {
+fn string_len(value: &str, character_set: CharacterSet) -> usize {
     match character_set {
-        CharacterSet::Ansi => u16::try_from(value.len()).unwrap(),
-        CharacterSet::Unicode => u16::try_from(value.encode_utf16().count() * 2).unwrap(),
+        CharacterSet::Ansi => value.len(),
+        CharacterSet::Unicode => value.encode_utf16().count() * 2,
     }
 }
```
CBenoit: Can you add a `TODO: UTF-16 helper`  at this place? Thank you 🙏

crates/ironrdp-pdu/src/geometry.rs
```diff
@@ -149,10 +149,12 @@ impl_rectangle!(InclusiveRectangle);
 impl_rectangle!(ExclusiveRectangle);
 
 impl Rectangle for InclusiveRectangle {
+    /// INVARIANT: `0 < output (width)`
     fn width(&self) -> u16 {
         self.right - self.left + 1
     }
 
+    /// INVARIANT: `0 < output (height)`
```
CBenoit, line 157: praise: Good style 👍

crates/ironrdp-connector/src/connection.rs
```diff
@@ -365,7 +366,7 @@ impl Sequence for ClientConnector {
                     warn!("Unexpected MultiTransportChannelData GCC block (not supported)");
                 }
 
-                let static_channel_ids = server_gcc_blocks.network.channel_ids;
+                let static_channel_ids = server_gcc_blocks.network.channel_ids.clone();
                 let io_channel_id = server_gcc_blocks.network.io_channel;
```
CBenoit, line 370: question: What is making this clone necessary?

crates/ironrdp-connector/src/connection.rs
```diff
@@ -347,9 +348,9 @@ impl Sequence for ClientConnector {
 
                 debug!(message = ?connect_response, "Received");
 
-                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks;
+                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks();
 
-                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks;
+                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks();
```
CBenoit: suggestion: Maybe add a `into_gcc_blocks` so we can retrieve the ownership of the internal data, and avoid cloning stuff 💡

crates/ironrdp-pdu/src/basic_output/fast_path.rs
CBenoit: Using `#[expect(clippy::as_conversions)]` for this specific kind of conversion is fine :+1:
For enum -> integer, `as` performs a value-preserving conversion of the discriminant.
Just make sure there is `#[repr(u8)]` when casting to `u8`, so it’s total, safe, and cannot produce an invalid value.
Indeed, there _is_ a pitfall, which is why we have this warning: silent narrowing, sign changes _if_ you cast to a different size/signedness (e.g., `#[repr(u16)]` but `as u8`). But if we encapsulate the logic at one place, like here, in a `as_u8` function, it’s perfectly satisfying :+1: 

Also add a reason such as `#[repr(u8)] guarantees discriminant layout, and as is the only way to cast enum -> primitive`

As a side note, the pitfall mentioned above is also _very_ unlikely, we could fall into it only for enums with more than 255 variants, and I don’t think we have even one such enum in the whole code base.

CBenoit: However, for the other way around (primitive -> enum), I agree that the match is the simplest, correct and good-enough approach.

crates/ironrdp-connector/src/connection.rs
RRRadicalEdward: Good suggestion. Done

crates/ironrdp-pdu/src/rdp/client_info.rs
RRRadicalEdward: Done

crates/ironrdp-pdu/src/gcc/conference_create.rs
RRRadicalEdward: Done

crates/ironrdp-pdu/src/gcc/conference_create.rs
RRRadicalEdward: Done

crates/ironrdp-pdu/src/gcc/conference_create.rs
RRRadicalEdward: Done

crates/ironrdp-pdu/src/basic_output/fast_path.rs
RRRadicalEdward: Done

crates/ironrdp-pdu/src/basic_output/fast_path.rs
```diff
@@ -328,6 +329,16 @@ pub enum UpdateCode {
     LargePointer = 0xc,
 }
 
+impl UpdateCode {
+    #[expect(
+        clippy::as_conversions,
+        reason = "guarantees discriminant layout, and as is the only way to cast enum -> primitive"
+    )]
+    pub fn as_u8(&self) -> u8 {
```
CBenoit: suggestion: Last small detail, but take it by value since it's a Copy type
```suggestion
    pub fn as_u8(self) -> u8 {
```

RRRadicalEdward: Done here and everything else

crates/ironrdp-connector/src/connection.rs
```diff
@@ -347,9 +348,9 @@ impl Sequence for ClientConnector {
 
                 debug!(message = ?connect_response, "Received");
 
-                let client_gcc_blocks = &connect_initial.conference_create_request.gcc_blocks;
+                let client_gcc_blocks = connect_initial.conference_create_request.gcc_blocks();
 
-                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks;
+                let server_gcc_blocks = connect_response.conference_create_response.gcc_blocks();
```
CBenoit:
```suggestion
                let server_gcc_blocks = connect_response.conference_create_response.into_gcc_blocks();
```

crates/ironrdp-connector/src/connection.rs
```diff
@@ -365,7 +366,7 @@ impl Sequence for ClientConnector {
                     warn!("Unexpected MultiTransportChannelData GCC block (not supported)");
                 }
 
-                let static_channel_ids = server_gcc_blocks.network.channel_ids;
+                let static_channel_ids = server_gcc_blocks.network.channel_ids.clone();
```
CBenoit:
```suggestion
                let static_channel_ids = server_gcc_blocks.network.channel_ids;
```

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -169,18 +190,33 @@ impl<'de> Decode<'de> for ConferenceCreateRequest {
...
+            "gcc blocks length",
+            gcc_blocks.size() + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE
+        )?;
+
+        Ok(Self { user_id, gcc_blocks })
+    }
+
+    pub fn gcc_blocks(&self) -> &ServerGccBlocks {
+        &self.gcc_blocks
+    }
```
CBenoit, line 225:
```suggestion
    }
    
    pub fn into_gcc_blocks(self) -> ServerGccBlocks {
        self.gcc_blocks
    }
```

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -169,18 +190,37 @@ impl<'de> Decode<'de> for ConferenceCreateRequest {
...
 
 impl ConferenceCreateResponse {
     const NAME: &'static str = "ConferenceCreateResponse";
+
+    pub fn new(user_id: u16, gcc_blocks: ServerGccBlocks) -> DecodeResult<Self> {
+        // INVARIANT: gcc_blocks.size() + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE <= u16::MAX
+        let _: usize = cast_length!(
+            "gcc blocks length",
+            gcc_blocks.size() + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE
+        )?;
```
CBenoit:
```suggestion
        // Ensure the invariant on gcc_blocks.size() is respected.
        if !(gcc_blocks.size() + CONFERENCE_RESPONSE_CONNECT_PDU_SIZE <= usize::from(u16::MAX)) {
            return Err(invalid_field_err!(
                "gcc_blocks",
                "gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE > u16::MAX"
            ));
        }
```

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -26,11 +26,32 @@ const CONFERENCE_NAME: &[u8] = b"1";
...
+    /// INVARIANT: `gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= u16::MAX`
+    gcc_blocks: ClientGccBlocks,
 }
 
 impl ConferenceCreateRequest {
     const NAME: &'static str = "ConferenceCreateRequest";
+
+    pub fn new(gcc_blocks: ClientGccBlocks) -> DecodeResult<Self> {
+        // Ensure the invariant on gcc_blocks.size() is respected
+        if gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE > usize::from(u16::MAX) {
```
CBenoit:
```suggestion
        // Ensure the invariant on gcc_blocks.size() is respected.
        if !(gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= usize::from(u16::MAX)) {
```

RRRadicalEdward: There was a warning if doing it this way, stating something like "expression can be simplified"

RRRadicalEdward: image

crates/ironrdp-pdu/src/rdp/suppress_output.rs
```diff
@@ -20,7 +20,10 @@ impl AllowDisplayUpdatesType {
             _ => None,
         }
     }
-
+    #[expect(
```
CBenoit, line 24:
```suggestion

    #[expect(
```

crates/ironrdp-pdu/src/gcc/conference_create.rs
```diff
@@ -26,11 +26,34 @@ const CONFERENCE_NAME: &[u8] = b"1";
...
+    pub fn new(gcc_blocks: ClientGccBlocks) -> DecodeResult<Self> {
+        // Ensure the invariant on gcc_blocks.size() is respected.
+        check_invariant(gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE <= usize::from(u16::MAX)).ok_or_else(
+            || {
+                invalid_field_err!(
+                    "gcc_blocks",
+                    "gcc_blocks.size() + CONFERENCE_REQUEST_CONNECT_PDU_SIZE > u16::MAX"
+                )
+            },
+        )?;
```
CBenoit: @RRRadicalEdward Here is how we can keep the invariant properly spelled in the check (https://github.com/Devolutions/IronRDP/pull/964#discussion_r2322718078)

CBenoit (approved):
LGTM! Thanks!!

PR #967, opened 2025-09-08
refactor(ironrdp-pdu)!: fix `as_conversions` clippy lint warnings

RRRadicalEdward:


crates/ironrdp-pdu/src/rdp/capability_sets/order.rs
```diff
@@ -9,6 +9,7 @@ const ORD_LEVEL_1_ORDERS: u16 = 1;
 const SUPPORT_ARRAY_LEN: usize = 32;
 const DESKTOP_SAVE_Y_GRAN_VAL: u16 = 20;
 
+#[repr(usize)]
```
CBenoit: issue: I think the real repr is u8, no?

For the array, you could use a helper or `usize::from(order.as_u8())`

RRRadicalEdward: Yeah. Make sense

RRRadicalEdward: Done

PR #971, opened 2025-09-09
fix(cliprdr): add missing std feature to ironrdp-code dependency

pacmancoder:
Fixes currently failing `Release crates` CI step

### DIFF
--- a/crates/ironrdp-cliprdr-format/Cargo.toml
+++ b/crates/ironrdp-cliprdr-format/Cargo.toml
@@ -16,9 +16,8 @@ doctest = false
 test = false
 
 [dependencies]
-ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
+ironrdp-core = { path = "../ironrdp-core", version = "0.1", features = ["std"] } # public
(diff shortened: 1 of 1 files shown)


crates/ironrdp-cliprdr-format/Cargo.toml
```diff
@@ -16,9 +16,8 @@ doctest = false
 test = false
 
 [dependencies]
-ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
+ironrdp-core = { path = "../ironrdp-core", version = "0.1", features = ["std"] } # public
```
pacmancoder, line 19: ironrdp-cliprdr-format can't be built in no_std currently (uses std::io:Cursor), so its fine to make ironrd-core/std non-optional

awakecoding (approved):

crates/ironrdp-cliprdr-format/Cargo.toml
CBenoit: Good catch!

PR #976, opened 2025-09-11
refactor: add `partial_pub_fields` clippy style and readability lint

RRRadicalEdward:
From the lint
[documentation](https://rust-lang.github.io/rust-clippy/master/index.html#/partial_pub_fields):

> Checks whether some but not all fields of a struct are public. Either make all fields of a type public, or make none of them public. 
> Most types should either be:
> * Abstract data types: complex objects with opaque implementation which guard interior invariants and expose intentionally limited API to the outside word.
> * Data: relatively simple objects which group a bunch of related attributes together, but have no invariants

The lint seems to be reasonable, but it adds some inconvenience, as we have to provide getters/setters. I'm not fully sure that we should add it. What do you think?

### DIFF
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -37,6 +37,7 @@ impl ConnectionActivationSequence {
         }
     }
 
+    /// Returns the current state as a district type, rather than `&dyn State` provided by [`Self::state`].
     pub fn connection_activation_state(&self) -> ConnectionActivationState {
         self.state
     }


RRRadicalEdward:
> I’m not sure either. I remember considering it, and ending up not enabling it, but I can’t remember exactly why.

Maybe we could roll with it, and then disable it in case we find it too inconvenient? I feel like this lint is good. I would rather have dedicated get/set methods than make some fields public, because the former provides a better way to control the data flow.

RRRadicalEdward:
> I’m not sure either. I remember considering it, and ending up not enabling it, but I can’t remember exactly why.

What about enabling it only for ironrdp-pdu crate? It seems you wanted to add it only this way

CBenoit:
> > > These are the Rust API Guidelines, and we could enable Clippy’s wrong_self_convention to enforce that.
> > 
> > 
> > That sounds great to me. I will add it to the lints list.
> 
> Seems like the lint is already abled by default - https://rust-lang.github.io/rust-clippy/master/index.html#/wrong_self_convention.

Indeed! I guess it does not suggest to use `into`, but rather suggest to not use it when it’s not appropriate :thinking:

crates/ironrdp-cliprdr/src/pdu/format_data/metafile.rs
```diff
@@ -38,11 +38,11 @@ bitflags! {
 /// NOTE: `Decode` implementation will read all remaining data in cursor as metafile contents.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PackedMetafile<'a> {
-    mapping_mode: PackedMetafileMappingMode,
-    x_ext: u32,
-    y_ext: u32,
+    pub mapping_mode: PackedMetafileMappingMode,
```
RRRadicalEdward, line 41: Noticed that we should make all fields public instead of providing getters, because this struct is PDU.

crates/ironrdp-connector/src/connection_activation.rs
```diff
@@ -37,6 +37,10 @@ impl ConnectionActivationSequence {
         }
     }
 
+    pub fn connection_activation_state(&self) -> ConnectionActivationState {
```
CBenoit, line 41: naming: `state()` is enough, given that the struct is called `ConnectionActivationSequence`.

CBenoit: Edited: remove `into` in my comment, because it’s not a consuming accessor, my bad

RRRadicalEdward: But, we don't consume the `self` here. Maybe, we could just have it as `state()`?

RRRadicalEdward: Done

RRRadicalEdward: Now, I remember why I named it as `connection_activation_state`:
image

RRRadicalEdward: Should we leave it as `connection_activation_state`? I can't think of a better name

CBenoit: Ah! I see. I can’t think of anything better either, let’s go with this then!

RRRadicalEdward: Done

crates/ironrdp-dvc/src/pdu.rs
```diff
@@ -447,6 +459,18 @@ impl DataPdu {
         }
     }
 
+    pub fn as_data(&self) -> &[u8] {
```
CBenoit: naming: This is not a conversion, we are accessing to an inner field, so data() would be more proper I think?

crates/ironrdp-mstsgu/src/proto.rs
```diff
@@ -66,7 +66,7 @@ impl TryFrom<u16> for PktTy {
 #[derive(Default, Debug)]
 pub(crate) struct PktHdr {
     pub ty: PktTy,
-    _reserved: u16,
+    pub _reserved: u16,
```
CBenoit, line 69: Not sure about that

crates/ironrdp-dvc/src/pdu.rs
RRRadicalEdward: Make sense

crates/ironrdp-mstsgu/src/proto.rs
RRRadicalEdward: The struct is PDU, which means that all fields should be public. But I do agree that it's not looking good

crates/ironrdp-dvc/src/pdu.rs
RRRadicalEdward: Done

CBenoit:
I let GitHub Copilot performs some double checks, I’ll merge afterwards!

crates/ironrdp-connector/src/connection_activation.rs
```diff
@@ -22,7 +22,7 @@ use crate::{
 /// [Server Deactivate All PDU]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/8a29971a-df3c-48da-add2-8ed9a05edc89
 #[derive(Debug, Clone)]
 pub struct ConnectionActivationSequence {
-    pub state: ConnectionActivationState,
+    state: ConnectionActivationState,
```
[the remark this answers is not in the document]
RRRadicalEdward: Done

CBenoit (approved):
LGTM!

PR #985, opened 2025-09-17
build(deps): replace opus by opus2

CBenoit:
opus is unmaintained and ponits to a 4-year old commit of the opus C library. This does not compile anymore on our CI, because their CMakeList.txt requires an older version of cmake that is not available in the runners we use. opus2 is a fork that points to a more recent version of it.


RRRadicalEdward:
I just saw that someone made a patch for libopus_sys - https://github.com/Lakelezz/audiopus_sys/pull/23 to fix CMake error. An alternative solution can be to use the patch instead of adding `opus2`.

RRRadicalEdward:
I managed to get it working with dynamic linking on macOS and Linux. But the issue on Windows is that no package manager (tried _choco_ and _scoop_) provides a way to install the opus DLL.  
Seems like other projects use prebuilt DLLs made by themself on Windows - https://github.com/discordrb/discordrb/wiki/Installing-libopus. 

I think the best solution right now would be to fallback to https://github.com/Devolutions/IronRDP/pull/985#issuecomment-3306923458 option, until https://github.com/cijiugechu/libopus_sys/issues/2 is addressed.

RRRadicalEdward:
> I think the best solution right now would be to fallback to https://github.com/Devolutions/IronRDP/pull/985#issuecomment-3306923458 option, until https://github.com/cijiugechu/libopus_sys/issues/2 is addressed.

https://github.com/Devolutions/IronRDP/pull/988

PR #987, opened 2025-09-17
fix(dvc-pipe-proxy): add blocking logic for sending dvc pipe messages

pacmancoder:


pacmancoder:
@CBenoit, This change is definitely controversial, but I did not find any easy fix to handle the channel backpressure inside the DVC processor. (Standard blocking `.send()` on `tokio::channel` is not allowed inside an async runtime (in which IronRDP is running), and try_send is not an option, because the channel fills up very fast when a lot of data is transferred.

I propose to use `block_in_place` for now, as it is critical to unblock further work on NowProto DVC support via IronRDP;

Some ideas for future implementation:
- Add `acknowledge` messages to NowProto to slow down transmission.
- Change DVC processor interface to allow reporting "blocked" DVC processor state to re-queue next DVC message processing

pacmancoder:
**UPD:**  Back to the drawing board.... tokio::task::block_in_place is definitely not what we want -- it will not work in FFI crate, there is no tokio 🥲

crates/ironrdp-dvc-pipe-proxy/src/worker.rs
```diff
@@ -132,21 +141,50 @@ async fn process_client<P: OsPipe>(ctx: &mut WorkerCtx) -> Result<NextWorkerStat
     }
 }
 
-async fn worker<P: OsPipe>(mut ctx: WorkerCtx) -> Result<(), DvcPipeProxyError> {
+async fn worker<P: OsPipe>(ctx: WorkerCtx) -> Result<(), DvcPipeProxyError> {
+    // Create a bridge between std::sync::mpsc and tokio for async compatibility.
+    let (async_tx, async_rx) = tokio::sync::mpsc::unbounded_channel();
```
CBenoit, line 149: suggestion: Add a comment about why it’s fine to use an unbounded channel here (= we are bounded by the to_pipe_rx channel).

PR #992, opened 2025-09-21
fix(web): fix `this.lastSentClipboardData` being null

RRRadicalEdward:
```js
await this.remoteDesktopService.onClipboardChanged(...)
```
Consumes the clipboard data we pass, so we need to clone the data to prevent `this.lastSentClipboardData` from being null.

### DIFF
--- a/crates/iron-remote-desktop/src/clipboard.rs
+++ b/crates/iron-remote-desktop/src/clipboard.rs
@@ -1,6 +1,6 @@
 use wasm_bindgen::JsValue;
 
-pub trait ClipboardData: Clone {
+pub trait ClipboardData {
     type Item: ClipboardItem;
 
     fn create() -> Self;
--- a/crates/iron-remote-desktop/src/lib.rs
+++ b/crates/iron-remote-desktop/src/lib.rs
@@ -160,7 +160,7 @@ macro_rules! make_bridge {
 
             #[wasm_bindgen(js_name = onClipboardPaste)]
             pub async fn on_clipboard_paste(&self, content: &ClipboardData) -> Result<(), IronError> {
-                $crate::Session::on_clipboard_paste(&self.0, content.0.clone())
+                $crate::Session::on_clipboard_paste(&self.0, &content.0)
                     .await
                     .map_err(IronError)
             }
--- a/web-client/iron-remote-desktop/src/services/clipboard.service.ts
+++ b/web-client/iron-remote-desktop/src/services/clipboard.service.ts
@@ -146,7 +146,6 @@ export class ClipboardService {
 
         if (!clipboardData.isEmpty()) {
             this.lastSentClipboardData = clipboardData;
-            // Explicitly clone the `clipboardData` object as `onClipboardChanged` consumes the value.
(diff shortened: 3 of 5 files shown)


CBenoit:
I think a better approach is to not consume the data in the callback, and to expose a more JavaScript friendly API in this case. How does that sound to you?

web-client/iron-remote-desktop/src/services/clipboard.service.ts
```diff
@@ -146,7 +146,7 @@ export class ClipboardService {
 
         if (!clipboardData.isEmpty()) {
             this.lastSentClipboardData = clipboardData;
-            // TODO(Fix): onClipboardChanged takes an ownership over clipboardData, so lastSentClipboardData will be nullptr.
+            // Explicitly clone the `clipboardData` object as `onClipboardChanged` consumes the value.
```
CBenoit: Verify leftover comments

crates/iron-remote-desktop/src/clipboard.rs
```diff
@@ -1,6 +1,6 @@
 use wasm_bindgen::JsValue;
 
-pub trait ClipboardData {
+pub trait ClipboardData: Clone {
```
CBenoit: suggestion: Do not require Clone at the trait level

crates/iron-remote-desktop/src/lib.rs
```diff
@@ -159,8 +159,8 @@ macro_rules! make_bridge {
             }
 
             #[wasm_bindgen(js_name = onClipboardPaste)]
-            pub async fn on_clipboard_paste(&self, content: ClipboardData) -> Result<(), IronError> {
-                $crate::Session::on_clipboard_paste(&self.0, content.0)
+            pub async fn on_clipboard_paste(&self, content: &ClipboardData) -> Result<(), IronError> {
+                $crate::Session::on_clipboard_paste(&self.0, content.0.clone())
```
CBenoit: At this place, pass a reference to on_clipboard_paste, and delegate the decision to the implementation itself (cloning the whole ClipboardData may not always be the best strategy)

RRRadicalEdward: Good suggestion

web-client/iron-remote-desktop/src/services/clipboard.service.ts
RRRadicalEdward: My bad. I will remove them

crates/iron-remote-desktop/src/clipboard.rs
RRRadicalEdward: Fixed

web-client/iron-remote-desktop/src/services/clipboard.service.ts
RRRadicalEdward: Cleaned up

crates/iron-remote-desktop/src/lib.rs
RRRadicalEdward: Done

PR #993, opened 2025-09-21
refactor(ironrdp-web): fix `indexing_slicing` clippy lint warnings

RRRadicalEdward:
[This lint](https://rust-lang.github.io/rust-clippy/master/index.html#/indexing_slicing)
> Checks for usage of indexing or slicing that may panic at runtime.

This one is quite pedantic and changes the way we write the code, but on the other hand, it makes the code more panic-resistant. Fixed the lint warnings only for _ironrdp-web_ to see early feedback on it. 
If we agree on adding it, I will add this lint for each crate separately, because there are too many warnings to do in 1 PR:
image

### DIFF
--- a/crates/ironrdp-web/src/canvas.rs
+++ b/crates/ironrdp-web/src/canvas.rs
@@ -45,10 +45,8 @@ impl Canvas {
         let region_height = region.height();
 
         let mut src = buffer.chunks_exact(4).map(|pixel| {
-            let r = *pixel.first().expect("index cannot be out of bounds");
-            let g = *pixel.get(1).expect("index cannot be out of bounds");
-            let b = *pixel.get(2).expect("index cannot be out of bounds");
-            u32::from_be_bytes([0, r, g, b])
+            let [r, g, b] = pixel.first_chunk::<3>().expect("cannot be out of bounds");
+            u32::from_be_bytes([0, *r, *g, *b])
         });
 
         let mut dst = self.surface.buffer_mut().expect("surface buffer");
--- a/crates/ironrdp-web/src/canvas.rs
+++ b/crates/ironrdp-web/src/canvas.rs
@@ -45,9 +45,9 @@ impl Canvas {
         let region_height = region.height();
 
         let mut src = buffer.chunks_exact(4).map(|pixel| {
-            let r = pixel[0];
-            let g = pixel[1];
-            let b = pixel[2];
+            let r = *pixel.first().expect("index cannot be out of bounds");
+            let g = *pixel.get(1).expect("index cannot be out of bounds");
+            let b = *pixel.get(2).expect("index cannot be out of bounds");
             u32::from_be_bytes([0, r, g, b])
         });
 
--- a/crates/ironrdp-web/src/image.rs
+++ b/crates/ironrdp-web/src/image.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::arithmetic_side_effects)]
 
-use anyhow::Context;
+use anyhow::Context as _;
 use ironrdp::pdu::geometry::{InclusiveRectangle, Rectangle as _};
 use ironrdp::session::image::DecodedImage;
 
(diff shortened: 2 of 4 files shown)


crates/ironrdp-web/src/canvas.rs
```diff
@@ -44,9 +44,9 @@ impl Canvas {
         let region_height = region.height();
 
         let mut src = buffer.chunks_exact(4).map(|pixel| {
-            let r = pixel[0];
-            let g = pixel[1];
-            let b = pixel[2];
+            let r = *pixel.first().expect("index cannot be out of bounds");
+            let g = *pixel.get(1).expect("index cannot be out of bounds");
+            let b = *pixel.get(2).expect("index cannot be out of bounds");
```
CBenoit: Looks a bit too verbose, although there may be a better way to write this with the new stabilized APIs

crates/ironrdp-web/src/image.rs
```diff
@@ -33,20 +40,31 @@ fn extract_smallest_rectangle(image: &DecodedImage, region: InclusiveRectangle)
...
         let src_begin = image_stride * (region_top + row) + region_left * pixel_size;
         let src_end = src_begin + region_stride;
-        let src_slice = &src[src_begin..src_end];
+        let src_slice = src.get(src_begin..src_end).with_context(|| {
+            format!(
+                "invalid region {region:?} for image with dimensions {}x{}",
+                image.width(),
+                image.height()
+            )
+        })?;
```
CBenoit, line 49: Preciously this was subject to panics upon bad input?

RRRadicalEdward: Yeah. I think it was possible to trigger panic if passing an empty `image` and any _valid_ region.

crates/ironrdp-web/src/canvas.rs
RRRadicalEdward: Do you have https://doc.rust-lang.org/std/primitive.slice.html#method.first_chunk in mind or something different? The perfect API here would be - https://doc.rust-lang.org/std/primitive.slice.html#method.as_array, but it's nightly only

CBenoit:
Sorry I actually forgot about it despite the TODO I wrote 😨

RRRadicalEdward:
> Sorry I actually forgot about it despite the TODO I wrote 😨

I just remembered about this one yesterday. We have a mental connect 😄

glamberson:
@RRRadicalEdward I took a look through this one. The fix itself looks solid, the slice-indexing to checked-access conversion is clean and well-scoped, and I don't see anything technically outstanding here. Looks like this has mostly just fallen through the cracks on review bandwidth rather than anything wrong with the change.

The branch is quite a few commits behind master at this point and will need a rebase before it can be merged. If you get it rebased, I'll advocate for getting this merged promptly.

CBenoit:
Thanks for exploring this, and sorry for the delay!

I’m leaning against enabling the lint workspace-wide, although some of the hardening in this PR is useful independently.

The distinction for me is whether a change actually improves error handling. Replacing source slicing with `.get(...).context(...)?` turns a potential panic into a propagated error. But replacing indexing with `.get_mut(...).expect(...)` mostly makes the same invariant more verbose. Similarly, the pixel accesses inside `chunks_exact(4)` are already guaranteed to be in bounds.

I do see an argument for stronger guardrails with the new AI-assisted development approach, but this lint is also easy to satisfy mechanically without improving correctness. We still need to review the bounds assumptions and decide whether an invalid input should produce an error or represents an internal invariant violation.

In that context, I think I would rather prioritize expanding fuzz coverage. Fuzzing isn’t a proof of panic freedom, but it can catch failures beyond indexing without imposing this much noise on otherwise straightforward code.

I would either close this PR as-is, or rework it to leave the already safe indexing alone, keeping only the genuinely improved error handling.

CBenoit:
I explored on my side: https://github.com/Devolutions/IronRDP/pull/1930
My conclusion is that it was not worth the hardening. Closing both PRs.

PR #996, opened 2025-09-23
fix(web): send first clipboard only after the connection is established

RRRadicalEdward:
When the clipboard is in automatic mode, `iron-remote-desktop` can send the first clipboard PDU to the server before the connection is established. Thus, this packet is ignored, and the clipboard on the server will not be synced with the current host's clipboard. 

This PR fixes it.


CBenoit (approved):
LGTM!

PR #1022, opened 2025-10-20
refactor: get rid of `lazy_static`

RRRadicalEdward:
The `lazy_static` was mostly used in tests. I think there were only 1 or 2 usages outside of the tests.

We still have it in the dependency tree because of _num-bigint-dig_ usage in _picky-asn1-x509_. I plan to remove _num-bigint-dig_  in _picky-asn1-x509_, as it's an old dependency and RustCrypto now uses _crypto-bigint_ instead.

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -371,7 +371,6 @@ dependencies = [
  "byteorder",
  "ironrdp-core",
  "ironrdp-pdu",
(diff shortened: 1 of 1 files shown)


RRRadicalEdward:
> Diff: +1.26%

🤩

CBenoit (approved):
LGTM! Thank you

PR #1028, opened 2025-11-04
build(deps): bump picky and sspi

### LINKED ISSUE #1029: Enabling connector feature gives build failures
I'm trying to use the `connector` feature like so:

```
ironrdp = { version = "0.13.0", default-features = true, features = ["connector"] }
```

But this leads to the following build errors:

```
the trait bound `rsa::RsaPrivateKey: TryFrom<&picky::key::PrivateKey>` is not satisfied
the following other types implement trait `From<T>`:
  `rsa::RsaPrivateKey` implements `From<BlindedSigningKey<D>>`
  `rsa::RsaPrivateKey` implements `From<rsa::pkcs1v15::SigningKey<D>>`
  `rsa::RsaPrivateKey` implements `From<rsa::pss::SigningKey<D>>`
required for `&picky::key::PrivateKey` to implement `Into<rsa::RsaPrivateKey>`
required for `rsa::RsaPrivateKey` to implement `TryFrom<&picky::key::PrivateKey>`
```

This looks similar to https://github.com/Devolutions/sspi-rs/issues/515, perhaps we need to update to the latest `sspi` crate in the `ironrdp` crate?

allan2:
Fixes #1029

This is a draft PR that uses currently uses git versions of picky and sspi.

It builds without any issues.

Waiting for:
1. new picky release
2. sspi to use the new picky versions
3. new sspi release
4. ironrdp to use the new sspi version

### DIFF
--- a/crates/ironrdp-connector/Cargo.toml
+++ b/crates/ironrdp-connector/Cargo.toml
@@ -33,7 +33,7 @@ rand = { version = "0.9", features = ["std"] } # TODO: dependency injection?
 tracing = { version = "0.1", features = ["log"] }
 picky-asn1-der = "0.5"
 picky-asn1-x509 = "0.15"
-picky = "=7.0.0-rc.20"
+picky = "=7.0.0-rc.20" # FIXME: We are pinning with = because the candidate version number counts as the minor number by Cargo, and will be automatically bumped in the Cargo.lock.
 
 [lints]
 workspace = true


crates/ironrdp-connector/src/credssp.rs
```diff
@@ -123,11 +123,11 @@ impl CredsspSequence {
                         certificate: cert,
                         reader_name: config.reader_name.clone(),
                         card_name: None,
-                        container_name: config.container_name.clone(),
+                        container_name: Some(config.container_name.clone()),
                         csp_name: config.csp_name.clone(),
                         pin: pin.as_bytes().to_vec().into(),
-                        private_key_file_index: None,
                         private_key: Some(key.into()),
+                        scard_type: sspi::SmartCardType::default(),
```
ymarcus93: Any reason we shouldn't use `SmartCardType::Emulated` here and pass in `pin`?

CBenoit: I think we should pass in `pin`

crates/ironrdp-connector/Cargo.toml
```diff
@@ -27,13 +27,13 @@ ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
...
-sspi = "0.16" # public
+sspi = { version = "0.18", features = ["scard"] }
 url = "2.5" # public
 rand = { version = "0.9", features = ["std"] } # TODO: dependency injection?
 tracing = { version = "0.1", features = ["log"] }
 picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky = "7.0.0-rc.17"
+picky-asn1-x509 = "0.15"
+picky = "7.0.0-rc.20"
```
CBenoit: suggestion: Not ideal, but to avoid breakages until picky 8 is released.
```suggestion
picky = "=7.0.0-rc.20"
```

CBenoit (approved):
LGTM.

Note: once the scard feature is fixed in sspi-rs to work again with WASM, we can look into cleaning up the feature flag.

PR #1031, opened 2025-11-16
fix(cliprdr)!: receiving a TemporaryDirectory PDU should not fail the svc

ymarcus93:
- Fixes the Cliprdr `SvcProcessor` impl. to support handling a `TemporaryDirectory` Clipboard PDU. 
- Removes `ClipboardError::UnimplementedPdu` since it is no longer used

Without this fix, I was unable to get the `ironrdp-server` crate to work with my custom`CliprdrBackend` impl. This change makes sense since this PDU is a valid clipboard PDU that we should process and is (optionally) sent by an RDP client. I've chosen not to add a callback in the `CliprdrBackend` trait since this PDU seems useless as there is nothing todo with it from the `ironrdp-server`'s perspective.


CBenoit (approved):
Looks reasonable. Thank you!

PR #1047, opened 2025-12-04
fix(cliprdr): prevent window class registration error on multiple sessions

thenextman:
When starting a second clipboard session, `RegisterClassA` would fail with `ERROR_CLASS_ALREADY_EXISTS` because window classes are global to the process. Now checks if the class is already registered before attempting registration, allowing multiple WinClipboard instances to coexist.

Also improved WinAPI error messages to include the actual error details for easier debugging.

### DIFF
--- a/crates/ironrdp-cliprdr-native/src/windows/mod.rs
+++ b/crates/ironrdp-cliprdr-native/src/windows/mod.rs
@@ -65,7 +65,7 @@ impl core::fmt::Display for WinCliprdrError {
             WinCliprdrError::Alloc => write!(f, "failed to allocate global memory"),
             WinCliprdrError::DataReceiveTimeout => write!(f, "failed to receive data from remote clipboard"),
             WinCliprdrError::RenderFormat => write!(f, "failed to render clipboard format"),
-            WinCliprdrError::WinAPI(error) => write!(f, "WinAPI error: {error}"),
+            WinCliprdrError::WinAPI(_error) => write!(f, "WinAPI error"),
         }
     }
 }


mamoreau-devolutions (approved):

crates/ironrdp-cliprdr-native/src/windows/mod.rs
```diff
@@ -64,7 +65,7 @@ impl core::fmt::Display for WinCliprdrError {
             WinCliprdrError::Alloc => write!(f, "failed to allocate global memory"),
             WinCliprdrError::DataReceiveTimeout => write!(f, "failed to receive data from remote clipboard"),
             WinCliprdrError::RenderFormat => write!(f, "failed to render clipboard format"),
-            WinCliprdrError::WinAPI(_error) => write!(f, "WinAPI error"),
+            WinCliprdrError::WinAPI(error) => write!(f, "WinAPI error: {error}"),
```
CBenoit: issue: This is intentional; the source error is already propagated via the `source()` method in `core::error::Error` implementation below

```suggestion
            WinCliprdrError::WinAPI(_error) => write!(f, "WinAPI error"),
```

CBenoit (approved):
LGTM

thenextman:
> > Also improved WinAPI error messages to include the actual error details for easier debugging.
> 
> As noted in [this comment](https://github.com/Devolutions/IronRDP/pull/1047#discussion_r2587867559), if you spotted a place where the error details are omitted, this must be fixed, but the right place for that is at the location where the error is reported.
> 
> For IronRDP errors, you can use the `.report()` method to obtain an error reporter which will properly display the error and all its sources.
> 
> Long explanation: [#525 (review)](https://github.com/Devolutions/IronRDP/pull/525#pullrequestreview-2231978013)

Ok, I understand; _but_ this probably needs to be fixed at a different level (that I don't fully grok right now).

My perspective is consuming the dotnet nuget package; and the `IronRdpException` (I think that's right) seems to simply be doing something like a `.ToString()` on the error. So, when I hit this bug (and this is not a bug that was hit until we pulled this code into RDM, where multiple sessions in the same process are a thing) all I got back from IronRDP was "WinAPI Error". Not very helpful 😢

I would guess there is work to do in the FFI crate to properly push the error context into the IronRdp exception. 

Further comment that, in this case, once I surfaced the actual error it was super obvious what was going on because there's really only one place in the code that could happen. But otherwise, context on where the error actually happened inside IronRDP would be extremely helpful. I imagine we will find more issues inside RDM in the future simply because we're spinning up multiple sessions which isn't something really done elsewhere (at least, with the native implementation, I don't know about web).

CBenoit:
@thenextman I opened a PR to address the problem at the right place: https://github.com/Devolutions/IronRDP/pull/1050

It’s best if we publish a new version of the NuGet with this change for simpler troubleshooting.

PR #1054, opened 2025-12-15
feat(ironrdp-tls)!: return x509_cert::Certificate from upgrade()

willwwarner:
This allows client applications to verify details of the certificate, possibly with the user, when connecting to a server using TLS.

Note that this does not use Rustls's certificate verification, and still gives it danger::NoCertificateVerification.

### DIFF
--- a/crates/ironrdp-tls/src/lib.rs
+++ b/crates/ironrdp-tls/src/lib.rs
@@ -25,21 +25,9 @@ compile_error!("a TLS backend must be selected by enabling a single feature out 
 #[cfg(any(feature = "stub", feature = "native-tls", feature = "rustls"))]
 pub use impl_::{upgrade, TlsStream};
 
-#[cfg(any(feature = "native-tls", feature = "rustls"))]
-pub(crate) fn extract_tls_server_public_key(cert: &[u8]) -> std::io::Result<Vec<u8>> {
-    use std::io;
-
-    use x509_cert::der::Decode as _;
-
-    let cert = x509_cert::Certificate::from_der(cert).map_err(io::Error::other)?;
-
-    let server_public_key = cert
-        .tbs_certificate
+pub fn extract_tls_server_public_key(cert: &x509_cert::Certificate) -> Option<&[u8]> {
+    cert.tbs_certificate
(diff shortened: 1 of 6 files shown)


CBenoit (commented):
Thank you for the PR!

My preference would go for parsing once and returning the certificate as-is, so that the consumer can decide exactly what to extract from it, and it does not impose the extra dependency on sha2 unless absolutely required.

```rust
pub async fn upgrade<S>(stream: S, server_name: &str) -> io::Result<(TlsStream<S>, x509_cert::Certificate)> { … }
```

Also, we can expose the helper for extracting the public key (without the `.to_owned()`:

```rust
pub fn extract_public_key(cert: &x509_cert::Certificate) -> Option<&[u8]> {
    cert
        .tbs_certificate
        .subject_public_key_info
        .subject_public_key
        .as_bytes()
}
```

But it’s almost useless at this point.

Finally, run `cargo xtask ci` and fix all the problems it reports.

willwwarner:
> Finally, run `cargo xtask ci` and fix all the problems it reports.

There was a formatting issue in ironrdp-connector/credssp.rs, but I am not sure if you want that to be fixed here or not.

CBenoit:
Also, can you update the Cargo.toml file to add `# public` on the right of `x509-cert`? With this change, the `x509-cert` crate is also part of the public interface

crates/ironrdp-tls/src/lib.rs
```diff
@@ -26,20 +26,9 @@ compile_error!("a TLS backend must be selected by enabling a single feature out
 pub use impl_::{upgrade, TlsStream};
 
 #[cfg(any(feature = "native-tls", feature = "rustls"))]
```
CBenoit: Also remove this line, it was there just to suppress the dead code warning when no backend was enabled, but now it’s part of the public API

CBenoit (approved):
LGTM! Thanks!

CBenoit:
> ```rust
> use sspi::Secret
> use sspi::Username;
> ```
> 
> but it looks like CI doesn't care :)

I wonder why this does not get formatted :thinking:

PR #1069, opened 2026-01-09
refactor: temporarily restore public Error fields for crate publishing

CBenoit:
Makes Error::context and Error::kind public again to fix a breaking change that wasn't properly reported. This is a temporary workaround to allow crate publishing to proceed.

### DIFF
--- a/crates/ironrdp-connector/src/lib.rs
+++ b/crates/ironrdp-connector/src/lib.rs
@@ -406,7 +406,8 @@ pub trait ConnectorResultExt {
 impl<T> ConnectorResultExt for ConnectorResult<T> {
     fn with_context(self, context: &'static str) -> Self {
         self.map_err(|mut e| {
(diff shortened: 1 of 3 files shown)


crates/ironrdp-connector/src/lib.rs
```diff
@@ -406,7 +406,8 @@ pub trait ConnectorResultExt {
 impl<T> ConnectorResultExt for ConnectorResult<T> {
     fn with_context(self, context: &'static str) -> Self {
         self.map_err(|mut e| {
-            e.set_context(context);
+            e.context = context;
+            // e.set_context(context);
```
vnikonov-devolutions, line 410: leftover comment?

CBenoit: Yes, but I’ll revert this commit immediately when the crates are published :+1:

Issue #140, opened 2023-05-16
Support for the RDP Multitransport Layer


CBenoit:
[TODO comment in code](https://github.com/Devolutions/IronRDP/blob/f5a6ac7751dce6125f8faada03b288ab600035bd/crates/ironrdp-connector/src/connection.rs#L871).

> clientMultitransportChannelData (8 bytes): Optional [Client Multitransport Channel Data](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/3801236b-b5ba-4b6e-bf0d-afbde1fe391c) structure (section 2.2.1.3.8). This field MUST NOT be included if the server did not advertise support for Extended Client Data Blocks by using the EXTENDED_CLIENT_DATA_SUPPORTED flag (0x00000001) as described in section 2.2.1.2.1.

[Source](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/db6713ee-1c0e-4064-a3b3-0fac30b4037b)

mamoreau-devolutions:
@glamberson I just saw your comments on https://github.com/Devolutions/IronRDP/pull/1583 and https://github.com/Devolutions/IronRDP/pull/1584 - I didn't know about your prior implementation, but I'd be more interested in getting yours merged upstream piece by piece instead of what I currently have on my own branches. The reason is that you've already gone through the pain of testing your implementation against real clients. I'm sorry it slipped through the cracks - as you've noticed I've become more active in the last week, trying to accelerate development in various big gaps of IronRDP. I hadn't been paying attention closely in a while, but I'll be more active in coming months.

So, yes, please rebase your UDP implementation, then start splitting it off into smaller PRs we can independently review one by one. As for the two PRs I still have opened, tell me if you'd rather have me get those changes fixed based on your comments, or if you'd rather have me close them so we can just start reviewing your UDP pull requests instead.

glamberson:
Ok I'll work on it a little later. It's been about 9 months since I worked on it but I'll get it in shape. Thanks.

glamberson:
@mamoreau-devolutions Rebased and ready to start splitting it up.

**On your question: I'd close #1583 and keep #1584.**

[...]

PR #1093, opened 2026-02-10
feat(svc): add SvcMessage::encode_unframed_pdu for headerless encoding

glamberson:
Add encode_raw() method that encodes the inner PDU without SVC channel headers.
Over UDP, DVC messages are sent as RDPEMT tunnel data without SVC framing, but
the pdu field on SvcMessage is private so there's no way to get the raw bytes.

See #140 for context.

### DIFF
--- a/crates/ironrdp-svc/src/lib.rs
+++ b/crates/ironrdp-svc/src/lib.rs
@@ -99,11 +99,11 @@ impl SvcMessage {
         self
     }
 
-    /// Encodes the inner PDU without SVC channel headers.
+    /// Encodes the inner PDU without SVC channel framing headers.
     ///
     /// Returns the raw PDU bytes for transports that handle their own framing
     /// (e.g. RDPEMT tunnel data for UDP transport).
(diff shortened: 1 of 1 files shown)


crates/ironrdp-svc/src/lib.rs
```diff
@@ -98,6 +98,14 @@ impl SvcMessage {
         self.flags |= flags;
         self
     }
+
+    /// Encodes the inner PDU without SVC channel headers.
+    ///
+    /// Returns the raw PDU bytes for transports that handle their own framing
+    /// (e.g. RDPEMT tunnel data for UDP transport).
+    pub fn encode_raw(&self) -> EncodeResult<Vec<u8>> {
```
CBenoit: naming: I think `encode_unframed_pdu` would be good to make the lack of channel framing explicit in the name, how does that sound?

glamberson: Good call, renamed to encode_unframed_pdu and updated the doc comment to reference framing headers. Pushed.

PR #1091, opened 2026-02-10
feat(pdu): add Initiate Multitransport Request/Response PDU types

glamberson:
Add MultitransportRequestPdu and MultitransportResponsePdu types for the
sideband UDP transport bootstrapping PDUs defined in [MS-RDPBCGR] 2.2.15.1
and 2.2.15.2. Needed to decode/encode the IO channel messages that initiate
UDP transport setup.

See #140 for context.

PR #1098, opened 2026-02-12
feat(connector)!: implement multitransport bootstrapping handshake

glamberson:
## Summary

Makes the `MultitransportBootstrapping` state functional instead of a no-op
pass-through. After licensing the server may send 0, 1, or 2 Initiate
Multitransport Request PDUs before capabilities exchange. Each one is surfaced
to the application, which establishes UDP transport (RDPEUDP2 + TLS + RDPEMT)
or declines, and the connector reports the outcome back to the server.

## API

Mirrors the existing `should_perform_X()` pause-point pattern used by TLS
upgrade and CredSSP, but uses `complete_X()` / `skip_X()` rather than
`mark_X_as_done()` because completion carries result data:

- `should_perform_multitransport()`: true while a request awaits an outcome
- `multitransport_request()`: the request awaiting an outcome, or `None`
- `complete_multitransport(result, output)`: report the outcome, resume
- `skip_multitransport(output)`: decline, resume

`complete_multitransport` accepts a `MultitransportResult` (a `Success` /
`Failure(hresult)` enum) rather than a caller-built response PDU. The connector
builds the response internally from the stored request ID.

Requests are surfaced one at a time rather than as a batch. There is no end
marker for the set, and MS-RDPBCGR 3.2.5.15.1 requires the client to act on a
request as soon as it decodes one, so waiting to learn how many are coming is
not an option the protocol offers. `should_perform_multitransport()` can
therefore come round twice; the caller answers reliable and lossy separately.

## Approach

**Routing.** Requests arrive on the negotiated MCS message channel
(2.2.15.1) and the Demand Active on the I/O channel, so the channel decides
which is which. The message channel also carries NetworkAutoDetect since #1348,
so a decode still confirms what arrived there, but the I/O channel is never
speculatively decoded as multitransport. A PDU on neither channel is an error.

For the decode to be a sound confirmation the request decoder must reject a
Demand Active, so this PR also tightens `MultitransportRequestPdu` to require
the exact `SEC_TRANSPORT_REQ` security-header flag.

**Yielding.** Each request is surfaced the moment it decodes. Responding
returns the connector to `MultitransportBootstrapping` to read whatever comes
next, which may be a second request or the Demand Active. Nothing is buffered
and nothing is replayed: when the request is surfaced the Demand Active has not
arrived yet.

**Soft-Sync.** The Initiate Multitransport Response is the Soft-Sync
signalling path (2.2.15.2), permitted only when both peers advertised
`SOFTSYNC_TCP_TO_UDP` in their GCC `MultiTransportChannelData`. The server's
block is retained from the GCC exchange and checked against the client's
configured flags. One rule covers both paths:

- Soft-Sync negotiated: always respond, `S_OK` or `E_ABORT`, including on
  `skip_multitransport()`, which 3.2.5.15.1 requires. Both the async and
  blocking drivers skip automatically, so without this every default client
  leaves a compliant server waiting.
- Not negotiated: never respond. The outcome is reported in band on the new
  transport, and putting anything on the main channel would be the violation.

The response goes on the message channel per 2.2.15.2 and 3.2.5.15.2. If
Soft-Sync was negotiated but no message channel exists the connector errors
rather than falling back to the I/O channel, and that check runs before the
pending state is taken, so the caller is left with a connector it can still
inspect or decline from.

## Wire behaviour

On the wire TCP and UDP negotiation happen in parallel: the UDP transport is
established alongside the ongoing TCP handshake, and its completion signals the
dynamic-channel layer that subsequent channels may migrate to UDP. The
connector's API yield point here is a Rust affordance, not a spec-mandated TCP
pause. Thanks to @hardening for the correction.

## Tests

Connector state-machine tests in `ironrdp-testsuite-core` drive the public API
with the shared `SERVER_DEMAND_ACTIVE` fixture:

- a request is surfaced on arrival, without waiting for a following PDU
  (regression test for the stall);
- responding returns to bootstrapping so a second request is read normally;
- a third request is rejected per the 2.2.15.1 cap;
- a Demand Active on the I/O channel ends bootstrapping;
- the response targets the message channel, decoded back off the wire;
- a `Failure` result is carried through;
- `skip` sends `E_ABORT` under Soft-Sync, and nothing without it;
- `complete` emits nothing without Soft-Sync but still resumes;
- a failed response leaves the connector in `MultitransportPending`, still able
  to report or decline, rather than `Consumed`;
- `complete` / `skip` outside `MultitransportPending` error;
- a Demand Active's user data does not decode as a `MultitransportRequestPdu`
  (regression test for the decoder tightening above).

Fuzzing: `MultitransportRequestPdu` and `MultitransportResponsePdu` are added
to the `pdu_decode` and `pdu_round_trip` oracles, so the decode surface is
fuzzed against arbitrary input (auto-discovered into the fuzz CI matrix). Both
targets smoke-run clean.

Full end-to-end coverage against a live server still awaits server-side
`MultitransportRequestPdu` emission, which `ironrdp::server::RdpServer` does
not implement yet.

## Breaking changes

`cargo semver-checks` reports three, so the title carries the marker:

- `ClientConnector` gains the public field `server_multitransport_flags`, so struct literals need updating (`constructible_struct_adds_field`).
- `ClientConnectorState::MultitransportBootstrapping` gains `message_channel_id` and `requests_seen` (`enum_struct_variant_field_added`).
- Three `ClientConnectorState` discriminants shift, because the new variants are inserted mid-enum (`enum_no_repr_variant_discriminant_changed`).

The third I would call a false positive worth naming rather than designing around: `ClientConnectorState` is `#[non_exhaustive]`, has no `repr`, and every variant carries fields, so there is no `as` cast a downstream crate could have depended on. The first two are real.

## Verification

## References

Builds on: #1091
Related: #140, #1348

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -19,6 +19,11 @@ use crate::{
     NegotiationFailure, Sequence, State, Written, encode_x224_packet, general_err, reason_err,
 };
 
+/// Maximum number of `Initiate Multitransport Request` PDUs the server is
+/// permitted to send during bootstrapping, per MS-RDPBCGR 2.2.15.1 (one per
+/// transport protocol: reliable + lossy UDP).
+const MAX_MULTITRANSPORT_REQUESTS: usize = 2;
+
 /// Outcome of a single multitransport bootstrapping request, passed to
 /// [`ClientConnector::complete_multitransport()`].
 ///
@@ -418,7 +423,7 @@ fn advance_licensing_exchange(
         ClientConnectorState::MultitransportBootstrapping {
             io_channel_id,
             user_channel_id,
-            requests: Vec::new(),
+            requests: Vec::with_capacity(MAX_MULTITRANSPORT_REQUESTS),
         }
     } else {
         ClientConnectorState::LicensingExchange {
@@ -832,10 +837,10 @@ impl Sequence for ClientConnector {
             //== Optional Multitransport Bootstrapping ==//
             //
             // The server may send 0, 1, or 2 Initiate Multitransport Request PDUs
-            // after licensing. We distinguish them from the Demand Active PDU by
-            // attempting to decode as MultitransportRequestPdu first — it has a
-            // distinctive structure (SEC_TRANSPORT_REQ flag + request_id + protocol
-            // + cookie). If decode fails, this is the Demand Active.
+            // after licensing (MS-RDPBCGR 2.2.15.1). We distinguish them from the
+            // Demand Active PDU by attempting to decode as MultitransportRequestPdu
+            // first; the decoder validates SEC_TRANSPORT_REQ internally, so a Demand
+            // Active PDU fails cleanly without false positives.
             ClientConnectorState::MultitransportBootstrapping {
                 io_channel_id,
                 user_channel_id,
@@ -843,11 +848,17 @@ impl Sequence for ClientConnector {
             } => {
                 let ctx = mcs::decode_send_data_indication(input).map_err(ConnectorError::decode)?;
 
-                // Try decoding as a multitransport request. The decoder validates
-                // the SEC_TRANSPORT_REQ flag, so a Demand Active PDU will fail
-                // cleanly without false positives.
                 match decode::<rdp::multitransport::MultitransportRequestPdu>(ctx.user_data) {
                     Ok(pdu) => {
+                        if requests.len() >= MAX_MULTITRANSPORT_REQUESTS {
+                            return Err(reason_err!(
+                                "MultitransportBootstrapping",
+                                "server sent more than {} multitransport requests (MS-RDPBCGR 2.2.15.1 caps the count at {})",
+                                MAX_MULTITRANSPORT_REQUESTS,
+                                MAX_MULTITRANSPORT_REQUESTS,
+                            ));
+                        }
+
                         debug!(
                             request_id = pdu.request_id,
                             protocol = ?pdu.requested_protocol,
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -7,7 +7,7 @@ use ironrdp_core::{Encode, WriteBuf, decode, encode_vec};
 use ironrdp_pdu::x224::X224;
 use ironrdp_pdu::{PduHint, gcc, mcs, nego, rdp};
 use ironrdp_svc::{StaticChannelSet, StaticVirtualChannel, SvcClientProcessor};
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info};
 
 use crate::channel_connection::{ChannelConnectionSequence, ChannelConnectionState};
 use crate::connection_activation::{
@@ -27,8 +27,10 @@ const MAX_MULTITRANSPORT_REQUESTS: usize = 2;
 /// Outcome of a single multitransport bootstrapping request, passed to
 /// [`ClientConnector::complete_multitransport()`].
 ///
-/// The connector uses this to build the response PDU internally, paired with
-/// the request ID and cookie from the server's original request.
+/// The connector uses this to build the response PDU internally, carrying the
+/// request ID from the server's original request. The request's 16-byte
+/// security cookie is not echoed here: it binds the UDP transport itself, not
+/// the main-channel response.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum MultitransportResult {
     /// UDP transport was established successfully (`S_OK`).
@@ -109,27 +111,37 @@ pub enum ClientConnectorState {
         user_channel_id: u16,
         license_exchange: LicenseExchangeSequence,
     },
-    /// Reading the server's optional Initiate Multitransport Request PDU(s).
+    /// Waiting for either an Initiate Multitransport Request or the Demand
+    /// Active that ends the optional bootstrapping phase.
     ///
-    /// The server may send 0, 1, or 2 requests (one per transport protocol).
-    /// If the first PDU on the IO channel after licensing is a Demand Active
-    /// (capabilities exchange), the server sent no multitransport requests and
-    /// the connector transitions directly to `CapabilitiesExchange`.
+    /// The server may send 0, 1, or 2 requests, and there is no end marker for
+    /// the set: it is not obliged to send Demand Active before the client acts
+    /// on a request it has already sent. So each request is surfaced to the
+    /// caller the moment it decodes, and the connector comes back here for the
+    /// next PDU rather than trying to collect a batch it cannot know the size
+    /// of. A PDU on the I/O channel is the Demand Active and ends the phase.
     MultitransportBootstrapping {
         io_channel_id: u16,
         user_channel_id: u16,
-        /// Multitransport requests received from the server so far.
-        requests: Vec<rdp::multitransport::MultitransportRequestPdu>,
+        /// MCS message channel negotiated during GCC, when the server offered
+        /// one. Multitransport travels on it in both directions per
+        /// MS-RDPBCGR 2.2.15.1 and 2.2.15.2, so it is both the channel inbound
+        /// requests arrive on and the channel responses go out on. `None` means
+        /// the server never offered one, in which case no request can be valid.
+        message_channel_id: Option<u16>,
+        /// How many requests have been surfaced so far, to enforce the cap in
+        /// MS-RDPBCGR 2.2.15.1 now that they are handled one at a time.
+        requests_seen: usize,
     },
-    /// State the connector enters after the server has sent multitransport
-    /// request(s) and before resuming with the Demand Active PDU. The
-    /// connector surfaces this as an API yield point so the caller can run
-    /// UDP transport setup and report the outcome.
+    /// A single Initiate Multitransport Request has been surfaced and the
+    /// connector is waiting for the caller to establish that UDP transport
+    /// (RDPEUDP2 + TLS + RDPEMT) or decline it.
     ///
     /// Call [`ClientConnector::complete_multitransport()`] or
-    /// [`ClientConnector::skip_multitransport()`] to advance. The buffered
-    /// Demand Active PDU is replayed internally on resume; the caller does
-    /// not need to re-feed it.
+    /// [`ClientConnector::skip_multitransport()`] to advance. Either returns
+    /// the connector to [`Self::MultitransportBootstrapping`] to read whatever
+    /// the server sends next, which may be a second request or the Demand
+    /// Active.
     ///
     /// On the wire, TCP and UDP negotiation happen in parallel: the UDP
     /// transport is established alongside the ongoing TCP handshake, and
@@ -139,11 +151,18 @@ pub enum ClientConnectorState {
     MultitransportPending {
         io_channel_id: u16,
         user_channel_id: u16,
-        requests: Vec<rdp::multitransport::MultitransportRequestPdu>,
-        /// The raw Demand Active PDU bytes that arrived after the last
-        /// multitransport request. Replayed through the activation sequence
-        /// when the application completes or skips multitransport.
-        buffered_demand_active: Vec<u8>,
+        /// MCS message channel the Initiate Multitransport Response must go out
+        /// on; see the same field on [`Self::MultitransportBootstrapping`].
+        message_channel_id: Option<u16>,
+        /// The request awaiting the caller's outcome.
+        request: rdp::multitransport::MultitransportRequestPdu,
+        /// Carried through so the cap survives the round trip through the
+        /// caller.
+        requests_seen: usize,
+        /// Whether both peers advertised Soft-Sync, captured when this state was
+        /// entered. Decides whether the completion and skip paths may emit an
+        /// Initiate Multitransport Response at all.
+        soft_sync: bool,
     },
     CapabilitiesExchange {
         connection_activation: ConnectionActivationSequence,
@@ -200,6 +219,11 @@ pub struct ClientConnector {
     pub static_channels: StaticChannelSet,
     /// MCS message channel ID assigned by the server, once negotiated.
     pub message_channel_id: Option<u16>,
+    /// Multitransport flags the server advertised in its GCC
+    /// `MultiTransportChannelData` block, if it sent one. Retained because the
+    /// Initiate Multitransport Response is only permitted once both peers have
+    /// advertised Soft-Sync; see [`Self::soft_sync_negotiated`].
+    pub server_multitransport_flags: Option<gcc::MultiTransportFlags>,
 }
 
 impl ClientConnector {
@@ -210,7 +234,26 @@ impl ClientConnector {
             client_addr,
             static_channels: StaticChannelSet::new(),
             message_channel_id: None,
+            server_multitransport_flags: None,
         }
+    }
+
+    /// Whether Soft-Sync (`SOFTSYNC_TCP_TO_UDP`) was mutually advertised.
+    ///
+    /// Per [\[MS-RDPBCGR\] 2.2.15.2] the Initiate Multitransport Response is
+    /// the Soft-Sync signalling path, and is only sent when *both* peers
+    /// advertised the flag in their GCC `MultiTransportChannelData` block.
+    /// Without Soft-Sync the client reports the outcome in band on the new
+    /// transport and sends nothing back on the main channel, so emitting a
+    /// response there would be a protocol violation rather than a courtesy.
+    ///
+    /// [\[MS-RDPBCGR\] 2.2.15.2]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/44044233-e498-46f8-8e16-1ffa595a8e8b
+    fn soft_sync_negotiated(&self) -> bool {
+        fn advertised(flags: Option<gcc::MultiTransportFlags>) -> bool {
+            flags.is_some_and(|flags| flags.contains(gcc::MultiTransportFlags::SOFT_SYNC_TCP_TO_UDP))
+        }
+
+        advertised(self.config.multitransport_flags) && advertised(self.server_multitransport_flags)
     }
 
     #[must_use]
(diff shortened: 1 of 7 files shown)


glamberson:
I really need feedback on this so I'll take the draft status off and see where we get.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -201,6 +221,117 @@ impl ClientConnector {
...
+    ) -> ConnectorResult<Written> {
+        assert!(self.should_perform_multitransport());
+
+        let ClientConnectorState::MultitransportPending {
+            io_channel_id,
+            user_channel_id,
+            ..
+        } = self.state
+        else {
+            unreachable!()
```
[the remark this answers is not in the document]
glamberson: Fixed. Both methods now use mem::replace(&mut self.state, Consumed) to take ownership cleanly, matching the pattern used by step().

crates/ironrdp-connector/src/connection.rs
```diff
@@ -201,6 +221,117 @@ impl ClientConnector {
...
+    pub fn skip_multitransport(&mut self) {
+        assert!(self.should_perform_multitransport());
+
+        let ClientConnectorState::MultitransportPending {
+            io_channel_id,
+            user_channel_id,
+            ..
+        } = self.state
+        else {
+            unreachable!()
```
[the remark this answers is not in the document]
glamberson: Fixed alongside thread 1 with same mem::replace approach.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -535,20 +668,99 @@ impl Sequence for ClientConnector {
...
+                let is_multitransport = ctx.user_data.len() >= 2 && {
+                    let flags_raw = u16::from_le_bytes([ctx.user_data[0], ctx.user_data[1]]);
+                    let flags = BasicSecurityHeaderFlags::from_bits_truncate(flags_raw);
+                    flags.contains(BasicSecurityHeaderFlags::TRANSPORT_REQ)
+                };
+
+                if is_multitransport {
+                    let pdu = decode::<rdp::multitransport::MultitransportRequestPdu>(ctx.user_data)
+                        .map_err(ConnectorError::decode)?;
+
```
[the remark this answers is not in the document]
glamberson: Reworked. Detection now tries decode::<MultitransportRequestPdu>() first — the decoder validates the SEC_TRANSPORT_REQ flag internally, so a Demand Active PDU fails cleanly without false positives. No more flag-peeking.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -201,6 +221,117 @@ impl ClientConnector {
...
+        else {
+            unreachable!()
+        };
+
+        let mut total_written = 0;
+
+        for response in responses {
+            let written = encode_send_data_request(user_channel_id, io_channel_id, response, output)?;
+            total_written += written;
+        }
```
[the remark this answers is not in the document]
glamberson: Reworked. The caller now passes &[MultitransportResult] (a Success/Failure(hresult) enum) instead of building response PDUs manually. The connector validates results.len() == requests.len() and builds the response PDUs internally using the stored request IDs.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -535,20 +668,99 @@ impl Sequence for ClientConnector {
...
+                    }
+                }
+            }
+
+            // MultitransportPending: application should call complete_multitransport()
+            // or skip_multitransport() instead of step()
+            ClientConnectorState::MultitransportPending { .. } => {
+                return Err(general_err!(
+                    "multitransport pending: call complete_multitransport() or skip_multitransport()"
+                ));
```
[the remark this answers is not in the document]
glamberson: Reworked. The connector now buffers the Demand Active PDU internally in MultitransportPending and replays it through ConnectionActivationSequence::step() when complete_multitransport() or skip_multitransport() is called. Existing connect_finalize loops in ironrdp-async/blocking work unchanged — no multitransport requests from server means the MultitransportBootstrapping state forwards the Demand Active directly, so no regression.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -72,9 +86,32 @@ pub enum ClientConnectorState {
...
+    /// (capabilities exchange), the server sent no multitransport requests and
+    /// the connector transitions directly to `CapabilitiesExchange`.
     MultitransportBootstrapping {
         io_channel_id: u16,
         user_channel_id: u16,
+        /// Multitransport requests received from the server so far.
+        requests: Vec<rdp::multitransport::MultitransportRequestPdu>,
+    },
+    /// The server sent multitransport request(s) and the connector is paused
+    /// waiting for the application to establish UDP transport or decline.
```
hardening, line 102: I know that the spec is vague about that, but that statement is wrong. With recent servers / client there's no pause until the UDP transport is established, when you look at captures with mstsc the RDPUDP stack is established in parallel of the TCP negotiation. And basically when the RDPUDP transport is established it's just a signal that the dynamic channel can consider moving the channels to UDP...

glamberson:
@CBenoit, polite ping on this one. The option-b reshape (connector-internal Demand Active replay, MultitransportResult enum, try-decode detection, mem::replace for state extraction) has been in place since 2026-02-23. Just rebased on master to keep CI fresh; happy to address any new review whenever you have time.

crates/ironrdp-connector/src/connection.rs
glamberson: Apologies for the slow turnaround on this. I appreciate your thoughtful read and correction; if I hadn't been in the middle of a move and a bunch of other things I'd have been more responsive. Rewrote the doc-comment per your point. State machine semantics unchanged.

Your FreeRDP UDP implementation is the closest open-source reference for what mstsc actually does on the wire, so reviews like this are exactly what catches drift between spec-flavored prose and actual behavior.

glamberson:
Rebased onto current master. Force-pushed \`eee839fa\` -> \`e1d03960\`.

20 days behind master since the 2026-05-19 doc-comment fix; rebase was clean (no conflicts, despite master having picked up #1254 / #1250 / #1273 / #1272 in the connector vicinity). David Fort's review thread (\`discussion_r3269401135\`) remains resolved.

All five xtask gates green on the new head:
- \`cargo xtask check fmt -v\`
- \`cargo xtask check lints -v\` (workspace, all-targets)
- \`cargo xtask check tests -v\`
- \`cargo xtask check typos -v\`
- \`cargo xtask check locks -v\`

glamberson:
Per improved internal prechecks I'm running on my PRs, took the following tightening actions:

- Cap incoming multitransport requests at the MS-RDPBCGR 2.2.15.1 limit of 2 with a clean `ConnectorError` on overflow.
- Drivers (`ironrdp-async`, `ironrdp-blocking`) `connect_finalize` now auto-skip multitransport instead of aborting when a server advertises it.
- Rewrote the PR body for accuracy on the `complete_X`/`skip_X` naming and the new driver behavior.

glamberson:
I just refreshed this and added testing, fuzzing, and rebased it for whenever it bubbles up to the surface.

glamberson:
@CBenoit This PR is creeping up particularly due to other contributors' requirements, so perhaps this needs to get done.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -211,6 +263,141 @@ impl ClientConnector {
...
+                }
+                MultitransportResult::Failure(hr) => rdp::multitransport::MultitransportResponsePdu {
+                    security_header: rdp::headers::BasicSecurityHeader {
+                        flags: rdp::headers::BasicSecurityHeaderFlags::TRANSPORT_RSP,
+                    },
+                    request_id: request.request_id,
+                    hr_response: *hr,
+                },
+            };
+            total_written += encode_send_data_request(user_channel_id, io_channel_id, &response, output)?;
```
mamoreau-devolutions: MS-RDPBCGR sections 2.2.15.2 and 3.2.5.15.2 require the Initiate Multitransport Response to use the negotiated MCS **message** channel. `io_channel_id` targets the I/O channel instead, so a Soft-Sync server will not receive this response. Carry `message_channel_id` into the bootstrap state, validate inbound requests use it, and use it here.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -636,20 +825,94 @@ impl Sequence for ClientConnector {
...
+
+                        debug!(
+                            request_id = pdu.request_id,
+                            protocol = ?pdu.requested_protocol,
+                            "Received Initiate Multitransport Request"
+                        );
+
+                        requests.push(pdu);
+
+                        // Stay in this state to read more requests (server may send a second)
```
mamoreau-devolutions: There is no end marker for the set of requests: a server can request only one transport and is not required to send Demand Active before the client acts. MS-RDPBCGR section 3.2.5.15.1 requires the client to attempt sideband creation upon decoding a request, but this state waits for another main-channel PDU before exposing it. That can stall the handshake. Yield the request immediately, then return to bootstrap processing after its result.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -211,6 +263,141 @@ impl ClientConnector {
...
+            return Err(general_err!(
+                "multitransport results count does not match requests count"
+            ));
+        }
+
+        let mut total_written = 0;
+
+        for (request, result) in requests.iter().zip(results) {
+            let response = match result {
+                MultitransportResult::Success => {
```
mamoreau-devolutions: `S_OK` is only permitted when the server advertised `SOFTSYNC_TCP_TO_UDP` (MS-RDPBCGR section 2.2.15.2). The server MultiTransportChannelData is currently discarded, so this emits a success response even if Soft-Sync was not negotiated. Retain and check both peers' capability flags before constructing this response.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -211,6 +263,141 @@ impl ClientConnector {
...
+    /// Skip multitransport bootstrapping without sending any responses.
+    ///
+    /// Use this when the application doesn't support or doesn't want UDP
+    /// transport. The server will continue with TCP-only operation.
+    ///
+    /// The buffered Demand Active PDU is replayed internally.
+    ///
+    /// Returns an error if the connector is not in `MultitransportPending`
+    /// state.
+    pub fn skip_multitransport(&mut self, output: &mut WriteBuf) -> ConnectorResult<Written> {
```
mamoreau-devolutions, line 384: With mutually negotiated Soft-Sync, MS-RDPBCGR section 3.2.5.15.1 requires an Initiate Multitransport Response regardless of success or failure. This path emits nothing, and both standard drivers call it automatically, leaving a compliant server waiting for the required failure response. Send `E_ABORT` for every pending request when Soft-Sync is negotiated.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -211,6 +263,141 @@ impl ClientConnector {
...
+        &mut self,
+        results: &[MultitransportResult],
+        output: &mut WriteBuf,
+    ) -> ConnectorResult<Written> {
+        let ClientConnectorState::MultitransportPending {
+            io_channel_id,
+            user_channel_id,
+            requests,
+            buffered_demand_active,
+        } = mem::replace(&mut self.state, ClientConnectorState::Consumed)
```
mamoreau-devolutions: The documented result-count validation error is recoverable by the caller, but this `mem::replace` has already discarded all pending requests and the buffered Demand Active. On a length mismatch at line 325, the connector remains `Consumed`, so the caller cannot retry with the correct results or call `skip_multitransport`. Validate the count before taking the state, or restore the pending state on this error.

mamoreau-devolutions (changes requested):
This implementation has protocol-correctness issues that prevent interoperable multitransport bootstrapping.

crates/ironrdp-connector/src/connection.rs
glamberson: Fixed. You're right that the block was being discarded, and the `warn!` sitting there still said multitransport was "not supported", which this PR rather changes.

The server's flags are now retained on the connector alongside `message_channel_id`, which is captured off the same `server_gcc_blocks` three lines above, and a `soft_sync_negotiated()` helper checks the flag on both sides. The result gets frozen into `MultitransportPending` when the state is entered rather than read back at completion time, so `complete_multitransport` can't observe a field that moved in between.

One judgment call worth your eyes, since it also decides the shape of your `E_ABORT` point below. When Soft-Sync isn't negotiated I now send no response PDU at all, rather than rejecting `Success` or quietly downgrading it to a failure. Reading 2.2.15.2, the response is the Soft-Sync signalling path, and without it the outcome is reported in band on the new transport, so putting anything on the main channel is itself the violation. That gives one rule covering both paths: with Soft-Sync, always respond, success or failure; without it, never respond. Results are still consumed and the buffered Demand Active still replays, so a caller that established the transport doesn't have to know which mode is in play.

Worth noting the acceptor already does the retaining half of this at `ironrdp-acceptor/src/connection.rs:527`, so the server side was keeping exactly what the client side was throwing away.

Added `complete_multitransport_omits_responses_without_soft_sync`. The existing tests assert `written != Written::Nothing`, which the replayed Demand Active satisfies on its own, so none of them would have caught this in either direction.

crates/ironrdp-connector/src/connection.rs
glamberson: Done, all three parts.

`message_channel_id` now flows from the connector through `advance_licensing_exchange` into `MultitransportBootstrapping` and on into `MultitransportPending`. Inbound requests are validated against it: `ctx.channel_id` was already coming back from `decode_send_data_indication` and being dropped, so a request on any other channel is now rejected, as is one arriving when no message channel was negotiated at all, since a server doing that contradicts its own GCC block.

For the response I unwrap the `Option` with an error rather than falling back to `io_channel_id`. Soft-Sync presupposes the message channel, and that fallback is precisely the bug you're describing, so failing loudly seemed better than continuing to write somewhere the server isn't reading.

Added `complete_multitransport_responds_on_the_message_channel`, which decodes the first PDU back off the output buffer and asserts the channel. Both channels are valid MCS targets, so nothing downstream notices the wrong one, which is presumably how this got as far as it did.

One thing deliberately left for the next fix rather than done here. Channel ID would be a better discriminator between a request and the Demand Active than the current try-decode, and the comment above that state already concedes the try-decode is a heuristic. I'm folding that into the fix for your request-yielding point, since that reshapes the same lines and the change belongs in one place rather than split across two.

Same shape as #1348 for NetworkAutoDetect, so the connector now has the message channel wired on both paths.

crates/ironrdp-connector/src/connection.rs
glamberson: Done. Skip now sends `E_ABORT` for every pending request on the message channel when Soft-Sync is negotiated, and stays silent when it isn't, which completes the rule from the Soft-Sync thread: with Soft-Sync always respond, without it never respond.

Your point about the drivers checks out and was the part I'd underweighted. `ironrdp-async/src/connector.rs:89` and `ironrdp-blocking/src/connector.rs:93` both skip unconditionally, so this wasn't an edge case. Every default client meeting a Soft-Sync server was leaving it waiting.

Two smaller things fell out. `skip_multitransport` now accumulates into `total_written` and returns `Written::from_size`, since it can write before the replay rather than only after it. And the response-building both paths were doing inline moved into a private `multitransport_response(request_id, hr_response)` helper. I stopped short of adding a `failure()` constructor beside the existing `success()` in `ironrdp-pdu`, since that's public surface in another crate to serve one caller.

The drivers keep skipping automatically, which reads correct to me now. A client with no UDP transport should decline, and `E_ABORT` is how you say so. The bug was declining silently, not declining.

Tests: `skip_multitransport_declines_with_e_abort_under_soft_sync` decodes the response back off the wire and checks both the channel and the HRESULT, and `skip_multitransport_stays_silent_without_soft_sync` is the mirror. The existing skip test asserts `written != Written::Nothing`, which the Demand Active replay satisfies on its own, so it passed either way.

crates/ironrdp-connector/src/connection.rs
glamberson: Reshaped. The yield now happens on decode rather than on the arrival of a following PDU, and `complete`/`skip` return to `MultitransportBootstrapping` instead of running on to capabilities exchange, so a second request is simply the next PDU.

Three things fell out of that which go beyond what you asked for, so worth calling out.

`buffered_demand_active` is gone, along with the replay. It only existed because the old design couldn't yield until a non-request PDU arrived, so it had to hold the Demand Active while the application worked. Yielding on decode means the Demand Active hasn't arrived yet, so there's nothing to buffer.

The API went singular: `complete_multitransport(MultitransportResult, output)`, and `multitransport_requests()` became `multitransport_request() -> Option<&_>`. Handling one request at a time makes a slice of results meaningless. It reads better for the two-transport case too, since the caller answers reliable and lossy separately rather than being asked to bring both up before reporting either.

The two public methods collapsed into one private body. Declining is just an `E_ABORT` outcome, so they differed only in the HRESULT.

On the channel-ID fold I promised on the other thread, it's in, with one correction to what I said there. Channel ID narrows but doesn't replace the decode, because #1348 put NetworkAutoDetect on the message channel as well, so "arrived on the message channel" doesn't uniquely mean multitransport. The gain is in the other direction: an I/O-channel PDU now routes straight to capabilities exchange instead of being speculatively decoded as a multitransport request first, which is the heuristic the old comment was apologising for. An unrecognised channel is now an explicit error rather than being treated as a Demand Active by default.

The new tests turned up a bug of mine. `Written::from_size(0)` returns an error rather than `Nothing`, and zero only became reachable once the no-Soft-Sync path stopped writing anything at all. Three tests failed on it before I fixed it.

One consequence for your `mem::replace` point. Removing the slice removes the result-count mismatch, so that specific error is gone and its test with it. The defect you identified is not gone: the `message_channel_id` unwrap I added on the other two threads fires after the `mem::replace` and leaves the connector `Consumed` exactly as you described. I'll deal with it there rather than close that thread as obsolete.

Tests: `multitransport_request_is_surfaced_without_waiting_for_another_pdu` drives a real request frame through `step` and asserts the yield happens on arrival, which is the regression test for the stall itself and something the old design simply couldn't express. Plus `responding_returns_to_bootstrapping_for_the_next_request`, `third_multitransport_request_is_rejected`, and `demand_active_on_the_io_channel_ends_bootstrapping`.

glamberson:
One cross-PR interaction worth flagging, since it touches a decoder this PR tightened.

#1458 switches `BasicSecurityHeader` decode from `from_bits()` to `from_bits_truncate()`, so unknown flag bits are masked away rather than rejecting the PDU. That looks right to me on its own terms: Windows Server 2019 with RDS licensing really does send them, and FreeRDP tolerates them.

It does interact with this PR. `MultitransportRequestPdu::decode` calls `BasicSecurityHeader::decode` and then requires the flags to equal `SEC_TRANSPORT_REQ` exactly, because a successful decode is what tells the connector that what arrived really is a request. Today a header of `TRANSPORT_REQ | <unknown>` fails inside the header decode. With truncation it reaches that comparison already equal to `TRANSPORT_REQ` and passes, so the set of byte patterns that can pass as a request gets wider.

The exposure is small. The connector now narrows by MCS channel before decoding, so a Demand Active on the I/O channel never reaches this decoder at all. What remains is auto-detect traffic on the message channel, which is the same channel multitransport uses since #1348.

I've rewritten the comment on that check accordingly. It previously justified the exact match by pointing at Demand Active aliasing, which the channel routing here made obsolete. It now says what the guard is actually for, and that its strength depends on what `BasicSecurityHeader` admits.

No change requested on either side. #1458 is a fix I'd want. Better that whoever merges second knows the two touch.

mamoreau-devolutions:
@glamberson I have merged https://github.com/Devolutions/IronRDP/pull/1458

crates/ironrdp-connector/src/connection.rs
```diff
@@ -221,12 +316,190 @@ impl ClientConnector {
...
+        let total_written = if let Some(response_channel) = response_channel {
+            let response = match result {
+                MultitransportResult::Success => {
+                    rdp::multitransport::MultitransportResponsePdu::success(request.request_id)
+                }
+                MultitransportResult::Failure(hr) => multitransport_response(request.request_id, hr),
+            };
+
+            encode_send_data_request(user_channel_id, response_channel, &response, output)?
+        } else {
```
mamoreau-devolutions, line 490: This suppresses `E_ABORT` as well as `S_OK` when Soft-Sync is absent. MS-RDPBCGR 3.2.5.15.1 says that when the client cannot initiate the sideband channel, it **SHOULD** send an Initiate Multitransport Response; only `S_OK` is restricted to servers advertising `SOFTSYNC_TCP_TO_UDP` (2.2.15.2). `skip_multitransport()` always supplies `E_ABORT`, so both default drivers silently drop the specified failure report. Keep the no-Soft-Sync suppression for success, but send `E_ABORT` for failures on the message channel.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -221,12 +316,190 @@ impl ClientConnector {
...
+        let response_channel = self.multitransport_response_channel()?;
+
+        let ClientConnectorState::MultitransportPending {
+            io_channel_id,
+            user_channel_id,
+            message_channel_id,
+            request,
+            requests_seen,
+            ..
+        } = mem::replace(&mut self.state, ClientConnectorState::Consumed)
```
mamoreau-devolutions: `multitransport_response_channel()` returns `Ok(None)` outside `MultitransportPending`, so this `mem::replace` consumes *any* other connector state before returning the documented outside-state error. For example, calling `complete_multitransport` while in `CapabilitiesExchange` leaves the connector `Consumed`, not in capabilities exchange. Validate/take the pending state before mutating it (or restore it on the error), and extend the existing outside-pending test to assert the state is preserved.

crates/ironrdp-pdu/src/rdp/multitransport.rs
```diff
@@ -115,8 +115,23 @@ impl<'de> Decode<'de> for MultitransportRequestPdu {
...
+        // PDU whose leading bytes happen to alias onto this field with the
+        // TRANSPORT_REQ bit set. The connector narrows by MCS channel first,
+        // but the message channel also carries auto-detect traffic, so this
+        // decode is still the thing that has to tell them apart.
+        //
+        // Note this guard is only as strong as what `BasicSecurityHeader`
+        // admits: a decoder that masks unknown bits away rather than rejecting
+        // them widens the set of byte patterns that reach the comparison
+        // already equal to TRANSPORT_REQ.
+        if security_header.flags != BasicSecurityHeaderFlags::TRANSPORT_REQ {
```
mamoreau-devolutions: MS-RDPBCGR 2.2.15.1 requires the flags to **contain** `SEC_TRANSPORT_REQ`, not equal it. More importantly, 2.2.8.1.1.2.1 says `SEC_RESET_SEQNO` and `SEC_IGNORE_SEQNO` **MUST be ignored** by clients and servers, but a request containing either alongside `SEC_TRANSPORT_REQ` is rejected here and aborts the connection. Preserve the discriminator without rejecting ignorable flags, and add a decode test for `SEC_TRANSPORT_REQ | SEC_RESET_SEQNO`.

mamoreau-devolutions (changes requested):
The rework resolves the original message-channel and request-yielding defects, but the response rules, security-header acceptance, and error-state preservation still need correction.

crates/ironrdp-pdu/src/rdp/multitransport.rs
glamberson: Correct on both counts, and the comment above the check was worse than the check: it asserted the flags "must be EXACTLY SEC_TRANSPORT_REQ (MS-RDPBCGR 2.2.15.1), not merely contain the bit", citing the section that says the opposite. 2.2.15.1 says "MUST contain" in both the field description and again in 3.3.5.15.1. That is now stated the right way round.

The two seqno flags are masked off before the comparison rather than treated as a mismatch, per 2.2.8.1.1.2.1. Worth noting they were never unknown bits: `RESET_SEQNO` and `IGNORE_SEQNO` are both defined in `BasicSecurityHeaderFlags`, so `from_bits` admitted them before #1458 too and this decoder rejected the result either way. My earlier comment on this PR framed the risk as #1458 widening what could pass; that was the wrong way round, and the real defect was rejecting traffic the spec says to accept, which predates it.

What remains after masking is still compared for equality. That is stricter than the spec asks, and I have kept it deliberately rather than dropping to `contains`, because a successful decode is what tells the connector that what arrived really is a request. The connector narrows by MCS channel first, but the message channel also carries auto-detect, and `SEC_AUTODETECT_REQ` alongside the transport bit is exactly what a bare `contains` would wave through. The comment now presents that as a local choice rather than a spec requirement, which is what it always was.

Tests: the request decodes with `SEC_RESET_SEQNO`, with `SEC_IGNORE_SEQNO`, and with both, since `0x0038` in the spec's own worked examples shows them arriving together; and a request carrying `SEC_AUTODETECT_REQ` alongside `SEC_TRANSPORT_REQ` is still rejected, so the discriminator is pinned rather than assumed.

crates/ironrdp-connector/src/connection.rs
glamberson: Right, and the comment sitting directly above that line claimed the opposite. It said the channel was resolved first so a failure could not leave the connector `Consumed`. That ordering only ever guarded the missing-message-channel error. `multitransport_response_channel()` returns `Ok(None)` for every state that is not `MultitransportPending`, so the outside-state case went straight past the `?` and into the `mem::replace`, exactly as you describe, and `CapabilitiesExchange` is the state the existing test happens to use.

Rather than reorder the check or restore on the error path, the state is no longer taken at all. Everything the body needs is a scalar, and the request PDU was only ever read for its `request_id`, so the state is borrowed, the values copied out, and `self.state` assigned once at the end when nothing further can fail.

I preferred that to restoring because it makes the property structural rather than procedural. There is now no point in the function at which the connector has been mutated but the work is incomplete, so any error added later preserves the state by construction. Restore-on-error stays correct only while every future error path remembers to do it, which is the same shape as the bug you just found.

`complete_multitransport_outside_pending_state_errors` now asserts the connector is still in `CapabilitiesExchange` afterwards, not just that the call errored. It fails against the previous code, since `mem::replace` had already stored `Consumed` by the time the error was returned.

crates/ironrdp-connector/src/connection.rs
glamberson: Correct, and I owe you an apology for this one in particular: I had written a test to enforce the wrong behavior. `skip_multitransport_stays_silent_without_soft_sync` asserted "no E_ABORT may be emitted when Soft-Sync was never negotiated", which is the defect stated as a requirement, and it made the mistake look verified rather than assumed.

The reading I had was that 2.2.15.2 makes the response the Soft-Sync signalling path in general. It does not; it restricts one value. `S_OK` "MUST only be sent to a server that advertises the SOFTSYNC_TCP_TO_UDP flag", and nothing there constrains `E_ABORT`. Meanwhile 3.2.5.15.1 asks for the response whenever the client could not initiate the sideband channel, with no Soft-Sync condition in that sentence at all, and the Soft-Sync clause is a separate obligation that adds the success case rather than gating both. So the split is by outcome, not by mode. Had I read those two sections against each other rather than against my own earlier comment, this would not have reached you.

`multitransport_response_channel` now takes the result and decides on that basis. Soft-Sync behaves as before and responds either way, with the message channel presupposed so its absence stays an error. Without Soft-Sync, success returns no channel and a failure returns one.

One judgment call worth naming: without Soft-Sync and with no message channel, it stays silent rather than erroring. Sending is a SHOULD there, so suppressing conforms, and failing the connection over an optional report would be worse than the bug being fixed. It should not arise anyway, since 2.2.15.1 requires the request itself to have arrived on that channel.

Tests: the silent-skip test is replaced by one asserting `E_ABORT` reaches the message channel with the right request ID, and the success test is narrowed to success with its rationale corrected to cite the actual restriction. The new test fails against the previous code.

crates/ironrdp-connector/src/connection.rs
glamberson: This one predates the reshape and I should have said so at the time rather than letting a force-push answer it silently. You posted at 01:22Z; I pushed the singular-API rework at 04:26Z without replying here. Sorry for the wasted read.

Both specifics are gone with the batch API. There is no longer a result count to validate, since each request is completed on its own, and `buffered_demand_active` is no longer carried in the state at all.

The general point stands, and it is the same one you made again at 15:46 against the current code, which is where I have addressed it: the state is no longer taken before the work that can fail, so an error leaves the connector exactly as it was.

I did check whether anything else in the new shape has the same property. The request cap does leave the connector `Consumed`, because it errors from inside `step()`, which has already taken the state. I have left that as-is on the grounds that it follows the convention every other state in `step()` uses, and a server exceeding the 2.2.15.1 cap is a protocol violation rather than something the caller can retry. Say the word if you would rather that one were recoverable too.

glamberson:
@mamoreau-devolutions all four addressed and pushed as `96520093`, rebased onto current master.

Three of the four were mine from this morning's round, and two of them shipped with a comment or a test asserting they were safe. The exact-flags check carried a comment claiming 2.2.15.1 required exactness, citing the section that says the opposite; the `mem::replace` had a comment directly above it stating that the ordering prevented exactly the state destruction you found; and the suppressed `E_ABORT` had a test named `skip_multitransport_stays_silent_without_soft_sync` enforcing it. That is worse than being wrong, because it made all three look checked. Sorry for the review time that cost you.

The common thread was that I read my own earlier comments as authority instead of reading the spec sections against each other. For this round I went back to the text for every claim before it went into a comment: "MUST contain" at 2.2.15.1 and again at 3.3.5.15.1, "MUST be ignored" for both seqno flags at 2.2.8.1.1.2.1 with the `0x0038` worked example, the unconditional SHOULD at 3.2.5.15.1, and the `S_OK`-only restriction at 2.2.15.2.

Two things I decided rather than derived, both flagged in the threads so you can overrule them. Past the ignorable flags the decoder still compares for equality, which is stricter than the spec, because the message channel also carries auto-detect and a bare `contains` would take `SEC_AUTODETECT_REQ` for a request; the comment now says that is a local choice rather than a spec requirement. And without Soft-Sync and with no message channel the response is dropped rather than erroring, since it is a SHOULD there.

The older thread from 01:22 I have answered and resolved: it predated the singular-API reshape by three hours and I let a force-push answer it silently instead of replying, which I should not have done.

One change you did not ask for: the title now carries `!`. `cargo semver-checks` reports three major failures against the merge-base, all from the original PR rather than today, and it would have gone red under the new gate. Details are in the body, including one of the three I read as a false positive.

crates/ironrdp-pdu/src/rdp/multitransport.rs
```diff
@@ -229,8 +246,13 @@ impl<'de> Decode<'de> for MultitransportResponsePdu {
 
         let security_header = BasicSecurityHeader::decode(src)?;
 
-        if !security_header.flags.contains(BasicSecurityHeaderFlags::TRANSPORT_RSP) {
-            return Err(invalid_field_err!("securityHeader", "expected TRANSPORT_RSP flag"));
+        // Must be EXACTLY SEC_TRANSPORT_RSP (MS-RDPBCGR 2.2.15.2), not merely
+        // contain the bit; see the note on `MultitransportRequestPdu::decode`.
+        if security_header.flags != BasicSecurityHeaderFlags::TRANSPORT_RSP {
```
mamoreau-devolutions: The request decoder now correctly masks `SEC_RESET_SEQNO` and `SEC_IGNORE_SEQNO`, but the response decoder still rejects both. MS-RDPBCGR 2.2.15.2 says the flags **MUST contain** `SEC_TRANSPORT_RSP`, while 2.2.8.1.1.2.1 says clients and servers **MUST ignore** both sequence-number flags. Thus a valid `SEC_TRANSPORT_RSP | SEC_RESET_SEQNO` response fails here. Apply the same mask before comparison and add the response-side variants to `decode_request_ignores_the_seqno_flags` (renamed accordingly).

crates/ironrdp-connector/src/connection.rs
```diff
@@ -158,7 +234,26 @@ impl ClientConnector {
...
+            server_multitransport_flags: None,
+        }
+    }
+
+    /// Whether Soft-Sync (`SOFTSYNC_TCP_TO_UDP`) was mutually advertised.
+    ///
+    /// Per [\[MS-RDPBCGR\] 2.2.15.2] the Initiate Multitransport Response is
+    /// the Soft-Sync signalling path, and is only sent when *both* peers
+    /// advertised the flag in their GCC `MultiTransportChannelData` block.
+    /// Without Soft-Sync the client reports the outcome in band on the new
```
mamoreau-devolutions: This Rustdoc still asserts the behavior just corrected: it says non-Soft-Sync outcomes are reported in-band and that sending a main-channel response is a protocol violation. MS-RDPBCGR 3.2.5.15.1 asks for the failure response without that condition; only `S_OK` is restricted by 2.2.15.2. The same obsolete claim remains in `multitransport_response_channel` at lines 391-394. Please remove or revise both so the public source docs agree with the result-specific implementation.

mamoreau-devolutions (changes requested):
The three prior behavioral findings are fixed. One mirrored protocol decoder still rejects a valid security-header form, and the connector docs retain the superseded no-Soft-Sync rule.

crates/ironrdp-pdu/src/rdp/multitransport.rs
glamberson: Fixed, and you were right to check the other decoder rather than take the first fix at face value.

The response side had all three parts of the same defect: the comment asserted "Must be EXACTLY SEC_TRANSPORT_RSP (MS-RDPBCGR 2.2.15.2), not merely contain the bit", the comparison was `!=` against the bare flag, and the error message said "exactly". The spec says the opposite in both places it mentions the field: "The flags field of the security header MUST contain the SEC_TRANSPORT_RSP (0x0004) flag" at 2.2.15.2, and again in 3.2.5.15.2. It now masks the two ignorable flags and compares what remains, identically to the request side.

I went looking for any further instances rather than fixing only the one you named. Across the workspace there were exactly two equality comparisons on `BasicSecurityHeaderFlags`, both introduced by this PR and both in this file; every other check in the codebase already uses `contains`. So that class is closed rather than merely reduced.

The test is renamed and now drives both decoders, since the two are symmetric and there was no reason to cover only one. I verified it the useful way round: reverting only the response-side mask, leaving the request fix in place, fails with "response with SEC_RESET_SEQNO must decode", so it genuinely exercises the half you found rather than passing on the strength of the other.

crates/ironrdp-connector/src/connection.rs
glamberson: Fixed, and there were more than the two you pointed at. Rather than correct those and stop, I swept every doc and comment in this PR that makes a claim about when a response is sent. Six were stale:

- `soft_sync_negotiated`, the one you named
- `multitransport_response_channel` at 391-394, the one you named
- `complete_multitransport`, which said the response goes out "when Soft-Sync was mutually negotiated"
- `skip_multitransport`, which said "Under Soft-Sync this sends `E_ABORT`". That one mattered most of the four you did not name, since it is the method both bundled drivers call and `E_ABORT` is now sent either way
- the `soft_sync` field on `MultitransportPending`, which claimed the flag alone decides whether a response is emitted
- the public `server_multitransport_flags` field, which said the response "is only permitted once both peers have advertised Soft-Sync"

All six now say the same thing the code does: a failure is reported whenever there is a channel to report it on, and only `S_OK` is withheld, per 2.2.15.2 restricting that one value and 3.2.5.15.1 asking for the failure report unconditionally.

The last of those surfaced only because I ran `cargo doc` to check some intra-doc links I had just added. It turned out to carry a second defect as well: three public items in this PR linked to private methods, producing `private_intra_doc_links` warnings, one of which predates tonight. Public docs no longer link to private helpers; they state the rule instead. `cargo doc -p ironrdp-connector` is now clean, where it was emitting warnings before this change.

I also checked the test comments for the same drift and found one adjacent case that is not stale: the note in `complete_multitransport_responds_on_the_message_channel` is about which channel carries the response, not about whether one is sent, so it stands.

glamberson:
Done. Sorry for the repeated mises. I think I have it now.

mamoreau-devolutions (approved):
The final revision resolves the outstanding multitransport protocol and state-safety findings. Approved.

mamoreau-devolutions:
@glamberson I don't know if you're using my windows-protocols agent skill, but it contains the AI-optimized markdown corpus of the entire 400+ Windows protocol specifications. That's what I'm using to review PRs for protocol correctness, it makes a huge difference: https://github.com/awakecoding/openspecs

PR #1100, opened 2026-02-16
fix(server): make MultifragmentUpdate max_request_size configurable

### LINKED ISSUE #318: server: determine negotiated value for MultifragmentUpdate::max_request_size
https://github.com/Devolutions/IronRDP/blob/f5f5e9c087162b39dabb6af39e4cb1252f26beff/crates/ironrdp-server/src/capabilities.rs#L77

is rejected by msctc, the client will disconnect. (presumably, it tries to allocate that much buffer space upfront)

The value was also adjusted in FreeRDP: https://github.com/FreeRDP/FreeRDP/pull/1313

There doesn't seem to be a way to negotiate that value with the client beforehand.

glamberson:
The hardcoded `max_request_size` of 16,777,215 in the server's MultifragmentUpdate capability causes mstsc to reject the connection (it likely tries to allocate that buffer upfront). FreeRDP hit the same problem and adjusted their value in FreeRDP/FreeRDP#1313.

This adds a configurable `max_request_size` field to `RdpServerOptions` with a default of 8 MB (matching what `ironrdp-connector` already uses on the client side) and exposes it through the builder via `with_max_request_size()`.

Reference: [MS-RDPBCGR 2.2.7.2.7](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/01717954-716a-424d-af35-28fb2b86df89) (TS_MULTIFRAGMENTUPDATE_CAPABILITYSET)

Closes #318

### DIFF
--- a/crates/ironrdp-server/src/builder.rs
+++ b/crates/ironrdp-server/src/builder.rs
@@ -177,6 +177,12 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
+    /// Sets the [MultifragmentUpdate] maximum reassembly buffer size advertised
+    /// during capability exchange.
+    ///
+    /// Defaults to [`RdpServerOptions::DEFAULT_MAX_REQUEST_SIZE`] (8 MB).
+    ///
+    /// [MultifragmentUpdate]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/01717954-716a-424d-af35-28fb2b86df89
     pub fn with_max_request_size(mut self, max_request_size: u32) -> Self {
         self.state.max_request_size = max_request_size;
         self
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -54,7 +54,7 @@ impl RdpServerOptions {
     /// to reject the connection.
     ///
     /// [MultifragmentUpdate]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/01717954-716a-424d-af35-28fb2b86df89
-    pub const DEFAULT_MAX_REQUEST_SIZE: u32 = 8 * 1024 * 1024;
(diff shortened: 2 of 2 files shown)


crates/ironrdp-server/src/builder.rs
```diff
@@ -159,12 +162,18 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
+    pub fn with_max_request_size(mut self, max_request_size: u32) -> Self {
+        self.state.max_request_size = max_request_size;
+        self
+    }
```
CBenoit, line 189: suggestion: Add a small docstring explaining what it is and maybe including what you explain in the PR description

crates/ironrdp-server/src/server.rs
```diff
@@ -40,9 +40,20 @@ pub struct RdpServerOptions {
...
 impl RdpServerOptions {
+    /// Default [MultifragmentUpdate] max reassembly buffer size (8 MB).
+    ///
+    /// Advertised to the client during capability exchange as the largest
+    /// reassembled Fast-Path Update the server can accept.
+    /// Values that are too large cause certain clients (notably mstsc)
+    /// to reject the connection.
+    ///
+    /// [MultifragmentUpdate]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/01717954-716a-424d-af35-28fb2b86df89
+    pub const DEFAULT_MAX_REQUEST_SIZE: u32 = 8 * 1024 * 1024;
```
CBenoit: suggestion: Should be `pub(crate)` to avoid making this part of the public API

crates/ironrdp-server/src/builder.rs
glamberson: done, thanks.

crates/ironrdp-server/src/server.rs
glamberson: done, thanks.

PR #1101, opened 2026-02-16
fix(session): fix pixel format handling in bitmap decoders

### LINKED ISSUE #1030: Using the Bgra32 pixel format, color channels are swapped when connecting to Windows
When using `Bgra32` as the pixel format for `DecodedImage`, then connecting to XRDP works fine (colors are correct) but for Windows RDP servers, the color channels are swapped (are actually RGBA). This doesn't happen the other way around when using `Rgba32`, then the XRDP colors are still fine.

I'm unsure if this is actually a bug in ironrdp or just a "quirk"/limitation of the Windows RDP server.

An example screenshot using my app that uses ironrdp can be found in this issue: https://github.com/theCapypara/field-monitor/issues/69

ramnes:
Hey there, thanks for the great project!

So far, `apply_rgb16_bitmap`, `apply_rgb24_iter`, and `apply_rgb32_bitmap` all hardcode an RGBA byte order, ignoring the configured `PixelFormat`. This caused me an R/B channel swap when I tried to use `BgrA32` instead of `RgbA32`.

This change replaces the hardcoded byte writes with `PixelFormat::write_color()` to arrange bytes according to the configured format.

I'm still new with IronRDP so not entirely sure, but I guess it should fix #1030. (cc. @theCapypara)

### DIFF
--- a/crates/ironrdp-graphics/src/image_processing.rs
+++ b/crates/ironrdp-graphics/src/image_processing.rs
@@ -1,7 +1,6 @@
 use core::{cmp, fmt};
 use std::io;
 
-use byteorder::WriteBytesExt as _;
 use ironrdp_pdu::geometry::{InclusiveRectangle, Rectangle as _};
 
 const ALPHA_OPAQUE: u8 = 0xff;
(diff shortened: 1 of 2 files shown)


CBenoit (commented):
Thank you for opening this PR!
The behavior itself seems correct, but I’m a little bit afraid that the performance cost of this implementation is non negligible: `write_color` is performing a match internally; this means that this code is effectively adding conditional code execution inside hot paths (the loops), and this is notoriously bad for performance.
I suggest measuring the performance cost and/or modifying into a more performance friendly pattern (e.g.: push conditions outside of the loops).

ramnes:
@CBenoit This latest commit should address the performance concern, let me know what you think!

PR #1109, opened 2026-02-17
Add ECHO virtual channel support across IronRDP

mamoreau-devolutions:
## Summary
- add new ironrdp-echo crate with MS-RDPEECO PDUs and DVC client/server processors
- wire feature-gated cho support through ironrdp, ironrdp-client, and ironrdp-server
- add server-side runtime ECHO handle/bridge for request dispatch and RTT measurement tracking
- add end-to-end Echo virtual channel test in ironrdp-testsuite-extra
- update server docs and fix fmt/clippy/doctest issues for CI parity

## Validation
- cargo xtask check fmt -v
- cargo xtask check lints -v
- cargo xtask check tests -v
- cargo test -p ironrdp-server --doc --locked

### DIFF
--- a/crates/ironrdp-client/Cargo.toml
+++ b/crates/ironrdp-client/Cargo.toml
@@ -29,6 +29,7 @@ rustls = ["ironrdp-tls/rustls", "tokio-tungstenite/rustls-tls-native-roots", "ir
 native-tls = ["ironrdp-tls/native-tls", "tokio-tungstenite/native-tls", "ironrdp-mstsgu/native-tls"]
 qoi = ["ironrdp/qoi"]
 qoiz = ["ironrdp/qoiz"]
+echo = ["ironrdp/echo"]
 
 [dependencies]
 # Protocols
--- a/crates/ironrdp-dvc/src/server.rs
+++ b/crates/ironrdp-dvc/src/server.rs
@@ -74,6 +74,19 @@ impl DrdynvcServer {
         }
     }
 
+    pub fn get_dvc_channel_id_by_type<T>(&self) -> Option<u32>
+    where
+        T: DvcServerProcessor + 'static,
+    {
+        self.dynamic_channels.iter().find_map(|(id, channel)| {
+            if channel.state != ChannelState::Opened || !channel.processor.as_any().is::<T>() {
+                return None;
+            }
+
+            id.try_into().ok()
+        })
+    }
+
     // FIXME(#61): it’s likely we want to enable adding dynamic channels at any point during the session (message passing? other approach?)
 
     #[must_use]
--- a/crates/ironrdp-echo/Cargo.toml
+++ b/crates/ironrdp-echo/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "ironrdp-echo"
+version = "0.1.0"
+readme = "README.md"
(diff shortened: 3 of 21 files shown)


crates/ironrdp-client/Cargo.toml
```diff
@@ -29,6 +29,7 @@ rustls = ["ironrdp-tls/rustls", "tokio-tungstenite/rustls-tls-native-roots", "ir
 native-tls = ["ironrdp-tls/native-tls", "tokio-tungstenite/native-tls", "ironrdp-mstsgu/native-tls"]
 qoi = ["ironrdp/qoi"]
 qoiz = ["ironrdp/qoiz"]
+echo = ["ironrdp/echo"]
```
CBenoit, line 32: question: Do we really want to put that behind a feature flag?

crates/ironrdp-dvc/src/server.rs
```diff
@@ -74,6 +74,19 @@ impl DrdynvcServer {
...
+    pub fn get_dvc_channel_id_by_type<T>(&self) -> Option<u32>
+    where
+        T: DvcServerProcessor + 'static,
+    {
+        self.dynamic_channels.iter().find_map(|(id, channel)| {
+            if channel.state != ChannelState::Opened || !channel.processor.as_any().is::<T>() {
+                return None;
+            }
+
+            id.try_into().ok()
```
CBenoit, line 86: note: This does not follow any existing pattern and can be confusing. I’ll send a follow up.

crates/ironrdp-echo/tests/echo.rs
CBenoit, line 1: issue: Tests should go into a testsuite crate.

crates/ironrdp-echo/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+categories.workspace = true
+
+[lib]
+doctest = false
+test = false
+
+[dependencies]
+ironrdp-core = { path = "../ironrdp-core", version = "0.1" } # public
+ironrdp-dvc = { path = "../ironrdp-dvc", version = "0.5" } # public
+ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.7" } # public
```
CBenoit, line 21: note: I didn’t check, but `# public` should be used only for dependencies whose types are used in the public API

crates/ironrdp-server/src/lib.rs
```diff
@@ -10,6 +10,8 @@ mod builder;
 mod capabilities;
 mod clipboard;
 mod display;
+#[cfg(feature = "echo")]
+mod echo;
```
CBenoit, line 14: thought: I think there is a better way to keep the crate modularized that is not using feature flags. We should look into this before adding too much of these.

crates/ironrdp-echo/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+authors.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[lib]
+doctest = false
+test = false
```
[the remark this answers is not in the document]
CBenoit: This is intended

Issue #1121, opened 2026-02-24
Fix `arbitrary` feature flag plumbing so it compiles


CBenoit:
The `arbitrary` feature is currently broken / not consistently compiled. We want it working again and wired correctly across crates that define PDUs.

## Scope
- Add/repair `arbitrary` feature flags in crates that own PDU structs/enums
- Ensure the dependency is optional and feature-gated
- Ensure it works in both `std` and `no_std + alloc` contexts where applicable
- Document “how to build with arbitrary” in the relevant .md documents

## Tasks
- Identify target crates (likely `ironrdp-pdu` and any other protocol crates that expose PDU types)
- Add `arbitrary = { version = "1", features = ["derive"], optional = true }`
- Add features wiring (`arbitrary = ["dep:arbitrary", ...]`)
- Check `cargo check -p <crate> --features arbitrary` passes locally

```toml
[dependencies]
arbitrary = { version = "1", features = ["derive"], optional = true }

[features]
# Pick one:
# - If the crate is std-only: arbitrary = ["std", "dep:arbitrary"]
# - If the crate supports no_std + alloc: arbitrary = ["alloc", "dep:arbitrary"]
arbitrary = ["alloc", "dep:arbitrary"]
````

## Acceptance criteria
- `cargo check` for the target crates succeeds with `--features arbitrary`
- No breaking change for default feature sets

Issue #1124, opened 2026-02-24
Add advanced fuzz targets to verify more invariants (round-trip, framing, state-machine…)


CBenoit:
Move beyond “bytes-only decode fuzzing” to higher-signal fuzzing that asserts invariants and exercises state machines.

## Proposed fuzz targets
1) **round_trip** (typed):
- Generate a structured PDU (`Arbitrary`)
- Encode -> Decode -> Compare (plus size invariants)

2) **message_decoding_invariants** (bytes):
- Feed raw bytes into decoder(s)
- If decoding succeeds, assert framing/size/consumption invariants

3) **connector_state_machine** (optional in this issue; can be a follow-up):
- Generate sequences / frames and step connector state machines
- Validate buffer bookkeeping and hint-based framing invariants

## Tasks
- Add `round_trip` fuzz target wired to an oracle module
- Add at least one non-trivial oracle invariant (not just “no panic”)
- Ensure targets can be run via `cargo xtask fuzz run <target>`
- Add regression test mechanism: replay minimized crashers in unit tests

### “Round-trip” oracle (typed fuzzing)

This is a *high signal* harness: generate a structured PDU, encode it, decode it back, compare.

**Fuzz target** (tiny):

```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

use ironrdp_fuzzing::{generators::AnyPdu, oracles};

fuzz_target!(|msg: AnyPdu<'_>| {
    oracles::round_trip(msg);
});
```

**Generator idea** (`AnyPdu<'a>`):

* an enum that contains a curated set of PDUs we want to round-trip (start small; expand over time).
* it should provide:

  * `encode()` (to `Vec<u8>`)
  * `decode_owned(bytes)` (back into an owned representation)
  * `into_owned()` for comparison

**Oracle idea**:

* encode
* decode_owned
* `assert_eq!(decoded, original.into_owned())`
* also assert size invariants (`encoded.len() == original.size()`) where applicable

This is the harness that typically finds:

* wrong length computations
* missing fields in encode/decode
* decode accepting invalid encodings (or encode producing something decode can’t parse)

### “Message decoding” oracle (raw bytes, but with invariants)

We keep a raw-bytes decoder too, but it should assert invariants, e.g.:

* if decoding succeeds, then “frame-size detection” should succeed and be consistent
* if decode consumes N bytes, that should match computed/expected sizes
* avoid silent under/over-reads

This is especially valuable for:

* frame boundary logic
[...]

Issue #1122, opened 2026-02-24
Implement `Arbitrary` for more PDU structs


CBenoit:
Provide `arbitrary::Arbitrary` (derive or manual impl) for a first batch of PDUs so structured fuzzing can start.

## Approach
- Prefer `#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]` for simple types
- Use manual `impl Arbitrary` for types with invariants (length fields, constraints, dependent fields)
- Bias toward “mostly valid” generation so fuzzing reaches deeper logic
- Add a `generators::AnyPdu` (or similar) enum used by fuzz targets which can produce a curated set of PDUs

### Add `Arbitrary` derivations (or manual impls)

For most “plain data” structs/enums:

```rust
#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct SomePdu { /* … */ }
```

### Handle invariants (manual `Arbitrary` where needed)

Some PDUs have constraints that `derive(Arbitrary)` can’t express well (length fields, restricted enums, “must be non-empty”, etc.). In those cases:

* either implement `Arbitrary` manually and *reject* invalid values by returning `Err(arbitrary::Error::IncorrectFormat)`,
* or generate a “raw” value and then *sanitize* it (clamp lengths, fix flags, etc.).

The goal is: **generated PDUs should be “mostly valid”**, so we reach deeper code paths (not fail immediately on trivial validation).

glamberson:
The \"first batch\" of PDU `Arbitrary` impls has landed:

- **#1272** (merged 2026-05-14) — `Arbitrary` derive across ~87 PDU structs / enums in `ironrdp-pdu`, plus the explicit cascade and per-type opt-outs documented in the design SDS at `analysis/IRONRDP-FUZZING-INFRASTRUCTURE-SDS-2026-05-13.md`.
- **#1284** (merged 2026-05-20) — extended to `ironrdp-egfx` types (with the reshape from 49 `#[non_exhaustive]` markings down to 13 consumer-facing types only per your 2026-05-19 review, and `Arbitrary` retained on most).

Structured fuzzing has started against the `Arbitrary`-derived surface:

[...]

Issue #1120, opened 2026-02-24
Advanced fuzzing with `arbitrary` + CI feature-matrix


CBenoit:
We want to move from shallow “bytes-only” fuzzing to **advanced/structured fuzzing** using `arbitrary`-generated PDUs and stronger oracles (round-trip, invariants, state-machine checks). This is expected to find many small problems which we will fix incrementally.

This issue replaces the earlier discussion-only issue (#365).

## Goals
- Restore and support `arbitrary` feature across the relevant crates
- Expand `Arbitrary` coverage over key PDU structs
- Add higher-signal fuzzing harnesses/oracles (round-trip, decoder invariants, connector state-machine)
- Improve CI compilation verification over feature sets (`arbitrary`, `std` vs `no_std`, etc.)

## Non-goals (for initial rollout)
- Perfect coverage of all PDUs on day 1
- Eliminating all findings before merging the harness improvements
- Rewriting protocol code purely to satisfy fuzzing (we’ll fix issues incrementally)

## Notes / links
- Prior discussion: #365
- Existing fuzz scaffold: `fuzz/` + `crates/ironrdp-fuzzing/` + `cargo xtask fuzz ...`

glamberson:
I wanted to flag a coordination point on the structured-fuzzing roadmap here. I have an open issue at #1257 proposing three small additive improvements to `ironrdp-error`: the thiserror migration in #1258, #1260, and #1261; the automatic `core::panic::Location` capture in #1262; and the shared `bail!` / `ensure!` macros in #1263. Together these reduce per-crate error boilerplate by roughly a factor of three and capture diagnostic `file:line` automatically.

The reason I am surfacing this on the fuzzing umbrella issue is that the two efforts compose well, and landing the error work first makes the structured-fuzzing follow-on materially easier.

The thiserror migration removes the hand-written `impl Display` and `impl core::error::Error` blocks on every `*ErrorKind` enum. After it lands, adding new Kind variants (which will be necessary as structured fuzzing surfaces hardening opportunities) becomes a single attribute on a single enum variant rather than touching three separate impl blocks per crate.

The Location capture means fuzz crashes report not just the kind variant but the source location at which the error was constructed. For triage on libFuzzer crash artifacts this is the difference between knowing that an `InvalidField` was returned somewhere and knowing that it was returned at `crates/ironrdp-pdu/src/foo.rs:142`.

[...]

PR #1128, opened 2026-02-26
Fix/arbitrary failing on feature flag

razzat008:
working on #1121

### DIFF
--- a/crates/ironrdp-connector/src/license_exchange.rs
+++ b/crates/ironrdp-connector/src/license_exchange.rs
@@ -56,7 +56,6 @@ impl State for LicenseExchangeState {
 ///
 /// [3.1.5.3.1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpele/8f9b860a-3687-401d-b3bc-7e9f5d4f7528
 #[derive(Debug)]
-#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct LicenseExchangeSequence {
     pub state: LicenseExchangeState,
     pub io_channel_id: u16,
@@ -64,6 +63,7 @@ pub struct LicenseExchangeSequence {
     pub domain: Option<String>,
     pub hardware_id: [u32; 4],
     pub license_cache: Arc<dyn LicenseCache>,
+    pub _phantom: std::marker::PhantomData<u8>,
 }
 
 // Use RefUnwindSafe so that types that embed LicenseCache remain UnwindSafe
@@ -100,7 +100,28 @@ impl LicenseExchangeSequence {
             domain,
             hardware_id,
             license_cache,
+            _phantom: std::marker::PhantomData,
         }
+    }
+}
+
+impl<'a> arbitrary::Arbitrary<'a> for LicenseExchangeSequence {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(Self {
+            state: LicenseExchangeState::arbitrary(u)?,
+            io_channel_id: u16::arbitrary(u)?,
+            username: String::arbitrary(u)?,
+            domain: Option::<String>::arbitrary(u)?,
+            hardware_id: [
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+            ],
+            // MIGHT NEED CHANGE, currently it's not using arbitrary function
+            license_cache: Arc::new(NoopLicenseCache),
+            _phantom: std::marker::PhantomData,
+        })
     }
 }
 
--- a/crates/ironrdp-connector/src/license_exchange.rs
+++ b/crates/ironrdp-connector/src/license_exchange.rs
@@ -63,7 +63,6 @@ pub struct LicenseExchangeSequence {
     pub domain: Option<String>,
     pub hardware_id: [u32; 4],
     pub license_cache: Arc<dyn LicenseCache>,
-    pub _phantom: std::marker::PhantomData<u8>,
 }
 
 // Use RefUnwindSafe so that types that embed LicenseCache remain UnwindSafe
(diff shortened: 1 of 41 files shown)


razzat008:
I was not confident with some of the impls so I've added `MIGHT NEED CHANGE` for those parts.
Also, the `StaticChannelSet` had an attribute which used the `TypeId` which didn't implement the `Arbitrary` trait, so it just calls the constructor for now.


Mannual impls are done for:

- LicenseExchangeSequence: the `license_cache` attribute had dyn trait
- ChannelName: a valid null terminated sequence was required

crates/ironrdp-connector/src/license_exchange.rs
```diff
@@ -56,14 +56,14 @@ impl State for LicenseExchangeState {
...
 #[derive(Debug)]
-#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct LicenseExchangeSequence {
     pub state: LicenseExchangeState,
     pub io_channel_id: u16,
     pub username: String,
     pub domain: Option<String>,
     pub hardware_id: [u32; 4],
     pub license_cache: Arc<dyn LicenseCache>,
+    pub _phantom: std::marker::PhantomData<u8>,
```
CBenoit: question: Why did you add this PhantomData?

crates/ironrdp-connector/src/license_exchange.rs
```diff
@@ -100,10 +100,31 @@ impl LicenseExchangeSequence {
...
+            username: String::arbitrary(u)?,
+            domain: Option::<String>::arbitrary(u)?,
+            hardware_id: [
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+                u32::arbitrary(u)?,
+            ],
+            // MIGHT NEED CHANGE, currently it's not using arbitrary function
+            license_cache: Arc::new(NoopLicenseCache),
```
CBenoit: I think this is perfectly fine as we don’t provide any LicenseCache implementation as part of this crate. Implementers wishing to fuzz using their license cache should do so on their side.

crates/ironrdp-connector/src/license_exchange.rs
razzat008: I initially thought it was required for the dyn trait object to compile successfully, so I left it in there and forgot about it, even after the manual impl for LicenseExchangeSequence. 
It compiles fine without it and isn't needed, my bad:))

crates/ironrdp-connector/src/license_exchange.rs
razzat008: noted

CBenoit (commented):
I’m not entirely sure what the `Arbitrary` trait is meant to do in `ironrdp-svc`: we’re probably not going to deal with "arbitrary static virtual channels", so maybe a better way would be to manually implement the trait at some places more in `ironrdp-connector` 🤔

razzat008:
I understand, the structs in the ironrdp-svc crate were used in some other places which had to implement arbitrary, so without those checks wouldn't compile.
I will look into manual implementation for the connector.
Thanks

glamberson:
Hi @razzat008,

Thanks for taking this on. The `arbitrary` feature is a meaningful unblocker for the fuzzing roadmap @CBenoit outlined in #1120 and the four sub-issues stacked under it.

I have been reading through the diff in preparation for picking up some of the structured-fuzzing follow-on work, and I wanted to consolidate the open items in one place so this PR has a clear path to land.

## Outstanding technical items

1. **Architectural placement of `Arbitrary` in `ironrdp-svc`.** @CBenoit's last review suggested moving these impls into manual implementations in `ironrdp-connector` instead, since `ironrdp-svc` is unlikely to need fuzz support directly. This is the substantive directional change still pending.

2. **`StaticChannelSet` Arbitrary constructor fallback.** The current `Ok(StaticChannelSet::new())` approach produces only the empty channel set under fuzzing, which means the connector state machine never sees a populated channel set during structured fuzz runs. If the intent is to exercise channel-dependent code paths under fuzz, this would need to populate the set with some plausible channels. If not, leaving it as a documented limitation in the PR description is fine.

3. **`Option<Arc<dyn LicenseCache>>` using `#[arbitrary(default)]`.** Same trade-off shape as item 2: license-cache-dependent paths never execute under fuzz with this approach. Worth recording the decision in the PR description either way.

4. **`PhantomData` in `license_exchange.rs`.** You already acknowledged in the thread that this is unnecessary. Worth removing in the next push.

5. **The `// MIGHT NEED CHANGE` markers.** Useful to keep visible during review, but should be resolved before merge.

6. **Feature flag wiring for `no_std + alloc`.** The umbrella in #1121 specifies the wiring should be `arbitrary = ["alloc", "dep:arbitrary"]` for crates that support `no_std + alloc`. The current wiring in `ironrdp-connector/Cargo.toml` is `arbitrary = ["dep:arbitrary", "ironrdp-pdu/arbitrary"]`, which omits the `alloc` feature dependency. Worth checking whether `ironrdp-connector` actually supports `no_std + alloc` (it looks std-only on a quick read) and adjusting the feature line if needed.

7. **Rebase against master and `cargo fmt`.** The branch is in `CONFLICTING` state due to drift since February, and the format check is failing. These would need to be cleared.

## Coordination

If you are still actively working on this and just bandwidth-limited, that is fine on my side. If you would welcome a collaborator on the architectural reshape in item 1 and the rebase in item 7 while you stay as the main author, I can chip in there. Either way works for me. The structured-fuzzing infrastructure work in #1122 and #1124 is downstream of this PR landing.

Let me know what would be most useful.

glamberson:
Hi @razzat008,

I went ahead and filed #1272 to address #1121 directly. The downstream fuzzing work in #1122 through #1124 was blocked on this landing, and the 2026-05-11 ping here did not get a response, so I treated the path as clear to step in.

Your exploration in this PR was the starting point. The decisions in #1272 about the `ChannelName` manual impl shape, the `Arc<dyn LicenseCache>` trade-off, the `StaticChannelSet` skip, and the feature-cascade wiring all build on the questions you and @CBenoit raised in this thread. Thanks for putting the original work in.

CBenoit:
Closing this PR since https://github.com/Devolutions/IronRDP/pull/1272 was merged earlier today.

Thank you for the initial exploration!

PR #1132, opened 2026-02-27
feat(session): handle slow-path graphics and pointer updates

glamberson:
## Summary

- Add slow-path update parsing module (`ironrdp-pdu::slow_path`) that decodes the framing headers for graphics (`updateType` u16) and pointer (`messageType` u16 + pad u16) slow-path updates
- Extract `process_bitmap_update()` and `process_pointer_update()` as public methods on `fast_path::Processor`, eliminating duplication between delivery paths
- Route `ShareDataPdu::Update` and `ShareDataPdu::Pointer` through the shared bitmap/pointer pipeline instead of hitting the catch-all error in `x224::Processor::process_io_channel()`

Previously, any server sending slow-path graphics or pointer updates (e.g. XRDP login screen before autologon, or servers that don't negotiate fast-path output) would trigger `unhandled PDU: Update PDU` and kill the session. The payload structures are byte-for-byte identical between slow-path and fast-path; only the framing differs.

Unsupported slow-path types (drawing orders, palette) log warnings instead of terminating the connection.

## Relates to

- #314 (Cannot connect to XRDP): slow-path updates are one of several interop gaps preventing XRDP connections. This PR addresses the session-layer update handling; other issues (TLS negotiation, bitmap row padding, DVC version) are covered by separate PRs.

## Test plan

- `cargo xtask check fmt -v` passes
- `cargo xtask check lints -v` passes
- `cargo xtask check tests -v` passes (all existing tests, no regressions)
- Tested against XRDP 0.10.1 (Debian 13): master crashes with `unhandled PDU: "Update PDU"`, this branch handles the slow-path Update PDU and progresses further into the session. A subsequent crash (`BitmapData::decode` size mismatch) is a pre-existing XRDP interop issue with compressed bitmaps, unrelated to this PR.

### DIFF
--- a/crates/ironrdp-pdu/src/basic_output/slow_path.rs
+++ b/crates/ironrdp-pdu/src/basic_output/slow_path.rs
@@ -11,7 +11,7 @@
 
 use ironrdp_core::{Decode as _, DecodeResult, ReadCursor, ensure_size, invalid_field_err};
 
-use super::bitmap::BitmapUpdateData;
+use super::bitmap::{BitmapData, BitmapUpdateData};
 use super::pointer::{
     CachedPointerAttribute, ColorPointerAttribute, LargePointerAttribute, PointerAttribute, PointerPositionAttribute,
     PointerUpdateData,
@@ -50,7 +50,14 @@ pub fn read_graphics_update_type(src: &mut ReadCursor<'_>) -> DecodeResult<Graph
 /// The cursor must be positioned right after the `updateType` field
 /// (i.e. already consumed by [`read_graphics_update_type`]).
 pub fn decode_slow_path_bitmap<'a>(src: &mut ReadCursor<'a>) -> DecodeResult<BitmapUpdateData<'a>> {
-    BitmapUpdateData::decode(src)
+    // Read numberRectangles directly; updateType was already consumed by the dispatcher.
+    ensure_size!(in: src, size: 2);
+    let rectangle_count = usize::from(src.read_u16());
+    let mut rectangles = Vec::with_capacity(rectangle_count);
+    for _ in 0..rectangle_count {
+        rectangles.push(BitmapData::decode(src)?);
+    }
+    Ok(BitmapUpdateData { rectangles })
 }
 
 // --- Pointer updates ([MS-RDPBCGR] 2.2.9.1.1.4) ---
--- a/crates/ironrdp-session/src/active_stage.rs
+++ b/crates/ironrdp-session/src/active_stage.rs
@@ -416,6 +416,8 @@ fn process_slow_path_graphics(
             warn!("Slow-path palette update not supported (8bpp)");
             Ok(Vec::new())
         }
+        // Synchronize is an artifact from the T.128 multipoint protocol
+        // and carries no data. Safe to ignore.
         GraphicsUpdateType::Synchronize => {
             debug!("Ignoring slow-path synchronize update");
             Ok(Vec::new())


CBenoit (commented):
Thank you! I think the direction is good. I’ll review for deeply next week. Also waiting on the manual tests 👀

glamberson:
> Thank you! I think the direction is good. I’ll review for deeply next week. Also waiting on the manual tests 👀

This is one of the identified residual issues from #314 which is resolved.  I completed my testing and noted it above. Thanks.

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -334,6 +349,11 @@ impl TryFrom<x224::ProcessorOutput> for ActiveStageOutput {
             }
             x224::ProcessorOutput::DeactivateAll(cas) => Ok(Self::DeactivateAll(cas)),
             x224::ProcessorOutput::MultitransportRequest(pdu) => Ok(Self::MultitransportRequest(pdu)),
+            // GraphicsUpdate and PointerUpdate are consumed in ActiveStage::process()
+            // before reaching this conversion.
+            x224::ProcessorOutput::GraphicsUpdate(_) | x224::ProcessorOutput::PointerUpdate(_) => {
+                unreachable!("slow-path updates are handled directly in ActiveStage::process()")
+            }
```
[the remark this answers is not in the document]
glamberson: Yup. Fixed by replacing unreachable!() with a SessionError return.. The invariant holds today but a defensive error is better than a panic if future refactors break the assumption.

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -362,3 +382,43 @@ impl core::fmt::Display for GracefulDisconnectReason {
...
+    data: &[u8],
+) -> SessionResult<Vec<UpdateKind>> {
+    let mut src = ReadCursor::new(data);
+    let update_type = slow_path::read_graphics_update_type(&mut src).map_err(SessionError::decode)?;
+
+    match update_type {
+        GraphicsUpdateType::Bitmap => {
+            let bitmap = slow_path::decode_slow_path_bitmap(&mut src).map_err(SessionError::decode)?;
+            fast_path_processor.process_bitmap_update(image, bitmap)
+        }
```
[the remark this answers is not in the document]
glamberson: This an interesting pre-existing issue.  Slow-path bulk compression isn't wired in IronRDP today.. SharedDataHeader::decode exposes compression flags but discards them before the payload reaches this layer. To stay in scope and avoid not coding something I'd like to think about more carefully, adding a TODO at the x224 dispatch site where the decompression wold need ot occur. The main slow-path servers (XRDP) don't compress these updates in practice, so this is safe to defer.

crates/ironrdp-pdu/src/basic_output/slow_path.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/// Decode a slow-path bitmap update.
+///
+/// The cursor must be positioned right after the `updateType` field
+/// (i.e. already consumed by [`read_graphics_update_type`]).
+pub fn decode_slow_path_bitmap<'a>(src: &mut ReadCursor<'a>) -> DecodeResult<BitmapUpdateData<'a>> {
+    BitmapUpdateData::decode(src)
+}
+
```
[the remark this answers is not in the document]
glamberson: Yup. I'm adding slow-path decode tests in the testsuite covering GraphicsUpdateType parsing, bitmap decode, and pointer messageType dispatch including error cases. for unknown types and short buffers.

crates/ironrdp-pdu/src/basic_output/slow_path.rs
```diff
@@ -0,0 +1,126 @@
...
+        )),
+    }
+}
+
+/// Decode a slow-path bitmap update.
+///
+/// The cursor must be positioned right after the `updateType` field
+/// (i.e. already consumed by [`read_graphics_update_type`]).
+pub fn decode_slow_path_bitmap<'a>(src: &mut ReadCursor<'a>) -> DecodeResult<BitmapUpdateData<'a>> {
+    BitmapUpdateData::decode(src)
```
CBenoit: issue: It seems `BitmapUpdateData::decode` expects and decodes the `updateType` field… so we’re consuming that field twice I think?

```rust
impl<'de> Decode<'de> for BitmapUpdateData<'de> {
    fn decode(src: &mut ReadCursor<'de>) -> DecodeResult<Self> {
        ensure_fixed_part_size!(in: src);

        let update_type = BitmapFlags::from_bits_retain(src.read_u16()); // <-- Here
        if !update_type.contains(BitmapFlags::BITMAP_UPDATE_TYPE) {
            return Err(invalid_field_err!("updateType", "invalid update type"));
        }

        let rectangle_count = usize::from(src.read_u16());
        let mut rectangles = Vec::with_capacity(rectangle_count);

        for _ in 0..rectangle_count {
            rectangles.push(BitmapData::decode(src)?);
        }

        Ok(Self { rectangles })
    }
}
```

CBenoit: suggestion: A simple fix could be to modify `decode_slow_path_bitmap` so it does not call `BitmapUpdateData::decode`. After `read_graphics_update_type` consumes `updateType`, a dedicated inner decoder should read `numberRectangles` directly and iterate `BitmapData::decode` calls.

```rust
pub fn decode_slow_path_bitmap<'a>(src: &mut ReadCursor<'a>) -> DecodeResult<BitmapUpdateData<'a>> {
    // Cursor is already past updateType; read numberRectangles directly.
    ensure_size!(in: src, size: 2);
    let rectangle_count = usize::from(src.read_u16());
    let mut rectangles = Vec::with_capacity(rectangle_count);
    for _ in 0..rectangle_count {
        rectangles.push(BitmapData::decode(src)?);
    }
    Ok(BitmapUpdateData { rectangles })
}
```

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -345,6 +360,11 @@ impl TryFrom<x224::ProcessorOutput> for ActiveStageOutput {
             x224::ProcessorOutput::DeactivateAll(cas) => Ok(Self::DeactivateAll(cas)),
             x224::ProcessorOutput::MultitransportRequest(pdu) => Ok(Self::MultitransportRequest(pdu)),
             x224::ProcessorOutput::AutoDetect(request) => Ok(Self::AutoDetect(request)),
+            // GraphicsUpdate and PointerUpdate are consumed in ActiveStage::process()
+            // before reaching this conversion.
+            x224::ProcessorOutput::GraphicsUpdate(_) | x224::ProcessorOutput::PointerUpdate(_) => Err(
+                SessionError::general("slow-path graphics/pointer updates should be handled before this conversion"),
+            ),
```
CBenoit, line 367: nitpick: `TryFrom` invariant is structurally unenforced.

The comment says these are "consumed before reaching this conversion.", which is true inside `ActiveStage::process()`, but it is not enforced structurally. `GraphicsUpdate` and `PointerUpdate` are public variants. Maybe `ProcessorOutput` should be made `pub(crate)`?

crates/ironrdp-session/src/active_stage.rs
```diff
@@ -373,3 +393,43 @@ impl core::fmt::Display for GracefulDisconnectReason {
...
+            Ok(Vec::new())
+        }
+        GraphicsUpdateType::Palette => {
+            warn!("Slow-path palette update not supported (8bpp)");
+            Ok(Vec::new())
+        }
+        GraphicsUpdateType::Synchronize => {
+            debug!("Ignoring slow-path synchronize update");
+            Ok(Vec::new())
+        }
```
CBenoit, line 424: suggestion: No functional issue, but I think it wouldbe good to add a doc comment noting that this is "an artifact from the T.128 protocol and should be ignored." (as per FreeRDP)

crates/ironrdp-pdu/src/basic_output/slow_path.rs
glamberson: You're right. The slow-path dispatcher consumes updateType for routing, but BitmapUpdateData::decode reads it again internally. Applied your suggested fix.

crates/ironrdp-session/src/active_stage.rs
glamberson: ProcessorOutput is referenced directly in ironrdp-testsuite-core (autodetect tests), so pub(crate) would break compilation there. The defensive error arm is the pragmatic option given the current visibility. 

Happy to revisit if the test suite access changes.

crates/ironrdp-session/src/active_stage.rs
glamberson: Added.

PR #1142, opened 2026-03-03
feat(dvc): add DvcChannelListener for multi-instance DVC support

### LINKED ISSUE #1135: [ironrdp-dvc] Support multiple DVC channel instances with the same name
MS-RDPEUSB (and potentially other protocols) requires opening multiple DVC channels that share the same channel name but have distinct channel IDs. The current DynamicChannelSet in ironrdp-dvc uses channel name as its primary key, which makes this impossible.

**Depends on:** nothing

### Tasks
- Add a DvcChannelListener trait (factory) to lib.rs that produces a new DvcClientProcessor instance per DYNVC_CREATE_REQ:
```rust
pub trait DvcChannelListener: Send {
  fn channel_name(&self) -> &str;

  /// Called for each new DYNVC_CREATE_REQ for this name.
  fn on_channel_creation(&mut self, channel_id: u32) -> Box<dyn DvcClientProcessor + Send>;
}
```
- Refactor DynamicChannelSet to store entries keyed by DynamicChannelId rather than DynamicChannelName; maintain a name → listener map for creation
- Update DrdynvcClient::process() to call the listener factory on DYNVC_CREATE_REQ instead of looking up a pre-registered processor by name
- Add DrdynvcClient::with_listener() / attach_listener() alongside the existing with_dynamic_channel() / attach_dynamic_channel() (keep the old API for single-instance channels)
- Add a take_new_channels() drain hook to DvcServerProcessor in server.rs; update DrdynvcServer::process() to drain it after each process() call and emit the corresponding DYNVC_CREATE_REQ PDUs

```rust
// Add a post-process drain hook on DvcServerProcessor:

pub trait DvcServerProcessor: DvcProcessor {
    /// Called by DrdynvcServer after each process().
    /// Return new processor instances to be opened as new DVC channels.
    fn take_new_channels(&mut self) -> Vec<Box<dyn DvcServerProcessor + Send>> {
        Vec::new()
    }
}

// Then in DrdynvcServer::process(), after calling c.processor.process(...):

// Drain any new channels the processor wants to open.
for new_proc in c.processor.take_new_channels() {
    let id = self.dynamic_channels.insert(DynamicChannel::new_boxed(new_proc));
    let channel_id = u32::try_from(id).expect("...");
    let channel = &self.dynamic_channels[id];
    let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(
        channel_id,
        channel.processor.channel_name().into(),
    ));
    self.dynamic_channels[id].state = ChannelState::Creation;
    resp.push(as_svc_msg_with_flag(req)?);
}

// The URBDRC control processor would then:

fn process(&mut self, channel_id: u32, payload: &[u8]) -> PduResult<Vec<DvcMessage>> {
    // ...
    self.pending_new_channels.push(Box::new(UsbDeviceChannel::new()));
    Ok(vec![])
}

fn take_new_channels(&mut self) -> Vec<Box<dyn DvcServerProcessor + Send>> {
    self.pending_new_channels.drain(..).collect()
}
```

EDIT: See this comment: https://github.com/Devolutions/IronRDP/issues/1135#issuecomment-3992215033

### Acceptance criteria
- A channel name can have multiple simultaneous open instances, each with its own independent processor and fragmentation buffer
- Existing single-instance channels (displaycontrol, cliprdr, etc.) continue to work unchanged via with_dynamic_channel()
- Server can open new DVC channels mid-session in response to processor requests
- Unit tests covering multi-instance creation, data routing by channel ID, and close handling

uchouT:
Resolves #1135 
- implement the client-side without breaking the current API signatures by introducing `OnceListener` wrapper for pre-registered DVCs.
- implement server side with a new method `create_channel` for `DrdynvcServer`.
---

## older discussion
I try to implement the client-side without breaking the current API signatures by introducing `OnceListener` wrapper for pre-registered DVCs. But `get_dvc_by_type_id`'s behavior has changed, it returns None until CreateRequest arrives.

This is unavoidable — the old 1:1 name→channel mapping cannot support multiple channels with the same name. The ChannelId is assigned by the server at CreateRequest time, so the DVC cannot exist before that point. And this behavior follows the RDPEDYC's expectation.

@elmarco @CBenoit Does this approach acceptable and make sense?

### DIFF
--- a/crates/ironrdp-dvc/src/client.rs
+++ b/crates/ironrdp-dvc/src/client.rs
@@ -22,6 +22,7 @@ pub trait DvcClientProcessor: DvcProcessor {}
 
 pub trait DvcChannelListener: Send {
     fn channel_name(&self) -> &str;
+
     /// Called for each incoming DYNVC_CREATE_REQ matching this name.
     /// Return `None` to reject (NO_LISTENER).
     fn create(&mut self) -> Option<Box<dyn DvcProcessor>>;


crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,44 @@ use crate::pdu::{
...
+}
+
+impl DvcChannelListener for OnceListener {
+    fn channel_name(&self) -> &str {
+        self.inner
+            .as_ref()
+            .expect("channel name called after created")
+            .channel_name()
+    }
+    fn create(&mut self) -> Option<Box<dyn DvcProcessor + Send>> {
```
CBenoit: nitpick/style: We usually insert an extra newline between functions

```suggestion
    }

    fn create(&mut self) -> Option<Box<dyn DvcProcessor + Send>> {
```

crates/ironrdp-dvc/src/client.rs
```diff
@@ -51,22 +88,30 @@ impl DrdynvcClient {
         }
     }
 
-    // FIXME(#61): it’s likely we want to enable adding dynamic channels at any point during the session (message passing? other approach?)
-
     #[must_use]
+    /// with a pre-registered dynamic virtual channel
     pub fn with_dynamic_channel<T>(mut self, channel: T) -> Self
```
CBenoit: style: The docstring generally comes first (keep the #[…] attributes close to the `pub fn foo(…)` definition)
```suggestion
    /// Inserts a pre-registered dynamic virtual channel
    #[must_use]
    pub fn with_dynamic_channel<T>(mut self, channel: T) -> Self
```

crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,44 @@ use crate::pdu::{
...
+    fn channel_name(&self) -> &str {
+        self.inner
+            .as_ref()
+            .expect("channel name called after created")
+            .channel_name()
+    }
+    fn create(&mut self) -> Option<Box<dyn DvcProcessor + Send>> {
+        self.inner.take()
+    }
+}
```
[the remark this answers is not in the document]
uchouT: There's no need to cache for safe. Because `OnceListener` is private, and is guaranteed by me to not panic.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -121,6 +119,27 @@ impl DrdynvcServer {
...
+        let mut dvc = DynamicChannel::new(channel);
+        dvc.state = ChannelState::Creation;
+
+        let id = self.dynamic_channels.insert(dvc);
+        // The slab index is used as the DVC channel ID (a u32).
+        let channel_id = u32::try_from(id).expect("DVC channel count should not exceed u32::MAX");
+
+        let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
+        as_svc_msg_with_flag(req)
+    }
```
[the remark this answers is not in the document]
uchouT: This is by design. Newly created channel shouldn't map type_id to channel_id, because the newly added channel's type may conflict with the pre registered channels.

crates/ironrdp-dvc/src/client.rs
```diff
@@ -182,6 +219,97 @@ impl SvcProcessor for DrdynvcClient {
...
+        self.active_channels.get(&id)
+    }
+
+    fn get_by_channel_id_mut(&mut self, id: DynamicChannelId) -> Option<&mut DynamicVirtualChannel> {
+        self.active_channels.get_mut(&id)
+    }
+
+    fn remove_by_channel_id(&mut self, id: DynamicChannelId) {
+        self.active_channels.remove(&id);
+    }
```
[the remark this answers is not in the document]
uchouT: Fixed in da1b438

uchouT (commented):
Addressed Copilot's review

crates/ironrdp-dvc/src/client.rs
```diff
@@ -51,22 +88,39 @@ impl DrdynvcClient {
         }
     }
 
-    // FIXME(#61): it’s likely we want to enable adding dynamic channels at any point during the session (message passing? other approach?)
-
+    /// with a pre-registered dynamic virtual channel
```
CBenoit: nitpick: Docstrings could be polished a little bit more.

Example:
```rust
/// Registers a pre-initialized dynamic virtual channel with the DrdynvcClient,
/// making it available for immediate use when the session starts.
```

Note that the first paragraph is the summary of the item, and is displayed in top pages of documentation.

Also an interesting read: https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments

But no need to overdo it 🙂

crates/ironrdp-dvc/src/client.rs
```diff
@@ -51,22 +88,39 @@ impl DrdynvcClient {
...
+    #[must_use]
+    pub fn with_listener<T>(mut self, listener: T) -> Self
+    where
+        T: DvcChannelListener + 'static,
+    {
+        self.dynamic_channels.insert_listener(listener);
         self
     }
 
+    /// Doesn't support type id look up
```
CBenoit: nitpick: Related to the previous comment, but this kind of docstring sounds like they’re lacking a bit of context, although that’s not a big deal here.

crates/ironrdp-dvc/src/client.rs
```diff
@@ -51,22 +88,39 @@ impl DrdynvcClient {
...
+    {
+        self.dynamic_channels.insert_listener(listener);
+    }
+
     pub fn attach_dynamic_channel<T>(&mut self, channel: T)
     where
         T: DvcProcessor + 'static,
     {
-        self.dynamic_channels.insert(channel);
+        self.dynamic_channels.insert_dvc(channel);
```
CBenoit: naming: Is there a reason for renaming `insert` to `insert_dvc`? I think this is a bit redundant since the type is already `DynamicChannelSet` and typically the variable name is something like `dynamic_channels`.

uchouT: Hi, because `DynamicChannelSet` now supports two kinds of insert manipulation: insert a listener, insert a pre-registered dynamic channel. And as you've mentioned, the current variable name `dynamic_channels` is kind of confusing, if I should rename it to `dynamic_channel_sets` or something else?

crates/ironrdp-dvc/src/client.rs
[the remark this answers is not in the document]
uchouT: Updated in d9f90f40

uchouT: Updated in d9f90f40

crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,44 @@ use crate::pdu::{
...
+
+pub type DynamicChannelListener = Box<dyn DvcChannelListener>;
+
+/// For pre-registered Dvc
+struct OnceListener {
+    inner: Option<Box<dyn DvcProcessor + Send>>,
+}
+
+impl OnceListener {
+    fn new(dvc_processor: impl DvcProcessor) -> Self {
```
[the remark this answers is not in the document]
uchouT: Updated in d9f90f40

crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,44 @@ use crate::pdu::{
...
+pub trait DvcChannelListener: Send {
+    fn channel_name(&self) -> &str;
+    /// Called for each incoming DYNVC_CREATE_REQ matching this name.
+    /// Return `None` to reject (NO_LISTENER).
+    fn create(&mut self) -> Option<Box<dyn DvcProcessor + Send>>;
+}
+
+pub type DynamicChannelListener = Box<dyn DvcChannelListener>;
+
+/// For pre-registered Dvc
```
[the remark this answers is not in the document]
uchouT: Updated in d9f90f40

CBenoit:
Are you sure [this](https://github.com/Devolutions/IronRDP/pull/1142#discussion_r2926137311) was a problem? It seems the compilation was successful

crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,45 @@ use crate::pdu::{
...
+    /// Return `None` to reject (NO_LISTENER).
+    fn create(&mut self) -> Option<Box<dyn DvcProcessor + Send>>;
+}
+
+pub type DynamicChannelListener = Box<dyn DvcChannelListener>;
+
+/// For pre-registered DVC
+struct OnceListener {
+    inner: Option<Box<dyn DvcProcessor + Send>>,
+}
```
CBenoit: nitpick: The `Send` bound is already implied by `DvcProcessor`

uchouT:
> Are you sure [this](https://github.com/Devolutions/IronRDP/pull/1142#discussion_r2926137311) was a problem? It seems the compilation was successful

Yes, the compilation was actually fine. I updated it mostly to keep the signature consistent with other methods. I removed the Send bound as you mentioned, and added the 'static bound (which Copilot also suggested, and it makes sense here).

Also, regarding [this](https://github.com/Devolutions/IronRDP/pull/1142/changes#r2926283479) discussion, does my explanation make sense to you?

crates/ironrdp-dvc/src/client.rs
```diff
@@ -182,6 +230,106 @@ impl SvcProcessor for DrdynvcClient {
...
+    fn insert_dvc<T: DvcProcessor + 'static>(&mut self, channel: T) {
+        let name = channel.channel_name().to_owned();
+        self.listeners.insert(
+            name,
+            ListenerEntry {
+                listener: Box::new(OnceListener::new(channel)),
+                type_id: Some(TypeId::of::<T>()),
+            },
+        );
+    }
```
CBenoit: Upon thinking on this again, I think I would suggest the naming `register_listener` and `register_once` for these functions.

`register_once` makes it immediately clear the processor will only be used for a single channel creation, while `register_listener` signals the repeatable factory pattern.

crates/ironrdp-dvc/src/client.rs
CBenoit: Oh, I see! I think `dynamic_channels` is fine, the `set` part is purely technical (set = no duplicate, list = may have duplicates).

crates/ironrdp-dvc/src/client.rs
```diff
@@ -13,10 +16,45 @@ use crate::pdu::{
     CapabilitiesResponsePdu, CapsVersion, ClosePdu, CreateResponsePdu, CreationStatus, DrdynvcClientPdu,
     DrdynvcServerPdu,
 };
-use crate::{DvcProcessor, DynamicChannelSet, DynamicVirtualChannel, encode_dvc_messages};
+use crate::{DvcProcessor, DynamicChannelId, DynamicChannelName, DynamicVirtualChannel, encode_dvc_messages};
 
 pub trait DvcClientProcessor: DvcProcessor {}
 
+pub trait DvcChannelListener: Send {
+    fn channel_name(&self) -> &str;
```
CBenoit, line 24:
```suggestion
    fn channel_name(&self) -> &str;

```

PR #1143, opened 2026-03-03
fix(acceptor): upgrade sspi, use NTLM when no Kerberos

ramnes:
Upgrade sspi to incorporate latest changes. The new version introduces a real SPNEGO implementation in `ServerMode::Negotiate`, but some RDP clients seem to send raw NTLM tokens in CredSSP rather than SPNEGO-wrapped ones, which doesn't seem supported by sspi yet? In the meantime, we can use `ServerMode::Ntlm` directly when Kerberos is disabled, so that we maintain compatibility with previous sspi behavior.

Note: I can't get the dependencies right, looks like there's a picky rc.22 / getrandom conflict; any help appreciated here!

### DIFF
--- a/crates/ironrdp-acceptor/src/credssp.rs
+++ b/crates/ironrdp-acceptor/src/credssp.rs
@@ -3,7 +3,7 @@ use ironrdp_connector::sspi::credssp::{
     CredSspServer, CredentialsProxy, ServerError, ServerMode, ServerState, TsRequest,
 };
 use ironrdp_connector::sspi::generator::{Generator, GeneratorState};
-use ironrdp_connector::sspi::negotiate::ProtocolConfig;
+use ironrdp_connector::sspi::ntlm::NtlmConfig;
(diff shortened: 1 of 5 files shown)


TheBestTvarynka:
Hi, yes, you are right. Let me add more context for others and for historical reasons :upside_down_face:

> but some RDP clients seem to send raw NTLM tokens in CredSSP rather than SPNEGO-wrapped ones

Yep. `mstsc` and `FreeRDP` do so. I was the one who refactored SPNEGO implementation. We needed to improve SPNEGO to fix these issues: https://github.com/Devolutions/sspi-rs/issues/433 and https://github.com/Devolutions/sspi-rs/issues/476. `sspi-rs` is not only used for RDP but also for SMB and other protocols. So, we needed proper SPNEGO support.

>  which doesn't seem supported by sspi yet?

It is supported, but the user (the caller) must specify it explicitly via `ClientMode::Ntlm` rather than `ClientMode::Negotiate`. And I see you do exactly this :+1: 

> We can use `ServerMode::Ntlm` directly when Kerberos is disabled

Currently, that's the only way to do so.

PR #1144, opened 2026-03-03
fix(pdu): replace all from_bits_truncate with from_bits_retain

### LINKED ISSUE #217: Preserve unknown bitmask bits values in PDUs (eliminate use of 'from_bits_truncate')

glamberson:
Completes the cleanup tracked in #217 by replacing all 65 remaining uses of from_bits_truncate with from_bits_retain across 8 crates (ironrdp-pdu, ironrdp-cliprdr, ironrdp-egfx, ironrdp-graphics, ironrdp-rdpsnd, ironrdp-server, ironrdp-testsuite-core).

from_bits_truncate silently discards unknown bits, which breaks the encode/decode round-trip property. This matters for fuzzing because a PDU that decodes and re-encodes should produce identical bytes. from_bits_retain preserves all bits, including those not yet defined in our bitflags types, so the round-trip property holds.

All 65 replacement sites were individually audited. Every one is a PDU parsing site where lenient, lossless decoding is the correct behavior. Pre-existing from_bits (strict) call sites were reviewed and left unchanged. Those handle security-sensitive fields (EncryptionMethod, BasicSecurityHeaderFlags, HttpExtendedAuth, license preamble, etc.) where rejecting unknown bits is intentional, and each already has documentation explaining why strictness is required there.

Notable sites reviewed in detail:

handler.rs:145, the cross-type cast from SyncToggleFlags (u32) to SynchronizeFlags (u8). The as u8 truncation is intentional and documented with a clippy suppression. from_bits_retain on the resulting u8 is correct because all four defined lock flags fit in the low byte.

x224.rs:173, a test literal 0x1F for ResponseFlags. All five defined flags are set with no unknown bits present. from_bits_retain is consistent with the rest of the codebase and behaves identically to from_bits for this value.

nego.rs (4 sites), security protocol negotiation for RequestFlags, ResponseFlags, and SecurityProtocol. These are PDU parsing sites. The RDP spec expects forward compatibility with extension bits in these fields. Security decisions are made by the connector layer, not the PDU decoder.

Resolves #217

### DIFF
--- a/benches/src/perfenc.rs
+++ b/benches/src/perfenc.rs
@@ -59,7 +59,7 @@ async fn main() -> Result<(), anyhow::Error> {
         OptCodec::QoiZ => update_codecs.set_qoiz(Some(0)),
     };
 
-    let mut encoder = UpdateEncoder::new(DesktopSize { width, height }, flags, update_codecs)
+    let mut encoder = UpdateEncoder::new(DesktopSize { width, height }, flags, update_codecs, 8 * 1024 * 1024)
         .context("failed to initialize update encoder")?;
 
     let mut total_raw = 0u64;
--- a/crates/ironrdp-rdpdr-native/src/nix/backend.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/backend.rs
@@ -631,14 +631,12 @@ fn make_create_drive_resp(
     file_id: u32,
 ) -> PduResult<Vec<SvcMessage>> {
     let io_response = DeviceIoResponse::new(device_io_request, NtStatus::SUCCESS);
-    let information = match create_disposation {
-        CreateDisposition::FILE_CREATE
-        | CreateDisposition::FILE_SUPERSEDE
-        | CreateDisposition::FILE_OPEN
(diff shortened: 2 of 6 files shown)


glamberson:
Thanks, done.

Cheers,
Greg

PR #1145, opened 2026-03-03
fix(rdpdr): model CreateDisposition as enum instead of bitflags

glamberson:
Found while auditing all from_bits_truncate sites for #217.

CreateDisposition values (FILE_SUPERSEDE through FILE_OVERWRITE_IF) are mutually exclusive integers 0 through 5, not combinable bit flags. Modeling them with the bitflags macro caused subtle correctness issues:

FILE_SUPERSEDE = 0 was equivalent to empty(), so any call to contains(FILE_SUPERSEDE) would always return true regardless of the actual disposition value. FILE_OPEN_IF = 3 appeared to bitwise-contain both FILE_OPEN (1) and FILE_CREATE (2), which is not what the protocol intends.

This converts CreateDisposition to a repr(u32) enum with TryFrom<u32> for decoding, matching the pattern already used by MajorFunction, DeviceType, and other protocol enums in the same crate. Unknown values are now rejected at decode time with a clear error instead of being silently retained as meaningless bit combinations.

The match in make_create_drive_resp (backend.rs) becomes exhaustive with this change, so the unreachable wildcard arm is removed.

Variant names are kept as SCREAMING_SNAKE to match the MS protocol spec constants (MS-SMB2 2.2.13) with an allow(non_camel_case_types) attribute.

### DIFF
--- a/crates/ironrdp-rdpdr-native/src/nix/backend.rs
+++ b/crates/ironrdp-rdpdr-native/src/nix/backend.rs
@@ -631,14 +631,12 @@ fn make_create_drive_resp(
     file_id: u32,
 ) -> PduResult<Vec<SvcMessage>> {
     let io_response = DeviceIoResponse::new(device_io_request, NtStatus::SUCCESS);
-    let information = match create_disposation {
-        CreateDisposition::FILE_CREATE
-        | CreateDisposition::FILE_SUPERSEDE
-        | CreateDisposition::FILE_OPEN
-        | CreateDisposition::FILE_OVERWRITE => Information::FILE_SUPERSEDED,
-        CreateDisposition::FILE_OPEN_IF => Information::FILE_OPENED,
-        CreateDisposition::FILE_OVERWRITE_IF => Information::FILE_OVERWRITTEN,
-        _ => Information::empty(),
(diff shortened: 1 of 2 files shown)


CBenoit (commented):
I’m sorry, this should be documented at a more obvious place, but I actually recommend we _don’t_ use enums for this kind of stuff, and the "parsing" should be infallible (no `TryFrom`): see [the full rationale here](https://github.com/Devolutions/IronRDP/tree/db2f40b5b0af66a4c83e0e075e2814467c060b1d/crates/ironrdp-pdu#enumeration-like-types-should-allow-resilient-parsing). (At least when it comes to stuff we receive on the wire; it’s otherwise better to restrict the number of states by using enums in typical Rust code.)

`MajorFunction`, `DeviceType` and other similar protocol enums in `ironrdp-pdu` or `ironrdp-rdpdr` are actually bad examples that should not be followed. (We should convert most of these to the NewType pattern too).

EDIT: There is an exception to this rule:

> Don’t forget that in some cases, the protocol specification explicitly states that other values MUST always be rejected. In such cases, an enum is deemed appropriate.

But make it clear with some inline documentation if that’s the case.

Issue #447, opened 2024-04-24
Devilish resize bug


ibeckermayer:
## Background
After adding protocol level support for dynamic resizes via the `DisplayControl` DVC in 

- https://github.com/Devolutions/IronRDP/pull/418
- https://github.com/Devolutions/IronRDP/pull/419
- https://github.com/Devolutions/IronRDP/pull/425

this was all actually hooked up in the `ironrdp-client` in https://github.com/Devolutions/IronRDP/pull/430. The initial attempt succeeded in allowing for dynamic resize, however after resizes performance of the client noticeably degraded after a resize or two. **This issue does not reproduce with FreeRDP**, meaning that it's a bug in IronRDP.

## Pointer settings and FreeRDP alignment
We tried many things to resolve the issue including:
- all variations of pointer settings in IronRDP
- Meticulously aligning IronRDP's [Capability Sets](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/7cc40be0-ceb0-4983-85d0-e08af4bec2a6) with FreeRDP's
- Meticulously aligning IronRDP's [Client MCS Connect Initial PDU with GCC Conference Create Request](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/db6713ee-1c0e-4064-a3b3-0fac30b4037b) with FreeRDP's

(IronRDP rev `f7b4f546650231ce345e9ee67f6ad29b2b93f937` is aligned with FreeRDP `ba8cf8cf2158018fb7abbedb51ab245f369be813`)

all to no avail.

We know the problem is related to the RemoteFX codepath -- when we switched to using bitmaps, it went away.

## Profiling

Comparing [flamegraphs](https://crates.io/crates/flamegraph) between an IronRDP session with a resize vs without did not reveal anything of relevance to this issue: [Download](https://github.com/Devolutions/IronRDP/files/15100279/ironrdp-profiling.zip).

[...]

PR #1148, opened 2026-03-04
fix(pdu): correct ShareDataHeader uncompressedLength calculation

glamberson:
## Summary

`ShareDataHeader::encode()` was calculating `uncompressedLength` as
`payload_size + 4` (adding pduType2, compressedType, and compressedLength
field sizes). Per [MS-RDPBCGR] 2.2.8.1.1.1.2, this field should contain
only the uncompressed payload size.

@pacmancoder confirmed this as a bug in May 2024
(https://github.com/Devolutions/IronRDP/issues/447#issuecomment-2093879507):
"pretty sure this is a bug on our side"

FreeRDP encodes only the payload size. Servers tolerate both values since
the decode path does not validate `uncompressedLength` against actual
payload size, but the current encoding does not match the spec.

## Changes

| File | Change |
|------|--------|
| `ironrdp-pdu/src/rdp/headers.rs` | Remove extra field sizes from uncompressedLength |
| `ironrdp-testsuite-core/src/rdp.rs` | Update 7 byte buffer fixtures to match corrected encoding |

## Verification

`cargo xtask check fmt`, `cargo xtask check lints`, and `cargo xtask check tests` all pass.

[MS-RDPBCGR]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/

Fixes part of #447

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -253,13 +253,7 @@ impl Encode for ShareDataHeader {
 
             write_padding!(dst, 1);
             dst.write_u8(self.stream_priority.as_u8());
-            dst.write_u16(cast_length!(
-                "uncompressedLength",
-                self.share_data_pdu.size()
-                    + PDU_TYPE_FIELD_SIZE
-                    + COMPRESSION_TYPE_FIELD_SIZE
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
LGTM!

PR #1150, opened 2026-03-05
fix(acceptor): skip credential check when server credentials are None

### LINKED ISSUE #1149: Acceptor rejects TLS connections when server credentials are None
### Summary

In TLS-only (non-NLA) connections, the acceptor's credential check in `connection.rs` unconditionally rejects clients when the server has not set credentials (`self.creds == None`). This makes it impossible for a server to operate in a "no authentication required" mode over TLS, because `None != Some(client_creds)` always evaluates to `true`, triggering `ServerDeniedConnection`.

A fix is submitted in the accompanying PR.

### Current Behavior

[`connection.rs:530-545`](https://github.com/Devolutions/IronRDP/blob/master/crates/ironrdp-acceptor/src/connection.rs#L530-L545):

```rust
if !protocol.intersects(SecurityProtocol::HYBRID | SecurityProtocol::HYBRID_EX) {
    let creds = client_info.client_info.credentials;

    if self.creds.as_ref() != Some(&creds) {
        // FIXME: How authorization should be denied with standard RDP security?
        // ...
        return Err(ConnectorError::general("invalid credentials"));
    }
}
```

This comparison has three cases:

| `self.creds` (server) | Client sends | Result | Expected? |
|---|---|---|---|
| `Some(creds_A)` | `creds_A` | ✅ Accept | Yes |
| `Some(creds_A)` | `creds_B` | ❌ Reject | Yes |
| `None` | anything | ❌ **Reject** | **No** — should accept |

When the server passes `None` to `set_credentials()`, the intent is "no credential requirement." But the comparison treats `None` as "no valid credentials exist," rejecting every client.

### Expected Behavior

When `self.creds` is `None`, the server has no credential requirement and the acceptor should skip the credential check entirely, accepting any client.

### MS-RDPBCGR Spec Context

The Client Info PDU credentials (UserName, Password, Domain) are defined in [MS-RDPBCGR §2.2.1.11](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/772d618e-b7d6-4cd0-b735-fa08af558f9d) as part of the Secure Settings Exchange phase. Per the spec:

- These fields carry **auto-logon data**, not protocol-level authentication tokens
- The `INFO_AUTOLOGON` flag controls whether the server attempts automatic logon using these credentials
- **The spec does not require the server to validate or reject credentials at the RDP protocol level** in non-NLA mode — validation is deferred to the OS logon subsystem (e.g., Winlogon on Windows)

In a Windows RDP server, these credentials are passed through to the logon screen; they are never validated by the RDP stack itself. The RDP connection proceeds regardless of what the client sends.

### How Other Implementations Handle This

**FreeRDP server** (`libfreerdp/core/peer.c`): Parses Client Info PDU credentials and passes them to an application callback (`Authenticate` / `PostConnect`). **No protocol-level rejection occurs.** The server application decides what to do with the credentials — accept, reject, or ignore. The RDP connection proceeds to licensing and capability exchange regardless.

**Windows RDP Server**: Credentials flow through to the Windows logon subsystem. The RDP protocol layer does not reject connections based on credential content. Even with wrong credentials, the RDP handshake completes and the user sees a logon failure at the Windows login screen.

### Current Workaround

Servers can work around this by passing `Some(Credentials { username: "", password: "", domain: None })` — but this only accepts clients that send *empty* credential strings. Any real RDP client (FreeRDP, mstsc, rdesktop) that sends a username/password will be rejected, which defeats the purpose.

### Related

- The existing `FIXME` comment on this code block ("How authorization should be denied with standard RDP security?") suggests this area was known to be incomplete
- This does not affect NLA/CredSSP connections, which have a separate authentication path

glamberson:
Fixes #1149

### Problem

In TLS-only (non-NLA) connections, the credential check at [`connection.rs:533`](https://github.com/Devolutions/IronRDP/blob/master/crates/ironrdp-acceptor/src/connection.rs#L533) unconditionally rejects clients when `self.creds` is `None`:

```rust
if self.creds.as_ref() != Some(&creds) {
    // ServerDeniedConnection
}
```

`None != Some(anything)` is always `true`, so every client is rejected when the server has no credential requirement. This makes it impossible to run a TLS-mode server that accepts connections without authentication.

### Fix

Guard the check so `None` means "no credential requirement — accept any client":

```rust
if self.creds.is_some() && self.creds.as_ref() != Some(&creds) {
    // ServerDeniedConnection
}
```

One-line change. Preserves existing behavior when credentials are set (exact match required). No API surface change. No impact on NLA/CredSSP connections.

### Spec justification

MS-RDPBCGR does not require the server to validate Client Info PDU credentials at the protocol level in non-NLA mode. These fields carry auto-logon data, not authentication tokens. Both Windows RDP Server and FreeRDP's server pass credentials through to the application layer without protocol-level rejection. See #1149 for full analysis.

### DIFF
--- a/crates/ironrdp-acceptor/src/connection.rs
+++ b/crates/ironrdp-acceptor/src/connection.rs
@@ -567,18 +567,20 @@ impl Sequence for Acceptor {
                 if !protocol.intersects(SecurityProtocol::HYBRID | SecurityProtocol::HYBRID_EX) {
                     let creds = client_info.client_info.credentials;
 
-                    if self.creds.is_some() && self.creds.as_ref() != Some(&creds) {
-                        // FIXME: How authorization should be denied with standard RDP security?
-                        // Since standard RDP security is not a priority, we just send a ServerDeniedConnection ServerSetErrorInfo PDU.
-                        let info = ServerSetErrorInfoPdu(ErrorInfo::ProtocolIndependentCode(
-                            ProtocolIndependentCode::ServerDeniedConnection,
-                        ));
+                    if let Some(expected) = &self.creds {
+                        if expected != &creds {
+                            // FIXME: How authorization should be denied with standard RDP security?
+                            // Since standard RDP security is not a priority, we just send a ServerDeniedConnection ServerSetErrorInfo PDU.
+                            let info = ServerSetErrorInfoPdu(ErrorInfo::ProtocolIndependentCode(
+                                ProtocolIndependentCode::ServerDeniedConnection,
+                            ));
 
-                        debug!(message = ?info, "Send");
+                            debug!(message = ?info, "Send");
 
-                        util::encode_send_data_indication(self.user_channel_id, self.io_channel_id, &info, output)?;
+                            util::encode_send_data_indication(self.user_channel_id, self.io_channel_id, &info, output)?;
 
-                        return Err(ConnectorError::general("invalid credentials"));
+                            return Err(ConnectorError::general("invalid credentials"));
+                        }
                     }
 
                     // Store credentials for later retrieval via AcceptorResult.


crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -530,7 +530,7 @@ impl Sequence for Acceptor {
                 if !protocol.intersects(SecurityProtocol::HYBRID | SecurityProtocol::HYBRID_EX) {
                     let creds = client_info.client_info.credentials;
 
-                    if self.creds.as_ref() != Some(&creds) {
+                    if self.creds.is_some() && self.creds.as_ref() != Some(&creds) {
```
CBenoit: nitpick: Just suggesting to use if-let chains which seems more appropriate in this case:
```suggestion
                    if let Some(creds) = self.creds.as_ref() && creds != &creds {
```
(Not sure if the `&` is required, but something close should works)

glamberson:
@CBenoit   Good question. The credentials field on Acceptor defaults to None via AcceptorBuilder, so yes — not calling set_credentials at all has the same effect as calling set_credentials(None).

I think this is the right default for a protocol library. The acceptor's job is connection establishment, not authorization.
That's the application's responsibility. MS-RDPBCGR Client Info credentials are auto-logon hints, not authentication tokens. Both Windows RDS and FreeRDP's server implementation pass them through to the application layer without protocol-level rejection.

The previous behavior (rejecting when None) was arguably worse: it made TLS-mode servers impossible without setting dummy empty credentials, which is confusing and doesn't actually provide any security.

That said, if you think an explicit opt-in would be cleaner, we could add something like AcceptorBuilder::allow_any_credentials() that must be called to get the None behavior. Happy to do that as a follow-up if you think it's warranted, though I'd argue the current semantics (None = no requirement, Some(creds) = exact match) are already clear.

Thanks.

crates/ironrdp-acceptor/src/connection.rs
glamberson: Sounds ok, but using nested instead of chain syntax per rust-lang/rust#53667...

Also the suggestion as written has a name collision (creds shadows the local), but the chain works with a rename: 

if let Some(expected) = &self.creds && expected != &creds {    

Will push that.                                                                                                            
   
Thanks.

PR #1152, opened 2026-03-06
fix(acceptor): send RDP_NEG_FAILURE on security protocol mismatch

glamberson:
## Summary

When the client and server have no common security protocol, the acceptor now sends a proper `RDP_NEG_FAILURE` PDU before returning an error, instead of dropping the TCP connection.

## Problem

FreeRDP 3.15+ changed its default X.224 Connection Request to only propose `PROTOCOL_HYBRID_EX | PROTOCOL_HYBRID` (no `PROTOCOL_SSL`). When connecting to a TLS-only IronRDP server (`RdpServerSecurity::Tls`), the acceptor finds no overlap between client and server protocols and returns a Rust error, which drops the connection without sending any X.224 response.

Per [MS-RDPBCGR 2.2.1.2.2], the server SHOULD respond with an `RDP Negotiation Failure` structure containing the appropriate failure code so the client can adjust and retry. Windows Server does this; IronRDP did not.

This affects any IronRDP-based server that doesn't support NLA/CredSSP when a FreeRDP 3.15+ client connects with default settings. Windows MSTSC still includes `PROTOCOL_SSL` in its request, so it's unaffected.

## Fix

In `InitiationSendConfirm`, when the bitwise AND of `requested_protocol & self.security` is empty, encode a `ConnectionConfirm::Failure` with the appropriate `FailureCode` into the output buffer before returning the error. The failure code is selected based on what the server requires:

- `SSL_REQUIRED_BY_SERVER` when the server has `PROTOCOL_SSL`
- `HYBRID_REQUIRED_BY_SERVER` when the server has `PROTOCOL_HYBRID`/`HYBRID_EX`

No new types needed. `ConnectionConfirm::Failure` and `FailureCode` were already fully implemented in `ironrdp-pdu` but unused by the acceptor.

## Tests

Three tests added to `ironrdp-acceptor`:

- `neg_failure_on_protocol_mismatch` -- TLS server vs HYBRID-only client, verifies `RDP_NEG_FAILURE` with `SSL_REQUIRED_BY_SERVER`
- `neg_failure_hybrid_required` -- HYBRID server vs SSL-only client, verifies `HYBRID_REQUIRED_BY_SERVER`
- `neg_success_when_protocols_match` -- confirms normal negotiation still works

## Reproduction

```
# FreeRDP 3.15 default (sends HYBRID|HYBRID_EX, no SSL)
xfreerdp3 /v:SERVER /u:user /p:pass /cert:ignore
# Server log: "failed to negotiate security protocol"
# Client sees: TCP reset instead of negotiation failure
```


crates/ironrdp-acceptor/src/connection.rs
```diff
@@ -758,3 +782,135 @@ fn create_gcc_blocks(
         multi_transport_channel: None,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use ironrdp_connector::{DesktopSize, Sequence as _, Written};
+    use ironrdp_core::{decode, WriteBuf};
```
CBenoit: suggestion: Place the tests in the appropriate testsuite crate

CBenoit (commented):
LGTM! Just need to address the comment about the tests

crates/ironrdp-acceptor/src/connection.rs
glamberson: Sure thing. Thanks.

PR #1153, opened 2026-03-06
chore: upgrade toolchain to 1.89.0 and migrate to Rust edition 2024

CBenoit:
- Bump pinned toolchain from 1.88.0 to 1.89.0 in rust-toolchain.toml
- Migrate all workspace crates from edition 2021 to edition 2024
- Run `cargo fix --edition` to auto-migrate source code for edition 2024 compatibility (unsafe in unsafe fn, tail expr drop order, etc.)
- Replace all `expr_2021` macro fragment specifiers with `expr` now that the workspace is fully on edition 2024
- Remove `keyword_idents_2024` lint (now enforced by the edition itself)
- Remove removed `clippy::string_to_string` lint (covered by implicit_clone)
- Fix new clippy lints from 1.93.0: `manual_is_multiple_of`, `useless_let_if_seq`, `manual_map`, `derivable_impls`, `redundant_imports`, `unused_crate_dependencies` (but didn’t bump MSRV)
- Drop `core::future::Future` unnecessary qualifications (Future is now in the Rust 2024 prelude)
- Remove direct `bytes` deps from ironrdp-futures and ironrdp-tokio (re-exported by ironrdp-async)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -128,7 +128,7 @@ Do not modify them unless specifically working on fixing their compilation.
 ## Coding Standards (The "Gold Standard")
 
 - **Language:** Rust (Edition 2024; toolchain pinned via `rust-toolchain.toml`)
-- **Toolchain baseline:** Rust `1.93.0`
+- **Toolchain baseline:** Rust `1.89.0`
 - **Formatter:** `rustfmt` (workspace config in `rustfmt.toml`)
 - **Lints:** Strict workspace lint policy (`[workspace.lints.rust]` and `[workspace.lints.clippy]` in root `Cargo.toml`)
 - **Error handling:** Prefer explicit, composable error messages following `STYLE.md`
--- a/xtask/src/check.rs
+++ b/xtask/src/check.rs
@@ -7,6 +7,7 @@ pub fn fmt(sh: &Shell) -> anyhow::Result<()> {
 
     if !output.status.success() {
         print!("{}", String::from_utf8_lossy(&output.stdout));
+        eprint!("{}", String::from_utf8_lossy(&output.stderr));
         anyhow::bail!("Bad formatting, please run 'cargo +stable fmt --all'");
     }
 


glamberson:
Hey, I want to flag a real downstream impact here.

**This PR as written would make lamco-rdp-server unbuildable on both Debian 13 and Fedora 43.** Those are my two most important distribution targets and I cannot lose either of them.

| Distribution | Rust version |
|-------------|-------------|
| Debian 13 (trixie) stable | 1.85.0 |
| Debian 13 backports | 1.90.0 |
| Fedora 43 (stable) | 1.90.0 |

No current non-rolling distribution ships Rust 1.93+. If the effective MSRV goes there, I lose Debian entirely (Debian 12 has Rust 1.63, no fallback) and I lose Fedora 43. I have packages awaiting sponsorship on both mentors.debian.net and RPM Fusion. Months of packaging work becomes dead on arrival.

To be clear about my priorities: I don't mind losing RHEL 9 (1.88). I've been headed in that direction myself, and my own project already declares `rust-version = "1.88"`. I review my MSRV weekly and push it as high as I can while still providing a product that works for the large majority of the Linux community. I'm not trying to be conservative for its own sake. But I absolutely cannot lose Fedora 43 or Debian 13. Debian is by far the more important of the two for me.

I went through the diff and the actual MSRV of this code is **1.85** (or **1.87** if you want to keep `is_multiple_of()`, which is the only post-1.85 feature I can find, ~10 call sites from clippy suggestions). Both are well within what every current distribution can handle. But without a `rust-version` field, downstream consumers have to treat the toolchain pin as the floor, and 1.94 breaks everyone.

I've opened #1156 to discuss MSRV policy more broadly. But for this PR specifically, I'd ask:

1. Add a `rust-version` field to the workspace `Cargo.toml` reflecting the actual minimum.
2. Either keep the modulo pattern (`rust-version = "1.85"`) or keep `is_multiple_of()` (`rust-version = "1.87"`). Either way, Fedora 43 and Debian 13 backports (both 1.90) remain viable, and Debian 13 base works for non-GUI consumers.
3. The toolchain pin to 1.94 for CI is fine. Just please separate it from the declared MSRV.

The edition 2024 migration itself is welcome. I just need the MSRV to reflect what the code actually requires so that my distribution packaging stays viable.

Happy to submit a PR adding `rust-version` if that helps.

Greg Lamberson
Lamco Development
greg@lamco.io

PR #1162, opened 2026-03-10
fix(web): disable clipboard polling loop on Firefox v127+

ymarcus93:
### Problem

On Firefox v127+, the clipboard monitoring loop repeatedly calls `navigator.clipboard.read()` without user activation, which causes Firefox to show a "Paste" permission dialog on every attempt. This dialog popup blocks _all_ user interaction with the remote desktop session, making the client effectively unusable. We can see the error spammed in the console:

```
Clipboard read operation is not allowed
Clipboard read request was blocked due to lack of user activation
```

### Root Cause

PR #935 introduced a clipboard polling loop that calls `navigator.clipboard.read()` every 100ms to detect local clipboard changes. This works on Chromium, which grants persistent clipboard-read permission via the Permissions API after a one-time user prompt.

A follow-up PR (#951) added the `ClipboardApiSupported` enum to handle older Firefox versions (<v127) that don't expose `clipboard.read()`/`clipboard.write()` at all (see [here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations) and [here](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#permission-aware_apis)), routing them to text-only fallback paths. 

The issue is that Firefox v127+ _does_ expose these APIs, so it gets classified as `ClipboardApiSupported.Full` and enters the _polling_ loop. 

The problem is that Firefox requires transient user activation for _every_ `clipboard.read()` call; there is no persistent permission grant. So if we enter the polling loop on v127+, every iteration will fail with a `NotAllowedError` as described above.

### Solution

We introduce a two-layer fix:

#### (1) Upfront detection via Permissions API probe

Before starting the monitoring loop, query `navigator.permissions.query({name: 'clipboard-read'})`. Chromium supports this and returns a valid `PermissionStatus`. Firefox will throw a `TypeError` that this permission is is not a valid value.

If the query throws, fall back to manual clipboard mode **immediately**; the polling loop is never started and no errors are shown.

#### (2) Runtime safety net in the polling loop

If a browser somehow passes the permission probe but still blocks reads at runtime, the existing catch block now detects `NotAllowedError` specifically, stops the loop via a flag, and switches to manual mode.

### DIFF
--- a/web-client/iron-remote-desktop/src/services/clipboard.service.ts
+++ b/web-client/iron-remote-desktop/src/services/clipboard.service.ts
@@ -34,7 +34,7 @@ export class ClipboardService {
         this.module = module;
     }
 
-    initClipboard() {
+    async initClipboard() {
         // Clipboard API is available only in secure contexts (HTTPS).
         if (!window.isSecureContext) {
             this.remoteDesktopService.emitWarningEvent('Clipboard is available only in secure contexts (HTTPS).');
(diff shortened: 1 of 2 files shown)


web-client/iron-remote-desktop/src/services/clipboard.service.ts
CBenoit, line 59: thought: It seems to me the fix should be here, instead of introducing a new `probeClipboardReadPermissionAndStartMonitoring`.

```ts
        try {
            await navigator.permissions.query({ name: 'clipboard-read' as PermissionName });
            // Full
        } catch {
            // TextOnly or TextOnlyServerOnly based on whether the functions are defined or not.
        }
```

With this in place, do you think it would still be necessary to keep the safety net?

web-client/iron-remote-desktop/src/services/clipboard.service.ts
```diff
@@ -326,7 +362,7 @@ export class ClipboardService {
                 this.lastClipboardMonitorLoopError = err;
             }
         } finally {
-            if (!get(isComponentDestroyed)) {
+            if (!this.clipboardMonitoringStopped && !get(isComponentDestroyed)) {
```
CBenoit: suggestion: Use a local variable instead of a field (which could be accessed and modified from anywhere in this class)

web-client/iron-remote-desktop/src/services/clipboard.service.ts
ymarcus93: Agreed, that seems like a better fix since we'll use the `ClipboardApiSupported` enum as I think it was originally intended. Let me try it out and test on our end before I push.

> With this in place, do you think it would still be necessary to keep the safety net?

The safety net is still valuable. I'm thinking about an edge case where the permission query succeeds, but then later `clipboard.read()` throws a `NotAllowedError` at clipboard invocation. I believe this could happen if the user denies the prompt or if the permission was already denied in site settings.

Without the safety net, the loop would spin every 100ms hitting the same error.

ymarcus93: I've pushed again and removed `probeClipboardReadPermissionAndStartMonitoring`. Also, I've noticed when testing in Firefox there is a way to get clipboard to work with the polling loop, if you set a special flag in `about:config`. I've added a comment about this in the latest commit.

CBenoit (approved):
LGTM! Thank you for improving the logic here!

Issue #1138, opened 2026-03-02
[ironrdp-rdpeusb] Server-side protocol state machine and extensibility traits


CBenoit:
Implement the server-side RDPEUSB protocol logic. The server creates virtual USB device representations and forwards IO requests. No specific backend (e.g. kernel USB, QEMU USB) is implemented here: the crate exposes traits so that downstream servers (e.g. qemu-rdp) can wire up their own USB stack.

**Depends on:**
- #1135
- #1136

### Tasks
- Define UsbServerBackend trait:

```rust
pub trait UsbServerBackend: Send {
  fn device_connected(&mut self, info: UsbDeviceInfo) -> PduResult<()>;
  fn device_disconnected(&mut self, interface_id: u32);
  fn submit_urb(&mut self, interface_id: u32, request: UsbRequest) -> PduResult<()>;
  fn cancel_urb(&mut self, interface_id: u32, request_id: u32);
}
```

- UrbdrcServerControl: implements DvcServerProcessor
- On CHANNEL_CREATED received: sends back CHANNEL_CREATED, then RIM_EXCHANGE_CAPABILITY_REQUEST
- On ADD_VIRTUAL_CHANNEL received: queues a new UsbDeviceServerChannel in take_new_channels()
- UsbDeviceServerChannel: implements DvcServerProcessor
- On start(): sends CHANNEL_CREATED, then REGISTER_REQUEST_CALLBACK
- On ADD_DEVICE received: decodes UsbDeviceInfo, calls backend.device_connected()
- Sends server IO requests via IO_CONTROL, TRANSFER_IN_REQUEST, etc. (driven by UsbServerBackend)
- On completions received: forwards to UsbServerBackend
- On RETRACT_DEVICE received: calls backend.device_disconnected()
- NoopUsbServerBackend: trivial no-op implementation (for testing / minimal setups)

### Acceptance criteria
- UsbServerBackend trait is sufficient for qemu-rdp to implement USB passthrough without depending on internals of ironrdp-rdpeusb
- Protocol-level unit tests with NoopUsbServerBackend
- Successful channel setup and device connection sequence tested end-to-end against the client state machine (#1137)

Issue #1137, opened 2026-03-02
[ironrdp-rdpeusb] Client-side protocol state machine


CBenoit:
Implement the client-side RDPEUSB protocol logic in ironrdp-rdpeusb. The client holds the physical USB device and responds to the server's IO requests.

**Depends on:**
- #1135
- #1136

### Tasks
- UrbdrcListener: implements DvcChannelListener; produces either a control channel or device channel processor depending on whether this is the first instance of "URBDRC" for the session
- UrbdrcControlChannel: implements DvcClientProcessor
- On start(): sends CHANNEL_CREATED, then RIM_EXCHANGE_CAPABILITY_REQUEST response
- On ADD_VIRTUAL_CHANNEL received: calls into the registered UsbClientBackend to detect available devices; queues device channel creation (surfaces via the listener factory on the next DYNVC_CREATE_REQ)
- UsbDeviceChannel: implements DvcClientProcessor
- On start(): sends CHANNEL_CREATED, then ADD_DEVICE
- Dispatches incoming server requests (IO_CONTROL, TRANSFER_IN_REQUEST, etc.) to UsbClientBackend
- Returns completions (URB_COMPLETION, IOCONTROL_COMPLETION) as DvcMessage
- On CANCEL_REQUEST: forwards cancellation to backend
- On RETRACT_DEVICE / channel close: teardown

### Acceptance criteria
- Successful channel setup sequence (control channel up, capability exchange done)
- Successful new device sequence (device channel up, ADD_DEVICE sent)
- IO request dispatched to backend and completion returned correctly
- Protocol-level unit tests with mock backend

uchouT:
# MS-RDPEUSB client/server state machine design draft

I am working on the MS-RDPEUSB state machine. Both sides have two kinds of
dynamic virtual channels:

- one control DVC for RDPEUSB session-level negotiation and device-channel
  creation
- one device DVC per redirected USB device

This draft describes the current design direction and the open questions where
I would like feedback.

## Server control DVC

`UrbdrcControlServer` is pre-registered in `DrdynvcServer`.

Expected flow:

1. `start()` sends `RIM_EXCHANGE_CAPABILITY_REQUEST`.
2. `process()` handles `RIM_EXCHANGE_CAPABILITY_RESPONSE`.
3. The server sends `CHANNEL_CREATED`.
4. After the corresponding client-side channel-created step, the control DVC is
   ready to process `ADD_VIRTUAL_CHANNEL`.

[...]

PR #1165, opened 2026-03-12
feat(rdpeusb): initial PDU module

### LINKED ISSUE #1136: [ironrdp-rdpeusb] PDU encoding/decoding
Create a new crate ironrdp-rdpeusb and implement all wire-format structures from MS-RDPEUSB using ironrdp-core's Encode/Decode traits.

**Depends on:** nothing

### Tasks
- Create crates/ironrdp-rdpeusb/ with a pdu module
- SHARED_MSG_HEADER (InterfaceId, Mask, MessageId, FunctionId)
- Interface manipulation: RIM_EXCHANGE_CAPABILITY_REQUEST, RIM_EXCHANGE_CAPABILITY_RESPONSE
- Control messages: ADD_VIRTUAL_CHANNEL, ADD_DEVICE, CHANNEL_CREATED, RETRACT_DEVICE
- Server IO requests: IO_CONTROL, INTERNAL_IO_CONTROL, TRANSFER_IN_REQUEST, TRANSFER_OUT_REQUEST, CANCEL_REQUEST, REGISTER_REQUEST_CALLBACK, QUERY_DEVICE_TEXT
- Client completions: IOCONTROL_COMPLETION, URB_COMPLETION, URB_COMPLETION_NO_DATA, QUERY_DEVICE_TEXT_RSP
- All TS_URB and TS_URB_RESULT structures (§2.2.9 and §2.2.10)
- USB_DEVICE_CAPABILITIES (§2.2.11)
- IO control codes (§2.2.12, §2.2.13)

### Acceptance criteria
- Round-trip encode/decode tests for every PDU type
- Spec-reference comments on every field

playbahn:
First draft.

Figured reviewing should get started on the things that are more or less complete right now, as the whole PR will be pretty chonky.

ping @Ben-Devolutions @elmarco 

[EDIT]: Why the "redundant" docs that more or less reword the protocol? I heavily use LSP hover, and getting info in editor is certainly better than changing applications.

[EDIT]: Closes #1136

### DIFF
--- a/crates/ironrdp-rdpeusb/Cargo.toml
+++ b/crates/ironrdp-rdpeusb/Cargo.toml
@@ -12,7 +12,14 @@ categories.workspace = true
 
 publish = false
 
+[features]
+default = []
+std = []
+
 [dependencies]
+ironrdp-core = { path = "../ironrdp-core", version = "0.1", features = ["alloc"] } # public
+ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.7", features = ["alloc"] } # public
+ironrdp-str = { path = "../ironrdp-str", version = "0.1" }
 
 [lints]
 workspace = true
--- a/crates/ironrdp-rdpeusb/src/pdu/usb_dev/mod.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/usb_dev/mod.rs
@@ -1,0 +1,1400 @@
+//! Messages specific to the [USB Device][1] interface.
+//!
+//! The USB device interface is used by the server to send IO-related requests to the client.
+//!
+//! [1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/034257d7-f7a8-4fe1-b8c2-87ac8dc4f50e
+
+use alloc::format;
+use alloc::vec::Vec;
+
+use ironrdp_core::{
+    DecodeError, DecodeOwned as _, DecodeResult, Encode, EncodeResult, ReadCursor, WriteCursor, ensure_fixed_part_size,
+    ensure_size, invalid_field_err, other_err, unsupported_value_err,
+};
+use ironrdp_pdu::utils::strict_sum;
+use ironrdp_str::prefixed::Cch32String;
+
+use crate::pdu::header::{InterfaceId, SharedMsgHeader};
+use crate::pdu::usb_dev::ts_urb::{TransferDirection, TsUrb};
+use crate::pdu::utils::{HResult, RequestId, RequestIdIoctl};
+#[cfg(doc)]
+use crate::pdu::{
+    completion::{IoControlCompletion, UrbCompletion, UrbCompletionNoData},
+    header::{FunctionId, Mask},
+    sink::AddDevice,
+};
+
+pub mod ts_urb;
+
+/// [\[MS-RDPEUSB\] 2.2.6.1 Cancel Request Message (CANCEL_REQUEST)][1] message.
+///
+/// Sent from the server to the client to cancel an outstanding IO request.
+///
+/// [1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/93912b05-1fc8-4a43-8abd-78d9aab65d71
+#[doc(alias = "CANCEL_REQUEST")]
+#[derive(Debug, PartialEq, Clone)]
+pub struct CancelRequest {
+    pub header: SharedMsgHeader,
+    pub req_id: RequestId,
+}
+
+impl CancelRequest {
+    const PAYLOAD_SIZE: usize = size_of::<RequestId>();
+
+    const FIXED_PART_SIZE: usize = Self::PAYLOAD_SIZE + SharedMsgHeader::SIZE_REQ;
+
+    pub fn decode(src: &mut ReadCursor<'_>, header: SharedMsgHeader) -> DecodeResult<Self> {
+        ensure_size!(in: src, size: Self::PAYLOAD_SIZE);
+        let req_id = src.read_u32();
+
+        Ok(Self { header, req_id })
+    }
+}
+
+impl Encode for CancelRequest {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
+        ensure_fixed_part_size!(in: dst);
+        self.header.encode(dst)?;
+        dst.write_u32(self.req_id);
+        Ok(())
+    }
+
+    fn name(&self) -> &'static str {
+        "CANCEL_REQUEST"
+    }
+
+    fn size(&self) -> usize {
+        Self::FIXED_PART_SIZE
+    }
+}
+
+/// [\[MS-RDPEUSB\] 2.2.6.2 Register Request Callback Message (REGISTER_REQUEST_CALLBACK)][1] message.
+///
+/// Sent from the server to the client in order to provide an interface ID for Request Completion
+/// to the client. This interface ID is to be used by the subsequent [`IoControlCompletion`],
+/// [`UrbCompletion`] and [`UrbCompletionNoData`] messages.
+///
+/// [1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/8693de72-5e87-4b64-a252-101e865311a5
+#[doc(alias = "REGISTER_REQUEST_CALLBACK")]
+#[derive(Debug, PartialEq, Clone)]
+pub struct RegisterRequestCallback {
+    pub header: SharedMsgHeader,
+    pub request_completion: Option<InterfaceId>,
+}
+
+impl RegisterRequestCallback {
+    pub fn decode(src: &mut ReadCursor<'_>, header: SharedMsgHeader) -> DecodeResult<Self> {
+        ensure_size!(in: src, size: size_of::<u32>());
+        let request_completion = match src.read_u32() {
+            0x0 => None,
+            0x1 => {
+                ensure_size!(in: src, size: InterfaceId::FIXED_PART_SIZE);
+                let interface = InterfaceId::try_from(src.read_u32()).map_err(|source| {
+                    let e: DecodeError =
+                        invalid_field_err!("REGISTER_REQUEST_CALLBACK::RequestCompletion", "more than 30 bits");
+                    e.with_source(source)
+                })?;
+                Some(interface)
+            }
+            _ => {
+                return Err(invalid_field_err!(
+                    "REGISTER_REQUEST_CALLBACK::NumRequestCompletion",
+                    "is not 0x0 or 0x1"
+                ));
+            }
+        };
+        Ok(Self {
+            header,
+            request_completion,
+        })
+    }
+}
+
+impl Encode for RegisterRequestCallback {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
+        ensure_size!(in: dst, size: self.size());
+        self.header.encode(dst)?;
+        if let Some(request_completion) = self.request_completion {
+            dst.write_u32(0x1);
+            dst.write_u32(request_completion.into());
+        } else {
+            dst.write_u32(0x0);
+        }
+
+        Ok(())
+    }
+
+    fn name(&self) -> &'static str {
+        "REGISTER_REQUEST_CALLBACK"
+    }
+
+    fn size(&self) -> usize {
+        const NUM_REQUEST_COMPLETION: usize = size_of::<u32>();
+        let request_completion = match self.request_completion {
+            Some(_) => InterfaceId::FIXED_PART_SIZE,
+            None => 0,
+        };
+
+        strict_sum(&[SharedMsgHeader::SIZE_REQ + NUM_REQUEST_COMPLETION + request_completion])
+    }
+}
+
+/// [\[MS-RDPEUSB\] 2.2.6.3 IO Control Message (IO_CONTROL)][1] message.
+///
+/// Sent from the server to the client to submit an IO control request to the USB device.
+///
+/// [1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/021733cb-8e3b-49ac-b3e3-f7a764b11141
+#[doc(alias = "IO_CONTROL")]
+#[derive(Debug, PartialEq, Clone)]
+pub struct IoControl {
+    pub header: SharedMsgHeader,
+    pub ioctl_code: IoctlInternalUsb,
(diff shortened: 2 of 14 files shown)


crates/ironrdp-rdpeusb/src/pdu/ts_urb/header.rs
```diff
@@ -0,0 +1,473 @@
...
+// the network, so MDL's don't really make a point.
+#[expect(
+    clippy::cast_possible_truncation,
+    clippy::as_conversions,
+    reason = "MS-RDPEUSB 2.2.9.1.1 specifies URB Function as being 2 bytes"
+)]
+#[repr(u16)]
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum UrbFunction {
```
playbahn: Note on the copy pasta docs for the 40+ variants: I will reword most of it.

crates/ironrdp-rdpeusb/Cargo.toml
```diff
@@ -12,7 +12,15 @@ categories.workspace = true
...
 
+[features]
+default = []
+std = []
+
 [dependencies]
+ironrdp-core = { path = "../ironrdp-core", version = "0.1", features = ["alloc"] } # public
+ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.7", features = ["alloc"] } # public
+# just for constants, etc
+windows-sys = { version = "0.61.2", features = ["Win32_Devices_Usb"] }
```
CBenoit: issue: I agree with @uchouT here, I think we should redefine the values locally to reduce the number of dependencies. The main issue is that this will probably not compile at all on Linux or macOS.

CBenoit (commented):
Great work!

Let me just recommend reading the README of the ironrdp-pdu project: https://github.com/Devolutions/IronRDP/tree/9a1ac3092ee3eac3e81823349d8e027065f5b8f8/crates/ironrdp-pdu

I should move all of this somewhere else more obvious, but I especially want to recommend the section about resilient parsing for enumeration-like values: https://github.com/Devolutions/IronRDP/tree/9a1ac3092ee3eac3e81823349d8e027065f5b8f8/crates/ironrdp-pdu#enumeration-like-types-should-allow-resilient-parsing

This will also save you from writing the two-way conversion logic.

crates/ironrdp-rdpeusb/src/pdu/common/mod.rs
CBenoit, line 1: nitpick: I’m not a big fan of "common" (or "enums", "structs", etc) modules as it’s semantically weak. If you can’t think of something better we can roll with this for now though.

playbahn: How's this im thinking:
```
pdu
|- header.rs
|- ts_urb
    |- mod.rs
    |- header.rs
|- chan_notify.rs
|- exchange_caps.rs
|- dev_sink.rs
|- mod.rs
|- req_complete.rs
|- usb_dev.rs
```
Which means dropping my hacked-togther string stuff and using `ironrdp-str`. Going through the crate, I noticed MultiSzString errors out when the `cch` field is zero: https://github.com/Devolutions/IronRDP/blob/9a1ac3092ee3eac3e81823349d8e027065f5b8f8/crates/ironrdp-str/src/multi_sz.rs#L541-L544

But https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/a26bcb6d-d45d-48a9-b9bd-22e0107d8393 says:

> cchHwIds (4 bytes): A 32-bit unsigned integer. This field MUST contain the number of Unicode characters in the HardwareIds field. This field MAY be 0x00000000.
> HardwareIds (variable): An array of bytes. A variable-length field that specifies a multisz string representing the hardware IDs of the client-side device. If the value in the cchHwIds field is 0x00000000, the HardwareIds buffer MUST NOT be present.

Same for `cchCompatIds` and `CompatibilityIds`. This is only the case for MultiSz.

[EDIT] It would also be helpful adding a link to https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/ab35aee7-1cf7-42dc-ac74-d0d7f4ca64f7 for the `cch` stuff

> Unicode character: Unless otherwise specified, a 16-bit UTF-16 code unit.

I was unsure if `cch` is really WCHAR, which is not really spelled out in RDPEUSB or RDPEDYC, but in the bigger picture that is RDPBCGR

crates/ironrdp-rdpeusb/Cargo.toml
playbahn: Hi. The project does build on my (Arch) Linux system, I am only importing the constants.

But i do agree that we could drop windows-sys in favour of hardcoding the values. I guess the reason why I'm using a foreign crate is that I/we don't wanna be the "source of truth", and so that weight is beared by Microsoft. But again, even the MS-RDPEUSB's text is not perfect https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/a7ea1b33-80bb-4197-a502-ee62394399c0

Regarding the compilation issue, is there some way we can enforce: *dont "compile" code from windows-sys, but constants is okay* ?

[EDIT] cc @uchouT

crates/ironrdp-rdpeusb/src/pdu/common/mod.rs
CBenoit: Good catch. Can you confirm if there are other places where this is a thing? If that’s exceptional, I think we can interpret that as an "optional MULTI_SZ field" where the 0x00000000 value is not so much a MULTI_SZ thing than a way to convey that the value is not present for this PDU. This would translate into `Option<MultiSzString>`. A good solution could be to use the `peek_u32` method and check for the value `0x00000000` prior to decoding MULTI_SZ. If this is a common pattern, we may imagine a `MultiSzString::decode_opt` method which returns `Option<Self>` so you don’t have to rewrite the logic everywhere 🙂

CBenoit: Also, the updated module hierarchy looks nice!

crates/ironrdp-rdpeusb/Cargo.toml
CBenoit: I don’t think we can control further the compilation, but as it appears you’re not having problems on Linux, so it’s probably not failing the way I was expecting. But then, I don’t think it’s too much a problem to vendor a few constants, because they will NEVER change (that would be an insane compatibility hazard in the Windows ecosystem), and even if they were to change we’re better off vendoring them so we can control the migration on our own terms rather than whatever `windows-sys` is up to. I don’t dislike the idea itself though: it’s generally good to avoid taking all the burden of defining everything. In that case it seems to be ~50ish constants that you still ended up defining locally anyway:
image

crates/ironrdp-rdpeusb/src/pdu/common/mod.rs
playbahn: [EDIT] TLDR: `ironrdp-str` does not really need to be modified.

> Can you confirm if there are other places where this is a thing?

> If this is a common pattern...

Only MultiSz, and its used only in `ADD_DEVICE::HardwareIds` and `ADD_DEVICE::CompatibilityIds`, `decode_opt` is not worth it, we can map "cch_0_error" to `Ok(None)`.

crates/ironrdp-rdpeusb/Cargo.toml
playbahn: Agreed. Redefining.

crates/ironrdp-rdpeusb/src/pdu/common/mod.rs
CBenoit: Sounds good!

crates/ironrdp-rdpeusb/src/pdu/req_complete.rs
```diff
@@ -0,0 +1,91 @@
...
+impl Encode for IoctlCompletion {
+    fn encode(&self, dst: &mut ironrdp_core::WriteCursor<'_>) -> ironrdp_core::EncodeResult<()> {
+        ensure_size!(in: dst, size: self.size());
+
+        self.header.encode(dst)?;
+
+        dst.write_u32(self.req_id);
+        dst.write_u32(self.hresult);
+        dst.write_u32(self.info);
+        dst.write_u32(self.out_buf_bytes);
```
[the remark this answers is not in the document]
playbahn: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/b1722374-0658-47ba-8368-87bf9d3db4d4

> OutputBufferSize (4 bytes): A 32-bit unsigned integer. The size, in bytes, of the OutputBuffer field. The value of this field MUST not exceed the value of OutputBufferSize field from IO_CONTROL message. If the HResult field indicates success, this field and the Information field MUST be equal. If the HResult field is equal to HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER) then this field is set to the value of OutputBufferSize from IO_CONTROL message and the Information field MUST indicate the expected size of the OutputBuffer field. For any other case this field MUST be set to 0 and the value of the Information field MUST be ignored.

Let's see how the next version of the MR turns out

playbahn:
Working on docs (adding spec refs)

[EDIT] Apologies for the file deletion and addition (renames in actual).
`crates/ironrdp-rdpeusb/src/pdu/`:
`chan_notify.rs` -> `notify.rs`
`req_complete.rs` -> `completion/mod.rs`
`usb_dev.rs` -> `usb_dev/mod.rs`
`ts_urb/header.rs` -> `usb_dev/ts_urb/utils.rs`
`ts_urb/mod.rs` -> `usb_dev/ts_urb/mod.rs`

uchouT:
Hi @playbahn, just checking in, hope you're doing well. Is this PR still moving forward? CI looks green, so it seems mostly waiting on review. Happy to help out (testing, reviewing, or co-authoring follow-ups) if that would be useful.

CBenoit:
I too think that the PR got big enough that we could consider reviewing and merging even if it’s not perfect yet (we can work on that in subsequent PRs). Is there something you wish to change before we do that? @playbahn

playbahn:
Hi @uchouT I should have mentioned @CBenoit when i was done with the PR.

> Is there something you wish to change before we do that?

I don't wanna take up more time, but just a suggestion for future work: top-level RDPEUSB messages could just encode `FunctionId`s, `Mask`s and default `InterfaceId`s, but that means the structures of the messages is non-uniform. At one point I was adding macros that would return errors if wrong InterfaceId/FunctionId were present in a message, but that's weak compared to just encoding by default, taking in only `MessageId` of `SHARED_MSG_HEADER` (and also `InterfaceId` for the *usb devices* and *request completion* interfaces). This would require discussion, if its really beneficial.

crates/ironrdp-rdpeusb/src/pdu/usb_dev/mod.rs
```diff
@@ -0,0 +1,1400 @@
...
+            0x220_007 => IoctlInternalUsb::ResetPort,
+            0x220_013 => IoctlInternalUsb::GetPortStatus,
+            0x220_01B => IoctlInternalUsb::GetHubCount,
+            0x220_01F => IoctlInternalUsb::CyclePort,
+            0x220_020 => IoctlInternalUsb::GetHubName,
+            0x220_420 => IoctlInternalUsb::GetBusInfo,
+            0x220_424 => IoctlInternalUsb::GetControllerName,
+            value => return Err(unsupported_value_err!("IoControlCode", format!("{value}"))),
+        };
+        if let size @ 1.. = src.read_u32(/* InputBufferSize */) {
```
uchouT, line 209: The validation is shaped by the semantics of [MS-RDPEUSB §2.2.12.1](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/8f13c014-2ece-481d-a843-9ae9b03d45fe), which deviates from the wire-format validation defined in [MS-RDPEUSB §2.2.6.3](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/021733cb-8e3b-49ac-b3e3-f7a764b11141).

crates/ironrdp-rdpeusb/src/pdu/usb_dev/ts_urb/mod.rs
```diff
@@ -0,0 +1,2249 @@
...
+                }
+                GetConfiguration => return Err(invalid_tsurb_err!("TS_URB_CONTROL_GET_CONFIGURATION_REQUEST")),
+                GetInterface => return Err(invalid_tsurb_err!("TS_URB_CONTROL_GET_INTERFACE_REQUEST")),
+                GetMsFeatureDescriptor => return Err(invalid_tsurb_err!("TS_URB_OS_FEATURE_DESCRIPTOR_REQUEST")),
+                _ => (),
+            }
+        }
+
+        let mut src = ReadCursor::new(
+            src.read_slice(usize::from(ts_urb_size) - size_of::<u16>(/* ts_urb_size */) - header.size()),
```
uchouT, line 152: Small `ts_urb_size` may cause integer underflow.

uchouT:
I've read through this PR against the MS-RDPEUSB spec. The PDU coverage looks solid as a foundation — flagging a few directions for follow-up work:

- `SharedMsgHeader` doesn't need to be embedded in every PDU. Extracting only the fields each PDU actually needs would let the type system make invalid (InterfaceId, FunctionId) combinations unrepresentable. (@playbahn touched on this in the last comment as well.)

- Each PDU's encode/decode should be relatively self-contained. Per-message validation shouldn't pull in semantics from unrelated spec sections. For example, https://github.com/Devolutions/IronRDP/pull/1165#discussion_r3250961454

- The outer enum wrapper's `decode` shouldn't take on state-machine responsibilities — e.g. the `usb_dev_s` / `completion_s` interface-ID matching it currently performs. I think it makes more sense for the PDU layer to expose the relevant fields and let the USB client/server state machine handle that validation.

- A few robustness issues. Unless guarded by the type system, we shouldn't make assumptions about input values, for example, https://github.com/Devolutions/IronRDP/pull/1165#discussion_r3250983579

These are broad directions rather than blockers. Since @playbahn mentioned

> I don't wanna take up more time,

@CBenoit  The cleanest path looks like landing it as-is and addressing the above in follow-ups. I'd like to pick up issues #1137 and #1138, they're the natural place for me to revisit the PDUs and fold these refinements in during implementation.

cc @elmarco

elmarco:
@CBenoit uchouT will take over the USB support effort as part of his GSoC (https://summerofcode.withgoogle.com/programs/2026/projects/ssoc2JtG). Merging playbahn work first makes sense, uchouT will then iterate on top.

CBenoit (approved):
@elmarco I agree!

@uchouT Thank you for the review. I totally agree with everything you said. Feel free to apply the suggested changes yourself in a future PRs 🙂 

@playbahn Thank you for all the work so far!

PR #1166, opened 2026-03-13
feat(cliprdr,web,ffi): implement clipboard file transfer support

gabrielbauman:
Add end-to-end clipboard file transfer (upload and download) across the CLIPRDR channel per MS-RDPECLIP. This spans the protocol layer, WASM/web backend, FFI bindings, and adds a TypeScript FileTransferManager.

This is a big PR. I've squashed the history because it was... a journey. 

Key changes:
- File list format negotiation, capability exchange, and lock management
- File contents request/response handling with streaming chunk transfer
- Proactive locking strategy with timeout, expiry, and cleanup logic
- Path traversal sanitization and wire-format validation
- WASM clipboard backend integration with drag-and-drop folder uploads
- FFI accessors for file transfer message variants and .NET bindings
- Fuzz targets for PackedFileList, FileContentsRequest, and channel state
- Comprehensive test coverage for delayed rendering, capabilities, and file contents validation

I suggest you start with `crates/ironrdp-cliprdr/README.md` - it gives a complete overview of the design, API, and lock lifecycle.

Some other review entry points that might help:

- Protocol layer - crates/ironrdp-cliprdr/src/lib.rs is the core. Focus on handle_format_list(), lock_clipboard(), request_file_contents(), and handle_file_contents_response() for the main flows
- PDU changes - crates/ironrdp-cliprdr/src/pdu/ for wire format changes (file_contents.rs, format_data/file_list.rs, capabilities.rs)
- Backend trait - crates/ironrdp-cliprdr/src/backend.rs defines the contract between the protocol processor and platform implementations. New methods all have default impls.
- WASM bridge - crates/iron-remote-desktop/src/session.rs and src/lib.rs for how Rust traits map to WASM
- TypeScript frontend - web-client/iron-remote-desktop/src/FileTransferManager.ts for the high-level user-facing API
- FFI - ffi/src/clipboard/ for .NET binding changes (I think I did this right)
- Tests - The #[cfg(test)] module in lib.rs and the .test.ts

There are a bunch of breaking changes here. I can summarize if need be. ironrdp-cliprdr, -ffi and web stuff will want new releases.

You probably want tests in the test crate, but the cliprdr tests refer to private structures so that was going to be hairy. I do see some inline tests throughout the project. Let me know what you'd like moved.

This code is working well with our internal browser RDP product. I hope you like it.


crates/ironrdp-cliprdr/Cargo.toml
```diff
@@ -14,7 +14,6 @@ categories.workspace = true
 
 [lib]
 doctest = false
-test = false
```
CBenoit, line 17: As an architectural invariant, we avoid adding unit tests directly into each crates (see ARCHITECTURE.md)

If you really need to access internals and you think there is value to that, one pattern we’ve been using is to expose them with a feature flag (I think there is another crate where we’re doing that somewhere)

gabrielbauman: OK, I'll re-evaluate and see what I can do here.

gabrielbauman: So my issue here is mostly state machine testing.

A lot of tests involve shortcutting `cliprdr.state` to `Ready` so I don't have to move it through the pub API. Convenience.

But then there are things like `test_format_list_preserves_in_flight_file_contents_requests` which directly inserts into `cliprdr.sent_file_contents_requests` to simulate an in-flight request, then processes a `FormatList` to verify it doesn't clobber the tracking. Setting this up through the public API would require a multi-step sequence (process capabilities -> monitor_ready -> process remote `FormatList` -> initiate_paste -> request_file_contents -> then process a second `FormatList`). The direct setup isolates the exact behavior being tested.

Some tests verify internal state after public API calls. They verify implementation that could be tested indirectly (like, "does a response arrive at the backend after the request tracking was set up?"). 

And there are a couple of critical private functions that absolutely do need unit testing: `sanitize_file_path` and `is_absolute_path`. These will def be feature flagged.

I do understand the "test from the boundaries" ideal, and yes, I can adapt a lot of these tests to drive the public API. There would be a *lot* more setup code and they may not be as precisely focused. But maybe they'd be "better" tests since they'd be fully exercising actual code paths.

What's your preference?

web-client/iron-remote-desktop/src/interfaces/Session.ts
```diff
@@ -20,4 +21,109 @@ export interface Session {
...
+     * the file list remains valid throughout the download.
+     *
+     * @returns Promise resolving to clipDataId for use in subsequent operations
+     * @example
+     * ```typescript
+     * const clipDataId = await session.lockClipboard();
+     * session.requestFileContents(streamId, fileIndex, 0x1, 0, 0, clipDataId);
+     * // ... download file ...
+     * session.unlockClipboard(clipDataId);
+     * ```
```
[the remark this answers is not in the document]
gabrielbauman: Fixed.

crates/ironrdp-cliprdr/README.md
```diff
@@ -1,14 +1,450 @@
...
+        last_write_time: Some(132489216000000000), // FILETIME format
+        file_size: Some(1024),
+        name: "document.pdf".to_owned(),
+    },
+    FileDescriptor {
+        attributes: Some(ClipboardFileAttributes::ARCHIVE),
+        last_write_time: Some(132489216000000000),
+        file_size: Some(2048),
+        name: "spreadsheet.xlsx".to_owned(),
+    },
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

web-client/iron-remote-desktop/src/services/enableFileTransfer.test.ts
```diff
@@ -0,0 +1,344 @@
...
+    });
+
+    it('should implicitly enable clipboard', () => {
+        // Disable clipboard first
+        service.setEnableClipboard(false);
+        service.enableFileTransfer();
+        // enableFileTransfer sets enableClipboard = true internally
+        // We verify this indirectly: connect() should register FileTransferManager callbacks
+        // (which only happens when enableClipboard is true)
+    });
```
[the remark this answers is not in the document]
gabrielbauman: Removed the useless test since the following test actually confirms the behaviour.

web-client/iron-remote-desktop/README.md
```diff
@@ -123,3 +119,123 @@ connect(
...
+
+```typescript
+import { FileContentsFlags } from '@devolutions/iron-remote-desktop';
+
+// Lock clipboard for file transfer
+const clipDataId = await session.lockClipboard();
+
+// Request file size
+session.requestFileContents(streamId, fileIndex, FileContentsFlags.SIZE, 0, 0, clipDataId);
+
```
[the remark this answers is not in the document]
gabrielbauman: Fixed.

crates/iron-remote-desktop/src/lib.rs
```diff
@@ -196,6 +196,71 @@ macro_rules! make_bridge {
...
+                    ));
+                }
+
+                #[expect(
+                    clippy::as_conversions,
+                    clippy::cast_possible_truncation,
+                    clippy::cast_sign_loss
+                )]
+                let position_u64 = position as u64;
+                $crate::Session::request_file_contents(
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

crates/ironrdp-web/src/session.rs
```diff
@@ -867,6 +1195,129 @@ impl iron_remote_desktop::Session for Session {
...
+                .ok()
+                .and_then(|v| v.as_f64())
+                .unwrap_or(0.0);
+            // Store as JS timestamp (ms since Unix epoch). FileMetadata::to_file_descriptor()
+            // handles the conversion to Windows FILETIME for the wire format.
+            #[expect(clippy::as_conversions, clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+            let last_modified = if last_modified_f64.is_finite() && 0.0 <= last_modified_f64 {
+                last_modified_f64 as u64
+            } else {
+                0
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

crates/ironrdp-cliprdr/README.md
```diff
@@ -1,14 +1,450 @@
...
+// Lock the clipboard before file transfer - returns unique clipDataId
+let (clip_data_id, messages) = cliprdr.lock_clipboard()?;
+// Send lock messages on the CLIPRDR virtual channel
+
+// Perform file operations...
+// Multiple locks can be active simultaneously with different clipDataIds
+
+// Unlock when transfer completes
+let messages = cliprdr.unlock_clipboard(clip_data_id)?;
+// Send unlock messages on the CLIPRDR virtual channel
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

crates/ironrdp-cliprdr/src/backend.rs
```diff
@@ -145,15 +141,89 @@ pub trait CliprdrBackend: AsAny + core::fmt::Debug + Send {
...
+    /// This is an incoming unlock request - the server is done with the locked clipboard snapshot.
+    ///
+    /// **Note**: For unlocking **remote clipboard data** after downloads, call
+    /// [`crate::Cliprdr::unlock_clipboard`] directly instead.
     fn on_unlock(&mut self, data_id: LockDataId);
+
+    /// [2.2.5.2] Processes remote file list metadata
+    ///
+    /// Called by [crate::Cliprdr] when file list metadata is received from the remote.
+    /// This is automatically requested when a FormatList contains FileGroupDescriptorW.
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

crates/ironrdp-web/src/session.rs
```diff
@@ -867,6 +1195,129 @@ impl iron_remote_desktop::Session for Session {
...
+                .context("File name is required")?;
+            let size_f64 = js_sys::Reflect::get(&file_obj, &JsValue::from_str("size"))
+                .ok()
+                .and_then(|v| v.as_f64())
+                .context("file size is required")?;
+            if !size_f64.is_finite() || size_f64 < 0.0 {
+                return Err(anyhow::anyhow!("file size must be a finite non-negative number (got: {size_f64})").into());
+            }
+            #[expect(clippy::as_conversions, clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+            let size = size_f64 as u64;
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

web-client/iron-remote-desktop/src/interfaces/Session.ts
```diff
@@ -20,4 +21,109 @@ export interface Session {
...
+     * @param size - Number of bytes requested for RANGE requests
+     * @param clipDataId - Optional lock ID from lockClipboard()
+     *
+     * @example
+     * ```typescript
+     * // Request file size
+     * session.requestFileContents(1, 0, 0x1, 0, 0, clipDataId);
+     * // Request file data chunk
+     * session.requestFileContents(1, 0, 0x2, 0, 65536, clipDataId);
+     * ```
```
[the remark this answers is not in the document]
gabrielbauman: Fixed

typos.toml
```diff
@@ -20,6 +20,11 @@
 [files]
 extend-exclude = ["*.bin", "*.bmp", "package-lock.json"]
 
+[default.extend-words]
+# "look-alikes" and "mis-attributed" are legitimate hyphenated words
+alikes = "alikes"
+mis = "mis"
+
```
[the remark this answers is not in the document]
gabrielbauman: I've removed the "mis" and just changed the spelling in the comment to "misattributed" haha

crates/ironrdp-cliprdr/Cargo.toml
CBenoit: Thanks, this is helpful context.

I still think the default should be to test from the boundaries / public API when we can, because that has been a useful architectural constraint for the project and it keeps the crates easier to evolve internally.

That said, I don’t want to force a lot of awkward setup just to preserve the rule mechanically. If there are a few cases where the behavior is genuinely much clearer to validate with narrowly scoped internal tests, I’m okay making room for that.

So my preference would be:

* move the tests that can reasonably be expressed through the public API in that direction;
* keep direct-internal tests only for the cases that are really hard to express otherwise, especially where the test would become much more indirect or much more verbose;
* for pure/private helpers like `sanitize_file_path` and `is_absolute_path`, feature-gated exposure for tests sounds fine.

I would lean toward keeping the architectural guideline, but applying it pragmatically rather than absolutely here.

gabrielbauman: Okay, give me a few hours to work this out. Thanks for the clarity!

crates/ironrdp-fuzzing/src/oracles/mod.rs
CBenoit, line 1: praise: Even the state machine was fuzzed 👏

web-client/iron-remote-desktop/src/FileTransferManager.test.ts
CBenoit, line 1: praise: Hey, thank you for also testing the TypeScript. We were missing this kind of stuff.

thought: Looks like it’s not handled in xtask and not run in the CI. Could be a good follow up.

crates/ironrdp-cliprdr/Cargo.toml
gabrielbauman: Right. I've moved the existing tests to the test crate using __test gating. Next I'll rewrite them to be a bit less nosy. Here's the strategy:

1. Behavior assertions -> verify via public API (returned SvcMessages, backend callbacks, PduResult errors)                 
2. Bookkeeping assertions (counts, key existence in maps) -> internal state via __test is fine
3. Test Setup -> feature-gated mutation is fine where the alternative is a verbose multi-PDU handshake                                                        
4. Pure helpers (sanitize_file_path, is_absolute_path) -> feature-gated exposure

Does that sound like a fair compromise?

gabrielbauman: I've gone ahead and moved the tests to testsuite-core following the approach above. Hope that looks ok.

web-client/iron-remote-desktop/src/FileTransferManager.test.ts
gabrielbauman: Wired it up, should run in CI now.

glamberson:
@gabrielbauman No concerns from the server side. This is the right move.

The per-snapshot semantics are the key point. With Proactive, the lock is already held when `on_remote_file_list()` fires, so backends can start issuing FileContentsRequests against a stable snapshot immediately. No lock-acquisition round-trip, no window where the clipboard could change between seeing the file list and requesting content.

One thing worth considering: server-side backends may not request file contents immediately. A backend might expose the file list to the host OS and defer the actual FileContentsRequests until an application reads the data, potentially minutes later. The lock needs to stay alive until the backend explicitly releases it or the clipboard changes via a new FormatList. As long as the auto-detach on new FormatList keeps the old lock alive for in-flight transfers (which it sounds like it does) this works well for deferred-read patterns.

Agreed on nuking the enum. Proactive matches the spec's Figure 3 sequence, matches FreeRDP, and removes a configuration surface that nobody should need to change.

Greg

gabrielbauman:
Okay, I've nuked the enum and made proactive locking the default and only lock strategy. I've also tested with our internal browser RDP client and the new locking works flawlessly against a Windows rdp server. Nice drop in complexity in the wasm and typescript layers.

I've also done some drive-by fixes for a few minor style and quality-of-life issues etc. I'm done tweaking until I hear back on the review front.

Sorry for the churn! We're in good shape now, I think. Thanks for the discussion, @glamberson! Have a nice weekend, everyone.

fuzz/Cargo.toml
```diff
@@ -48,3 +48,10 @@ test = false
 doc = false
 bench = false
 
+[[bin]]
+name = "cliprdr_channel_processing"
```
CBenoit, line 52: suggestion: Can you add this target in the fuzz workflow too?

crates/ironrdp-cliprdr/README.md
```diff
@@ -1,14 +1,410 @@
...
+- **Maximum lifetime** (2h default): Force cleanup regardless of activity
+
+This prevents resource leaks from stalled file transfers without requiring background threads or explicit periodic calls.
+
+##### Manual Cleanup (Optional)
+
+For advanced use cases where you need explicit control, you can call `cleanup_expired_locks()` manually:
+
+```rust
+// Optional: Force cleanup check before automatic interval
```
CBenoit: suggestion: I think this wording is easier to understand (at least for me)

```suggestion
// Optional: Trigger cleanup check before automatic interval
```

crates/ironrdp-cliprdr/README.md
```diff
@@ -1,14 +1,410 @@
...
+
+#### Automatic Lock Behavior
+
+When a FormatList containing `FileGroupDescriptorW` is received and `CAN_LOCK_CLIPDATA` was negotiated, the cliprdr processor automatically:
+
+1. Sends a Lock Clipboard Data PDU with a new `clipDataId`
+2. Passes the `clipDataId` to the backend via `on_remote_file_list(files, clip_data_id)`
+3. Manages the lock lifecycle (expiry, cleanup, Unlock PDUs) internally
+
+Backends receive the `clipDataId` and can pass it to `request_file_contents()` via the `data_id` field. No explicit lock/unlock calls are needed.
```
CBenoit, line 174: issue: The lazy piggyback pattern is a pragmatic shortcut, but it has real costs in my opinion.

rationale:
1. It forces `&mut self` on semantically read-only operations (`initiate_paste`, `submit_format_data`), so callers and type system can no longer distinguish between "mutating" and "just builds a PDU".
2. The return type becomes semantically muddled: `initiate_paste` may return `[UNLOCK_CLIPDATA, FORMAT_DATA_REQUEST]`, which can be surprising, it’s documented but it’s unrelated side-effects bundled in.
3. Cleanup is only triggered by user activity, and is unpredictable, although I understand it’s not a big issue in practice.

suggestion: I think there is a more idiomatic alternative here. I’m thinking about a dedicated `poll`/`drive` called from the event loop. I’ve seen this pattern in crates such as `rustls`, `quinn` or `tokio`. Some return a `Duration` hint to let the caller scheduler the next drive without polling blindly. In our case, I think we don’t really need much more than an interval though:

```rust
// A persistent interval stream.
let cleanup_interval_ms = 5_000;
let mut cleanup_timer = gloo_timers::future::IntervalStream::new(cleanup_interval_ms);

let disconnect_reason = 'outer: loop {
   let outputs = select! {
       /* ... */

       _ = cleanup_timer.next() => {
           if let Some(cliprdr) = active_stage.get_svc_processor_mut::<CliprdrClient>() {
               match cliprdr.cleanup_expired_locks() {
                   Ok(svc_messages) => {
                       let frame = active_stage.process_svc_processor_messages(svc_messages)?;
                       vec![ActiveStageOutput::ResponseFrame(frame)]
                   }
                   Err(e) => {
                       error!(error = %e, "Clipboard lock cleanup failed");
                       Vec::new()
                   }
               }
           } else {
               Vec::new()
           }
       }
   };
};
```

That way `﻿Cliprdr` can drop the internal `auto_cleanup_interval` / `last_cleanup_check` / `maybe_cleanup_locks` machinery entirely and the scheduling responsibility moves to the caller, making it configurable in the session setup code without coupling it to the protocol processor.

With that in place:

- `initiate_copy`, `initiate_paste`, `submit_format_data` go back to `&self` (they just encode PDUs)
- `maybe_cleanup_locks()` disappears entirely
- `cleanup_expired_locks()` is the single, explicit, testable cleanup path
- `get_svc_processor` suffices for most call sites; only genuinely mutating operations (`request_file_contents`, `initiate_file_copy`) need `get_svc_processor_mut`

crates/ironrdp-cliprdr/src/backend.rs
```diff
@@ -145,15 +135,102 @@ pub trait CliprdrBackend: AsAny + core::fmt::Debug + Send {
...
+    ///
+    /// ## Parameters
+    ///
+    /// - `clip_data_ids`: List of clipDataIds for locks that were automatically cleared.
+    ///
+    /// **Note**: Unlock PDUs are automatically sent for each cleared lock.
+    fn on_outgoing_locks_cleared(&mut self, clip_data_ids: &[LockDataId]) {
+        let _ = clip_data_ids;
+        // Default implementation does nothing - backends can override to handle lock cleanup
+    }
```
CBenoit: issue: ﻿The trigger description here is wrong, and it conflicts with the `on_outgoing_locks_expired` doc directly below.

It says "triggered when a new FormatList PDU is received", but that is precisely what triggers `on_outgoing_locks_expired`. `on_outgoing_locks_cleared` is triggered later, by `cleanup_expired_locks()`, after the inactivity/max-lifetime timeout elapses on an expired lock.

Additionally, the spec citation `[MS-RDPECLIP] 2.2.4.2` is the structure definition of `CLIPRDR_UNLOCK_CLIPDATA`, just a PDU layout, and it contains no directive about when locks must be released on clipboard change.

crates/ironrdp-cliprdr/README.md
```diff
@@ -1,14 +1,410 @@
...
+**Inactivity Timeout (default: 60 seconds)**
+- Locks are cleaned up after 60 seconds without `FileContentsRequest` activity
+- Each `FileContentsRequest` for the lock resets the inactivity timer
+- This handles abandoned downloads while allowing active transfers to continue
+- Configurable via `Cliprdr::with_lock_timeouts(backend, inactivity_duration, max_lifetime_duration)`
+
+**Maximum Lifetime (default: 2 hours)**
+- Locks are force-cleaned after 2 hours regardless of activity
+- This prevents indefinite resource accumulation from slow or stalled transfers
+- Measured from when the lock transitioned to Expired state (not from lock creation)
```
CBenoit: issue: It seems this line contradicts the code, `cleanup_expired_locks_impl` computes `total_lifetime_ms = now.saturating_sub(lock.created_at_ms)`.

crates/iron-remote-desktop/src/session.rs
```diff
@@ -96,6 +114,134 @@ pub trait Session {
...
+    /// # Example: Download a file
+    /// ```text
+    /// // When filesAvailableCallback receives files and clipDataId:
+    /// // First request size
+    /// requestFileContents(1, 0, 0x1, 0, 8, clipDataId);
+    /// // Wait for fileContentsResponseCallback with size
+    /// // Then request data in chunks
+    /// requestFileContents(1, 0, 0x2, 0, 4096, clipDataId);
+    /// // Wait for fileContentsResponseCallback with data
+    /// unlockClipboard(clipDataId);
```
CBenoit: issue: I don’t see this function (or maybe I’m missing it?) (EDIT: I think it was part of the initial implementation)

web-client/iron-remote-desktop/README.md
```diff
@@ -123,3 +119,123 @@ connect(
...
+const clipDataId = await session.lockClipboard();
+
+// Request file size
+session.requestFileContents(streamId, fileIndex, FileContentsFlags.SIZE, 0, 8, clipDataId);
+
+// Request file data
+session.requestFileContents(streamId, fileIndex, FileContentsFlags.RANGE, offset, chunkSize, clipDataId);
+
+// Unlock clipboard when done
+session.unlockClipboard(clipDataId);
```
CBenoit: issue:  Same as above, lockClipboard() and unlockClipboard() were part of the initial iteration, before the proactive locking refactor.

web-client/iron-remote-desktop/src/interfaces/Session.ts
```diff
@@ -20,4 +21,81 @@ export interface Session {
...
+     *
+     * Per MS-RDPECLIP 2.2.5.3.1, sends FileContentsRequest PDU to remote.
+     * Response arrives via fileContentsResponseCallback.
+     *
+     * @param streamId - Unique stream identifier for this transfer
+     * @param fileIndex - Index in file list (0-based)
+     * @param flags - FileContentsFlags: 0x1 (SIZE) or 0x2 (RANGE)
+     * @param position - Byte offset for RANGE requests
+     * @param size - Number of bytes requested for RANGE requests
+     * @param clipDataId - Optional lock ID from lockClipboard()
```
CBenoit: issue: Another reference to lockClipboard()

fuzz/Cargo.toml
CBenoit: https://github.com/Devolutions/IronRDP/blob/a6b41093ce4ece081d2538c157f6bc547c3b2607/.github/workflows/fuzz.yml#L44

crates/iron-remote-desktop/src/session.rs
CBenoit, line 1: issue: ﻿The `Session`/`SessionBuilder` traits (both Rust and TypeScript) now expose raw RDPECLIP wire concepts: `streamId`, `FileContentsFlags`, `clipDataId`, etc. These don't exist in any other remote protocol and violate the design invariant that `iron-remote-desktop` is protocol-agnostic.

suggestion: Push `requestFileContents`, `submitFileContents`, etc, into the RDP backend layer and wire them through the existing `invokeExtension` / `extension` mechanism. For that, you can refer to existing extensions such as `displayControl` or `kdcProxyUrl`.

You should use the `extension_match!` macro to match on the extension, and extract the value. The current extensions are trivial, only using primitive types. In this case, you will need to use [`js_sys::Reflect`](https://docs.rs/js-sys/latest/js_sys/Reflect/index.html) module in order to extract the parameters from the JavaScript object / `JsValue`, and [`Function::try_from_js_value`](https://docs.rs/js-sys/latest/js_sys/struct.Function.html#impl-TryFromJsValue-for-T) to convert the `JsValue` into a `js_sys::Function`.

Example:

```rust
    fn invoke_extension(&self, ext: Extension) -> Result<JsValue, Self::Error> {
        use js_sys::{JsString, Object};
        use wasm_bindgen::convert::TryFromJsValue as _;

        iron_remote_desktop::extension_match! {
            match ext;
            |request_file_contents: JsValue| {
                let obj = into_object(request_file_contents)?;
                let stream_id = get_u32(&obj, "stream_id")?;
                let file_index = get_i32(&obj, "file_index")?;
                let flags = get_u32(&obj, "flags")?;
                let position = get_u64(&obj, "position")?;
                let size = get_u32(&obj, "size")?;
                let clip_data_id = get_u32_opt(&obj, "clip_data_id")?;

                self.input_events_tx
                    .unbounded_send(RdpInputEvent::ClipboardBackend(
                        WasmClipboardBackendMessage::FileContentsRequestSend {
                            stream_id,
                            index: file_index,
                            flags: FileContentsFlags::from_bits_truncate(flags),
                            position,
                            size,
                            clip_data_id,
                        },
                    ))
                    .context("send file contents request")
                    .map_err(Into::into)?;

                return Ok(JsValue::NULL);
            };
        }

        return Err(
            IronError::from(anyhow::Error::msg(format!("unknown extension: {}", ext.ident())))
                .with_kind(IronErrorKind::General),
        );

        fn into_object(val: JsValue) -> Result<Object, IronError> {
            let obj = Object::try_from_js_value(val).ok().context("invalid value")?;
            Ok(obj)
        }

        fn get_u32(obj: &Object, key: &str) -> Result<u32, IronError> {
            let property = js_sys::Reflect::get_str(obj, &JsString::from(key))
                .map_err(|e| anyhow::anyhow!("get property `{key}`: {e:?}"))?
                .with_context(|| format!("object is missing property `{key}`"))?;

            let value = property
                .as_f64()
                .map(f64_to_u32_saturating_cast)
                .with_context(|| format!("invalid type for property `{key}`"))?;

            Ok(value)
        }

        fn get_u64(obj: &Object, key: &str) -> Result<u64, IronError> {
            let property = js_sys::Reflect::get_str(obj, &JsString::from(key))
                .map_err(|e| anyhow::anyhow!("get property `{key}`: {e:?}"))?
                .with_context(|| format!("object is missing property `{key}`"))?;

            let value = property
                .as_f64()
                .map(f64_to_u64_saturating_cast)
                .with_context(|| format!("invalid type for property `{key}`"))?;

            Ok(value)
        }

        fn get_i32(obj: &Object, key: &str) -> Result<i32, IronError> {
            let property = js_sys::Reflect::get_str(obj, &JsString::from(key))
                .map_err(|e| anyhow::anyhow!("get property `{key}`: {e:?}"))?
                .with_context(|| format!("object is missing property `{key}`"))?;

            let value = property
                .as_f64()
                .map(f64_to_i32_saturating_cast)
                .with_context(|| format!("invalid type for property `{key}`"))?;

            Ok(value)
        }

        fn get_u32_opt(obj: &Object, key: &str) -> Result<Option<u32>, IronError> {
            match js_sys::Reflect::get_str(obj, &JsString::from(key))
                .map_err(|e| anyhow::anyhow!("get property `{key}`: {e:?}"))?
            {
                Some(property) => {
                    let value = property
                        .as_f64()
                        .map(f64_to_u32_saturating_cast)
                        .with_context(|| format!("invalid type for property `{key}`"))?;

                    Ok(Some(value))
                }
                None => Ok(None),
            }
        }
    }
```

We may refactor the helpers, but it would be good enough as-is for now!

crates/ironrdp-web/src/session.rs
```diff
@@ -867,6 +1145,119 @@ impl iron_remote_desktop::Session for Session {
...
+                WasmClipboardBackendMessage::FileContentsRequestSend {
+                    stream_id,
+                    index: file_index,
+                    flags: FileContentsFlags::from_bits_truncate(flags),
+                    position,
+                    size,
+                    clip_data_id,
+                },
+            ))
+            .context("Send file contents request")
```
CBenoit: style:

```suggestion
            .context("send file contents request")
```

web-client/iron-remote-desktop-rdp/src/RdpFileTransferProvider.ts
CBenoit, line 1: praise: I really like this high-level abstraction.

suggestion: Based on my previous suggestion, this should go into `iron-remote-desktop-rdp` because it will rely on the new RDP extensions.

thought: At some point in the future we may consider writing a common abstraction using some form of dependency injection again, but I think it’s not necessary for now.

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
```diff
@@ -56,11 +57,58 @@ export interface SessionBuilder {
...
+     * Optional
+     *
+     * Called when remote sends file contents to client (download). This is the response
+     * to a previous file contents request initiated by the client.
+     */
+    fileContentsResponseCallback(callback: FileContentsResponseCallback): SessionBuilder;
+    /**
+     * Optional
+     *
+     * Called when remote locks their clipboard for file transfer. The dataId associates
```
CBenoit: issue: I think this is the other way around?

```suggestion
     * Called when remote locks our clipboard for file transfer. The dataId associates
```

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -140,8 +597,25 @@ impl<R: Role> Cliprdr<R> {
...
+        //
+        // SAFETY (ordering): `handle_server_capabilities()` runs
+        // synchronously during `process()` for the Capabilities PDU,
+        // which always precedes the Monitor Ready PDU in the server's
+        // initialization sequence. By the time any backend callback
+        // triggers `initiate_copy()`, `self.capabilities` has already
+        // been downgraded against the server's capabilities. This holds
+        // regardless of whether the backend responds synchronously or
+        // asynchronously, because `process()` for the Capabilities PDU
+        // completes before `process()` for Monitor Ready begins.
```
CBenoit: nitpick: I think it may be best to keep `SAFETY:` comments for `unsafe` blocks, as a convention.

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -309,12 +1264,336 @@ impl<R: Role> Cliprdr<R> {
...
+    /// - The file index (lindex) must be obtained from a prior file list exchange
+    /// - For SIZE requests: cbRequested must be 8, position must be 0
+    /// - For RANGE requests: the specified range must be within file bounds
+    ///
+    /// The streamId is tracked to validate the corresponding FileContentsResponse.
+    ///
     /// [2.2.5.3]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeclip/cbc851d3-4e68-45f4-9292-26872a9209f2
-    pub fn request_file_contents(&self, request: FileContentsRequest) -> PduResult<CliprdrSvcMessages<R>> {
-        ready_guard!(self, request_file_contents);
+    pub fn request_file_contents(&mut self, mut request: FileContentsRequest) -> PduResult<CliprdrSvcMessages<R>> {
```
CBenoit, line 1202: suggestion: It could be good to enforce the `CB_HUGE_FILE_SUPPORT_ENABLED` (0x00000020) constraint on `nPositionHigh` and `nPositionLow`. What do you think?

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -309,12 +1264,336 @@ impl<R: Role> Cliprdr<R> {
...
+                trace!(
+                    clip_data_id,
+                    stream_id = request.stream_id,
+                    "Updated lock activity timestamp"
+                );
+            }
+        }
+
+        // [MS-RDPECLIP] 2.2.5.3 - Validate flags are spec-compliant
+        if let Err(e) = request.flags.validate() {
```
CBenoit, line 1256: nitpick: Just a note about "Parse Don’t Validate" pattern (https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/); but it’s probably fine as-is as well

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -410,3 +1966,91 @@ impl Role for Server {
         true
     }
 }
+
+/// Test-only accessors for `Cliprdr` internal state.
+///
+/// These methods are gated behind the `__test` feature and exist solely
+/// so that tests in `ironrdp-testsuite-core` can set up / inspect internal
+/// fields without making them part of the public API.
+#[cfg(feature = "__test")]
```
CBenoit, line 1883: suggestion: Also add ﻿`#[doc(hidden)]` so it’s never shown in docs.rs
```suggestion
#[cfg(feature = "__test")]
﻿#[doc(hidden)]
```

fuzz/Cargo.toml
gabrielbauman: Will do

crates/ironrdp-cliprdr/README.md
gabrielbauman: I've gone ahead and done this. Will push shortly.

crates/iron-remote-desktop/src/session.rs
gabrielbauman: I see what you mean. Let me poke some more.

web-client/iron-remote-desktop-rdp/src/RdpFileTransferProvider.ts
gabrielbauman: Noted, will moveit. I'm removing the AbortController and cancellation support because we're finding some RDP servers hate it and have user-facing crashes on cancellation.

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Makes sense!

crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -227,18 +810,45 @@ impl<R: Role> Cliprdr<R> {
...
+                let mut merged = available_formats.to_vec();
+                merged.push(file_format);
+                merged
+            } else {
+                // No file upload active - clear file list state as before
+                self.local_file_list = None;
+                self.local_file_list_format_id = None;
+                available_formats.to_vec()
+            };
+
```
ymarcus93: I think there is a bug here. I've been running into an issue when testing policy enforcement on our end.

**Observations:**
- C2S text clipboard deny text stops appearing after a successful file upload under a mixed policy (file transfer allowed, text clipboard blocked).

It seems that `local_file_list` (set by `initiate_file_copy()`), but is only cleared in two places: 
1. `FormatListResponse::Fail` handling in `handle_format_list_response`
2. The `else` branch of the highlighted code block (which only runs when `local_file_list` is already `None`). 

**Issue**: After a _successful_ file upload completes, `local_file_list` remains `Some(...)` indefinitely. This causes issues in our proxy when dealing with hybrid policy case since we have to send `FileGroupDescriptorW` for correctness even though user is not really doing a file upload. This prevents the server from seeing the `CF_UNICODETEXT` format in the `FormatList` and requesting a paste (to see the deny text).


**Two possible fixes:**
1. Clear `local_file_list` in `initiate_copy` unconditionally, removing the merge branch entirely. When the user copies new non-file content, the file upload is effectively "done" from the user's perspective. The old file list should be discarded. This is the simplest fix but does mean a concurrent file download-in-progress would be interrupted by a text copy.

2. Add transfer-completion detection; track whether the server has finished downloading (not sure how though? do we have a signal for this?). Clear `local_file_list` once the transfer is complete.

Fix (1) seems the least complex and feels correct since the CLIPRDR spec [states](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeclip/680788b3-2bd8-4e2b-9806-589aba7cf814): "The Clipboard Format ID Map is cleared and initialized whenever a Format List PDU is processed (section 3.1.5.2.2)." The user explicitly copying new content is a strong signal that the previous clipboard state (including file lists) should be replaced.

web-client/iron-remote-desktop/src/FileTransferManager.ts
```diff
@@ -0,0 +1,1807 @@
...
+                            }
+                        },
+                        { signal: internalAc.signal },
+                    );
+                }
+            }
+
+            // Suppress clipboard monitoring while the upload is in flight so the
+            // 100ms polling loop does not clobber our FormatList with a text/image update.
+            this.onUploadStarted?.();
```
ymarcus93: I think this code is causing issues in our policy enforcement for the case of a deny-all policy (both text and file transfer blocked in both directions).

**Observations:**
- Under a deny-all policy (both text and file transfer blocked in both directions), C2S text clipboard deny text stops appearing after a file upload attempt. This also affects any mixed policy where file transfer is blocked in the relevant direction (see other thread).
- The same issue occurs whether files are uploaded via `showFilePicker()` (upload button) or drag-and-drop; both paths go through `uploadFiles()`.

**Sequence of events:**
1. C2S text copy-paste: deny text appears correctly.
2. User uploads files: `uploadFiles()` calls `this.onUploadStarted?.()`, which calls `clipboardService.suppressMonitoring()`, setting `monitoringSuppressed = true`. The clipboard monitoring loop (`clipboard.service.ts::onMonitorClipboard()`) now returns immediately on every poll _without_ reading the browser clipboard.
3. `initiateFileCopy()` sends a `FormatList` with `FileGroupDescriptorW` to the server. Our proxy (enforcing deny-all) replaces the entire `FormatList` with a `CF_UNICODETEXT` deny placeholder. The server never learns about file formats.
4. Because the server doesn't know about file formats, it _never_ sends `FileContentsRequest`. So `completedFiles.size` never reaches `expectedFileCount` (see `handleFileContentsRequest()` in this file), and `onUploadFinished` never fires. **`monitoringSuppressed` stays true permanently.**
5. User copies text; the monitoring loop skips it. No `LocalClipboardChanged` event fires, no `FormatList` is sent to the server, deny text never appears again for the rest of the session.

The root issue is that `onUploadStarted` fires immediately and unconditionally, but `onUploadFinished` depends entirely on remote behavior (the remote requesting every single file). Under any blocking policy, _or if the remote user simply never pastes_, monitoring is suppressed forever.

**Possible solutions:**

1. **Timeout**: Resume monitoring if no `FileContentsRequest` arrives within N seconds after `initiateFileCopy()`, bounding the suppression window.
2. **Short debounce instead of indefinite suppression**: Suppress for a brief period (e.g., 500ms) after `initiateFileCopy()` to prevent the polling loop from racing with the `FormatList` send, then resume regardless of transfer state. The `uploadState` can remain active to handle incoming `FileContentsRequest` PDUs independently.
3. **Resume immediately after `initiateFileCopy()` succeeds**: The concern this suppression addresses is the monitoring loop clobbering the file `FormatList` with a text/image update. That race window is very short; only the time between `onUploadStarted` and the `FormatList` actually being sent on the wire. Resuming right after `initiateFileCopy()` returns (or after a small delay) would close the race without coupling monitoring lifetime to transfer completion.

crates/iron-remote-desktop/src/session.rs
gabrielbauman: Done. File transfer methods (`requestFileContents`, `submitFileContents`, `initiateFileCopy`) removed from the `Session` trait and routed through `invoke_extension()` with `js_sys::Reflect` parameter extraction. The six file transfer callbacks removed from `SessionBuilder` and routed through `extension()` with `JsValue` -> `js_sys::Function` conversion via `dyn_into`. Helper functions (`into_object`, `get_u32`, `get_u64`, etc.) follow the pattern from your example code.

web-client/iron-remote-desktop-rdp/src/RdpFileTransferProvider.ts
gabrielbauman: Moved. `FileTransferManager` is now `RdpFileTransferProvider` in `iron-remote-desktop-rdp`. It implements a new protocol-agnostic `FileTransferProvider` interface defined in `iron-remote-desktop`. Session interaction uses `invokeExtension()` exclusively. Extension factory functions (`filesAvailableCallback`, `requestFileContents`, etc.) are exported from the rdp package. The web component's `enableFileTransfer(provider)` accepts any `FileTransferProvider` via dependency injection - no direct import of the RDP package needed.

crates/ironrdp-web/src/session.rs
gabrielbauman: Fixed. Also lowercased the other `.context()` strings we introduced (`"cliprdr initiate copy"`, `"send clipboard backend event"`, etc.).

web-client/iron-remote-desktop/src/interfaces/SessionBuilder.ts
gabrielbauman: Fixed. The six callback methods have been removed from `SessionBuilder.ts` entirely as part of the extension refactor, so the incorrect doc is gone. The callbacks are now registered via `builder.extension(lockCallback(cb))` from the RDP package.

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Changed to `INVARIANT (ordering):` to match the project convention.

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Done. Positions >= 2^31 are rejected with a descriptive error when `CB_HUGE_FILE_SUPPORT_ENABLED` is not negotiated.

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Done. `FileTransferState.file_index` is now `usize`. The `i32` from the wire is validated via `try_from` and the resulting `usize` is stored directly. Doc comment explains the choice.

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Done.

crates/ironrdp-cliprdr/README.md
gabrielbauman: Done. `maybe_cleanup_locks`/`auto_cleanup_interval`/`last_cleanup_check` removed. `drive_timeouts()` is the dedicated public method, called from a 5-second interval timer in the web session. `submit_format_data` takes `&self`. 

`initiate_copy` and `initiate_paste` still take `&mut self` - this is not a cleanup side-effect but genuine state tracking: `initiate_copy` manages `local_file_list`, `initiate_paste` tracks `pending_format_data_request` for response correlation. Both have inline doc comments explaining why.

crates/ironrdp-cliprdr/src/backend.rs
gabrielbauman: Fixed.

crates/ironrdp-cliprdr/README.md
gabrielbauman: Fixed.

crates/iron-remote-desktop/src/session.rs
gabrielbauman: Removed.

web-client/iron-remote-desktop/README.md
gabrielbauman: Removed.

web-client/iron-remote-desktop/src/interfaces/Session.ts
gabrielbauman: Removed.

fuzz/Cargo.toml
gabrielbauman: Done (was addressed in a previous push).

crates/ironrdp-cliprdr/src/lib.rs
gabrielbauman: Fixed. `initiate_copy()` now clears `local_file_list` unconditionally. The old merge branch that preserved the file list during active uploads is removed. Per [MS-RDPECLIP] 3.1.1.1, each FormatList completely replaces the previous. A text/image copy during an active file download ends the upload's visibility to the remote - an acceptable tradeoff since the user explicitly chose new content. Inline comment documents this decision.

web-client/iron-remote-desktop/src/FileTransferManager.ts
gabrielbauman: Fixed. Monitoring now resumes immediately after `initiateFileCopy()` (in a `finally` block), rather than waiting for all file contents requests to complete. The suppression window is intentionally brief - just long enough to prevent the clipboard polling loop from racing with the FormatList send. All scattered `onUploadFinished` calls in upload completion/error paths have been removed. Upload state tracking continues independently for progress and error reporting.

crates/iron-remote-desktop/src/lib.rs
```diff
@@ -196,6 +196,10 @@ macro_rules! make_bridge {
                 <<$api as $crate::RemoteDesktopApi>::Session as $crate::Session>::invoke_extension(&self.0, ext)
                     .map_err(IronError)
             }
+
+            // File transfer methods (requestFileContents, submitFileContents,
+            // initiateFileCopy) are protocol-specific and routed through
+            // invoke_extension() rather than living on the Session trait directly.
```
CBenoit: Refactor leftovers (probably left by the agent)

gabrielbauman: Oh, Mr. Agent. *pulls out memory wiping flashy device thing*

gabrielbauman: Removed the Session-side leftover. Kept a trimmed version on the SessionBuilder side since it explains why `extension()` is the entry point for file transfer callbacks - someone reading the macro would otherwise wonder where they went.

CBenoit: We can mention them as examples, but I think that’s fine as it’s not a concern for `iron-remote-desktop`

PR #1179, opened 2026-03-19
fix(graphics): correct three RLGR encoder bugs per MS-RDPRFX spec

elmarco:
## Summary

Fixes three bugs in the RLGR entropy encoder (`ironrdp-graphics::rlgr::encode`) identified by cross-referencing the MS-RDPRFX §3.1.8.1.7.3 pseudocode and the FreeRDP reference implementation (`rfx_rlgr.c`).

- **RL mode trailing value**: the encoder skipped emitting sign bit + GR code when input was exhausted after a zero run. The decoder unconditionally reads these bits, so the encoder must always emit them. Restructured to match FreeRDP's `GetNextInput` pattern.
- **RLGR3 exhausted second value**: used `unwrap_or(1)` as default `twoMs` when the second value in a GR-mode pair was unavailable. FreeRDP's `GetNextInput` returns 0 when exhausted, and `Get2MagSign(0) = 0`, so the correct default is 0.
- **RLGR1 GR-mode kp update**: used `UP_GR` (4, the RL-mode constant) instead of `UQ_GR` (3, the GR-mode constant) when updating `kp` for zero symbols. Both the spec pseudocode and FreeRDP use `UQ_GR` here.

Each fix is in a separate commit with full rationale and spec/FreeRDP references.

### Spec errata note

The MS-RDPRFX §4.2.4.1 reference hex dump appears to have been generated with `UQ_GR=4` and `twoMs2=1` defaults — inconsistent with the normative pseudocode in §3.1.8.1.7.3. Our encoder follows the pseudocode (which is authoritative over example data) and matches FreeRDP. The Y test vector is updated accordingly (2 bytes differ from the spec hex dump at indices 939–940). Encoder→decoder roundtrip is verified correct.

## Test plan

- [x] All 12 existing RLGR unit tests pass (encode + decode, small vectors + full 4096-coefficient Y/Cb/Cr datasets)
- [x] Encoder→decoder roundtrip verified for Y, Cb, Cr components
- [x] `cargo clippy -p ironrdp-graphics` clean
- [x] `cargo xtask check lints -v` passes


glamberson:
Independent confirmation from a downstream audit: all three bugs reproduce in our local copy of `ironrdp-graphics::rlgr::encode` against current master (`5f93a03f`).

- Line 106: `if let Some(val) = input.next()` skips emitting sign bit + GR code when input is exhausted after a zero run, matching your Bug 1.
- Line 133: `.unwrap_or(1)` for the exhausted RLGR3 second value, matching your Bug 2. FreeRDP's `GetNextInput` returns 0 here.
- Line 125: `kp + UP_GR` (the RL-mode constant, 4) where the spec pseudocode and FreeRDP use `UQ_GR` (3) for kp updates in GR mode after zero symbols, matching your Bug 3.

Audit notes saved at our end (`upstream/ironrdp/analysis/RLGR-AUDIT-2026-04-30.md`). Affects our merged #1196 (Progressive RFX primitives) and the open #1198/#1199 stack indirectly: the round-trip tests pass because both sides match each other, masking the divergence from spec/FreeRDP. We're holding our test vectors stable until this lands rather than rebasing twice.

Happy to help drive this to merge if any of the maintainer review or test-vector validation would benefit from another pair of hands.

elmarco:
@glamberson if you can confirm that mstsc is happy with the encoded result, let's mark it as ready for review/merge. thanks!

glamberson:
@elmarco I'm doing an infrastructure migration so as soon as that's done I'll get this taken care of. Thanks.

glamberson:
I ran a differential test of this PR's encoder against FreeRDP 3.15.0's `rfx_rlgr_encode` to corroborate the changes against an independent implementation in addition to the spec citations in the commit messages.

**Method.** Vendored FreeRDP's `rfx_rlgr_encode` and `rfx_bitstream.h` as a standalone C file (about 250 LOC after stripping winpr macros, Apache-2.0), wrapped in a tiny CLI. Rust harness calls `ironrdp_graphics::rlgr::encode` directly, shells out to the C harness on the same input, byte-diffs the outputs, then roundtrips via `ironrdp_graphics::rlgr::decode` to confirm any byte-count slack is inert.

**Coverage (25 cases).**
- 6 in-tree fixtures (Y/CB/CR × RLGR3/RLGR1) from `crates/ironrdp-testsuite-core/tests/graphics/rlgr.rs`.
- 8 adversarial cases targeting each bug path:
  - Bug 1 (RL trailing zero): single zero, eight zeros, nonzero then zeros, 256 zero run.
  - Bug 2 (RLGR3 GR 2nd input exhaustion): three patterns covering positive, negative, and zero first values.
  - Bug 3 (RLGR1 GR kp churn): zero then more, alternating zero/nonzero.
- 8 generic stressors: increasing, signed walk, dense and sparse PRNG.

**Result.** 25 of 25 byte-content equal between patched IronRDP and FreeRDP. Roundtrip passes everywhere it can.

Two observations from running this:

1. Trailing-zero inputs in RL mode encode the same way in both implementations, but the decoder produces a phantom magnitude of 1 in the position past the run. This is inherent to MS-RDPRFX (the RL terminator's `mag = code+1 >= 1` math), not specific to either implementation. Real RFX tiles after DWT and quantization don't end in trailing zeros, so it doesn't affect production.

2. The two implementations occasionally differ in `processed_bytes` by 1 (e.g., 981 vs 982 on `CB_DATA` RLGR1). The extra byte on the FreeRDP side is always trailing 0x00 from `rfx_bitstream_flush` writing `8 - bits_left` zeros (which carries past the byte boundary), so `processed_bytes` returns `byte_pos + 1`. The encoded bit content is identical and the IronRDP decoder produces identical coefficients from either side's output. Whether IronRDP should match FreeRDP's count exactly is a separate enhancement question, independent of the three correctness fixes here.

Net: the three fixes look correct.

glamberson:
Confirmed against mstsc on Windows 11. With these three commits applied mstsc renders RemoteFX correctly. Without them, on RLGR1, it renders nothing at all.

| | RLGR3 | RLGR1 |
|---|---|---|
| master | renders correctly | **black screen** |
| master + this PR | renders correctly | renders correctly |

The failure mode is worth describing precisely, because it's more definite than a corrupted image. The session connects, NLA succeeds, and audio keeps playing throughout. Only the graphics are gone. The server logs no error of any kind: it encodes every tile, frames it, and sends it believing it's valid. mstsc takes the stream and draws nothing. I disconnected and reconnected to check and got the same black screen both times. IronRDP's own client, incidentally, renders garbage rather than black on the same bytes, so the two decoders disagree about what to do with a desynchronised stream but agree it's broken.

## Which commit

I bisected the three cumulatively using IronRDP's own client, comparing the decoded frame against the source image by RMSE:

| build | RLGR1 | RLGR3 |
|---|---|---|
| master | 0.3912 | 0.02592 |
| + align RLGR encoder RL mode | 0.3912 | 0.02592 |
| + zero default for exhausted RLGR3 second value | 0.3912 | 0.02592 |
| + UQ_GR in RLGR1 GR-mode kp update | 0.02592 | 0.02592 |

0.02592 is the quantisation floor for this image, so the visible failure is entirely `UP_GR` versus `UQ_GR`. That tracks with the code: `compute_rlgr1_magnitude` in the same file already uses `UQ_GR`, so before that commit the encoder and IronRDP's own decoder disagreed about how `kp` adapts. `kp` sets `k = kp >> LS_GR`, which picks run-length versus Golomb-Rice mode and the run field width, so once the two ends drift nothing after that point decodes.

The other two commits do change the encoded bytes, just not enough to see here. I'd still keep both. The RLGR3 odd-input default can't fire from the RFX server path, since `rfx_encode_component` requires exactly 4096 coefficients, but a wrong default is still worth fixing and `progressive.rs` calls the same encoder. The run-length tail is a genuine divergence from the reference encoder regardless of whether this particular content reaches it.

## Why this went unnoticed

mstsc advertises both coders. It sends TS_RFX_ICAP with RLGR1 first and RLGR3 second, and the ICAP loop in `server.rs` is last-wins, so the server always ends up on RLGR3. In practice IronRDP's RLGR1 encoder never runs against a real Windows client, which is why a total failure has been sitting there unreported. I had to override the selection to reach it at all.

MS-RDPRFX 2.2.1.1.1.1 describes the ICAP array as the set the decoder supports and doesn't define the ordering as a ranking, so I don't read this as ignoring a stated preference. The point is just that the choice is accidental, right next to the FIXME already in that arm about picking the last parsed. I'll raise that separately rather than pull it into this PR.

## How it was tested

Two builds of `crates/ironrdp/examples/server.rs` differing only in `rlgr.rs`, running side by side on separate ports so the same client hits both. The example's random noise generator is replaced with a fixed test pattern (colour bars, luminance ramp, checkerboards at 1 to 8 px, a line-pitch ramp, seeded noise, and flat grey with markers) so the decoded result can be compared against a known image, and the source frame is byte-identical between the two builds. Plus a small override to force the entropy coder, since otherwise RLGR1 is unreachable. Happy to post that patch if it would be useful to anyone else.

Leaving the ready-for-review call to you.

CBenoit (approved):
LGTM. Sorry for not noticing sooner

glamberson:
The first of the three fixes here has a regression, and #1569 corrects it.

RL mode codes the following value's magnitude minus one, so it cannot express
zero. Coding a zero when the run consumed the rest of the input decodes as a
magnitude of one, so every input ending in a zero run gains a trailing
coefficient it never had. Quantized wavelet coefficients almost always end in a
zero run, so this reaches every RemoteFX tile.

It surfaces as `progressive_fractional_base_quantization_reconstructs_rgb` from
#1499 failing on master. That test landed after this branch was based, so CI
here never ran it.

The other two fixes are correct and untouched, including the `UQ_GR` one that
produced the mstsc result I posted above.

PR #1181, opened 2026-03-21
feat(server): make run_connection generic over stream type

glamberson:
## Summary

Make `run_connection` accept any `AsyncRead + AsyncWrite` stream instead of requiring `TcpStream` concretely.

- `run_connection<S>(&mut self, stream: S)` where `S: AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static`
- Replace `peer_addr()` call in Hybrid/CredSSP path with `local_addr` fallback
- Remove unused `TcpStream` import

## Motivation

RDP servers may accept connections from transports other than TCP: Unix domain sockets (for local WebSocket relay), VSOCK (VM-to-host), or in-process streams (integration tests). The current `TcpStream` parameter prevents this, even though everything downstream of `run_connection` already operates generically (`TokioFramed<S>`, TLS accept, `accept_finalize<S>`, `client_loop<R, W>`).

Concrete use case: a QEMU display server that listens on both TCP (native RDP clients) and a Unix domain socket (browser clients via WebSocket relay) using the same `RdpServer` instance.

## Changes

**1 file, +8/-5 lines:**

- Generalize `run_connection` signature with `AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static` bounds (matching `accept_finalize`)
- Replace `peer_addr()` in the Hybrid/CredSSP path: the existing comment noted "doesn't seem to matter yet" for NTLM auth. Use `local_addr` as fallback since generic streams don't expose peer address
- Remove now-unused `TcpStream` import (`TcpListener` remains for `run()`)

## Backward Compatibility

Fully backward compatible. Callers passing `TcpStream` continue to work unchanged via type inference. `run()` is unaffected.

## Test Plan

- [x] `cargo xtask check fmt -v` passes
- [x] `cargo xtask check lints -v` passes
- [x] `cargo xtask check typos -v` passes
- [x] Verified `run()` compiles with inferred `TcpStream`
- [x] Verified `UnixStream` accepted via `run_connection` in downstream project

(Replaces #1180, which was closed due to head branch move.)

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -417,8 +417,7 @@ impl RdpServer {
                 if let RdpServerSecurity::Hybrid((_, pub_key)) = &self.opts.security {
                     // Generic streams don't expose peer address. Use a neutral
                     // placeholder; it's unclear whether CredSSP/NTLM actually
-                    // uses this value in practice (the original code noted
-                    // "doesn't seem to matter yet").
+                    // uses this value in practice.
                     let client_name = "rdp-client".to_owned();
 
                     ironrdp_acceptor::accept_credssp(


crates/ironrdp-server/src/server.rs
```diff
@@ -412,9 +415,9 @@ impl RdpServer {
                 acceptor.mark_security_upgrade_as_done();
 
                 if let RdpServerSecurity::Hybrid((_, pub_key)) = &self.opts.security {
-                    // how to get the client name?
-                    // doesn't seem to matter yet
-                    let client_name = framed.get_inner().0.get_ref().0.peer_addr()?.to_string();
+                    let client_name = self
+                        .local_addr
+                        .map_or_else(|| "rdp-client".to_owned(), |a| a.to_string());
```
CBenoit: The new implementation uses `self.local_addr` as the "client name," but this actually refers to the server’s own address, not the remote client. Since we can’t reliably acquire a dynamic client identifier with the new generic stream interface, maybe it’s clearer to just use the fallback string (e.g., `"rdp-client"`) for now. I would recommend replacing `self.local_addr` with the fallback value and adding a comment explaining that we aren’t capturing any dynamic client information at this stage, and it’s currently unclear if that detail is even important for this flow. This should avoid confusion and accurately reflect the current behavior.

crates/ironrdp-server/src/server.rs
```diff
@@ -380,7 +380,10 @@ impl RdpServer {
         acceptor.attach_static_channel(dvc);
     }
 
-    pub async fn run_connection(&mut self, stream: TcpStream) -> Result<()> {
+    pub async fn run_connection<S>(&mut self, stream: S) -> Result<()>
+    where
+        S: AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static,
```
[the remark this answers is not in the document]
glamberson: OK done.

crates/ironrdp-server/src/server.rs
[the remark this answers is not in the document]
glamberson: OK,. I'll use the constant fallback and add a comment. The original code already noted this "doesn't seem to matter yet" for NTLM auth, so a neutral placeholder is honest about the current state. Will also drop the 'static bound per the other thread.

glamberson: Good point. Switched to the constant fallback with a comment explaining the context. Also dropped the `'static` bound since `accept_finalize` doesn't require it. Pushed as 33e64c8.

crates/ironrdp-server/src/server.rs
```diff
@@ -412,9 +415,11 @@ impl RdpServer {
                 acceptor.mark_security_upgrade_as_done();
 
                 if let RdpServerSecurity::Hybrid((_, pub_key)) = &self.opts.security {
-                    // how to get the client name?
-                    // doesn't seem to matter yet
-                    let client_name = framed.get_inner().0.get_ref().0.peer_addr()?.to_string();
+                    // Generic streams don't expose peer address. Use a neutral
+                    // placeholder; it's unclear whether CredSSP/NTLM actually
+                    // uses this value in practice (the original code noted
+                    // "doesn't seem to matter yet").
```
CBenoit: The original code is not relevant anymore

```suggestion
                    // uses this value in practice.
```

PR #1188, opened 2026-03-28
build: upgrade sspi to 0.19, picky to rc.22, fix NTLM fallback

### LINKED ISSUE #1186: Help needed: newer picky/sspi versions unblock downstream integration, but IronRDP workspace still   hits a dependency conflict
I’m investigating IronRDP 0.14.0 integration in a larger Rust workspace, and the main problem I ran into was
  a dependency conflict in the current published dependency chain around security-related crates.

  More specifically, the dependency chain around `ironrdp-connector -> picky -> ...` conflicts with
  other crates in my workspace. That conflict is the reason I tried moving IronRDP to newer `picky` /
  `sspi` versions.

  I am not familiar enough with the IronRDP codebase to judge the correct upstream solution or the full
  impact of these changes, so I’m opening this issue mainly to share what I found and ask for guidance.

  I tested the following two changes in a fork:

  - build: update `picky` to `7.0.0-rc.22` and `sspi` to `0.19`
  - fix: adapt `ironrdp-connector` and `ironrdp-tokio` to `sspi 0.19`

  Commits:
  - https://github.com/sfwwslm/IronRDP/commit/7696d0eec3fac94f737bf15eab6ab1dba5377893
  - https://github.com/sfwwslm/IronRDP/commit/38548a283b754bb7f5fdbea462aee3df455ead4e

  These changes were enough to unblock my downstream integration scenario: after applying them in my
  fork, my project was able to build and run successfully again.

  The two code-level fixes were:

  1. `ironrdp-connector`
     - `sspi::KerberosConfig.client_computer_name` is no longer `Option<String>`
     - the `KerberosConfig -> sspi::KerberosConfig` conversion needs adjustment

  2. `ironrdp-tokio`
     - the reqwest network client needs to use the `sspi` types re-exported by `ironrdp-connector`
     - otherwise trait signatures no longer match after the `sspi 0.19` upgrade

  However, when checking the IronRDP workspace itself, I still hit another dependency conflict:

  ```text
  error: failed to select a version for `rand_core`.
  ...
  required by package `getrandom v0.4.0`
  ... which satisfies dependency `getrandom = "^0.4"` of package `uuid v1.22.0`
  ... which satisfies dependency `uuid = "^1.21"` of package `ironrdp-client`
  ...
  previously selected package `rand_core v0.10.0-rc-3`
  ... which satisfies dependency `rand_core = "=0.10.0-rc-3"` of package `picky v7.0.0-rc.22`

  So from my point of view:

  - upgrading picky / sspi seems useful and already helps in a real downstream integration scenario
  - the two compatibility fixes above also seem necessary for that path
  - but the IronRDP workspace still appears to need additional dependency alignment, likely involving:
      - picky
      - rand_core
      - getrandom
      - uuid
      - possibly ironrdp-client

  I’m not claiming this is a complete fix. I’m mainly sharing a reproducible investigation path and
  asking:

  1. does this upgrade direction make sense upstream?
  2. do the two compatibility fixes above look conceptually correct?
  3. what would be the right next step for the remaining workspace conflict?

  My main motivation here is resolving the current security dependency conflict in downstream projects,
  and this upgrade path already appears to help with that.

glamberson:
## Summary

Upgrade sspi from 0.18 to 0.19 and picky from rc.20 to rc.22, resolving the dependency conflicts reported in #1186 and the NTLM/SPNEGO compatibility issue in #1143.

- Use `ServerMode::Ntlm` directly when Kerberos is disabled, since sspi 0.19's SPNEGO implementation expects SPNEGO-wrapped tokens but mstsc/FreeRDP send raw NTLM tokens in CredSSP (@ramnes identified this in #1143, confirmed by @TheBestTvarynka)
- Implement new `CredentialsProxy::auth_data` trait method added in sspi 0.19
- Fix `KerberosConfig` conversion (`hostname` changed from `Option<String>` to `String` in sspi 0.19, identified by @sfwwslm in #1186)
- Route sspi types through `ironrdp-connector::sspi` re-export in `ironrdp-tokio` and remove the direct sspi dependency to avoid version mismatch (also identified by @sfwwslm)
- Pin uuid to <1.21 as a temporary workaround for `picky rc.22`'s exact-pinned `rand_core = "=0.10.0-rc-3"` conflicting with uuid 1.21+'s `getrandom 0.4` -> stable `rand_core 0.10.0` (documented by @sfwwslm in #1186). This pin can be removed when picky ships with stable RustCrypto deps.

Closes #1186. Supersedes #1143.

## Credit

This PR synthesizes the investigations by @ramnes (#1143) and @sfwwslm (#1186). @ramnes identified the NTLM fallback fix and first attempted the sspi upgrade in March. @sfwwslm independently found the `KerberosConfig` conversion fix, the reqwest.rs re-export fix, and documented the remaining `rand_core` workspace conflict. The uuid pin workaround and `auth_data` trait implementation are the remaining pieces needed for a complete workspace build.

## Test plan

- [ ] All 5 xtask checks pass locally (fmt, lints, tests, typos, locks)
- [ ] Full workspace builds including FFI crate
- [ ] Verify CredSSP connection with mstsc (NTLM, no Kerberos)
- [ ] Verify CredSSP connection with FreeRDP (NTLM, no Kerberos)

### DIFF
--- a/crates/ironrdp-connector/src/credssp.rs
+++ b/crates/ironrdp-connector/src/credssp.rs
@@ -34,7 +34,7 @@ impl From<KerberosConfig> for sspi::KerberosConfig {
     fn from(val: KerberosConfig) -> Self {
         sspi::KerberosConfig {
             kdc_url: val.kdc_proxy_url,
-            client_computer_name: val.hostname,
+            client_computer_name: val.hostname.unwrap_or_default(),
         }
     }
 }
--- a/crates/ironrdp-connector/src/credssp.rs
+++ b/crates/ironrdp-connector/src/credssp.rs
@@ -11,14 +11,14 @@ use crate::{
     ConnectorError, ConnectorErrorKind, ConnectorResult, Credentials, ServerName, Written, custom_err, general_err,
 };
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct KerberosConfig {
     pub kdc_proxy_url: Option<url::Url>,
-    pub hostname: Option<String>,
+    pub hostname: String,
 }
 
 impl KerberosConfig {
-    pub fn new(kdc_proxy_url: Option<String>, hostname: Option<String>) -> ConnectorResult<Self> {
+    pub fn new(kdc_proxy_url: Option<String>, hostname: String) -> ConnectorResult<Self> {
         let kdc_proxy_url = kdc_proxy_url
             .map(|url| url::Url::parse(&url))
             .transpose()
@@ -34,7 +34,7 @@ impl From<KerberosConfig> for sspi::KerberosConfig {
     fn from(val: KerberosConfig) -> Self {
         sspi::KerberosConfig {
             kdc_url: val.kdc_proxy_url,
-            client_computer_name: val.hostname.unwrap_or_default(),
+            client_computer_name: val.hostname,
         }
     }
 }
--- a/crates/ironrdp-tokio/Cargo.toml
+++ b/crates/ironrdp-tokio/Cargo.toml
@@ -18,7 +18,7 @@ test = false
 
 [features]
 default = []
-reqwest = ["dep:reqwest", "dep:sspi", "dep:url", "dep:ironrdp-connector"]
+reqwest = ["dep:reqwest", "dep:url", "dep:ironrdp-connector"]
 reqwest-rustls-ring = ["reqwest", "reqwest?/rustls-tls-webpki-roots"]
 reqwest-native-tls = ["reqwest", "reqwest?/native-tls"]
 
@@ -27,7 +27,7 @@ ironrdp-async = { path = "../ironrdp-async", version = "0.8" } # public
 ironrdp-connector = { path = "../ironrdp-connector", version = "0.8", optional = true }
 tokio = { version = "1", features = ["io-util"] }
 reqwest = { version = "0.12", default-features = false, features = ["http2", "system-proxy"], optional = true }
-sspi = { version = "0.18", features = ["network_client", "dns_resolver"], optional = true }
+# sspi types accessed through ironrdp-connector::sspi re-export
 url = { version = "2.5", optional = true }
 
 [lints]
--- a/crates/ironrdp-tokio/Cargo.toml
+++ b/crates/ironrdp-tokio/Cargo.toml
@@ -27,7 +27,6 @@ ironrdp-async = { path = "../ironrdp-async", version = "0.8" } # public
 ironrdp-connector = { path = "../ironrdp-connector", version = "0.8", optional = true }
 tokio = { version = "1", features = ["io-util"] }
 reqwest = { version = "0.12", default-features = false, features = ["http2", "system-proxy"], optional = true }
-# sspi types accessed through ironrdp-connector::sspi re-export
 url = { version = "2.5", optional = true }
 
 [lints]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,19 +26,19 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
-version = "0.6.0-rc.2"
+version = "0.6.0-rc.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac8202ab55fcbf46ca829833f347a82a2a4ce0596f0304ac322c2d100030cd56"
+checksum = "67a578e7d4edaef88aeb9cdd81556f4a62266ce26601317c006a79e8bc58b5af"
 dependencies = [
- "crypto-common 0.2.0-rc.4",
+ "crypto-common 0.2.0-rc.8",
  "inout",
 ]
 
 [[package]]
 name = "aes"
-version = "0.9.0-rc.1"
+version = "0.9.0-rc.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e713c57c2a2b19159e7be83b9194600d7e8eb3b7c2cd67e671adf47ce189a05"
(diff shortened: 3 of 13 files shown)


crates/ironrdp-tokio/Cargo.toml
```diff
@@ -27,7 +27,7 @@ ironrdp-async = { path = "../ironrdp-async", version = "0.8" } # public
 ironrdp-connector = { path = "../ironrdp-connector", version = "0.8", optional = true }
 tokio = { version = "1", features = ["io-util"] }
 reqwest = { version = "0.12", default-features = false, features = ["http2", "system-proxy"], optional = true }
-sspi = { version = "0.18", features = ["network_client", "dns_resolver"], optional = true }
+# sspi types accessed through ironrdp-connector::sspi re-export
```
CBenoit:
```suggestion
```

crates/ironrdp-connector/src/credssp.rs
```diff
@@ -34,7 +34,7 @@ impl From<KerberosConfig> for sspi::KerberosConfig {
     fn from(val: KerberosConfig) -> Self {
         sspi::KerberosConfig {
             kdc_url: val.kdc_proxy_url,
-            client_computer_name: val.hostname,
+            client_computer_name: val.hostname.unwrap_or_default(),
```
CBenoit: question: Is the empty string a correct value at this place?

CBenoit: cc @TheBestTvarynka

glamberson: Good question. I traced the code path through sspi and the answer is **no, empty string is not correct here**, but the fix belongs in sspi rather than in this conversion.

**What happens with `""`:** In sspi 0.19, `client_computer_name` flows unconditionally into `generate_as_req_kdc_body()` ([generators.rs:272-274](https://github.com/Devolutions/sspi-rs/blob/v0.19.2/src/kerberos/client/generators.rs#L272-L274)), which always constructs a `HostAddress` with `addr_type = NET_BIOS_ADDR_TYPE` and `address = hostname.as_bytes()`. There is no empty check. An empty string produces a 0-byte NetBIOS address in the AS-REQ `addresses` field, which is malformed per RFC 4120 section 5.2.5 (NetBIOS addresses should be 16 octets). It also ends up in `EncKrbPrivPart.s_address` for password change flows.

**What sspi 0.18 did:** The field was `Option<String>`, and `None` was guarded by [`unwrap_hostname()`](https://github.com/Devolutions/sspi-rs/blob/v0.18.7/src/kerberos/utils.rs#L98-L104) which returned `Err("the hostname is not provided")`, a clear, fast failure. sspi 0.19 made the field non-optional but removed that guard without adding empty-string validation, so the safety net is gone.

**Why `unwrap_or_default()` here:** IronRDP's `KerberosConfig.hostname` is `Option<String>` and sspi 0.19's `client_computer_name` is now `String`. The `From` impl needs to bridge the types. `unwrap_or_default()` is the minimal change that compiles, but it silently degrades from "clear error" to "malformed protocol message."

**Practical scope:** This conversion only runs when Kerberos is explicitly configured. Users who configure Kerberos but omit the hostname would have gotten an error with sspi 0.18 and now get silent protocol violations with 0.19. Most IronRDP users use NTLM (no Kerberos config), so the blast radius is small.

Since @TheBestTvarynka is already tagged, I think the right fix is in sspi: either validate/reject empty `client_computer_name`, or skip the `addresses` field in AS-REQ when it is empty (the field is OPTIONAL per RFC 4120). Happy to adjust the IronRDP side to whatever approach the sspi team prefers: making `hostname` non-optional in `KerberosConfig`, using `TryFrom` instead of `From`, etc.

crates/ironrdp-connector/src/credssp.rs
```diff
@@ -34,7 +34,7 @@ impl From<KerberosConfig> for sspi::KerberosConfig {
...
         }
```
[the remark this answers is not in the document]
glamberson: See above....

crates/ironrdp-connector/src/credssp.rs
CBenoit: I cc'd Pavlo on Slack to make sure he sees this. I’m suspecting the intention is to make sure the user is providing a hostname by removing the `None` case. In that case, it may make sense to also make IronRDP’s KerberosConfig.hostname `String` instead of `Option<String>`, as we were mirroring sspi-rs in the first place. But I would like Pavlo to confirm

TheBestTvarynka: Hi, @glamberson and @CBenoit 

Sorry for the late response. GH did not send me a notification for some reason (or I just missed it).

Previously (even in sspi 0.18.7), we always expected the hostname to be present: `sspi-rs` returned an error if no hostname was specified: https://docs.rs/crate/sspi/0.18.7/source/src/kerberos/utils.rs#98. So, during the last refactorings, I made it required in the `KerberosConfig`.

From RFC 4210:

> These are the addresses from which the ticket can be used. If there are no addresses, the ticket can be used from any location.  The decision by the KDC to issue or by the end server to accept addressless tickets is a policy decision and is left to the Kerberos and end-service administrators;
> Network addresses are included in the ticket to make it harder for an attacker to use stolen credentials.

I decided to keep the hostname required because it makes the ticket _more specific_. As I remember, the default KDC policies allow requests without the `addresses` field, but, as I understood from the RFC, it can be changed in the domain policies.

If it is not hard to provide the `hostname`, I would like to keep it. Otherwise, it's easy to make it optional in `sspi`.

glamberson: Thanks @TheBestTvarynka . That confirms it. I'll update this PR to make KerberosConfig.hostname a required String instead of Option<String>, matching sspi 0.19's intent. The unwrap_or_default() goes away.

Thanks.

CBenoit (approved):
LGTM

Issue #1158, opened 2026-03-09
Multi-Codec Graphics Pipeline: Adaptive Content Encoding via EGFX


glamberson:
# Multi-Codec Graphics Pipeline for IronRDP: Adaptive Content Encoding via EGFX

## A Case for Implementing the Full EGFX Codec Suite

**Presented by Lamco Development**

---

### Introduction

This document examines the RDP Graphics Pipeline Extension [1] as it was designed: a multi-codec graphics system. The EGFX wire format carries per-PDU codec identifiers. The framing model supports mixed-codec frames. The capability exchange negotiates codec availability per session. Microsoft's Azure Virtual Desktop uses this multi-codec model in production [11], encoding text with ClearCodec, video with H.264, and photographic content with progressive refinement -- all within single frame boundaries. This is not a theoretical protocol capability; it is how the world's largest RDP deployment delivers graphics today.

Most open-source RDP implementations treat EGFX as an H.264 delivery mechanism. This works, but it leaves most of the protocol's design unused. ClearCodec [8] -- the lossless text codec that the spec declares mandatory for all EGFX implementations -- is not implemented in any open-source RDP server. Since approximately 80% of typical remote session content is text [11], this means every open-source implementation is either applying lossy video compression to its dominant content type or falling back to raw bitmap.

Lamco Development has been building on IronRDP [16] for over a year, contributing EGFX infrastructure, server-side extensions, and Linux-specific integration. Through this work, we have developed a production health monitoring and stream observability system that illustrates the operational dimension of multi-codec encoding: when each tile carries a codec tag, the telemetry is richer and the adaptation space is wider. We describe this system in Section 11 because we believe session observability deserves consideration as a first-class interface in the library.

We have also developed implementations of the missing EGFX codecs (ClearCodec, NSCodec, RemoteFX Progressive) and the mixed-frame server API, and are ready to contribute these to IronRDP as upstream open-source submissions.

What follows is a technical overview of the EGFX protocol architecture, the codec suite it defines, and how the pieces fit together.

### 1. The Three Eras of RDP Graphics

The evolution of RDP's graphics subsystem tells a story about the progressive abandonment of monolithic encoding in favor of content-aware pipelines. Understanding this history is necessary to understand why multi-codec EGFX is the only viable path forward for a serious RDP implementation.

[...]

PR #1196, opened 2026-04-01
feat(pdu,graphics): add RemoteFX Progressive codec primitives

glamberson:
Part of the multi-codec EGFX implementation described in #1158 (Section 4).

## Summary

Add wire-format types for RemoteFX Progressive Codec (MS-RDPRFX Progressive Extension) and the computational primitives required for progressive refinement.

**ironrdp-pdu**: Progressive codec block types (SYNC, CONTEXT, REGION, TILE_SIMPLE, TILE_FIRST, TILE_UPGRADE, FRAME_BEGIN, FRAME_END), quantization values with band-level accessors, tile header parsing.

**ironrdp-graphics**: Reduce-extrapolate discrete wavelet transform (DWT) for 64x64 tile decomposition into 10 subbands. SRL (Subband Reconstruction Layer) entropy codec for progressive coefficient transmission with zero-run encoding and sign-magnitude packing.

## Changes

5 files, 2,326 lines. No changes to existing code.

## Test plan

- [ ] All 5 xtask gates pass
- [ ] DWT and SRL have inline unit tests

### DIFF
--- a/crates/ironrdp-graphics/src/dwt_extrapolate.rs
+++ b/crates/ironrdp-graphics/src/dwt_extrapolate.rs
@@ -1,0 +1,590 @@
+//! Reduce-extrapolate variant of the LeGall 5/3 DWT for progressive RFX.
+//!
+//! When `RFX_DWT_REDUCE_EXTRAPOLATE` (0x01) is set in the progressive context,
+//! the DWT uses boundary extrapolation instead of symmetric extension. This
+//! produces asymmetric subbands that avoid wraparound artifacts at tile edges.
+//!
+//! For a 64x64 tile with 3-level decomposition:
+//!
+//!   Level 1: HL1(31x33), LH1(33x31), HH1(31x31) — LL1(33x33) feeds level 2
+//!   Level 2: HL2(16x17), LH2(17x16), HH2(16x16) — LL2(17x17) feeds level 3
+//!   Level 3: HL3(8x9), LH3(9x8), HH3(8x8), LL3(9x9)
+//!
+//! Buffer layout (4096 coefficients total):
+//!   [HL1:1023][LH1:1023][HH1:961][HL2:272][LH2:272][HH2:256][HL3:72][LH3:72][HH3:64][LL3:81]
+
+/// Subband position and dimensions within the coefficient buffer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BandInfo {
+    /// Width (columns) of the subband.
+    pub width: usize,
+    /// Height (rows) of the subband.
+    pub height: usize,
+    /// Starting index in the linearized coefficient array.
+    pub offset: usize,
+}
+
+impl BandInfo {
+    /// Total number of coefficients in this subband.
+    pub const fn count(&self) -> usize {
+        self.width * self.height
+    }
+}
+
+/// Low-pass output count for the reduce-extrapolate split.
+///
+/// For even N: `N/2 + 1` (extrapolated sample).
+/// For odd N: `(N + 1) / 2` (standard ceiling division).
+fn low_count(n: usize) -> usize {
+    // Even: (64+2)/2=33. Odd: (33+2)/2=17, (17+2)/2=9.
+    (n + 2) / 2
+}
+
+/// High-pass output count: remainder after low-pass.
+fn high_count(n: usize) -> usize {
+    n - low_count(n)
+}
+
+/// Band layout for the reduce-extrapolate 3-level DWT on a 64x64 tile.
+///
+/// Returns 10 subbands in buffer order:
+/// `[HL1, LH1, HH1, HL2, LH2, HH2, HL3, LH3, HH3, LL3]`
+///
+/// Band indices for progressive quantization (`ComponentCodecQuant::for_band()`):
+///   0=LL3, 1=HL3, 2=LH3, 3=HH3, 4=HL2, 5=LH2, 6=HH2, 7=HL1, 8=LH1, 9=HH1
+#[expect(clippy::similar_names, reason = "lw/hw/lh/hh are standard DWT band dimensions")]
+pub fn band_layout() -> [BandInfo; 10] {
+    let (lw1, hw1) = (low_count(64), high_count(64)); // (33, 31)
+    let (lw2, hw2) = (low_count(lw1), high_count(lw1)); // (17, 16)
+    let (lw3, hw3) = (low_count(lw2), high_count(lw2)); // (9, 8)
+
+    // Vertical dimensions are the same (square tile)
+    let (lh1, hh1) = (lw1, hw1);
+    let (lh2, hh2) = (lw2, hw2);
+    let (lh3, hh3) = (lw3, hw3);
+
+    let mut off = 0;
+    let mut b = |w: usize, h: usize| {
+        let info = BandInfo {
+            width: w,
+            height: h,
+            offset: off,
+        };
+        off += w * h;
+        info
+    };
+
+    [
+        b(hw1, lh1), // HL1: 31x33 = 1023
+        b(lw1, hh1), // LH1: 33x31 = 1023
+        b(hw1, hh1), // HH1: 31x31 = 961
+        b(hw2, lh2), // HL2: 16x17 = 272
+        b(lw2, hh2), // LH2: 17x16 = 272
+        b(hw2, hh2), // HH2: 16x16 = 256
+        b(hw3, lh3), // HL3: 8x9  = 72
+        b(lw3, hh3), // LH3: 9x8  = 72
+        b(hw3, hh3), // HH3: 8x8  = 64
+        b(lw3, lh3), // LL3: 9x9  = 81
+    ]
+}
+
+/// Inverse 3-level reduce-extrapolate DWT (coefficients to 64x64 tile).
+///
+/// Reconstructs the tile in-place in `buffer`. Both slices must have at least
+/// 4096 elements.
+///
+/// # Panics
+///
+/// Panics if either slice has fewer than 4096 elements.
+pub fn decode(buffer: &mut [i16], temp: &mut [i16]) {
+    assert!(buffer.len() >= 4096, "buffer must hold 4096 coefficients");
+    assert!(temp.len() >= 4096, "temp must hold 4096 elements");
+
+    // Inner-to-outer: level 3 first (smallest subbands), then 2, then 1
+    decode_block(&mut buffer[3807..], temp, 9, 8);
+    decode_block(&mut buffer[3007..], temp, 17, 16);
+    decode_block(buffer, temp, 33, 31);
+}
+
+/// Forward 3-level reduce-extrapolate DWT (64x64 tile to coefficients).
+///
+/// Transforms the tile in-place in `buffer`. Both slices must have at least
+/// 4096 elements.
+///
+/// # Panics
+///
+/// Panics if either slice has fewer than 4096 elements.
+pub fn encode(buffer: &mut [i16], temp: &mut [i16]) {
+    assert!(buffer.len() >= 4096, "buffer must hold 4096 coefficients");
+    assert!(temp.len() >= 4096, "temp must hold 4096 elements");
+
+    // Outer-to-inner: level 1 first (full tile), then 2, then 3
+    encode_block(buffer, temp, 33, 31);
+    encode_block(&mut buffer[3007..], temp, 17, 16);
+    encode_block(&mut buffer[3807..], temp, 9, 8);
+}
+
+// ---------------------------------------------------------------------------
+// Inverse (decode) implementation
+// ---------------------------------------------------------------------------
+
+/// Inverse DWT for one decomposition level.
+///
+/// `n_l` and `n_h` are the low-pass and high-pass band counts.
+/// Buffer contains `[HL, LH, HH, LL]` contiguously.
+#[expect(clippy::similar_names, reason = "hl/lh/hh/ll are standard DWT subband names")]
+fn decode_block(buffer: &mut [i16], temp: &mut [i16], n_l: usize, n_h: usize) {
+    let dst_w = n_l + n_h;
+
+    // Subband offsets within this block's buffer region
+    let hl_off = 0;
+    let lh_off = n_h * n_l;
+    let hh_off = lh_off + n_l * n_h;
+    let ll_off = hh_off + n_h * n_h;
+
+    // Temp: L half (n_l rows x dst_w cols) then H half (n_h rows x dst_w cols)
+    let l_off = 0;
+    let h_off = n_l * dst_w;
+
+    // Horizontal inverse: (LL + HL) -> L rows
+    for row in 0..n_l {
+        let ll_start = ll_off + row * n_l;
+        let hl_start = hl_off + row * n_h;
+        let l_start = l_off + row * dst_w;
+        idwt_row(
+            &buffer[ll_start..ll_start + n_l],
+            &buffer[hl_start..hl_start + n_h],
+            &mut temp[l_start..l_start + dst_w],
+        );
+    }
+
+    // Horizontal inverse: (LH + HH) -> H rows
+    for row in 0..n_h {
+        let lh_start = lh_off + row * n_l;
+        let hh_start = hh_off + row * n_h;
+        let h_start = h_off + row * dst_w;
+        idwt_row(
+            &buffer[lh_start..lh_start + n_l],
+            &buffer[hh_start..hh_start + n_h],
+            &mut temp[h_start..h_start + dst_w],
+        );
+    }
+
+    // Vertical inverse: (L + H) -> reconstructed columns in buffer
+    for col in 0..dst_w {
+        idwt_col(temp, l_off + col, h_off + col, dst_w, buffer, col, dst_w, n_l, n_h);
+    }
+}
+
+/// 1D inverse DWT on a contiguous row.
+///
+/// Reconstructs `n_l + n_h` output samples from `n_l` low-pass and `n_h`
+/// high-pass coefficients.
+fn idwt_row(low: &[i16], high: &[i16], dst: &mut [i16]) {
+    let n_l = low.len();
+    let n_h = high.len();
+
+    let mut h0 = i32::from(high[0]);
+    let mut x0 = t(i32::from(low[0]) - h0);
+    let mut x2 = x0;
+
+    let mut di = 0;
+
+    for j in 0..n_h - 1 {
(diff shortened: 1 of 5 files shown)


crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+//! blocks forming a progressive bitmap stream.
+
+use core::iter;
+
+use ironrdp_core::{
+    Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteCursor, cast_length, ensure_fixed_part_size,
+    ensure_size, invalid_field_err,
+};
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive as _;
```
CBenoit: issue: We are phasing out num_derive and num_traits, so I would rather not add more code using these traits.

CBenoit: **suggestion (non-blocking):** The sibling files in this module already use `num-derive`, so this is consistent, but the project is phasing these crates out (see root `Cargo.toml` comment). For new code, prefer an explicit match on the raw discriminant:

```rust
fn from_u16(raw: u16) -> Option<Self> {
    match raw {
        0xCCC0 => Some(Self::Sync),
        0xCCC1 => Some(Self::FrameBegin),
        0xCCC2 => Some(Self::FrameEnd),
        0xCCC3 => Some(Self::Context),
        0xCCC4 => Some(Self::Region),
        0xCCC5 => Some(Self::TileSimple),
        0xCCC6 => Some(Self::TileFirst),
        0xCCC7 => Some(Self::TileUpgrade),
        _ => None,
    }
}
```

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+    fn decode(src: &mut ReadCursor<'de>) -> DecodeResult<Self> {
+        ensure_size!(ctx: Self::NAME, in: src, size: Self::HEADER_SIZE);
+
+        let tile_size = src.read_u8();
+        if tile_size != 0x40 {
+            return Err(invalid_field_err!("tileSize", "only 64x64 tiles supported"));
+        }
+
+        let num_rects = usize::from(src.read_u16());
+        let num_quant = usize::from(src.read_u8());
```
CBenoit, line 868: **issue:** Two guards present in FreeRDP (`progressive.c:2004-2020`) are missing here:

1. `numRects` must be >= 1 — a region with zero rectangles is malformed; FreeRDP returns `-1013` for this case.
2. `numQuant` must be <= 7 — FreeRDP hard-rejects values > 7 (`-1014`). Without this bound, a tile's `quant_idx_*` (a `u8`) can reference an index beyond the parsed `quant_vals` vec, causing a panic downstream when the codec layer indexes into it.

```rust
if num_rects == 0 {
    return Err(invalid_field_err!("numRects", "region must contain at least one rectangle"));
}
if num_quant > 7 {
    return Err(invalid_field_err!("numQuant", "quant count exceeds maximum of 7"));
}
```

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+            let tile = match header.block_type {
+                ProgressiveBlockType::TileSimple => ProgressiveTile::Simple(TileSimple::decode(tile_src)?),
+                ProgressiveBlockType::TileFirst => ProgressiveTile::First(TileFirst::decode(tile_src)?),
+                ProgressiveBlockType::TileUpgrade => ProgressiveTile::Upgrade(TileUpgrade::decode(tile_src)?),
+                _ => {
+                    return Err(invalid_field_err!("blockType", "expected tile block inside region"));
+                }
+            };
+            tiles.push(tile);
+        }
```
CBenoit, line 924: **issue:** After collecting `tiles`, there is no check that `quant_idx_y/cb/cr` are within bounds of `quant_vals`. FreeRDP validates this for every tile in `progressive_decompress_tile_first` and `_tile_upgrade` (lines 908-929 and 1340-1360). Without the `numQuant <= 7` guard above, any `quant_idx_*` value of 0-255 can arrive here, and callers that do `quant_vals[quant_idx_y as usize]` will panic rather than return an error. Add a pass after the tile loop:

```rust
for tile in &tiles {
    let (iy, icb, icr) = match tile {
        ProgressiveTile::Simple(t)  => (t.quant_idx_y, t.quant_idx_cb, t.quant_idx_cr),
        ProgressiveTile::First(t)   => (t.quant_idx_y, t.quant_idx_cb, t.quant_idx_cr),
        ProgressiveTile::Upgrade(t) => (t.quant_idx_y, t.quant_idx_cb, t.quant_idx_cr),
    };
    if usize::from(iy) >= quant_vals.len()
        || usize::from(icb) >= quant_vals.len()
        || usize::from(icr) >= quant_vals.len()
    {
        return Err(invalid_field_err!("quantIdx", "tile quant index out of range"));
    }
}
```

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+
+    while src.len() >= ProgressiveBlockHeader::SIZE {
+        let header = ProgressiveBlockHeader::decode(&mut src)?;
+        let body_len = header
+            .block_len
+            .checked_sub(BLOCK_HEADER_SIZE_U32)
+            .ok_or_else(|| invalid_field_err!("blockLen", "block length too small"))?;
+        let body_len: usize = cast_length!("bodyLen", body_len)?;
+        ensure_size!(ctx: "ProgressiveStream", in: src, size: body_len);
+        let body_src = &mut ReadCursor::new(src.read_slice(body_len));
```
CBenoit, line 995: **issue:** MS-RDPEGFX §2.2.4.2.1.1-4 is normative about exact `blockLen` values for fixed-size blocks (SYNC=12, FRAME_BEGIN=12, FRAME_END=6, CONTEXT=10). FreeRDP rejects mismatches (`progressive_wb_frame_begin` returns `-1005` if `blockLen != 12`). Right now, a sender with a wrong `blockLen` that still leaves enough bytes decodes silently. Consider adding a size check before dispatching:

```rust
let expected_body: Option<usize> = match header.block_type {
    ProgressiveBlockType::Sync      => Some(ProgressiveSyncPdu::FIXED_PART_SIZE),
    ProgressiveBlockType::FrameBegin => Some(ProgressiveFrameBeginPdu::FIXED_PART_SIZE),
    ProgressiveBlockType::FrameEnd  => Some(0),
    ProgressiveBlockType::Context   => Some(ProgressiveContextPdu::FIXED_PART_SIZE),
    _ => None,
};
if let Some(expected) = expected_body {
    if body_len != expected {
        return Err(invalid_field_err!("blockLen", "unexpected size for fixed-size block"));
    }
}
```

crates/ironrdp-graphics/src/srl.rs
```diff
@@ -0,0 +1,383 @@
...
+                writer.write_bits(zero_count, k);
+                idx += usize::try_from(zero_count).unwrap_or(usize::MAX);
+
+                if zero_count > 0 {
+                    continue;
+                }
+            } else {
+                // k == 0, each zero is individually handled
+                // (zero coefficient in unary mode is skipped, we move to next)
+                idx += usize::try_from(zero_count).unwrap_or(usize::MAX);
```
CBenoit: **question:** When `k == 0`, the encoder silently advances past the entire zero run without writing any bits for it. On the decode side, `k == 0` means the loop falls through to unary mode for every slot — it never emits a zero in that mode either. So if an encoder starts with `kp = 0` and a leading run of zeros, those zeros are swallowed, and the decoder will read unary-coded values for positions that should be zero. Is this intentional? The `encode_all_zeros` test accepts any output ending in `0x00`, which would not catch this. A test that round-trips a mixed slice starting with zeros while `kp` is 0 would be a good addition.

crates/ironrdp-graphics/src/srl.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    #[test]
+    fn round_trip_nonzero_only() {
+        // SRL is designed for coefficients already identified as non-zero
+        // by the DAS (delta-analysis state). Test with only non-zero values.
+        let original = vec![1, -1, 2, -3, 1];
+        let encoded = encode_srl(&original, 4);
+        let decoded = decode_srl(&encoded, original.len(), 4);
+        for (i, (&orig, &dec)) in original.iter().zip(decoded.iter()).enumerate() {
+            assert_eq!(orig.signum(), dec.signum(), "index {i}: sign mismatch");
```
CBenoit: **issue:** Only the sign is checked, not the magnitude. A decoder bug that consistently returns the wrong absolute value would pass this test. Please assert exact equality:

```rust
assert_eq!(decoded, original, "SRL round-trip must preserve exact coefficient values");
```

crates/ironrdp-graphics/src/srl.rs
```diff
@@ -0,0 +1,383 @@
...
+            let zeros = reader.read_bits(k);
+            if zeros > 0 {
+                nz = zeros;
+                nz -= 1;
+                output[out_idx] = 0;
+                out_idx += 1;
+                continue;
+            }
+            // Fall through to unary mode (no more zeros)
+        }
```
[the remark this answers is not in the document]
glamberson: You're right on all three threads (1, 2, 3). They're the same underlying issue: the encoder and decoder disagree on the wire format when k=0, and additionally the unary magnitude encoding has an off-by-one between encode and decode.

Traced through the actual bit-level round-trip and found three bugs:

1. Decoder skipped zero-run mode entirely at k=0, producing non-zero values for zero positions.
2. Encoder silently dropped zeros at k=0 (wrote nothing), and also batched zero chunks with a stale k value, causing kp to diverge from the decoder's kp after mixed zero/non-zero sequences.
3. Magnitude coding: decoder started quotient at 1 (should be 0), encoder wrote (quotient-1) zeros (should be quotient). The existing test only checked signum() so it didn't catch the value mismatch.

Fixed by: removing the k>0 guard in the decoder (k=0 degenerates to single-zero chunks naturally), rewriting the encoder to process zero chunks one at a time recomputing k each iteration, and aligning both sides to magnitude = (quotient << extra_bits) | remainder with quotient = count of zero-bits before terminator.

Added exact-value round-trip tests including mixed zero/non-zero and all-zeros inputs.

crates/ironrdp-graphics/src/srl.rs
```diff
@@ -0,0 +1,383 @@
...
+                idx += usize::try_from(zero_count).unwrap_or(usize::MAX);
+
+                if zero_count > 0 {
+                    continue;
+                }
+            } else {
+                // k == 0, each zero is individually handled
+                // (zero coefficient in unary mode is skipped, we move to next)
+                idx += usize::try_from(zero_count).unwrap_or(usize::MAX);
+                continue;
```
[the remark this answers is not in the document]
glamberson: See above.

crates/ironrdp-graphics/src/srl.rs
```diff
@@ -0,0 +1,383 @@
...
+            magnitude += 1;
+        }
+
+        // Read remaining magnitude bits (num_bits - 1 bits)
+        let extra_bits = u32::from(num_bits).saturating_sub(1);
+        if extra_bits > 0 && extra_bits < 16 {
+            let extra = reader.read_bits(extra_bits);
+            magnitude = (magnitude << extra_bits) | extra;
+        }
+
```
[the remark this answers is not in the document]
glamberson: See above.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
glamberson: Done. Replaced FromPrimitive derive with a manual from_u16() match, same pattern as RequestedProtocol in multitransport.rs and others.

glamberson: Done. Replaced FromPrimitive derive with a manual from_u16() match, same pattern as RequestedProtocol in multitransport.rs and others.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+                ProgressiveBlock::FrameBegin(ProgressiveFrameBeginPdu::decode(body_src)?)
+            }
+            ProgressiveBlockType::FrameEnd => ProgressiveBlock::FrameEnd(ProgressiveFrameEndPdu::decode(body_src)?),
+            ProgressiveBlockType::Context => ProgressiveBlock::Context(ProgressiveContextPdu::decode(body_src)?),
+            ProgressiveBlockType::Region => ProgressiveBlock::Region(ProgressiveRegion::decode(body_src)?),
+            // Tile blocks should only appear inside regions; skip at top level
+            ProgressiveBlockType::TileSimple | ProgressiveBlockType::TileFirst | ProgressiveBlockType::TileUpgrade => {
+                return Err(invalid_field_err!("blockType", "tile block outside of region"));
+            }
+        };
```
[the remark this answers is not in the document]
glamberson: Same situation as the ClearCodec trailing bytes discussion on #1174. The cursor is sliced to exactly body_len bytes via read_slice(), so trailing bytes stay within the length-delimited boundary and can't affect the outer stream. The spec is silent on trailing bytes within blocks and FreeRDP handles them identically. The PDU crate is side-effect-free with no logging, so strict errors would risk rejecting valid server output. No change.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+            let tile_src = &mut ReadCursor::new(src.read_slice(body_len));
+
+            let tile = match header.block_type {
+                ProgressiveBlockType::TileSimple => ProgressiveTile::Simple(TileSimple::decode(tile_src)?),
+                ProgressiveBlockType::TileFirst => ProgressiveTile::First(TileFirst::decode(tile_src)?),
+                ProgressiveBlockType::TileUpgrade => ProgressiveTile::Upgrade(TileUpgrade::decode(tile_src)?),
+                _ => {
+                    return Err(invalid_field_err!("blockType", "expected tile block inside region"));
+                }
+            };
```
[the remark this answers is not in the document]
glamberson: See block-level comment above, same reasoning for tiles.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+        }
+
+        // Each tile is wrapped in a block header
+        for tile in &self.tiles {
+            let (block_type, body_size) = match tile {
+                ProgressiveTile::Simple(s) => (ProgressiveBlockType::TileSimple, s.size()),
+                ProgressiveTile::First(f) => (ProgressiveBlockType::TileFirst, f.size()),
+                ProgressiveTile::Upgrade(u) => (ProgressiveBlockType::TileUpgrade, u.size()),
+            };
+            let block_len = u32::try_from(ProgressiveBlockHeader::SIZE + body_size).expect("tile size fits u32");
```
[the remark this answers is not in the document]
glamberson: Switched both to cast_length! to match the rest of the file.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
```diff
@@ -0,0 +1,1350 @@
...
+
+    for block in blocks {
+        let (block_type, body_size) = match block {
+            ProgressiveBlock::Sync(s) => (ProgressiveBlockType::Sync, s.size()),
+            ProgressiveBlock::FrameBegin(f) => (ProgressiveBlockType::FrameBegin, f.size()),
+            ProgressiveBlock::FrameEnd(f) => (ProgressiveBlockType::FrameEnd, f.size()),
+            ProgressiveBlock::Context(c) => (ProgressiveBlockType::Context, c.size()),
+            ProgressiveBlock::Region(r) => (ProgressiveBlockType::Region, r.size()),
+        };
+        let block_len = u32::try_from(ProgressiveBlockHeader::SIZE + body_size).expect("block size fits u32");
```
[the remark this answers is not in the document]
glamberson: Same as above.

crates/ironrdp-graphics/src/dwt_extrapolate.rs
```diff
@@ -0,0 +1,590 @@
...
+    n - low_count(n)
+}
+
+/// Band layout for the reduce-extrapolate 3-level DWT on a 64x64 tile.
+///
+/// Returns 10 subbands in buffer order:
+/// `[HL1, LH1, HH1, HL2, LH2, HH2, HL3, LH3, HH3, LL3]`
+///
+/// Band indices for progressive quantization (`ComponentCodecQuant::for_band()`):
+///   0=LL3, 1=HL3, 2=LH3, 3=HH3, 4=HL2, 5=LH2, 6=HH2, 7=HL1, 8=LH1, 9=HH1
```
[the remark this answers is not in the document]
glamberson: The doc comment had the band order reversed. 

The actual for_band() implementation uses 0=HL1 through 9=LL3, which matches the buffer return order of band_layout(). Corrected the comment.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
glamberson: Added both guards.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
glamberson: Added the bounds check after tile collection.

crates/ironrdp-pdu/src/codecs/rfx/progressive.rs
glamberson: Added the fixed-size block length validation.

crates/ironrdp-graphics/src/srl.rs
glamberson: Same as above. 

Fixed. The encoder now processes zero runs at all k values (including k=0) through the same chunk loop, recomputing k after each kp update to stay in sync with the decoder. Added round_trip_mixed_zeros test with leading zeros while kp=0.

crates/ironrdp-graphics/src/srl.rs
glamberson: Also same as above... Fixed. Test now asserts exact value equality.

CBenoit (approved):
LGTM

PR #1208, opened 2026-04-08
refactor(server): replace wildcard re-exports with explicit item list

glamberson:
Replace all `pub use module::*` in ironrdp-server with explicit item lists, addressing the existing `FIXME: too much wildcards here` comment.

This makes the public API surface auditable and prevents accidental exposure of internal types when new items are added to modules.

No functional change. All previously exported items remain exported at the same paths.


CBenoit (approved):
Thank you!

PR #1214, opened 2026-04-10
feat(connector): add `allow_encryption_level_none` config option

maurerdietmar:
Add a new `Config` flag that permits the connector to accept `ENCRYPTION_LEVEL_NONE` / `ENCRYPTION_METHOD_NONE` from the server. This does not imply full Standard RDP Security (RC4) support; it only allows the degenerate case where the server selects `PROTOCOL_RDP` with no encryption at all. When enabled, the connector skips the TLS/CredSSP upgrade and proceeds directly to basic settings exchange.

This is intended for setups where the transport is already secured by other means (e.g., a TLS WebSocket proxy or SSH tunnel).

### DIFF
--- a/crates/ironrdp-client/src/config.rs
+++ b/crates/ironrdp-client/src/config.rs
@@ -477,6 +477,7 @@ impl Config {
             domain: args.domain,
             enable_tls: !args.no_tls,
             enable_credssp: !args.no_credssp,
+            allow_encryption_level_none: false,
             keyboard_type: KeyboardType::parse(args.keyboard_type),
             keyboard_subtype: args.keyboard_subtype,
             keyboard_layout: 0, // the server SHOULD use the default active input locale identifier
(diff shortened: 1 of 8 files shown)


CBenoit (approved):
This is looking good to me! Thank you!

CBenoit:
There is a compilation error at crates/ironrdp-testsuite-core/tests/session/autodetect.rs:23:5, you need to add the new field to the struct initialization

maurerdietmar:
just fixed the compilation error.

CBenoit:
The public Config field and changed security-negotiation behavior warrant breaking-change and high-risk labels; coverage is missing for the newly reachable standard-RDP no-encryption path, including absent Server Security Data, skipped Security Exchange, and unencrypted Client Info.

glamberson:
@maurerdietmar Following up here. CBenoit's April approval still stands on the design, but his August 3rd comment raised a real gap that hasn't been addressed yet: the newly-reachable no-encryption path needs test coverage for Server Security Data handling, the skipped Security Exchange, and unencrypted Client Info. Given this touches connection security negotiation, that coverage matters.

The branch is also well behind master at this point and will need a rebase. If you can add that coverage and get it rebased, I'll advocate for getting this merged.

Issue #535, opened 2024-08-20
keyboard left shift and direction keys don't work


zombiehoffa:
I have a keychron k10 pro keyboard connected via usb and the left shift and arrow keys don't work. I've tried the other ibm keyboard layouts via passing the cli option but it doens't appear to do anything. Any help would be much appreciated. I"m running latest gity code.

CBenoit:
Hello!

Maybe our input event handling logic is lacking. Since you are building from master, I think you could easily provide us with more troubleshooting information.

At this place:
https://github.com/Devolutions/IronRDP/blob/af7deae70ab21435e9f0a864210b17c63efdd9e3/crates/ironrdp-client/src/app.rs#L159-L172

Can you modify that in such way?

```rust
                if let Some(scancode) = event.physical_key.to_scancode() {
                    let scancode = ironrdp::input::Scancode::from_u16(u16::try_from(scancode).unwrap());

                    let operation = match event.state {
                        event::ElementState::Pressed => ironrdp::input::Operation::KeyPressed(scancode),
                        event::ElementState::Released => ironrdp::input::Operation::KeyReleased(scancode),
                    };

                    let input_events = self.input_database.apply(std::iter::once(operation));

                    send_fast_path_events(&self.input_event_sender, input_events);
                } else {
                    warn!(?event, "Unhandled keyboard event");
                }
```

And then, let’s see what kind of warnings we get.

manio:
@CBenoit ^^^

CBenoit:
I think you could add logs in the keyboard-related branches: https://github.com/Devolutions/IronRDP/blob/53d814e3aec3398a2a16931fb788ae971607d912/crates/ironrdp-client/src/app.rs#L168-L222
And see what is received on IronRDP end

I’m not sure how we can configure `winit`, but there is likely something to do, yes. First, we are not running on the latest version, updating could be a good step as well.

manio:
Still trying to figure this out...
I updated winit to latest one (https://github.com/Devolutions/IronRDP/pull/1225) but this doesn't help.
I also did this:
```diff
diff --git a/crates/ironrdp-client/src/app.rs b/crates/ironrdp-client/src/app.rs
index bad9597b..3d39f1e9 100644
--- a/crates/ironrdp-client/src/app.rs
+++ b/crates/ironrdp-client/src/app.rs
@@ -283,6 +283,8 @@ impl ApplicationHandler for App {
[...]

PR #1225, opened 2026-04-21
[WiP] bump winit crate

manio:
Hi,
I’m creating this PR mainly so that if anyone picks this up, we don’t end up duplicating work. I’ve started migrating winit to the latest beta version. There are quite a lot of changes and still some more work to be done (some parts are commented out/disabled), but I’m sharing the code in its current state anyway.
The winit beta will probably become stable at some point, and at least this way the groundwork is already there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -488,20 +488,11 @@ dependencies = [
 
 [[package]]
 name = "block2"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


CBenoit:
Thank you!!
I'm curious to know: is it helping with the issue you encountered the other day?

manio:
If you are asking about keyboard problem, no this update doesn't help, but I found the root cause (here: https://github.com/Devolutions/IronRDP/issues/535#issuecomment-4343286783)

PR #1230, opened 2026-04-26
feat(rdpdr): add Web RDPDR virtual printer support

matt020191:
## Summary

Adds RDPDR virtual printer support for web sessions. The web client can now announce an IronRDP printer, receive server-side print jobs over RDPDR, and return completed PostScript jobs to the browser callback for PDF conversion / user printing.

This uses `MS Publisher Imagesetter` as the default PostScript driver, matching the FreeRDP approach, while also exposing a configurable printer driver name for deployments that need to override it.

## Notable changes

- Advertise RDPSND alongside RDPDR when printer redirection is enabled.
- Announce printer devices after `RDPDR_USER_LOGGEDON_PDU`, which fixes unreliable printer creation on Windows Server guests.
- Align printer announce metadata with the FreeRDP behavior: default/network printer flags and empty PnP name.
- Add web printer configuration validation and driver override support.
- Avoid acknowledging writes for unknown printer handles.
- Add print job size protection for the web backend.
- Update the native RDPDR backend trait implementation.
- Move/add RDPDR printer coverage in the testsuite so the wire-layout tests actually run.
- Fix the portable fast-path test shim issue separately from the printer implementation.

## Validation

Tested successfully against Windows 11 and Windows Server 2012 R2 / 2016 / 2022 / 2025 VMs. Previously unreliable Server VMs now consistently create the redirected printer and complete print jobs.

Also ran the relevant Rust checks/tests for the touched crates, including RDPDR tests, web WASM check, formatting, clippy, and lockfile validation.


crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -515,6 +537,32 @@ pub const GENERAL_CAPABILITY_VERSION_02: u32 = 0x0000_0002;
...
+/// [MS-RDPEPC 2.2.2.3] RDPDR_PRINTER_ANNOUNCE flag: the server should expect XPS output.
+pub const RDPDR_PRINTER_ANNOUNCE_FLAG_XPSFORMAT: u32 = 0x0000_0010;
+
+/// Default server-side printer driver announced for PostScript virtual printers.
+///
+/// `MS Publisher Imagesetter` is the PostScript driver used by Guacamole-like
+/// RDP printer redirection flows. If a target host does not have this driver
+/// installed, callers can use the explicit-driver helpers to advertise a
+/// different server-side driver.
+pub const DEFAULT_PRINTER_DRIVER_NAME: &str = "MS Publisher Imagesetter";
```
CBenoit, line 594: I think a more correct attribution would be to say that it’s FreeRDP’s default:

https://github.com/FreeRDP/FreeRDP/blob/1a1c3aba8c39679e5f600b04bc471596cd1aea0b/channels/printer/client/cups/printer_cups.c#L326-L331

Additional things worth noting:

1. FreeRDP itself acknowledges the default is fragile on newer macOS, and it overrides to `"Microsoft Print to PDF"` on macOS 14 Sonoma+, because `MS Publisher Imagesetter` is no longer available there.
2. [The Windows backend (`printer_win.c`)](https://github.com/FreeRDP/FreeRDP/blob/master/channels/printer/client/win/printer_win.c) doesn't have a hardcoded default, it reads the actual driver name from the Windows spooler API, which is always correct on Windows clients.

I see we have an escape hatch in the form of `with_printer_driver`, and that is good. Maybe the next thing we could do here is to provide the right tools to be closer to FreeRDP’s logic.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -945,6 +1015,138 @@ impl DeviceAnnounceHeader {
         }
     }
 
+    /// Construct a printer announce with sensible defaults — see
+    /// [`Devices::add_printer`] for the policy. Callers with custom
+    /// driver requirements should use
+    /// [`Self::new_printer_with_driver`].
+    pub fn new_printer(device_id: u32, print_name: String) -> Self {
+        // `MS Publisher Imagesetter` is the PostScript driver used by
+        // Guacamole-like RDP printer redirection flows, so the announce
```
CBenoit: Is "Guacamole-like RDP printer redirection flow" something specific to Guacamole, or is it the flow defined by the RDP specification and/or something from FreeRDP?

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -1575,6 +1777,45 @@ impl From<ServerDriveLockControlRequest> for ServerDriveIoRequest {
...
+    /// Server pushed print-job bytes into the opened handle; answer with a
+    /// [`DeviceWriteResponse`] echoing the request length.
+    Write(DeviceWriteRequest),
+    /// Server finalized the print job; answer with a [`DeviceCloseResponse`]
+    /// and finalize whatever document buffer the backend accumulated.
+    Close(DeviceCloseRequest),
+    /// Any other major function. Backends should complete the IRP with
+    /// `STATUS_NOT_IMPLEMENTED`; printers don't legitimately see Read,
+    /// DirectoryControl, QueryInformation, etc.
+    Unsupported(DeviceIoRequest),
```
CBenoit: I’m not very convinced by the design decision here. It sounds a bit counter-intuitive to me that we need to pass an "Unsupported" variant to the backend just for it to return "STATUS_NOT_IMPLEMENTED". The Rdpdr SVC processor already has all the information needed to respond itself without the backend being involved. Then, the public enum becomes just Create, Write and Close.

crates/ironrdp-rdpdr/src/lib.rs
```diff
@@ -122,15 +152,54 @@ impl Rdpdr {
...
+        if device_list.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        Self::announce_devices(device_list)
+    }
+
+    fn is_pre_logon_device(device: &DeviceAnnounceHeader) -> bool {
+        matches!(device.device_type(), DeviceType::Smartcard)
+    }
```
CBenoit, line 205: It seems we are missing a RDPDR_VERSION_MINOR_RDP51 guard for pre-logon device annoucement.

Here is the logic in FreeRDP:

```c
 /**
  * 1. versionMinor 0x0005 doesn't send PAKID_CORE_USER_LOGGEDON
  *    so all devices should be sent regardless of user_loggedon
  * 2. smartcard devices should be always sent
  * 3. other devices are sent only after user_loggedon
  */
 if ((rdpdr->clientVersionMinor == RDPDR_VERSION_MINOR_RDP51) ||   // 0x0005
     (device->type == RDPDR_DTYP_SMARTCARD) || arg->userLoggedOn)
```

Since this is only useful for old servers (Windows XP / Server 2023), I’m okay with not supporting that, but let’s add a comment to explain why we diverge from FreeRDP. A good reason is that we may not support such servers at all just because they only support old, broken TLS cipher suites that are not even supported by rustls, and it’s not worth the effort.

crates/ironrdp-rdpdr/src/backend/mod.rs
```diff
@@ -6,12 +6,27 @@ use ironrdp_core::AsAny;
...
     fn handle_server_device_announce_response(&mut self, pdu: ServerDeviceAnnounceResponse) -> PduResult<()>;
     fn handle_scard_call(&mut self, req: DeviceControlRequest<ScardIoCtlCode>, call: ScardCall) -> PduResult<()>;
     fn handle_drive_io_request(&mut self, req: ServerDriveIoRequest) -> PduResult<Vec<SvcMessage>>;
+    /// Handle a server-initiated IRP addressed to a printer device.
+    ///
+    /// `req` carries the fully-decoded printer IRP. Printers only see
+    /// [`PrinterIoRequest::Create`] / [`PrinterIoRequest::Write`] /
+    /// [`PrinterIoRequest::Close`] on the happy path; any other major
+    /// function arrives as [`PrinterIoRequest::Unsupported`] and
+    /// should be NAK'd with `STATUS_NOT_IMPLEMENTED`.
```
CBenoit: I would recommend we use STATUS_NOT_SUPPORTED instead, mirroring FreeRDP. STATUS_NOT_SUPPORTED semantically means "the device does not support this operation" whereas STATUS_NOT_IMPLEMENTED is "the OS primitive is not implemented" (mostly used for unsupported device operations).

CBenoit (commented):
Hello,

Thank you for the PR!

The "Typos" job is failing. I suggest you add the `./crates/ironrdp-rdpdr/src/pdu/efs.rs` file to the exception list in the typos configuration file.

Generally speaking, I think referring to Guacamole is not the correct attribution, as Guacamole is essentially wrapping the FreeRDP library under the hood. Unless you are referring to something that is explicitly outside of FreeRDP (in the calling code), I would rather have FreeRDP mentioned as the reference.

I see we don’t handle IRP_MJ_DEVICE_CONTROL. FreeRDP returns STATUS_SUCCESS with OutputBufferLength = 0 for device-control IRPs on the printer device. I think it’s easy enough to update the behavior on our side to match FreeRDP.

crates/ironrdp-web/src/printer.rs
```diff
@@ -0,0 +1,292 @@
...
+impl WasmPrinterMessageProxy {
+    pub(crate) fn new(tx: mpsc::UnboundedSender<RdpInputEvent>) -> Self {
+        Self { tx }
+    }
+
+    fn send(&self, message: PrinterBackendMessage) {
+        if self.tx.unbounded_send(RdpInputEvent::Printer(message)).is_err() {
+            error!("Failed to queue printer backend message, event loop receiver is closed");
+        }
+    }
```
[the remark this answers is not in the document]
CBenoit: This is a very valid point

crates/ironrdp-web/src/printer.rs
```diff
@@ -0,0 +1,292 @@
...
+                document_bytes,
+            } => {
+                // Hand the bytes to JS as a Uint8Array. js_sys::Uint8Array::from
+                // copies the slice into the WASM→JS boundary; the backend's
+                // buffer is already consumed at this point.
+                let array = js_sys::Uint8Array::from(document_bytes.as_slice());
+                let this = JsValue::NULL;
+                if let Err(err) = self.callbacks.on_job_complete.call1(&this, &array) {
+                    error!(?err, "on_job_complete JS callback threw");
+                }
```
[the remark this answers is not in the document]
CBenoit: Streaming or zero-copy handoff would be ideal, is it achievable?

CBenoit (commented):
Also, two extra refactor points I noticed while reviewing this:

1. handle_server_capability ignores server's announced capabilities: `handle_server_capability` sends back all of self.capabilities regardless of what the server advertised in its CoreCapability PDU. FreeRDP gates each capability in its response on `rdpdr->capabilities[CAP_xxx_TYPE]` (set when parsing the server PDU). This pre-exists this PR (same issue for drive/smartcard), but the printer capability is newly added here.

2. `NoopRdpdrBackend::handle_printer_io_request` silently drops IRPs. I think it’s low risk in practice because printer device type is only registered when a non-noop printer backend is configured, but a defensive STATUS_NOT_IMPLEMENTED response would be safer.

crates/ironrdp-web/src/session.rs
```diff
@@ -351,6 +419,30 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
...
+            Some(callback) => {
+                let (backend, printer) = wasm_printer_pair(input_events_tx.clone(), callback);
+                (Some(backend), Some(printer))
+            }
+            None => (None, None),
+        };
+
+        // Device id 1 is reserved for the drive device; default to 2 here to
+        // avoid a collision if drive redirection is ever enabled alongside.
+        let printer_device_id = printer_device_id.unwrap_or(2);
```
CBenoit: That’s only true if the drive redirection is enabled alongside the printer. If it’s not, device ID 1 is free. It should say something like
```suggestion
        // Default to 2 to avoid a potential collision if drive redirection is enabled in the same session.
        let printer_device_id = printer_device_id.unwrap_or(2);
```

crates/ironrdp-rdpdr-native/src/nix/backend.rs
```diff
@@ -69,6 +69,22 @@ impl RdpdrBackend for NixRdpdrBackend {
...
+            PrinterIoRequest::Unsupported(req) => req,
+        };
+
+        warn!(?device_io_request, "Printer IRP reached drive-only backend");
+        Ok(vec![SvcMessage::from(RdpdrPdu::DeviceCloseResponse(
+            DeviceCloseResponse {
+                device_io_response: DeviceIoResponse::new(device_io_request, NtStatus::NOT_IMPLEMENTED),
+            },
+        ))])
+    }
```
CBenoit: That’s pure boilerplate that could be covered by a default implementation (both for nix and noop).

crates/ironrdp-rdpdr/src/pdu/mod.rs
```diff
@@ -37,6 +36,15 @@ pub enum RdpdrPdu {
...
     EmptyResponse,
+    /// A PDU whose `PacketId` decodes successfully but for which we have no
+    /// first-class handler. The body bytes are left unread in the source
+    /// cursor; the processor is expected to log-and-drop.
+    ///
+    /// Used for the printer-cache family (`PAKID_PRN_CACHE_DATA`,
+    /// `PAKID_PRN_USING_XPS`) that Server 2025's spooler sends after a
+    /// printer announce. Returning a decode error here would tear down the
+    /// SVC session, so we tolerate the PDU instead.
+    Unhandled(PacketId),
```
CBenoit: `RdpdrPdu::Unhandled` is a decode-only variant living inside an enum that also serves as the outbound PDU type. Its `Encode` impl silently succeeds with an empty body, which is a footgun, if it were ever accidentally queued as a response, a malformed PDU would be sent with no error on our side.

I think a cleaner pattern is to handle the "unrecognised packet ID" case in `SvcProcessor::process` before a `RdpdrPdu` is constructed: decode the `SharedHeader` first, check the `PacketId`, log-and-return early if unknown. That way every variant in `RdpdrPdu` is encode-safe, `Unhandled` disappears, and the drop policy lives in exactly one place.

crates/ironrdp-web/src/session.rs
```diff
@@ -260,6 +279,45 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
             |locks_expired_callback: JsValue| {
                 self.0.borrow_mut().locks_expired_callback = locks_expired_callback.dyn_into::<js_sys::Function>().ok();
             };
+            // Printer callbacks - protocol-specific, routed through extension()
+            // rather than dedicated trait methods to keep iron-remote-desktop protocol-agnostic.
```
CBenoit: This is probably LLM prompt leftover comments

crates/ironrdp-web/src/session.rs
```diff
@@ -260,6 +279,45 @@ impl iron_remote_desktop::SessionBuilder for SessionBuilder {
...
+            };
+            |printer_driver_name: String| {
+                let mut inner = self.0.borrow_mut();
+                inner.printer_requested |= !printer_driver_name.is_empty();
+                inner.printer_driver_name = if printer_driver_name.is_empty() {
+                    None
+                } else {
+                    Some(printer_driver_name)
+                };
+            };
```
CBenoit, line 315: printer_requested is set to true by three independent extension setters: printer_name, printer_device_id, and printer_driver_name, in addition to print_job_complete_callback. This means setting e.g. printer_name = "My Printer" without providing a callback will pass `printer_requested = true` into the validation check and then immediately fail with "printer redirection requires a valid print_job_complete_callback".

My understanding is that we need to set all of that at once for it to work. Maybe you meant to use an object so you can pass all the required arguments at once?

https://github.com/matt020191/IronRDP/blob/4a7c36da52e0d89f1ab399aa302361e49dbc8dfd/crates/ironrdp-web/src/session.rs#L1147-L1171

That way, `printer_requested` is probably not required anymore. You can store all the parameters into a `Option<PrinterCtx>` such that None means current `printer_requested = false`.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -945,6 +1015,138 @@ impl DeviceAnnounceHeader {
...
+        // [2.2.2.3]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpepc/2057a02f-57d5-47db-9a32-e337ac3f50e9
+
+        fn utf16le_with_nul(s: &str) -> Vec<u8> {
+            let mut out = Vec::with_capacity((s.len() + 1) * 2 /* 2 bytes per UTF-16 unit */);
+            for unit in s.encode_utf16() {
+                out.extend_from_slice(&unit.to_le_bytes());
+            }
+            out.extend_from_slice(&[0, 0] /* UTF-16 NUL terminator */);
+            out
+        }
```
CBenoit: Per the IronRDP style guide, nested helper functions should be placed at the end of their enclosing function, not the beginning. `utf16le_with_nul` is currently defined at the top of `new_printer_with_driver`, before any of its call sites. Please move it to the bottom of the function body.

crates/ironrdp-rdpdr/src/pdu/efs.rs
```diff
@@ -945,6 +1015,138 @@ impl DeviceAnnounceHeader {
...
+                + pnp_name_bytes.len()
+                + driver_name_bytes.len()
+                + print_name_bytes.len(),
+        );
+        device_data.extend_from_slice(&flags.to_le_bytes());
+        device_data.extend_from_slice(&code_page.to_le_bytes());
+        device_data.extend_from_slice(
+            &u32::try_from(pnp_name_bytes.len())
+                .expect("PnPName length fits in u32")
+                .to_le_bytes(),
```
CBenoit: `pnp_name_bytes` is unconditionally `Vec::new()` and thus its length is always 0.

CBenoit (commented):
A question I got while reviewing: did you consider hot-adding printers? I guess it’s not something very useful in the web?

crates/ironrdp-web/src/printer.rs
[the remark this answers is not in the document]
matt020191: That was a very good point, had missed that. Fixed in the most recent commit, print jobs remain capped, and completed queued jobs now have a bounded byte budget so browser/WASM memory cannot grow unbounded while jobs wait to be delivered.

crates/ironrdp-web/src/printer.rs
```diff
@@ -0,0 +1,292 @@
...
+    fn handle_drive_io_request(&mut self, _req: ServerDriveIoRequest) -> PduResult<Vec<SvcMessage>> {
+        warn!("Drive IRP reached printer-only backend; ignoring");
+        Ok(Vec::new())
+    }
+
+    fn handle_printer_io_request(&mut self, req: PrinterIoRequest) -> PduResult<Vec<SvcMessage>> {
+        match req {
+            PrinterIoRequest::Create(create) => {
+                let file_id = self.allocate_file_id();
+                self.open_files.insert(file_id, Vec::new());
```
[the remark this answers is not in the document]
matt020191: I added focused unit tests for the web printer backend covering the full Create -> Write -> Close delivery path, unknown file_id write rejection, oversized write rejection with partial-job cleanup, and completed-job queue budget enforcement. The tests assert the expected IRP completion status/byte counts and verify delivered job contents.

crates/ironrdp-rdpdr/src/pdu/efs.rs
matt020191: Here I was referring to when I rearranged the printer announce to match FreeRDP's implementation to fix the flaky Windows Server issue - you are right I should have attributed FreeRDP instead of Guacamole, I think I fixed all the references in the latest commit (PR description included)

crates/ironrdp-rdpdr/src/pdu/efs.rs
matt020191: You are right, I removed the Unsupported printer request variant and kept the public printer request enum limited to the backend-relevant operations: Create, Write, and Close. Unsupported printer IRPs are now completed directly by the RDPDR SVC processor with an unsuccessful status, so backends no longer need to round-trip an operation they can’t meaningfully handle. I also added testsuite coverage for this path.

crates/ironrdp-rdpdr/src/lib.rs
matt020191: Great catch, I had missed that. Windows XP / Server 2000 is indeed really old, but its a quick addition to support it so I went ahead and matched the FreeRDP behavior instead of just documenting a divergence. The client now treats RDPDR_VERSION_MINOR_RDP51 / 0x0005 as a special case and announces all devices immediately from PAKID_CORE_CLIENTID_CONFIRM, since those older servers do not send PAKID_CORE_USER_LOGGEDON. I also added testsuite coverage for it.

crates/ironrdp-rdpdr/src/backend/mod.rs
matt020191: Agreed, I switched the unsupported printer IRP completion path to use STATUS_NOT_SUPPORTED to match FreeRDP’s behavior and the proper meaning. The default backend fallback also returns STATUS_NOT_SUPPORTED, and I added testsuite coverage for it.

crates/ironrdp-web/src/session.rs
matt020191: Good point, implemented

crates/ironrdp-rdpdr-native/src/nix/backend.rs
matt020191: I moved this into a default RdpdrBackend::handle_printer_io_request implementation that returns STATUS_NOT_SUPPORTED, so both the nix backend and noop backend can use the shared fallback instead of carrying identical boilerplate.

crates/ironrdp-rdpdr/src/pdu/mod.rs
matt020191: Agreed. I removed RdpdrPdu::Unhandled entirely so every RdpdrPdu variant is encode-safe. SvcProcessor::process now decodes the SharedHeader first and handles the known printer-cache PDUs with a log-and-return path before constructing a typed RdpdrPdu; unsupported/unknown packet IDs now fail decode instead of creating a response-capable enum variant with an empty encode body.

crates/ironrdp-web/src/session.rs
matt020191: Yes it was, deleted those

crates/ironrdp-web/src/session.rs
matt020191: Good catch. I removed the separate printer_requested flag so optional metadata setters such as printerName, printerDeviceId, and printerDriverName no longer independently activate printer redirection. The activation point is now the printJobCompleteCallback, because without that callback there is nowhere safe to deliver completed jobs. The behavior now matches the intended model: optional printer settings are applied only if printer redirection is actually enabled by the callback.

crates/ironrdp-rdpdr/src/pdu/efs.rs
matt020191: You are right, I removed the unnecessary pnp_name_bytes = Vec::new() and now write an explicit pnp_name_len: u32 = 0, since the PostScript printer path intentionally leaves PnPName empty.

crates/ironrdp-web/src/printer.rs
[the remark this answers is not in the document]
matt020191: Changed this to a single streaming path. The web backend no longer buffers completed print jobs in Rust or queues a full-job payload; it emits per-job create/data/complete/abort events and only tracks byte counts. This is not true zero-copy across the WASM boundary because each chunk still becomes a JS Uint8Array, but the copy is now bounded to each write chunk rather than the full document.

crates/ironrdp-rdpdr/src/pdu/efs.rs
matt020191: You are right, FreeRDP is definitely the correct attribution, I changed it in all references. I completely missed the macOS 14+ change to Microsoft Print to PDF, I've implemented that to keep consistent. 

FreeRDP’s Windows behavior makes sense for native clients because printer_win.c can query the local Windows spooler and use the actual PRINTER_INFO_2::pDriverName. The web backend can’t do that as it runs in browser/WASM and has no access to WinSpool or installed local printer drivers, and this printer is synthetic rather than a redirected local Windows printer. I think the right split is to keep the web path explicit/heuristic with printerDriverName(...). We could do a future native Windows ironrdp-rdpdr-native backend to follow FreeRDP and read the real driver name from the spooler. 

You asked in another comment if I considered hot-adding printers and I did before going to the web-only path because of the above as well, browser/WASM cannot query or monitor the local Windows spooler. This could also be included in a future ironrdp-rdpdr-native backend.

crates/ironrdp-rdpdr/src/pdu/efs.rs
matt020191: Missed that, sorry :) Got it moved to the proper location

CBenoit:
I asked one last final review to Copilot to verify the details in your follow up commits, but otherwise it looks good to me :)

crates/ironrdp-rdpdr/src/pdu/mod.rs
```diff
@@ -120,15 +116,22 @@ impl Decode<'_> for RdpdrPdu {
             )),
             PacketId::CoreDeviceIoRequest => Ok(RdpdrPdu::DeviceIoRequest(DeviceIoRequest::decode(src)?)),
             PacketId::CoreUserLoggedon => Ok(RdpdrPdu::UserLoggedon),
-            _ => Err(unsupported_value_err!(
-                "RdpdrPdu",
+            _ => Err(invalid_field_err!(
+                "RdpdrPdu::decode_body",
                 "PacketId",
-                header.packet_id.to_string()
+                "unexpected packet id"
```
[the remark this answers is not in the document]
matt020191: The suggestion as written wouldn't have compiled, went ahead and patched with another commit. The low confidence suppressed comment was right, fixed that in another commit as well.

matt020191:
> I asked one last final review to Copilot to verify the details in your follow up commits, but otherwise it looks good to me :)

Copilot flagged two issues, went ahead and patched in another commit. I want to mention that I did re-test after all of the changes yesterday in my implementation (after a patch for the streaming change) and it tested ok as before. Thanks so much for your very through review. Rust is a new language for me (I primarily use .NET/C#) so this one was a learning experience.

CBenoit (approved):
Thank you! I think we're good to merge now =)

CBenoit:
Oops, it appears we have some conflicts that need to be fixed before we can actually merge

matt020191:
> Oops, it appears we have some conflicts that need to be fixed before we can actually merge

@CBenoit Missed that too! Just pushed a commit resolving the conflict. Should be good to go now

Issue #1232, opened 2026-04-27
Surface the actual Share Control PDU type when 'unexpected Share Control Pdu' fires


GlassOnTin:
## Problem

When ironrdp receives a Share Control PDU during Capabilities Exchange that isn't `ServerDemandActive`, the user-facing error is:

```
RDP connect finalize failed: Error { context: \"unexpected Share Control Pdu (expected ServerDemandActive)\", kind: General, source: None }
```

The actual PDU type that *was* received isn't included in the error string. From a downstream client's perspective (Haven, in this case) this means we can't distinguish between, e.g.:

- Server sent a `SetErrorInfo` PDU rejecting the session for a specific reason (the most informative case — that PDU has a numeric error code we'd want to surface)
- Server sent `SaveSessionInfo` or `Server Auto-Reconnect Status` instead of demand-active
- Server sent some other implementation-quirk PDU

## Reported via

[GlassHaven/Haven#109](https://github.com/GlassHaven/Haven/issues/109#issuecomment-4328969739) — agevlakh's GNOME Remote Desktop on Fedora 44 hits this after CredSSP succeeds. We were able to characterise everything *up to* this point (TLS ✓, NLA ✓ now after a separate sspi-rs interop session) but the actual server response that ironrdp rejects is opaque.

## Asks

1. Include the received PDU's discriminant (and, ideally for `SetErrorInfo`, the error code value) in the `context` string when raising this error. Even just `unexpected Share Control Pdu (got SetErrorInfo, expected ServerDemandActive)` would let us trace it back to MS-RDPBCGR §2.2.5.1 and explain the failure to users.
2. Optionally a structured variant of `ConnectorErrorKind` for "server rejected at capabilities phase" so downstream code can map to a friendlier message instead of substring-sniffing the `Reason` string.

Happy to PR if you'd like — would be a small change in `crates/ironrdp-connector/src/connection_activation.rs` (or wherever the demand-active wait is checked).

CBenoit:
I acknowledge the pain point, although I didn’t reread carefully the current code to verify what would be the best approach. At this point feel free to open a PR so we can see how it turns there.

GlassOnTin:
Following up on this — the Haven user with the GRD interop bug ([Haven#109](https://github.com/GlassHaven/Haven/issues/109)) ran journalctl on the GRD side. It's exactly the EGFX requirement:

```
gnome-remote-desktop-daemon[792]: [RDP] Client did not advertise support 
[...]

PR #1236, opened 2026-04-28
fix(connector): surface actual PDU type when an unexpected Share Control PDU arrives

GlassOnTin:
Closes part of #1232.

## Why

Three error sites in the connector previously printed only the *expected* PDU type, dropping the actual type the server sent on the floor. That made interop bugs harder to diagnose than necessary.

Concrete case from #1232 (originally surfaced via [Haven#109](https://github.com/GlassHaven/Haven/issues/109) on a GNOME Remote Desktop server): when a client connects to GRD without advertising EGFX, GRD's \`freerdp_peer::Capabilities()\` callback fails — confirmed via journalctl on the server:

\`\`\`
gnome-remote-desktop-daemon: [RDP] Client did not advertise support for the Graphics Pipeline, closing connection
[ERROR] [rdp_peer_handle_state_demand_active]: 
  [CONNECTION_STATE_CAPABILITIES_EXCHANGE_DEMAND_ACTIVE] freerdp_peer::Capabilities() callback failed
\`\`\`

GRD then sends a Server Deactivate All PDU instead of Server Demand Active. The connector saw \`ServerDeactivateAll\`, didn't match \`ServerDemandActive\`, and reported \"unexpected Share Control Pdu (expected ServerDemandActive)\" — with no hint that the server had sent Deactivate All (which itself is a strong signal the server gave up during caps validation, e.g. because of a missing capability).

## What

Replaces three \`general_err!\` sites in the connector with \`reason_err!\` calls that include \`pdu.as_short_name()\` (the helper already exists on \`ShareControlPdu\`):

| File | Function | Diagnosis context |
|---|---|---|
| \`connection_activation.rs:125\` | \`ConnectionActivation::CapabilitiesExchange\` | server's reply to \`ConfirmActive\` |
| \`legacy.rs:152\` | \`decode_share_data\` | wraps \`decode_share_control\` for callers expecting Data |
| \`legacy.rs:213\` | \`decode_io_channel\` | I/O channel dispatcher for the session loop |

Now diagnoses look like:

> unexpected Share Control PDU during capabilities exchange: got Server Deactivate All PDU (expected Server Demand Active PDU)

vs. the previous:

> unexpected Share Control Pdu (expected ServerDemandActive)

which immediately points at GRD-style \"I rejected your caps and deactivated the share\" rather than leaving callers to guess from network traces.

## Scope

- No behavior change beyond the error string.
- Existing matches on \`ServerDemandActive\` / \`Data\` / \`ServerDeactivateAll\` continue to dispatch the same way.
- The wider EGFX-as-client work (so IronRDP-based clients can actually interop with GRD instead of just diagnosing the failure better) is a much larger separate effort — flagged in the issue thread.

## Verification

- \`cargo build -p ironrdp-connector\` clean.
- \`cargo clippy -p ironrdp-connector\` clean (no new warnings; pre-existing duplicates persist).
- \`cargo test -p ironrdp-connector --lib\` — 0 tests in the lib, no regressions.
- No tests grep for the old error strings.

### DIFF
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -6,6 +6,7 @@ use tracing::{debug, warn};
 
 use crate::{
     Config, ConnectionFinalizationSequence, ConnectorResult, DesktopSize, Sequence, State, Written, general_err, legacy,
+    reason_err,
 };
 
 /// Represents the Capability Exchange and Connection Finalization phases
@@ -122,8 +123,10 @@ impl Sequence for ConnectionActivationSequence {
                 {
                     server_demand_active.pdu.capability_sets
                 } else {
-                    return Err(general_err!(
-                        "unexpected Share Control Pdu (expected ServerDemandActive)",
+                    return Err(reason_err!(
+                        "ConnectionActivation::CapabilitiesExchange",
+                        "unexpected Share Control PDU during capabilities exchange: got {} (expected Server Demand Active PDU)",
+                        share_control_ctx.pdu.as_short_name(),
                     ));
                 };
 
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
LGTM

PR #1237, opened 2026-04-30
feat(connector)!: expose SUPPORT_DYN_VC_GFX_PROTOCOL early-cap flag for EGFX clients

GlassOnTin:
Currently, `early_capability_flags` in the GCC core data is built from a fixed set in `connection.rs`. Clients that want to use the Graphics Pipeline Extension (MS-RDPEGFX) — by attaching a `DvcClientProcessor` for `Microsoft::Windows::RDS::Graphics` — have no way to set `SUPPORT_DYN_VC_GFX_PROTOCOL` without forking the connector, and modern Windows servers won't open the EGFX channel unless the client advertises support.

This PR adds an opt-in `Config.support_dyn_vc_gfx_protocol: bool` (default `false`). When set, the flag is OR'd into `early_capability_flags` alongside the existing `WANT_32_BPP_SESSION` conditional. Existing consumers are unaffected; the doc comment includes a safety note that setting this without an EGFX implementation will cause Windows to stop sending legacy bitmap updates, leaving the desktop blank.

Used downstream by [Haven](https://github.com/GlassHaven/Haven) (an Android RDP/VNC client) which implements EGFX with ClearCodec + RemoteFxProgressive decoders; this lets us drop a vendored fork of `ironrdp-connector` we currently carry just for this one flag.

Default `false` to preserve current behaviour. `cargo check --workspace` clean — six other in-tree `Config { … }` builders updated with the default-false field.


glamberson:
Reading through, this is the smallest possible fix and the doc-comment captures the foot-gun precisely. Without wiring a DvcClientProcessor for the Graphics channel, modern Windows servers stop sending legacy bitmap updates entirely and the desktop goes blank, which is exactly the GRD case in #1232 and the failure mode every downstream EGFX client hits.

No conflict with #1098 (multitransport bootstrapping) on my read of the GCC block area, just two independent flags being OR'd in. Looks good from the server-side perspective too: any IronRDP server that already speaks EGFX (lamco-rdp-server, GRD) sees this flag and routes graphics through DVC as expected.

glamberson:
@GlassOnTin Hi, could you rebase this? I'm trying to get this prioritized, and this would be a first step on the client side work I'm looking t. Thanks!

GlassOnTin:
Done — rebased onto current `master` and force-pushed. It's a single commit again and GitHub now shows it as mergeable.

The only conflicts were in the test builders, where `master` had restructured things since I opened this: `session/autodetect.rs`'s local `test_config()` and the monolithic `testsuite-extra/tests/main.rs` were removed/split, so I dropped my now-obsolete edits to those and instead added `support_dyn_vc_gfx_protocol: false` to the `Config { … }` literals in their new homes — `connector/autodetect.rs`, `session/connection_activation.rs`, and `e2e.rs`. The core change (the `early_capability_flags` OR in `connection.rs` and the field on `Config`) merged cleanly.

`cargo check --workspace --all-targets` is clean. Happy to squash/adjust further if you'd like it shaped differently for the client-side work.

glamberson:
Thank you! @CBenoit can you expedite this please? Thank you. My client issue will be posted later today.

GlassOnTin:
For reviewer context: I've opened #1472 as a sibling to this — a small `ironrdp-pdu` fix to tolerate an empty Server Font Map body (another VirtualBox VRDP quirk). Together these two are the last things a downstream Android RDP client needs to drop a vendored `ironrdp-connector` fork it currently carries: this PR removes the EGFX early-capability workaround, #1472 removes the Font Map one. Both are opt-in / strictly-more-lenient and don't change existing behaviour. No rush — just flagging they land the same cleanup.

CBenoit:
Adding `support_dyn_vc_gfx_protocol` to the exhaustively constructible public `Config` struct is source-incompatible for downstream struct literals. The EGFX capability bit is correctly opt-in and preserves the false default; enabled-wire-bit and end-to-end negotiation coverage remain valuable.

GlassOnTin:
Thanks — both points taken. Rebased onto current `master` and addressed them.

**On the source incompatibility:** you're right, and I've stopped pretending otherwise — the PR is now marked breaking (`feat(connector)!:`). `Config` is not `#[non_exhaustive]`, so any added field breaks downstream struct literals, and this one is no exception.

For what it's worth, #1501 added `pub auto_reconnect_cookie: Option<ServerAutoReconnect>` to this same struct and shipped as `feat(pdu,session,connector,server)!:` — so if that's the accepted shape for this kind of change, this now matches it. If you'd rather absorb the break once and be done with it, making `Config` `#[non_exhaustive]` with a `Default` impl or a builder would mean this field and every future one costs nothing downstream; I'm happy to do that as a separate PR ahead of this one, but it's a bigger call than mine to make.

**On the coverage:** added both.

*Wire bit*, in `ironrdp-connector`'s `create_gcc_blocks` (inline `mod tests` at the file end, per #1527) — three tests:

- `dyn_vc_gfx_protocol_bit_is_absent_by_default` — the default leaves the bit clear. This is the one that matters most: a client that sets the flag without wiring an EGFX-capable DVC processor gets modern Windows servers to stop sending legacy bitmap updates entirely, and the desktop goes blank.
- `dyn_vc_gfx_protocol_bit_is_set_when_enabled` — opting in sets it.
- `dyn_vc_gfx_protocol_opt_in_changes_exactly_one_bit` — asserts the symmetric difference between the enabled and disabled flag sets is exactly `SUPPORT_DYN_VC_GFX_PROTOCOL`, so the opt-in can't quietly disturb anything else.

These are checked to actually fail: replacing the `if config.support_dyn_vc_gfx_protocol` gate with `if false` reddens the second and third and correctly leaves the first green.

*End to end*, in `testsuite-extra`: `test_client_server_advertising_dyn_vc_gfx_protocol` runs the full client↔server connect with the bit advertised.

I want to be straight about that last one's scope rather than oversell it: it asserts negotiation **completes** with the bit set, not that the server acted on it. `ironrdp-server` doesn't surface the client's early capability flags, and plumbing that through purely to observe it in a test would have been a larger change than the feature itself. The real assertion lives in the unit tests above. If you'd like the server side to expose received early caps — useful beyond this PR — say so and I'll open it separately.

`cargo fmt --check`, `cargo check --workspace --all-targets` and `cargo clippy -p ironrdp-connector --all-targets -- -D warnings` are clean.

GlassOnTin:
Rebased onto current master and back to a single commit (d9111257). While doing so I found that a push on Aug 5 had accidentally stacked five unrelated commits onto this branch (they belonged with #1541 and with session work that has since landed upstream), so anyone opening the PR since then saw a six-commit mixture instead of the one-line capability flag. Sorry about that; it is cleaned up now.

Checked locally on the rebased branch: the fmt gate passes via cargo xtask check fmt, cargo check --workspace --all-targets is clean, and the connector unit tests pass including the three tests covering the flag (absent by default, set on opt-in, changes exactly one bit).

@CBenoit the state since your Aug 3 review is unchanged in substance: the PR is marked breaking (feat(connector)!:) for the added Config field, matching the shape #1501 shipped with, and the default keeps the wire bit clear. Is there anything else you would like changed, or can this go into the review queue? @glamberson mentioned in July this was a first step for their client-side work, so there is still downstream interest beyond mine.

crates/ironrdp-connector/src/lib.rs
```diff
@@ -292,6 +292,34 @@ pub struct Config {
...
+    /// Windows servers to stop sending legacy bitmap updates and
+    /// route everything over EGFX — which the connector consumer
+    /// can't decode, leaving the desktop blank. To enable EGFX:
+    ///
+    /// 1. Set this field to `true`.
+    /// 2. Attach a [`DvcClientProcessor`] for
+    ///    `Microsoft::Windows::RDS::Graphics` via
+    ///    [`ClientConnector::with_static_channel`] +
+    ///    [`ironrdp_dvc::DrdynvcClient::with_dynamic_channel`].
+    /// 3. Decode `ironrdp_pdu::rdp::vc::dvc::gfx::ServerPdu` and
```
mamoreau-devolutions: Could we replace this recipe with the actual EGFX client API? `ironrdp_pdu::rdp::vc::dvc::gfx::ServerPdu` does not exist, and strict rustdoc fails on the links to `ClientEarlyCapabilityFlags` and `ironrdp_dvc::DrdynvcClient::with_dynamic_channel`. The supported processor is `ironrdp_egfx::client::GraphicsPipelineClient`. Also, `false` only leaves EGFX unadvertised; it does not imply slow-path-only output because the connector still advertises fast-path output and surface commands. Please use valid links and describe the supported registration path.

> [!NOTE]

crates/ironrdp-testsuite-extra/tests/e2e.rs
```diff
@@ -62,6 +62,27 @@ async fn test_client_server() {
...
 
+/// Advertising the Graphics Pipeline early-capability bit must not disturb the
+/// connection sequence itself.
+///
+/// Scope, stated plainly: this asserts negotiation still completes end to end
+/// with the bit set. It does *not* assert the server acted on it —
+/// `ironrdp-server` does not surface the client's early capability flags, and
+/// plumbing that through only to observe it here would be a larger change than
+/// the feature. The bit's presence and absence on the wire are asserted
+/// directly in `ironrdp-connector`'s `create_gcc_blocks` unit tests.
```
mamoreau-devolutions: These bit assertions are not part of the normal `cargo test --workspace --locked` CI run because `ironrdp-connector` sets `[lib] test = false`; only this e2e smoke test is discovered, and it does not observe the generated flag. Also, "on the wire" overstates the unit tests-they inspect `ClientGccBlocks` before encoding. Please move the bit assertions into an enabled testsuite target (or otherwise include the library tests in CI) and adjust this wording.

> [!NOTE]

mamoreau-devolutions (approved):
Approved. The capability flag wiring is protocol-conformant and preserves existing behavior by default. The inline notes are nonblocking follow-ups.

> [!NOTE]

crates/ironrdp-testsuite-extra/tests/e2e.rs
mamoreau-devolutions: Moved the flag regression into the enabled core testsuite. It now decodes the emitted Connect Initial PDU and verifies that enabling the option changes exactly the EGFX bit; the e2e wording now reflects its narrower scope.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/ironrdp-connector/src/lib.rs
mamoreau-devolutions: Updated the rustdoc to point to `GraphicsPipelineClient` and the valid DRDYNVC registration APIs, added the relevant specification links, and clarified that `false` only leaves EGFX unadvertised. Strict broken-link rustdoc now passes.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Issue #1209, opened 2026-04-08
ironrdp-server: replace anyhow::Result with typed error in public API


glamberson:
The `ironrdp-server` public API uses `anyhow::Result` for trait signatures and method return types:

- `RdpServerDisplayUpdates::next_update() -> Result<Option<DisplayUpdate>>`
- `RdpServerDisplay::updates() -> Result<Box<dyn RdpServerDisplayUpdates>>`
- `RdpServer::run() -> Result<()>`
- `RdpServer::run_connection() -> Result<()>`
- `ConnectionHandler::on_disconnected(error: Option<&anyhow::Error>)`
- `TlsIdentityCtx::init_from_paths() -> anyhow::Result<Self>`
- `EchoServerHandle::send_request() -> Result<()>`

The rest of the IronRDP library crates use typed errors via `ironrdp_error::Error<Kind>` (e.g. `SessionError`, `ConnectorError`). Using `anyhow` in the server crate means consumers can't pattern-match on error kinds, and every consumer takes an implicit dependency on the anyhow crate.

A `ServerError` / `ServerErrorKind` type following the same pattern as `ironrdp-session` and `ironrdp-connector` would bring the server crate in line with the rest of the library. Rough sketch of what that would look like:

```rust
pub type ServerError = ironrdp_error::Error<ServerErrorKind>;
pub type ServerResult<T> = Result<T, ServerError>;

#[non_exhaustive]
pub enum ServerErrorKind {
    Io,
    Pdu,
    Encode,
    Security,
    Channel,
    General,
    Custom,
}
```

Files that would need changes:
- New `error.rs` module with the type definitions and `ServerErrorExt` convenience trait
- `display.rs`: trait signatures (`next_update`, `updates`)
- `server.rs`: `run()`, `run_connection()`, `ConnectionHandler::on_disconnected`
- `echo.rs`: `send_request`
- `helper.rs`: `init_from_paths`, `make_acceptor`
- `builder.rs`: noop trait impls (`NoopDisplayUpdates`, `NoopDisplay`)
- `encoder/mod.rs`: internal anyhow usage could stay or migrate
- Example server (`examples/server.rs`): trait impl return types
- Doc examples in `display.rs`

The main complication is the consumer-implemented traits (`RdpServerDisplay`, `RdpServerDisplayUpdates`). Changing their return types is semver-breaking and requires all downstream implementors to update. The `Custom` variant with `with_source()` preserves the flexibility consumers currently get from anyhow.

[...]

PR #1239, opened 2026-04-30
feat(server): handle SuppressOutput, RefreshRectangle, FrameAcknowledge PDUs

glamberson:
## Summary

These three slow-path share-data PDUs are well-defined parts of MS-RDPBCGR and modern Windows clients send them routinely, but the server core has been logging them as `Unexpected share data pdu` and dropping them on the floor:

- **SuppressOutput** ([2.2.11.3]): client signals it does not want display updates while minimized or fully occluded.
- **RefreshRectangle** ([2.2.11.2]): client asks the server to retransmit specific regions, typically after unminimize.
- **FrameAcknowledge** (slow-path): legacy flow-control acknowledgment for non-EGFX update paths.

This PR adds three default-implementation methods to `ConnectionHandler` so server consumers can react to these signals (pause capture, mark regions dirty, throttle frame rate). Default impls ignore the PDU, preserving current behavior for consumers that have not opted in.

## Why a `ConnectionHandler` extension rather than internal handling

The right behavior for SuppressOutput is to stop encoding frames at the capture source, not the wire. RefreshRectangle similarly needs to invalidate at the capture layer. The server core does not own the capture pipeline; the consumer (via `RdpServerDisplay` and friends) does. Putting hooks on the existing lifecycle trait keeps the consumer in control while the server core stops dropping useful signals.

The slow-path FrameAcknowledge is independent of EGFX's own DVC-channel FrameAcknowledge handled by `ironrdp-egfx`. Documented in the doc-comment to avoid confusion.

## Adjacent work

@maurerdietmar has a `add-server-session-handler-trait` branch in his fork (not yet a PR) that proposes a similar `RdpServerSessionHandler` trait with `connect()`, `disconnect()`, and `suppress_output(allow: bool)` callbacks. The shapes are different (this PR extends the existing `ConnectionHandler` rather than introducing a new trait, and passes the `desktop_rect` through rather than collapsing to a bool) but the underlying need is the same. Happy to align on naming or structure if @maurerdietmar prefers his shape, this PR is the smallest change that closes the dropped-PDU gap.

## Coordinates with #1244

#1244 changes the signature of `ConnectionHandler::on_disconnected` (from `Option<&anyhow::Error>` to `Option<&ServerError>`) as part of the typed-error migration (#1242 / #1209). The three new methods this PR adds are independent of `on_disconnected` and unaffected by that signature change. Either PR can land first; whichever lands second is a trivial rebase to the new trait shape.

## Test plan

- [x] `cargo xtask check lints -v` clean (workspace, all-targets, with helper + __bench features)
- [ ] Manual smoke against mstsc (suppress output observed when minimizing the RDP window) is downstream test plan; consumer of `ConnectionHandler` decides what to do with the signal.

## Out of scope

Spec-compliant behavior for the three PDUs (actually pausing capture under SuppressOutput, marking regions dirty under RefreshRectangle, throttling under FrameAcknowledge) lives in the consumer. This PR is "stop dropping the signal," not "implement the behavior."

[2.2.11.3]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/47b1ace4-f2ba-4d11-93ed-c0e1c8e62221
[2.2.11.2]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/15679f83-0e1e-4a36-a4e6-3cb7b6e0a6ce

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -11,6 +11,7 @@ use ironrdp_cliprdr::backend::ClipboardMessage;
 use ironrdp_core::{decode, encode_vec, impl_as_any};
 use ironrdp_displaycontrol::pdu::DisplayControlMonitorLayout;
 use ironrdp_displaycontrol::server::{DisplayControlHandler, DisplayControlServer};
+use ironrdp_pdu::geometry::InclusiveRectangle;
 use ironrdp_pdu::input::InputEventPdu;
 use ironrdp_pdu::input::fast_path::{FastPathInput, FastPathInputEvent};
 use ironrdp_pdu::mcs::{SendDataIndication, SendDataRequest};
@@ -58,11 +59,12 @@ pub enum PostConnectionAction {
 /// Hooks for connection lifecycle events in [`RdpServer::run`].
 ///
 /// Implement this trait to add pre-accept filtering (rate limiting,
-/// IP allowlists) and post-disconnect logic (cleanup, session validity
-/// checks, metrics).
+/// IP allowlists), post-disconnect logic (cleanup, session validity
+/// checks, metrics), and per-connection responses to client share-data
+/// PDUs that the server core does not consume directly.
 ///
-/// All methods have default implementations that accept all connections
-/// and continue unconditionally.
+/// All methods have default implementations that accept all connections,
+/// continue unconditionally, and ignore client share-data signals.
 pub trait ConnectionHandler: Send {
     /// Called after `accept()` returns but before `run_connection()`.
     ///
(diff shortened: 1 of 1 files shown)


glamberson:
Rebased onto current master. Force-pushed \`77b4e612\` -> \`a4898b78\`.

20 days behind master since 2026-04-30; rebase clean (no conflicts). Author normalized lamco-office -> Greg Lamberson <greg@lamco.io>. Body extended with a "Coordinates with #1244" section mirroring the note already on #1244 — either PR can land first, the other rebases trivially since the three methods this PR adds are independent of \`on_disconnected\`'s signature change.

All five xtask gates green on the new head.

glamberson:
I'm closing this in favor of #1319, which landed `SuppressOutput` and `RefreshRectangle` handling via an `Arc<AtomicBool>` exposed on `RdpServer`. The remaining gap is the slow-path `ShareDataPdu::FrameAcknowledge` variant, which master still drops as `Unexpected share data pdu`. Issue #447 documents a real-world workload (RFX clients) where this PDU is exercised, but it isn't currently driving server-side throttling work in any in-tree consumer. A focused follow-up PR makes more sense once a downstream consumer needs slow-path frame-count throttling and can supply the "who needs this" justification.

PR #1254, opened 2026-05-10
fix(connector): handle ServerDeactivateAll during CapabilitiesExchange

### LINKED ISSUE #1253: ConnectionActivationSequence fails with "unexpected Share Control Pdu" when server sends non-DemandActive PDU during CapabilitiesExchange
When connecting to GNOME Remote Desktop (gnome-remote-desktop / grd, GNOME 44+), the connection fails during the Capabilities Exchange phase with:

```
unexpected Share Control Pdu (expected ServerDemandActive)
The error originates in ConnectionActivationSequence::step() at the CapabilitiesExchange state, which unconditionally expects ShareControlPdu::ServerDemandActive as the first PDU after licensing completes.
```

**Root cause:**

In connection_activation.rs (line ~113 in 0.8.0):

```
let capability_sets = if let rdp::headers::ShareControlPdu::ServerDemandActive(server_demand_active) =
    share_control_ctx.pdu
{
    server_demand_active.pdu.capability_sets
} else {
    return Err(general_err!(
        "unexpected Share Control Pdu (expected ServerDemandActive)",
    ));
};
```

According to MS-RDPBCGR §1.3.1.3, the server MAY send a Server Deactivate All PDU before Server Demand Active PDU as part of the Deactivation-Reactivation Sequence. GNOME Remote Desktop appears to do this.

The decode_io_channel function in legacy.rs already handles ServerDeactivateAll correctly for the active session phase, but the initial CapabilitiesExchange state does not account for it.

**Expected behavior:**

If a ServerDeactivateAll (or other non-DemandActive Share Control PDU) is received during CapabilitiesExchange, the connector should either:

Ignore the DeactivateAll and wait for the next PDU (which should be ServerDemandActive), or
Reset the activation state and re-enter CapabilitiesExchange to handle the Deactivation-Reactivation Sequence

**Steps to reproduce:**

Set up GNOME Remote Desktop on a machine running GNOME 44+ (e.g., Fedora 39+, Ubuntu 24.04+)
Enable RDP sharing via Settings → Sharing → Remote Desktop
Connect using IronRDP client with NLA (NTLM) authentication
Connection fails after successful NLA with the error above

**Environment:**

ironrdp-connector: 0.8.0
ironrdp-tokio: 0.8.0
Server: GNOME Remote Desktop (gnome-remote-desktop), GNOME 44
Authentication: NLA/NTLM (succeeds — error occurs after CredSSP completes)

**Debug log excerpt:**

```
Starting X.224 connection negotiation host=brigid port=3389
X.224 negotiation complete, starting TLS upgrade
TLS upgrade complete, proceeding to NLA/capabilities
TLS certificate not validated (no CA verification) host=brigid port=3389
IronRDP connect_finalize: host=brigid, nla=true, has_username=true, has_password=true
CredSSp finished NLA stage.
ERROR IronRDP connect_finalize failed: Error { context: "unexpected Share Control Pdu (expected ServerDemandActive)", kind: General }

```
Suggested fix:

In ConnectionActivationSequence::step(), handle ShareControlPdu::ServerDeactivateAll during CapabilitiesExchange by consuming it and remaining in the same state to wait for the actual ServerDemandActive:

```
ConnectionActivationState::CapabilitiesExchange { io_channel_id, user_channel_id } => {
    let send_data_indication_ctx = legacy::decode_send_data_indication(input)?;
    let share_control_ctx = legacy::decode_share_control(send_data_indication_ctx)?;

    // Handle DeactivateAll during initial connection (some servers like
    // GNOME Remote Desktop send this before DemandActive)
    if matches!(share_control_ctx.pdu, rdp::headers::ShareControlPdu::ServerDeactivateAll(_)) {
        debug!("Received ServerDeactivateAll during CapabilitiesExchange, waiting for DemandActive");
        return Ok(Written::Nothing);  // Stay in same state, read next PDU
    }

    let capability_sets = if let rdp::headers::ShareControlPdu::ServerDemandActive(server_demand_active) =
        share_control_ctx.pdu
    { ... }
}
```
Note: Written::Nothing with the state unchanged would cause single_sequence_step to read the next PDU and call step() again, effectively skipping the DeactivateAll.

totoshko88:
Some RDP servers (notably GNOME Remote Desktop / grd) send a ServerDeactivateAll PDU before ServerDemandActive during the initial Capabilities Exchange phase. This is valid per MS-RDPBCGR §1.3.1.3 (Deactivation-Reactivation Sequence).

Previously this caused a hard error:
"unexpected Share Control Pdu (expected ServerDemandActive)"

Now the connector skips the DeactivateAll and waits for the next PDU.

Fixes [#1253](https://github.com/Devolutions/IronRDP/issues/1253)

### DIFF
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -122,7 +122,13 @@ impl Sequence for ConnectionActivationSequence {
                 // before ServerDemandActive as part of a Deactivation-Reactivation Sequence
                 // (MS-RDPBCGR §1.3.1.3). Skip it and stay in the same state to wait for
                 // the actual DemandActive PDU.
-                if matches!(share_control_ctx.pdu, rdp::headers::ShareControlPdu::ServerDeactivateAll(_)) {
+                //
+                // The decoded PDU is intentionally discarded: the DeactivateAll body carries
+                // no payload we need during initial activation.
+                if matches!(
+                    share_control_ctx.pdu,
+                    rdp::headers::ShareControlPdu::ServerDeactivateAll(_)
+                ) {
                     debug!("Received ServerDeactivateAll during CapabilitiesExchange, waiting for ServerDemandActive");
                     self.state = ConnectionActivationState::CapabilitiesExchange {
                         io_channel_id,
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -129,7 +129,9 @@ impl Sequence for ConnectionActivationSequence {
                     share_control_ctx.pdu,
                     rdp::headers::ShareControlPdu::ServerDeactivateAll(_)
                 ) {
-                    debug!("Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active");
+                    debug!(
+                        "Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active"
+                    );
                     self.state = ConnectionActivationState::CapabilitiesExchange {
                         io_channel_id,
                         user_channel_id,
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -118,6 +118,19 @@ impl Sequence for ConnectionActivationSequence {
                     );
                 }
 
+                // Some servers (e.g. GNOME Remote Desktop) send a ServerDeactivateAll PDU
+                // before ServerDemandActive as part of a Deactivation-Reactivation Sequence
+                // (MS-RDPBCGR §1.3.1.3). Skip it and stay in the same state to wait for
+                // the actual DemandActive PDU.
+                if matches!(share_control_ctx.pdu, rdp::headers::ShareControlPdu::ServerDeactivateAll(_)) {
+                    debug!("Received ServerDeactivateAll during CapabilitiesExchange, waiting for ServerDemandActive");
+                    self.state = ConnectionActivationState::CapabilitiesExchange {
+                        io_channel_id,
+                        user_channel_id,
+                    };
+                    return Ok(Written::Nothing);
+                }
+
                 let capability_sets = if let rdp::headers::ShareControlPdu::ServerDemandActive(server_demand_active) =
                     share_control_ctx.pdu
                 {
--- a/crates/ironrdp-connector/src/connection_activation.rs
+++ b/crates/ironrdp-connector/src/connection_activation.rs
@@ -129,7 +129,7 @@ impl Sequence for ConnectionActivationSequence {
                     share_control_ctx.pdu,
                     rdp::headers::ShareControlPdu::ServerDeactivateAll(_)
                 ) {
-                    debug!("Received ServerDeactivateAll during CapabilitiesExchange, waiting for ServerDemandActive");
+                    debug!("Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active");
                     self.state = ConnectionActivationState::CapabilitiesExchange {
                         io_channel_id,
                         user_channel_id,
--- a/crates/ironrdp-testsuite-core/tests/session/connection_activation.rs
+++ b/crates/ironrdp-testsuite-core/tests/session/connection_activation.rs
@@ -1,0 +1,122 @@
+use std::borrow::Cow;
+
+use ironrdp_connector::connection_activation::{ConnectionActivationSequence, ConnectionActivationState};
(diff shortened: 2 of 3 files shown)


glamberson:
Thanks for picking this up @totoshko88. The fix is correct as written: the explicit `self.state = ConnectionActivationState::CapabilitiesExchange { ... }` before the early return is necessary because `mem::take(&mut self.state)` at the top of `step()` has already replaced the state with `Consumed`, and the normal `self.state = next_state` assignment at the bottom of the function is bypassed by the early return.

The spec read is also reasonable. MS-RDPBCGR §1.3.1.3 frames the Deactivation-Reactivation Sequence as a post-completion server-initiated reactivation, so a server sending DeactivateAll during the initial connection is technically nonstandard. The "skip and wait" approach this PR takes is the conservative interoperable interpretation: treat the inbound DeactivateAll as the start of a reactivation, stay in CapabilitiesExchange, and let the next DemandActive arrive normally.

Two small suggestions a reviewer is likely to ask for:

**1. A test for the new behavior.** A connector-step test that feeds a `ServerDeactivateAll` PDU during `CapabilitiesExchange` and asserts the state stays at `CapabilitiesExchange { io_channel_id, user_channel_id }` and `Written::Nothing` is returned would lock this in. A follow-on assertion that feeds a normal `ServerDemandActive` immediately after and shows the connector transitions to `ConnectionFinalization` would cover the full GRD pattern end to end.

**2. A short comment noting that the decoded PDU is intentionally discarded.** Something like:

```rust
// Drop the decoded share_control_ctx: the DeactivateAll body carries
// no payload we need during initial activation, and the channel-id
// mismatch warning above has already fired if applicable.
```

This is downstream of relevance to Lamco's GNOME Remote Desktop interop work, where this connector-side error has been blocking client-side testing against GRD. Worth flagging that #1232 (GlassOnTin) reports a cousin failure mode with the same `unexpected Share Control PDU` error string family but a different root cause: the server closing the connection during capabilities exchange rather than sending DeactivateAll. That one is not addressed by this PR and remains a separate work item. Calling it out so a reader of #1232 finds their way here and does not assume the two cases overlap fully.

totoshko88:
Thanks for the review @glamberson! I've addressed both suggestions:
  
1. Added a comment noting the decoded PDU is intentionally discarded — clarifies that the DeactivateAll body carries no payload needed during initial activation.

2. Added tests for the new behavior:
- deactivate_all_during_capabilities_exchange_stays_in_same_state — feeds a ServerDeactivateAll PDU during CapabilitiesExchange and asserts the state remains unchanged with Written::Nothing returned.
- demand_active_after_deactivate_all_transitions_to_connection_finalization — feeds ServerDeactivateAll followed by ServerDemandActive and asserts the  connector transitions to ConnectionFinalization, covering the full GRD pattern end to end.

glamberson:
Thanks @totoshko88, both items look right. The comment captures the "body decoded but discarded" rationale cleanly, and the tests cover the case that was breaking GRD interop.

No further comments from my side.

crates/ironrdp-connector/src/connection_activation.rs
```diff
@@ -118,6 +118,25 @@ impl Sequence for ConnectionActivationSequence {
...
+                // (MS-RDPBCGR §1.3.1.3). Skip it and stay in the same state to wait for
+                // the actual DemandActive PDU.
+                //
+                // The decoded PDU is intentionally discarded: the DeactivateAll body carries
+                // no payload we need during initial activation.
+                if matches!(
+                    share_control_ctx.pdu,
+                    rdp::headers::ShareControlPdu::ServerDeactivateAll(_)
+                ) {
+                    debug!("Received ServerDeactivateAll during CapabilitiesExchange, waiting for ServerDemandActive");
```
CBenoit:
```suggestion
                    debug!("Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active");
```

CBenoit (approved):
LGTM! Thank you both!

crates/ironrdp-connector/src/connection_activation.rs
```diff
@@ -118,6 +118,25 @@ impl Sequence for ConnectionActivationSequence {
...
+                // (MS-RDPBCGR §1.3.1.3). Skip it and stay in the same state to wait for
+                // the actual DemandActive PDU.
+                //
+                // The decoded PDU is intentionally discarded: the DeactivateAll body carries
+                // no payload we need during initial activation.
+                if matches!(
+                    share_control_ctx.pdu,
+                    rdp::headers::ShareControlPdu::ServerDeactivateAll(_)
+                ) {
+                    debug!("Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active");
```
CBenoit:
```suggestion
                    debug!(
                        "Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active"
                    );
```

PR #1256, opened 2026-05-10
fix(rdpsnd-native): allocate Opus PCM buffer as Vec<i16> to avoid alignment panic

glamberson:
Closes #1202

## Problem

The Opus decode path in `DecodeStream::new` allocates the destination PCM buffer as `Vec<u8>`, then casts it to `&mut [i16]` for the Opus decoder via `bytemuck::cast_slice_mut`:

```rust
let mut pcm = vec![0u8; nb_samples * chan as usize * size_of::<i16>()];
if let Err(error) = dec.decode(&pkt, bytemuck::cast_slice_mut(pcm.as_mut_slice()), false) {
```

`Vec<u8>` has alignment 1. The Rust allocator returns a pointer that may or may not happen to be 2-byte aligned. When the pointer lands on an odd address, `bytemuck::cast_slice_mut::<u8, i16>` panics with `TargetAlignmentGreaterAndInputNotAligned`.

The latent bug surfaces in #1202 when the remote server reboots. The server emits a flood of malformed Opus packets during reboot, each triggering a fresh allocation in this loop. With enough allocations in quick succession, one lands on an unaligned address and the dec_thread crashes. The cascade of "channel is empty / sending half is closed" warnings that follow is the consequence: the panicked thread dropped its sender.

Reproduced from the panic.log attached to #1202:

```
thread '<unnamed>' panicked at .../bytemuck-1.25.0/src/internal.rs:33:3:
cast_slice_mut>TargetAlignmentGreaterAndInputNotAligned
```

## Fix

Allocate the PCM buffer as `Vec<i16>` directly. `Vec<i16>` has the required 2-byte alignment by construction, and `opus2::Decoder::decode` takes `&mut [i16]` natively with no cast needed.

The downstream channel still carries `Vec<u8>`, so we convert via `bytemuck::cast_slice(&pcm_i16).to_vec()`. Going from a larger alignment (i16 = 2) to a smaller one (u8 = 1) is always safe and never panics. The extra `Vec` copy is one small allocation per audio packet, negligible for audio (Opus packets are typically 2.5-60 ms of audio).

## Test

- `cargo xtask check fmt` clean
- `cargo xtask check lints` clean
- `cargo xtask check locks` clean
- `cargo xtask check typos` clean
- `cargo xtask check tests` clean


crates/ironrdp-rdpsnd-native/src/cpal.rs
```diff
@@ -164,11 +164,19 @@ impl DecodeStream {
...
                             continue;
                         }
+                        // Vec<u8> is what the channel carries downstream. Reinterpreting
+                        // Vec<i16> -> Vec<u8> via cast_slice is safe (smaller alignment).
+                        // Allocating as Vec<i16> in the first place avoids the alignment
+                        // hazard of `bytemuck::cast_slice_mut::<u8, i16>` panicking when
+                        // the allocator hands back a u8 buffer that is not 2-byte aligned
+                        // (which manifested as a hard crash in #1202 under the burst of
+                        // malformed Opus packets generated by a server reboot).
+                        let pcm = bytemuck::cast_slice(&pcm_i16).to_vec();
```
CBenoit, line 179: It seems to be performing an extra allocation, and I feel there must be a way to avoid that. There is a `cast_vec` function in the `bytemuck` crate: https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_vec.html

glamberson: Good instinct on the alloc, but `bytemuck::cast_vec::<i16, u8>` actually panics on these types:

```rust
let v: Vec<i16> = vec![0x4142, 0x4344, 0x4546];
let _: Vec<u8> = bytemuck::cast_vec(v);
// thread 'main' panicked at bytemuck-1.25.0/src/allocation.rs:242:
// called `Result::unwrap()` on an `Err` value: AlignmentMismatch
```

`cast_vec` requires `align_of::<A>() == align_of::<B>()` and `size_of::<A>() == size_of::<B>()`.
The deeper reason: `Vec` deallocates with `Layout::array::<T>(capacity)`, so reusing a `Vec<i16>`
(align=2) allocation as a `Vec<u8>` (align=1) would dealloc with a different Layout than was used
at alloc. That's UB per the `GlobalAlloc` contract.

The copy is structural to crossing into `Receiver<Vec<u8>>`. Making the channel typed plus using
cpal's typed output stream would skip it, but that's out of scope here.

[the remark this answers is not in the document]
glamberson: See above.

CBenoit (approved):
LGTM!

PR #1258, opened 2026-05-11
refactor(mstsgu): derive Display and Error for GwErrorKind via thiserror

glamberson:
Refs #1257.

First proof-of-concept for the broader `thiserror` migration proposed in #1257. `GwErrorKind` is the smallest hand-written `*ErrorKind` enum in the workspace at 7 variants, so it is suitable for verifying that the migration is byte-identity-preserving and that the workspace `thiserror` dep lands cleanly.

### What this changes

Replaces the hand-written `impl Display for GwErrorKind` and `impl core::error::Error for GwErrorKind` blocks with `#[derive(thiserror::Error)]` and per-variant `#[error("...")]` attributes. Adds `thiserror = "2"` to `crates/ironrdp-mstsgu/Cargo.toml` as a direct dependency, following the same pattern that `ironrdp-pdu` already uses for its internal `thiserror` types.

### Public API impact

None. The `pub type Error = ironrdp_error::Error<GwErrorKind>` alias and the `GwErrorExt::custom()` helper are unchanged. Consumers see no source-level break.

### Display byte-identity verification

Verified via `cargo expand`. Both the prior hand-written implementation and the `thiserror`-derived implementation call `formatter.write_str("literal")` with the same string for each of the 7 variants:

| Variant | String |
|---|---|
| `InvalidGwTarget` | `"invalid GW Target"` |
| `Connect` | `"connection error"` |
| `PacketEof` | `"PacketEOF"` |
| `UnsupportedFeature` | `"unsupported feature"` |
| `Custom` | `"custom"` |
| `Encode` | `"encode"` |
| `Decode` | `"decode"` |

### Verification

- `cargo xtask check fmt`: clean
- `cargo xtask check lints`: clean
- `cargo xtask check locks`: clean
- `cargo xtask check typos`: clean
- `cargo xtask check tests`: clean

### MSRV

Unaffected. `thiserror` 2.0 requires Rust 1.61, well below IronRDP's MSRV of 1.89.

Issue #1257, opened 2026-05-11
Modernize `ironrdp-error`: thiserror 2.0, Location capture, shared `bail!`/`ensure!`


glamberson:
> [!NOTE]
> **Edit 2026-05-11.** Change 1 below (the `thiserror` derive on Core Tier `*ErrorKind` enums) was withdrawn after @CBenoit cited the proc-macro architectural invariant at `ARCHITECTURE.md` line 36. The full reasoning is in [this comment](https://github.com/Devolutions/IronRDP/issues/1257#issuecomment-4426485616). Implementation PRs #1258, #1260, and #1261 are closed. Change 2 (PR #1262, Location capture via `#[track_caller]`) and Change 3 (PR #1263, shared `bail!`/`ensure!` via `macro_rules!`) remain open since neither introduces a proc-macro dependency. A new PR #1264 delivers the inverse migration per the maintainer's stated direction: it removes the existing `thiserror = "2.0"` line from `ironrdp-pdu/Cargo.toml` and hand-rolls the impls in that crate. Body below is preserved as originally filed.

---

The `ironrdp-error::Error<Kind>` design is sound, and the per-crate `*ErrorKind` pattern matches what other large Rust workspaces such as Polars and GreptimeDB have settled on for cross-crate error chains. Two ecosystem changes since `ironrdp-error` was first written open up a meaningful boilerplate reduction without changing the public type. I am proposing three small additive changes here, each as its own pull request. All three preserve `Error<Kind>`, so consumers of typed-error aliases would see no source-level break.

These changes would be valuable to land before two efforts already in flight: the structured-fuzzing series (#1120, #1121, #1122) and the library-side `anyhow` removal in `ironrdp-server` and `ironrdp-rdpsnd-native`. The per-PDU `Arbitrary` implementations would then land on top of cleaner error definitions, and the eventual `ServerError` would be roughly thirty lines of code rather than one hundred and fifty once the boilerplate is reduced.

### The two ecosystem changes that unblock this

1. `thiserror` 2.0 (released November 2024) added first-class `no_std` support via a `std` feature flag. The historical reason that `Kind` enums in Core Tier crates were hand-written is no longer in force. `ironrdp-pdu` already pins `thiserror = "2.0"` for some of its internal types.
2. `core::error::Error` was stabilized in Rust 1.81 (September 2024), well below IronRDP's MSRV of 1.89. `ironrdp-error` already depends on it, and the wider workspace can now do the same.

[...]

PR #1259, opened 2026-05-11
refactor: derive Display and Error via thiserror across core, pdu, connector, session

glamberson:
Refs #1257.

Second batch in the `thiserror` migration proposed in #1257. The first batch (#1258, mstsgu) verified the byte-identical Display approach on the smallest hand-written `*ErrorKind` enum. This batch applies the same pattern to the four remaining crates with hand-written `Display + core::error::Error` impls atop `ironrdp-error::Error<Kind>`.

### What this changes per crate

- **`ironrdp-core`**: adds `thiserror = { version = "2", default-features = false }` and wires `thiserror/std` into the existing `std` feature. Both `DecodeErrorKind` and `EncodeErrorKind` are derived. The cfg-gated `UnsupportedValue` variants (with/without `alloc`) each receive their own `#[error("...")]` attribute matching the prior format string. The `#[cfg(feature = "std")] impl core::error::Error` gate is removed since `core::error::Error` is in core (stabilised Rust 1.81, well below MSRV 1.89) and thiserror's derive emits an unconditional impl.
- **`ironrdp-pdu`**: thiserror was already pinned at `"2.0"` for some internal types. `PduErrorKind` is now derived. Hand-written `Display` and `Error` impls removed.
- **`ironrdp-connector`**: adds `thiserror = "2"`. `ConnectorErrorKind` is derived. Tuple variants (`Encode`, `Decode`, `Credssp`, `Negotiation`) receive `#[source]` attributes so the source chain that the prior hand-written `source()` impl established is preserved.
- **`ironrdp-session`**: same pattern as connector. `SessionErrorKind` is derived with `#[source]` on `Pdu`, `Encode`, `Decode` tuple variants. Removes a now-unused `use core::fmt;` import.

### Public API impact

None. The `pub type FooError = ironrdp_error::Error<FooErrorKind>` aliases and the `*ErrorExt` traits are unchanged. Consumers see no source-level break.

### Display byte-identity verification

Verified via `cargo expand`. Each derived impl calls `formatter.write_str()` or `write_fmt(format_args!())` with the same literal/arguments as the prior hand-written code. Unit variants emit identical `write_str` calls; tuple variants with formatted source emit `format_args!("prefix: {0}", _0.as_display())` which calls `Display` on the inner field, identical to the prior `write!(f, "prefix: {field_name}")`.

### no_std verification (ironrdp-core)

All three feature combinations build clean:
- `cargo check -p ironrdp-core --no-default-features` (no_std, no alloc)
- `cargo check -p ironrdp-core --no-default-features --features alloc` (no_std + alloc)
- `cargo check -p ironrdp-core --features std` (std)

### xtask verification

- `cargo xtask check locks`: clean

### Diff size

9 files changed, +50 −159 (net −109 lines of code removed).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2538,6 +2538,7 @@ dependencies = [
  "picky-asn1-x509",
  "rand 0.9.2",
  "sspi",
+ "thiserror 2.0.18",
  "tracing",
  "url",
 ]
@@ -2547,6 +2548,7 @@ name = "ironrdp-core"
 version = "0.1.5"
 dependencies = [
  "ironrdp-error",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -2842,6 +2844,7 @@ dependencies = [
  "ironrdp-pdu",
  "ironrdp-svc",
  "qoicoubeh",
+ "thiserror 2.0.18",
  "tracing",
  "zstd-safe",
 ]
--- a/crates/ironrdp-connector/Cargo.toml
+++ b/crates/ironrdp-connector/Cargo.toml
@@ -35,6 +35,7 @@ tracing = { version = "0.1", features = ["log"] }
 picky-asn1-der = "0.5"
 picky-asn1-x509 = "0.15"
 picky = "=7.0.0-rc.22" # FIXME: We are pinning with = because the candidate version number counts as the minor number by Cargo, and will be automatically bumped in the Cargo.lock.
+thiserror = "2"
 
 [lints]
 workspace = true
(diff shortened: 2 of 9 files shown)


glamberson:
Closing in favor of two smaller PRs per STYLE.md guidance ('Sending small cleanup PRs is encouraged'). Splitting along the natural boundary:

- PR A: `ironrdp-pdu`, `ironrdp-connector`, `ironrdp-session` — std-only crates with mechanical thiserror derive (same pattern across all three).
- PR B: `ironrdp-core` — qualitatively different because of the no_std-aware feature wiring (`thiserror = { default-features = false }` + `std = [..., "thiserror/std"]`), removing the `#[cfg(feature = "std")] impl core::error::Error` gate, and the cfg-gated `UnsupportedValue` variants.

Will also fix the missed `fuzz/Cargo.lock` update that broke the Fuzzing job's lockfile-check step here. Refs #1257.

PR #1260, opened 2026-05-11
refactor: derive Display and Error via thiserror in pdu, connector, session

glamberson:
Refs #1257. Replaces #1259 (closed for split).

Part of the thiserror migration proposed in #1257. The first proof-of-concept (#1258, mstsgu) verified the byte-identical Display approach. This PR applies the same mechanical pattern to the three std-only crates with hand-written `*ErrorKind` impls atop `ironrdp-error`. `ironrdp-core` is handled in a separate PR (incoming) because of its no_std-aware feature wiring.

### Per-crate notes

- **`ironrdp-pdu`**: `thiserror` was already pinned at `"2.0"` for some internal types. `PduErrorKind` is now derived. Hand-written `Display` and `Error` impls removed.
- **`ironrdp-connector`**: adds `thiserror = "2"` as a direct dep. `ConnectorErrorKind` is derived. Tuple variants (`Encode`, `Decode`, `Credssp`, `Negotiation`) receive `#[source]` attributes so the source chain that the prior hand-written `core::error::Error::source()` impl established is preserved.
- **`ironrdp-session`**: same pattern as connector. `SessionErrorKind` is derived with `#[source]` on `Pdu`, `Encode`, `Decode` tuple variants. Removes the now-unused `use core::fmt;` import.

### Public API impact

None. The `pub type FooError = ironrdp_error::Error<FooErrorKind>` aliases and the `*ErrorExt` traits are unchanged. Consumers see no source-level break.

### Display byte-identity verification

Verified via `cargo expand`. Each derived impl calls `formatter.write_str()` or `write_fmt(format_args!())` with the same literal/arguments as the prior hand-written code. Unit variants emit identical `write_str` calls; tuple variants with formatted source (e.g. `Reason(String)`, `Negotiation`) emit `format_args!("prefix: {0}", _0.as_display())` which calls `Display` on the inner field, identical to the prior `write!(f, "prefix: {field_name}")`.

### xtask verification

- `cargo xtask check locks`: clean (root `Cargo.lock` updated and included)

### Diff size

6 files changed, +31 −86 (net −55 lines of code removed).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2538,6 +2538,7 @@ dependencies = [
  "picky-asn1-x509",
  "rand 0.9.2",
  "sspi",
+ "thiserror 2.0.18",
  "tracing",
  "url",
 ]
@@ -2842,6 +2843,7 @@ dependencies = [
  "ironrdp-pdu",
  "ironrdp-svc",
  "qoicoubeh",
+ "thiserror 2.0.18",
  "tracing",
  "zstd-safe",
 ]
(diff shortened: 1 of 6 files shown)


glamberson:
Closing per #1257 (https://github.com/Devolutions/IronRDP/issues/1257#issuecomment-4426485616). Change 1's premise is withdrawn after the Core Tier proc-macro invariant correction.

PR #1261, opened 2026-05-11
refactor(core): derive Display and Error via thiserror in DecodeErrorKind and EncodeErrorKind

glamberson:
Refs #1257. Companion to #1260.

Companion to the std-only `thiserror` migration in #1260. `ironrdp-core` is split off because it is `no_std`-aware and the change requires Cargo.toml feature wiring that the other crates do not need.

### Changes

- Adds `thiserror = { version = "2", default-features = false }` to dependencies.
- Wires `thiserror/std` into the existing `std` feature: `std = ["alloc", "ironrdp-error/std", "thiserror/std"]`.
- Derives `thiserror::Error` on `DecodeErrorKind` and `EncodeErrorKind` with per-variant `#[error("...")]` attributes that match the prior `write!(f, "...")` format strings exactly.
- Each cfg-gated `UnsupportedValue` variant (with/without `alloc`) receives its own `#[error("...")]` attribute.
- Removes the `#[cfg(feature = "std")] impl core::error::Error for *ErrorKind {}` gates. `core::error::Error` lives in `core` (stabilized Rust 1.81, well below MSRV 1.89), and `thiserror`'s derive emits an unconditional impl. The behaviour change is from "Error impl available only with std" to "Error impl always available", which is a strict superset.

### Public API impact

None. The `pub type DecodeError = ironrdp_error::Error<DecodeErrorKind>` and `pub type EncodeError = ironrdp_error::Error<EncodeErrorKind>` aliases are unchanged. Consumers see no source-level break.

### no_std verification

All three feature combinations build cleanly:

- `cargo check -p ironrdp-core --no-default-features` (no_std, no alloc)
- `cargo check -p ironrdp-core --no-default-features --features alloc` (no_std + alloc)
- `cargo check -p ironrdp-core --features std`

### xtask verification

- `cargo xtask check locks`: clean (both root `Cargo.lock` and `fuzz/Cargo.lock` updated and included; the missing fuzz lock was the failure mode of the earlier #1259 attempt)

### Diff size

5 files changed, +20 −73 (net −53 lines of code removed).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2547,6 +2547,7 @@ name = "ironrdp-core"
 version = "0.1.5"
 dependencies = [
  "ironrdp-error",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/crates/ironrdp-core/Cargo.toml
+++ b/crates/ironrdp-core/Cargo.toml
@@ -18,8 +18,9 @@ test = false
 
 [features]
 default = []
-std = ["alloc", "ironrdp-error/std"]
+std = ["alloc", "ironrdp-error/std", "thiserror/std"]
 alloc = ["ironrdp-error/alloc"]
 
 [dependencies]
 ironrdp-error = { path = "../ironrdp-error", version = "0.1" } # public
+thiserror = { version = "2", default-features = false }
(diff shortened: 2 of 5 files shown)


glamberson:
Closing per #1257 (https://github.com/Devolutions/IronRDP/issues/1257#issuecomment-4426485616). Change 1's premise is withdrawn after the Core Tier proc-macro invariant correction.

PR #1264, opened 2026-05-12
refactor(pdu): hand-roll Display/Error impls and remove thiserror dependency

glamberson:
Closes the proc-macro architectural-invariant violation that issue #1257 surfaced and that I committed to addressing in https://github.com/Devolutions/IronRDP/issues/1257#issuecomment-4426485616.

## What changes

`ironrdp-pdu` is Core Tier per `ARCHITECTURE.md`. Core Tier carries the architectural invariant that no proc-macro dependency (`syn`) is allowed, so compilation stays fast for downstream consumers that build only the protocol-decode tree (the fuzzing infrastructure, the web client, `no_std` users).

`thiserror = "2.0"` was carried in the legacy block of `ironrdp-pdu/Cargo.toml` under the existing `# TODO: get rid of these dependencies` comment. This PR delivers on that TODO. Every `#[derive(thiserror::Error)]` in the crate is replaced with hand-written `impl Display`, `impl core::error::Error`, and `impl From<X>` blocks that preserve the exact `Display` output byte-for-byte and the exact `Error::source()` semantics that the derives generated.

## Enums converted (13 total across 13 files)

GCC family:
- `CoreDataError`, `GccError`, `NetworkDataError`, `ClusterDataError`, `SecurityDataError`

RDP family:
- `RdpError`, `CapabilitySetsError`, `ChannelError`, `SessionError`, `ServerLicenseError`, `ClientInfoError`

Other:
- `InputEventError`
- `McsError` (legacy submodule in `mcs.rs`)

Each enum keeps `#[derive(Debug)]`. The hand-written `Display` preserves every `#[error("...")]` format string verbatim, including `{0}` and `{0:?}` interpolations and the named-field interpolation in `ChannelError::InvalidDvcTotalMessageSize`. The `Error::source()` impl preserves the `#[from]`-driven source chain plus the named `source:` field on `ServerLicenseError::InvalidX509Certificate`. The existing manual `From<PduError>`, `From<RdpError> for io::Error`, `From<McsError> for io::Error`, `From<ChannelError> for io::Error`, and `From<LicensingErrorMessage> for ServerLicenseError` impls are preserved unchanged.

## Verification

- `cargo expand` against the post-change tree shows `Display` and `source` impls that are functionally equivalent to the thiserror-generated versions for every variant.
- `cargo check --workspace` clean.
- `cargo test --package ironrdp-pdu` 850 tests pass, 0 failures.
- `cargo clippy --package ironrdp-pdu --all-targets -- -D warnings` clean.
- `cargo xtask check fmt` clean.
- `cargo xtask check typos` clean.

## Downstream impact

External consumers of these types in the workspace (`ironrdp-connector` for `ServerLicenseError`, `ffi` and `ironrdp-session` for `RdpError` and `SessionError`) treat them as opaque error types. A workspace-wide grep finds no variant-level pattern matching on these enums outside `ironrdp-pdu` itself, so the preserved `From` conversions and `Display` output fully cover downstream expectations.

## Diff shape

15 files changed, +748 / -216 (net +532). Most of the added lines are the explicit `Display` match arms and the per-variant `From` and `source()` arms that the proc-macro previously generated at compile time.

Related: #1257.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2715,7 +2715,6 @@ dependencies = [
  "pkcs1 0.7.5",
  "sha1 0.10.6",
  "tap",
- "thiserror 2.0.18",
  "x509-cert",
 ]
 
--- a/crates/ironrdp-pdu/Cargo.toml
+++ b/crates/ironrdp-pdu/Cargo.toml
@@ -33,7 +33,6 @@ tap = "1"
 bit_field = "0.10"
 byteorder = "1.5" # TODO: remove
 der-parser = "10.0"
-thiserror = "2.0"
 md5 = { package = "md-5", version = "0.10" }
 num-bigint = "0.4"
 num-derive.workspace = true # TODO: remove
--- a/crates/ironrdp-pdu/src/gcc/cluster_data.rs
+++ b/crates/ironrdp-pdu/src/gcc/cluster_data.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use std::io;
 
 use bitflags::bitflags;
@@ -6,7 +7,6 @@ use ironrdp_core::{
 };
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive as _;
-use thiserror::Error;
 
 const REDIRECTION_VERSION_MASK: u32 = 0x0000_003C;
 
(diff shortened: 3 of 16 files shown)


CBenoit (approved):
Thank you, LGTM! Running Copilot just to double check

PR #1269, opened 2026-05-13
fix(error): box diagnostic metadata to shrink Error<Kind> size

CBenoit:
Move context, location, and source into a heap-allocated ErrorMeta struct so that Error<Kind> holds only kind on the stack. This eliminates the size cascade that caused ConnectorError to exceed clippy's 128-byte threshold (it was raised to 152 to accommodate the Location field).

Error construction is already #[cold], so the single Box allocation per error is acceptable.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -4,10 +4,3 @@ accept-comment-above-statement = true
 accept-comment-above-attributes = true
 allow-panic-in-tests = true
 allow-unwrap-in-tests = true
-# Raised from the default of 128 to accommodate the `&'static core::panic::Location<'static>`
-# field that `ironrdp_error::Error<Kind>` captures via `#[track_caller]` for diagnostic
-# reporting. The location field adds one usize (8 bytes) to every error type, which pushes
-# `Error<ConnectorErrorKind>` (whose largest variant nests another `Error<Kind>`) just over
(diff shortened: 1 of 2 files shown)


awakecoding (approved):

glamberson:
I saw this come in this morning. The Box<ErrorMeta> layout is the right move.

I'll rebase #1266 onto this. The Option<ErrorPosition> I proposed there fits as a fourth field on ErrorMeta, which dissolves the threshold bump that was the awkward part of that PR. I'll mark #1266 as depending on this once I push. If you don't want position metadata at all, withdrawal is fine.

vnikonov-devolutions (approved):

CBenoit:
> I saw this come in this morning. The Box layout is the right move.
> 
> I'll rebase #1266 onto this. The Option I proposed there fits as a fourth field on ErrorMeta, which dissolves the threshold bump that was the awkward part of that PR. I'll mark #1266 as depending on this once I push. If you don't want position metadata at all, withdrawal is fine.

Hey! Thank you; yes that’s exactly what I had in mind. I saw that we were growing the size of the main error type a bit too much, but I think this new approach will allow us to store more metadata without feeling bad about it 🙂

PR #1271, opened 2026-05-13
feat(fuzz): add egfx PDU decoders to pdu_decode oracle + fix two surfaced bugs

glamberson:
## Summary

Extends the existing `pdu_decode` oracle in `ironrdp-fuzzing::oracles` with EGFX coverage, satisfying the Core Tier "must be fuzzed" invariant (per `ARCHITECTURE.md`) for the graphics-pipeline DVC parsing. Two pre-existing bugs in `ironrdp-egfx`'s `Avc444BitmapStream` / `Avc420BitmapStream` decoders that the new coverage immediately surfaced are fixed in the same series.

## What the oracle exercises

The EGFX decode list added to `pdu_decode` runs the input through:

- `GfxPdu` (umbrella enum covering all 23 EGFX message types — primary entry point)
- `CacheToSurfacePdu` (sole sub-PDU with an independent `Decode` impl)
- `EgfxCapabilitySet` (aliased to disambiguate against `capability_sets::CapabilitySet` from `ironrdp-pdu`)
- `Avc420BitmapStream<'_>` and `Avc444BitmapStream<'_>`
- `QuantQuality`, `Point`, `Color`

## Shape

Folded into the existing `pdu_decode` oracle in `c5211e46` per your suggestion on the original filing ("I think it's fine to add this directly to `pdu_decode`, so we don't have an extra target for essentially the same logic"). No separate `egfx_pdu_decoding.rs` target / `[[bin]]` entry; the EGFX decoders sit alongside the existing PDU decoders in the same oracle and share the corpus.

## Bug fixes surfaced by the new coverage

Both pre-existing in `ironrdp-egfx`; surfaced by the EGFX decode additions before this PR could merge cleanly.

- **`f96ba08e fix(egfx): bound stream_len before split_at in Avc444BitmapStream`**: `Avc444BitmapStream::decode` read a 30-bit `streamLen` field and called `src.split_at(stream_len as usize)` without bounds-checking. `ReadCursor::split_at` panics on offset overflow. Fixed with `ensure_size!(ctx: Self::NAME, in: src, size: stream_len)` before the split, so a malformed `streamLen` returns `NotEnoughBytes` instead of panicking.
- **`f741fb26 fix(egfx): cap Avc420BitmapStream pre-allocation against remaining bytes`**: `Avc420BitmapStream::decode` read a u32 `num_regions` and called `Vec::with_capacity(num_regions as usize)` twice without bounding. ASan tripped on a ~33 GB allocation for `num_regions = 0xF7_00_AC_7E`. Fixed by capping `with_capacity` at `src.len() / (InclusiveRectangle::FIXED_PART_SIZE + QuantQuality::FIXED_PART_SIZE)`, matching the bounded-pre-allocation pattern in `cliprdr/format_data/file_list.rs`. The actual decode loop is unchanged.

## Verification

- `cargo xtask check fmt` ✓
- `cargo xtask check lints` ✓
- `cargo xtask check locks` ✓
- `cargo xtask check typos` ✓
- `cargo xtask check tests` ✓
- `cargo check --manifest-path fuzz/Cargo.toml` builds the `pdu_decoding` target cleanly
- CI on `f741fb26`: all 9 jobs green including Fuzzing 5m21s (the same 5-minute fuzz budget that previously surfaced both bugs now passes clean)

Refs #1120.

### DIFF
--- a/crates/ironrdp-fuzzing/src/oracles/mod.rs
+++ b/crates/ironrdp-fuzzing/src/oracles/mod.rs
@@ -14,6 +14,10 @@ use crate::generators::BitmapInput;
 
 pub fn pdu_decode(data: &[u8]) {
     use ironrdp_core::decode;
+    use ironrdp_egfx::pdu::{
+        Avc420BitmapStream, Avc444BitmapStream, CacheToSurfacePdu, CapabilitySet as EgfxCapabilitySet, Color, GfxPdu,
+        Point, QuantQuality,
+    };
     use ironrdp_pdu::mcs::{ConnectInitial, ConnectResponse, McsMessage};
     use ironrdp_pdu::nego::{ConnectionConfirm, ConnectionRequest};
     use ironrdp_pdu::rdp::{ClientInfoPdu, capability_sets, headers, server_error_info, server_license, vc};
@@ -80,6 +84,15 @@ pub fn pdu_decode(data: &[u8]) {
 
     let _ = decode::<ironrdp_rdpsnd::pdu::ServerAudioOutputPdu<'_>>(data);
     let _ = decode::<ironrdp_rdpsnd::pdu::ClientAudioOutputPdu>(data);
+
+    let _ = decode::<GfxPdu>(data);
+    let _ = decode::<CacheToSurfacePdu>(data);
+    let _ = decode::<EgfxCapabilitySet>(data);
+    let _ = decode::<Avc420BitmapStream<'_>>(data);
+    let _ = decode::<Avc444BitmapStream<'_>>(data);
+    let _ = decode::<QuantQuality>(data);
+    let _ = decode::<Point>(data);
+    let _ = decode::<Color>(data);
 }
 
 pub fn rle_decompress_bitmap(input: BitmapInput<'_>) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2643,6 +2643,7 @@ dependencies = [
  "ironrdp-cliprdr-format",
  "ironrdp-core",
  "ironrdp-displaycontrol",
+ "ironrdp-egfx",
  "ironrdp-graphics",
  "ironrdp-pdu",
  "ironrdp-rdpdr",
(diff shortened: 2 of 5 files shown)


crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -12,6 +12,29 @@
...
+
+    let _ = decode::<GfxPdu>(data);
+    let _ = decode::<CacheToSurfacePdu>(data);
+    let _ = decode::<CapabilitySet>(data);
+    let _ = decode::<Avc420BitmapStream<'_>>(data);
+    let _ = decode::<Avc444BitmapStream<'_>>(data);
+    let _ = decode::<QuantQuality>(data);
+    let _ = decode::<Point>(data);
+    let _ = decode::<Color>(data);
+}
```
CBenoit: suggestion: I think it’s fine to add this directly to `pdu_decode`, so we don’t have an extra target for essentially the same logic

CBenoit (commented):
Thank you! That was a blind spot

crates/ironrdp-fuzzing/src/oracles/mod.rs
glamberson: Folded into `pdu_decode` and dropped the separate target. Force-pushed as c5211e46. The egfx `CapabilitySet` is aliased to `EgfxCapabilitySet` in the imports since it collides with `capability_sets::CapabilitySet` from ironrdp-pdu.

CBenoit (approved):
LGTM

crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -14,6 +14,10 @@ use crate::generators::BitmapInput;
 
 pub fn pdu_decode(data: &[u8]) {
     use ironrdp_core::decode;
+    use ironrdp_egfx::pdu::{
+        Avc420BitmapStream, Avc444BitmapStream, CacheToSurfacePdu, CapabilitySet as EgfxCapabilitySet, Color, GfxPdu,
+        Point, QuantQuality,
```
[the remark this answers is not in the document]
glamberson: Title and body updated to describe what shipped: the EGFX decoders were folded into the existing `pdu_decode` oracle in `c5211e46` after CBenoit's suggestion ("don't add an extra target for essentially the same logic"). No separate `egfx_pdu_decode` symbol and no `fuzz_targets/egfx_pdu_decoding.rs` exist; the EGFX decode list lives alongside the other PDU decoders in `pdu_decode` and shares the corpus. PR description previously read as the original filing intent and is now corrected.

Issue #1123, opened 2026-02-24
CI: compile/test a feature matrix including `arbitrary` (std/no_std)


CBenoit:
Prevent regressions by ensuring CI compiles key crates under important feature combinations.

## Approach
Two acceptable options:
1) Use `cargo-hack` to run a selected matrix, OR
2) Add explicit `cargo check` commands for curated feature sets

## Tasks
- Decide matrix strategy (`cargo-hack` vs curated commands)
- Add CI job(s) to cover at least:
  - `--features arbitrary` for PDU/protocol crates
  - `std` enabled and `std` disabled (no_std/alloc) where supported
- Ensure runtime stays reasonable (can be a separate, parallel job)

## Acceptance criteria
- CI fails if `arbitrary` breaks
- CI validates at least one no_std/alloc build path if supported
- Matrix is documented (so future contributors know what to extend)

glamberson:
## Workspace feature-matrix state on `master @ 0282d182`

Workspace-wide `cargo hack check --feature-powerset --depth 2` run as preparation for the matrix CI gate this issue calls for. Posting the findings here so the gate's design and acceptance criteria are grounded in the actual workspace state rather than assumed-clean.

### Methodology

cargo-hack 0.6.44, run per-crate against each of the 50 workspace members:

```
cargo hack check -p <crate> \
  --feature-powerset --no-dev-deps \
  --exclude-features qoiz,__bench,__test \
  --depth 2
```

Per-crate isolation so each failure shows independently. Exclude rationale: `qoiz` is dependent on `qoi` and its combinatorics are covered by the `qoi` enumeration; `__bench` and `__test` are private (`__`-prefixed) features already exercised by the existing lints job's `--features helper,__bench`. `--depth 2` to bound combinatorial cost; depth-3 may surface further combinations but is out of this scan's scope.

### Top-line

| Outcome | Count |
|---|---|
| Pass | 41 |
| Fail | 6 |
| Skipped — already in workspace-root `[workspace] exclude` with `FIXME: fix compilation` | 3 |

The 3 pre-excluded crates (`ironrdp-client-glutin`, `ironrdp-glutin-renderer`, `ironrdp-replay-client`) are documented as broken in the workspace `Cargo.toml` and are not in scope here.

[...]

PR #1272, opened 2026-05-13
feat(pdu): add arbitrary feature for structure-aware fuzzing

### LINKED ISSUE #1121: Fix `arbitrary` feature flag plumbing so it compiles

glamberson:
## Summary

**Closes #1121.** Part of @CBenoit's fuzzing umbrella #1120; this PR unblocks the downstream sub-issues #1122 (Arbitrary impls for more PDU structs), #1123 (CI feature matrix), and #1124 (advanced fuzz targets). Supersedes #1128 by @razzat008 (stalled 75+ days; 2026-05-11 collaboration offer did not get a response).

Wires the `arbitrary` feature on `ironrdp-pdu` and adjusts the existing `ironrdp-connector` cascade so it composes correctly. This is the foundation for structure-aware fuzzing harnesses that generate valid-looking PDU inputs rather than raw bytes.

@razzat008's exploration in #1128 was the starting point. The decisions in this PR about the `ChannelName` manual impl shape, the `Arc<dyn LicenseCache>` trade-off, the `StaticChannelSet` skip, and the feature-cascade wiring all build on the questions raised in that thread.

## Approach

The Cargo.toml wiring uses the explicit cascade pattern:

```toml
[features]
arbitrary = ["alloc", "dep:arbitrary", "bitflags/arbitrary"]

[dependencies]
bitflags = "2.11"
arbitrary = { version = "1", features = ["derive"], optional = true }
```

This keeps `bitflags` itself free of an always-on `arbitrary` feature so consumers that do not opt into our `arbitrary` feature do not pull the `arbitrary` crate transitively. The cascade preserves the `no_std + alloc` build path the issue acceptance criterion calls for.

`#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]` is applied to PDU structs and enums across the gcc, mcs, nego, capability sets, client_info, server_license, autodetect, finalization_messages, session_info, suppress_output, refresh_rectangle, server_error_info, multitransport, headers, input, basic_output, codecs/rfx, and vc/dvc/gfx modules.

`ChannelName` gets a hand-rolled `Arbitrary` impl that preserves the wire-format invariant (8-byte null-terminated, up to seven ANSI characters). `LicenseExchangeSequence` in `ironrdp-connector` gets a hand-rolled impl with `NoopLicenseCache` as a placeholder so the license-exchange surface stays reachable under fuzz.

## Alternatives considered

- **Always-on `bitflags/arbitrary`** (as in #1128): pulls the `arbitrary` crate into every consumer of `ironrdp-pdu` unconditionally. Rejected because it defeats the optional-feature design.
- **`arbitrary` feature on `ironrdp-svc`** (as in #1128): rejected per the review comment that "we're probably not going to deal with 'arbitrary static virtual channels'." The trait-object types on the connector side are skipped via `#[arbitrary(default)]` instead.
- **Populating `StaticChannelSet` with plausible channels under fuzz**: considered but deferred. The skip path matches #1121's compile-gate scope; channel-state fuzzing coverage is a follow-up evaluated after the initial fuzz runs surface what coverage looks like.

## Trade-offs accepted

- `StaticChannelSet` fields on `ConnectionResult` and `ClientConnector` are empty under fuzz (skip via `#[arbitrary(default)]`). Connector code paths that branch on populated channel state are not exercised. Tracked for coverage iteration.
- `LicenseExchangeSequence` uses `NoopLicenseCache` under fuzz. Cache-dependent paths are not exercised. A fuzz-mode `LicenseCache` impl is a possible follow-up.
- Error enums carrying foreign types (`ServerLicenseError`, `ConnectorErrorKind`) are intentionally not annotated. They are not part of the wire-protocol surface fuzzing targets.

Each trade-off is recorded with an inline source comment naming the reason.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo check -p ironrdp-pdu --features arbitrary` succeeds
- `cargo check -p ironrdp-pdu --features arbitrary,std` succeeds
- `cargo check -p ironrdp-pdu --no-default-features --features arbitrary,alloc` succeeds
- `cargo check -p ironrdp-pdu` (default features) succeeds
- `cargo check -p ironrdp-connector --features arbitrary` succeeds
- `cargo xtask check fmt` clean
- `cargo xtask check lints` clean (pre-existing warnings unchanged)
- `cargo xtask check typos` clean

## References

- Closes #1121
- Supersedes #1128
- Part of the #1120 fuzzing umbrella; unblocks #1122, #1123, #1124


fuzz/README.md
```diff
@@ -34,3 +34,35 @@ Feeds random inputs to the RDP6 bitmap decoder.
...
+
+A handful of PDU types do not implement `Arbitrary`. They fall into two categories:
+
+- **Types with non-derivable fields** (e.g., `Arc<dyn LicenseCache>`,
+  `StaticChannelSet` keyed by `TypeId`). These are either skipped via
+  `#[arbitrary(default)]` on the containing struct's field, or hand-rolled
+  with a placeholder.
+- **Error enums carrying foreign types** (e.g., `ServerLicenseError`,
+  `ConnectorErrorKind`). These are not part of the wire-protocol surface
+  fuzzing targets and are excluded.
```
CBenoit: Generally speaking, anything that is not part of the wire-protocol surface, does not need to implement `Arbitrary`; this applies to most error types. The only exception are error PDUs parts of the wire-protocol (I think there is one in the mcs module?).

crates/ironrdp-pdu/src/rdp/server_license/mod.rs
```diff
@@ -201,6 +207,9 @@ impl BlobType {
 
 // FIXME: licensing logic and any code that is not purely about PDU
 // encoding/decoding concerns should be moved out of ironrdp-pdu.
+// Note: ServerLicenseError carries foreign error types (io::Error, FromUtf8Error,
+// x509_cert::der::Error, PduError) that do not implement Arbitrary, so this enum
+// is intentionally not annotated with #[cfg_attr(feature = "arbitrary", ...)].
```
CBenoit: I don’t think we need to annotate with a comment since the error type is not part of the wire-protocol

crates/ironrdp-pdu/src/rdp/client_info.rs
```diff
@@ -779,6 +793,7 @@ pub mod builder {
     // [`ExtendedClientOptionalInfo`] strictly requires to set all preceding optional fields before
     // setting the next one, therefore we use a state machine to enforce this during the compile time.
     #[derive(Debug, Clone, PartialEq, Eq)]
+    #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
```
CBenoit: We don’t need `Arbitrary` on a builder type, it adds zero fuzz coverage over the built type

CBenoit (commented):
I think I would drop `Arbitrary` from state machine types for now: `ClientConnector`, `ClientConnectorState`, `ConnectionActivationSequence`, `ConnectionActivationState` (connector, not PDU).
These are orchestration types, and the core problem is that both `*State::Consumed` variants are freely generated. Any fuzzer iteration landing in `Consumed` immediately gets back `"connector sequence state is finalized or consumed (this is a bug)"`. We could use `#[arbitrary(skip)]`, but I think we’re better off with waiting until we implement a good harness at which point it will be easier to design something relevant there.

glamberson:
Fixed. The bounded-collection invariant pattern recurs in two more places in this PR's cascade, so the fix is broader than just `FastPathInput`:

- **`FastPathInput`** (the named case): hand-rolled `Arbitrary` impl that picks `len = u.int_in_range(1..=255)?` then fills the Vec.
- **`ConferenceCreateRequest`** and **`ConferenceCreateResponse`**: same class of bug. Both have a `gcc_blocks.size() <= u16::MAX - ...` invariant whose violation panics via `u16::try_from(...).expect(...)` in `encode()`. Hand-rolled impls now generate the inner fields then call the existing validating `Self::new(...)` constructor, mapping the validation `Err` to `arbitrary::Error::IncorrectFormat` so the fuzzer discards size-overflowing inputs cleanly.

Other `try_from(...).expect(...)` sites in the cascade were audited and confirmed not panic-prone under derive: compile-time SIZE constants, bit-range extractions, and post-validated values do not depend on Arbitrary-generated Vec lengths.

fuzz/README.md
glamberson: OK. Reframed the bullet: principle is wire-protocol surface, not foreign-types specificity. Called out `DisconnectProviderUltimatum` as the wire-protocol error PDU exception you referenced.

crates/ironrdp-pdu/src/rdp/server_license/mod.rs
glamberson: OK. Dropped the comment.

crates/ironrdp-pdu/src/rdp/client_info.rs
glamberson: OK. Dropped the derive.

glamberson:
OK. Dropped derives on `ClientConnector`, `ClientConnectorState`, `ConnectionActivationSequence`, and `ConnectionActivationState`. Also dropped on `ConnectionResult` since it carries `connection_activation` as a field and falls in the same orchestration-output bucket. Will reconsider when #1124's state-machine oracle lands and we have a notion of valid state-sequence inputs.

CBenoit (approved):
Thank you!

CBenoit:
I opened a follow up PR: https://github.com/Devolutions/IronRDP/pull/1272

PR #1276, opened 2026-05-15
fix(server): keep newest queued waves on per-batch overflow

clintcan:
## Problem

`RdpServer::dispatch_server_events` in `crates/ironrdp-server/src/server.rs` keeps **the oldest 4** queued `RdpsndServerMessage::Wave` events per batch and drops the rest:

```rust
  let mut wave_limit = 4;
  for event in events.drain(..) {
      // ...
      RdpsndServerMessage::Wave(data, ts) => {
          if wave_limit == 0 {
              debug!("Dropping wave");
              continue;
          }
          wave_limit -= 1;
          rdpsnd.wave(data, ts)
      }
      // ...
  }
  ```

`events.drain(..)` iterates oldest-first; `wave_limit` counts down from 4. So the four waves that survive are always the four oldest in the batch, and any newer waves are dropped.

This means a one-time dispatch stall — e.g. a video encode holding the server mutex for longer than the audio sample interval — bakes itself in as a *permanent* audio-latency offset: the client gets the 4 stale waves and the fresh ones it should be playing next are gone. The intent of the limit appears to be the opposite — bound latency by discarding stale audio in favor of recent audio.

I hit this building a macOS RDP server on top of `ironrdp-server`: after any video-encode hiccup, audio playback would lag the source by  hundreds of milliseconds and never recover.

## Fix

Count the wave events up front, compute how many leading ones to skip (`total - WAVE_KEEP`), and skip those. Same `WAVE_KEEP = 4` budget — applied to the right end of the batch.

```rust
  const WAVE_KEEP: usize = 4;
  let wave_total = events
      .iter()
      .filter(|e| matches!(e, ServerEvent::Rdpsnd(RdpsndServerMessage::Wave(..))))
      .count();
  let mut wave_skip = wave_total.saturating_sub(WAVE_KEEP);
  for event in events.drain(..) {
      // ...
      RdpsndServerMessage::Wave(data, ts) => {
          if wave_skip > 0 {
              wave_skip -= 1;
              debug!("Dropping stale wave");
              continue;
          }
          rdpsnd.wave(data, ts)
      }
      // ...
  }
```

## Safety / scope

- **No behavior change in steady state.** When fewer than `WAVE_KEEP` waves are queued (the common case — dispatch keeps up with capture), `wave_skip == 0` and every wave is sent, exactly as today. 

- **Only the recovery path after a stall changes.** Instead of "keep the oldest 4, lose the rest" (permanent offset), it's now "skip the oldest, keep the newest 4" (transient gap, then back in sync).
- One extra `events.iter().filter(...).count()` per dispatch call. The dispatch is already proportional to batch size, so this is the same big-O.
- All comments and the `WAVE_KEEP = 4` value are preserved; just retargeted.

## Verification

Built and tested in the [Devolutions/IronRDP] workspace:

```
  cargo build -p ironrdp-server          # ok
  cargo clippy -p ironrdp-server         # ok (only pre-existing MSRV warning)
  cargo fmt -p ironrdp-server --check    # ok
```

End-to-end: my downstream project ran a vendored fork with exactly this patch through hundreds of mstsc / Microsoft Remote Desktop for Mac sessions during a long debugging cycle. Audio recovers cleanly from video-encode contention now; without the patch, latency grows monotonically across the session.

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -701,10 +701,22 @@ impl RdpServer {
         io_channel_id: u16,
         user_channel_id: u16,
     ) -> Result<RunState> {
-        // Avoid wave message queuing up and causing extra delays.
-        // This is a naive solution, better solutions should compute the actual delay, add IO priority, encode audio, use UDP etc.
-        // 4 frames should roughly corresponds to hundreds of ms in regular setups.
-        let mut wave_limit = 4;
+        // Avoid wave messages queuing up and causing extra delay. When a
+        // batch carries more than `WAVE_KEEP` waves, drop the OLDEST ones
+        // and keep the most recent — playing stale audio just bakes the
+        // latency in permanently, so a one-time dispatch stall (e.g. a video
+        // encode holding the server lock) would otherwise become a permanent
+        // audio offset.
+        //
+        // This is still a naive solution; better long-term: compute the
+        // actual delay, add IO priority, encode audio, use UDP, etc. 4 frames
+        // is roughly low hundreds of ms in regular setups.
+        const WAVE_KEEP: usize = 4;
+        let wave_total = events
+            .iter()
+            .filter(|e| matches!(e, ServerEvent::Rdpsnd(RdpsndServerMessage::Wave(..))))
+            .count();
+        let mut wave_skip = wave_total.saturating_sub(WAVE_KEEP);
         for event in events.drain(..) {
             trace!(?event, "Dispatching");
             match event {
@@ -725,11 +737,11 @@ impl RdpServer {
                     };
                     let msgs = match s {
                         RdpsndServerMessage::Wave(data, ts) => {
-                            if wave_limit == 0 {
-                                debug!("Dropping wave");
+                            if wave_skip > 0 {
+                                wave_skip -= 1;
+                                debug!("Dropping stale wave");
                                 continue;
                             }
-                            wave_limit -= 1;
                             rdpsnd.wave(data, ts)
                         }
                         RdpsndServerMessage::SetVolume { left, right } => rdpsnd.set_volume(left, right),


glamberson (approved):
Hi @clintcan, I'm Greg from Lamco Development. We ship `lamco-rdp-server` on top of `ironrdp-server`, so this code path is one I follow.

I'm very interested in the macOS RDP server work you mention. I've often wondered whether there's downstream infrastructure that would be instrumental to macOS development on this stack, but I haven't explored the possibility myself. If you've published anything or have a pointer to share, I'd love to learn more.

I've confirmed the bug analysis from the current master code, and I want to add a downstream-consumer perspective.

Reading `dispatch_server_events`, I see that `events.drain(..)` iterates oldest-first; `wave_limit` counts down on the first four waves, so subsequent (newer) waves hit `wave_limit == 0` and get dropped. That is the inverse of what the original comment claimed the limit was for, and it matches the bake-in-latency symptom you describe.

The fix preserves the right properties:

- **Steady-state unchanged**: when `wave_total <= WAVE_KEEP`, `wave_skip == 0` via `saturating_sub`; every wave goes through exactly as before.
- **Only the recovery path changes**: drop the oldest, keep the freshest four. This bounds latency as the comment originally promised.
- `saturating_sub` is the right defensive choice.
- The `WAVE_KEEP` constant plus the clearer comment is a nice readability bump.

This fix is useful for us as well, so I am approving.

CBenoit (approved):
Thank you! It may be possible to do something better, but that’s a net improvement and I appreciate the simplicity of the fix. LGTM

clintcan:
Hi @glamberson - yes, I've did a proof of concept just to scratch an itch of mine: control dev Macs from different rooms (useful for me). So far it's working dandy, and am even typing this from from a remotely accessed Mac in Wifi with pretty good response and egfx/h264 enabled. I've put up the code here if you want to look at it https://github.com/clintcan/macrdp. - there's another one that needs a PR but it's too clunky that I haven't made it yet, and that is file copy paste functionality (there are things that MacOS doesn't expose that I've done some ugly workarounds and tried to avoid kernel driver deps as much as possible - like the app switching).

By the way, nice RDP server you've got there. Will take a look at that for my linux boxes.

thenextman:
@clintcan The file paste thing is quite hard to get right without a virtual file system (which I think is what Windows App does). We (Devolutions) have a workaround in our main Mac client using `NSPasteboardItemProvider` to drop a temporary file on disk to serve as the paste "source" and then we're able to transfer the actual file(s) on top of it once the user actually does a paste. Let me know if you'd like more information.

glamberson:
@clintcan Nice work. I'll dig into it deeper when I can. It's nice to see something on MacOS like this. I hope you keep going with it.

Cheers.

clintcan:
@thenextman that's a good starting point.  Can you lead me how can this be done?  I currently also save on the filesystem, but I currently do eager copy (Cmd+V copy and pastes the file), as I am still looking into Mac API functions that can do this. The blocker with this is that we cannot do just lazy copy/paste using standard APIs as they are not geared for headless environments.

thenextman:
@clintcan Disclaimer that our stuff is client-side and you're obviously dealing with a server here, not sure if it makes a difference. Also, since we wrote our stuff, I note that there is now the File Provider API and FSKit which seem to provide better abstractions for this kind of thing. I have not investigated deeply, though, but it would seem to be a more modern approach.

Once we have the `FileGroupDescriptorW` (which is retrieved eagerly, of course) we create a temporary file (or files, if we're copying multiple/folders). The file is empty but we pre-size it; this matters because otherwise pasting to an external disk doesn't work (a zero-byte file gets committed before we can fill it). Then we write the temporary file URL(s) to the pasteboard, which now allows the user to Paste in Finder.app (or wherever).

The key is that each file is also a `NSFilePresenter` registered with `NSFileCoordinator`. When the paste actually happens, Finder coordinates the read and calls our `relinquishPresentedItem(toReader:)` delegate. Here we block and actually download the file bytes over the top of our temporary file. While that is happening, the user sees a "Preparing to paste" native progress dialog.

Once we have the file(s), we call the `reacquirer` callback which relinquishes control back to Finder and the paste can go ahead.

Then on top of that it's just bookkeeping to make sure we clean up temporary files when the pasteboard changes again or session disconnects or whatever.

clintcan:
@thenextman Thank you for the info, I'll take a look into this when I have the time.

Issue #1229, opened 2026-04-26
ironrdp-egfx: planned timeline for crates.io publish?


GlassOnTin:
First — thanks for the recent EGFX work, the client-side implementation in [#1103](https://github.com/Devolutions/IronRDP/pull/1103) is impressive. Surface management, capability negotiation, frame ack, ZGFX, AVC420 decode through the pluggable `H264Decoder` trait — really well-organised.

The crate currently has `publish = false # TODO: publish` in `crates/ironrdp-egfx/Cargo.toml`. Is there a planned timeline for publishing it to crates.io (presumably as part of an `ironrdp-*` 0.9.x release line)?

**Why I'm asking**: Haven (an Android RDP client built on `ironrdp-blocking` / `-connector` / `-session` 0.8) ships to users targeting modern Windows hosts (Server 2025, Win11) where the legacy bitmap path is line-by-line slow and EGFX/AVC444 is the modern fast path. Adopting the client-side EGFX work would solve that for a large fraction of our users. We could pin to `git` URLs against this repo, but mixing master-pulled `ironrdp-egfx` with crates.io-pinned `ironrdp-svc` etc. trips Cargo's same-source-multiple-versions check (the workspace cross-deps need to be patched together). A published 0.9.x line would let us adopt cleanly.

No pressure on the schedule — just trying to plan whether to wait for a release or invest in a forked-tree integration in the meantime. If there's an open tracker for the next release I should follow instead, happy to redirect.

Reported via context: [GlassHaven/Haven#109](https://github.com/GlassHaven/Haven/issues/109).

glamberson:
Adding a downstream voice. lamco-rdp-server has been carrying twelve patched ironrdp crates plus two bundled clipboard crates as path-deps for exactly this reason, and a planned lamco-rdp-wasm component is gated on the same publish landing. Independent of Haven's case, two more downstream consumers waiting to drop their patches when the batch lands.

Thanks for the kind words on the EGFX work, and full confidence in the sequencing of the publish. The dvc-com-plugin and dependency-graph constraints from #1215 are real, and untangling them in one batch is the right call rather than partial publishes that downstreams would have to special-case. Would love to see this happen expeditiously when the timing works on your end, since each week of delay is another week of patch-overrides downstream.

Anything we can help with on the dependent-crate side, happy to put hands on it.

PR #1284, opened 2026-05-18
refactor(egfx)!: mark consumer-facing types as `#[non_exhaustive]` for pre-publish API stability

glamberson:
Closes the API-stability item of #1229 (egfx pre-publish hardening). Marks consumer-facing aggregates and the top-level dispatch enum so future field/variant additions don't constitute breaking changes for downstream crates.

## Marked

- **Consumer-facing state with pub fields**: `Surface` (client + server), `BitmapUpdate`, `DecodedFrame`, `CodecCapabilities` (client + server), `FrameInfo`, `QoeMetrics`, `QoeSnapshot`.
- **Errors**: `DecoderError`.
- **Dispatch enum**: `GfxPdu` (consumers must add `_ =>` arms).

## Intentionally not marked

- **Spec-defined wire PDUs**: shape is fixed by MS-RDPEGFX; spec growth comes via new `GfxPdu` variants, not field-adds to existing PDUs.
- **Spec primitives** (`Timestamp`, `Point`, `Color`, `QuantQuality`, `Avc420Region`, `CacheEntryMetadata`, AVC bitmap streams): shape fixed by spec.
- **Spec-discriminant enums** (`CapabilitySet`, `Codec1Type`, `Codec2Type`, `PixelFormat`, `QueueDepth`): to be addressed separately via the wrapper-struct-around-integer pattern documented in `crates/ironrdp-pdu/README.md` (cf. `ironrdp_graphics::rle::Code`, `ironrdp_cliprdr::pdu::format_list::ClipboardFormatId`).
- **Tuple-struct PDU wrappers**: same logic as wire PDUs.
- **All-private-field aggregates** (`Surfaces`, `FrameTracker`, `GraphicsPipelineClient`, `GraphicsPipelineServer`): `#[non_exhaustive]` has no mechanical effect when fields are private.

## DecodedFrame::new

Adds `debug_assert_eq!` on the RGBA8888 size invariant (`data.len() == width * height * 4`) so contract-violating decoder impls surface in debug builds.

## Breaking change

External crates can no longer construct the marked structs via struct-literal syntax; exhaustive match on `GfxPdu` now requires a `_ =>` arm. Marked with `!` in the subject. Internal-to-workspace construction is unaffected.

## Local validation

`cargo xtask check fmt / lints / tests / typos / locks` — all good.

### DIFF
--- a/crates/ironrdp-egfx/src/decode.rs
+++ b/crates/ironrdp-egfx/src/decode.rs
@@ -27,6 +27,7 @@ use core::fmt;
 /// The pixel data is in RGBA format (4 bytes per pixel),
 /// row-major, top-to-bottom, left-to-right.
 #[derive(Clone)]
+#[non_exhaustive]
 pub struct DecodedFrame {
     /// RGBA pixel data (4 bytes per pixel)
     pub data: Vec<u8>,
@@ -34,6 +35,21 @@ pub struct DecodedFrame {
     pub width: u32,
     /// Frame height in pixels
     pub height: u32,
+}
+
+impl DecodedFrame {
+    #[expect(
+        clippy::as_conversions,
+        reason = "usize to u64 is lossless on all supported platforms (32/64-bit)"
+    )]
+    pub fn new(data: Vec<u8>, width: u32, height: u32) -> Self {
+        debug_assert_eq!(
+            data.len() as u64,
+            u64::from(width).saturating_mul(u64::from(height)).saturating_mul(4),
+            "DecodedFrame buffer must be RGBA8888 (width * height * 4 bytes)",
+        );
+        Self { data, width, height }
+    }
 }
 
 impl fmt::Debug for DecodedFrame {
@@ -52,6 +68,7 @@ impl fmt::Debug for DecodedFrame {
 
 /// Error type for decoder operations
 #[derive(Debug)]
+#[non_exhaustive]
 pub struct DecoderError {
     context: String,
     source: Option<Box<dyn core::error::Error + Send + Sync>>,
@@ -149,8 +166,9 @@ pub trait H264Decoder: Send {
 
 #[cfg(feature = "openh264")]
 mod openh264_impl {
-    use super::{DecodedFrame, DecoderError, DecoderResult, H264Decoder};
     use tracing::warn;
+
+    use super::{DecodedFrame, DecoderError, DecoderResult, H264Decoder};
 
     /// H.264 decoder backed by Cisco's OpenH264 library
     ///
--- a/crates/ironrdp-egfx/src/pdu/cmd.rs
+++ b/crates/ironrdp-egfx/src/pdu/cmd.rs
@@ -45,6 +45,7 @@ const RESET_GRAPHICS_PDU_SIZE: usize = 340 - GfxPdu::FIXED_PART_SIZE;
 ///
 /// INVARIANTS: size of encoded inner PDU is always less than `u32::MAX - Self::FIXED_PART_SIZE`
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum GfxPdu {
     WireToSurface1(WireToSurface1Pdu),
     WireToSurface2(WireToSurface2Pdu),
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -120,6 +120,7 @@ impl DvcEncode for ZgfxWrappedBytes {}
 /// Per MS-RDPEGFX, the server maintains an "Offscreen Surfaces ADM element"
 /// which is a list of surfaces created on the client.
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct Surface {
     /// Surface identifier (unique per session)
     pub id: u16,
@@ -229,6 +230,7 @@ impl Surfaces {
 /// Per MS-RDPEGFX, the server maintains an "Unacknowledged Frames ADM element"
 /// which tracks frames sent but not yet acknowledged.
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct FrameInfo {
     /// Frame identifier
     pub frame_id: u32,
@@ -242,6 +244,7 @@ pub struct FrameInfo {
 
 /// Quality of Experience metrics from client
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct QoeMetrics {
     /// Frame ID this relates to
     pub frame_id: u32,
@@ -423,6 +426,7 @@ impl Default for QoeCollector {
 ///
 /// Returned by [`GraphicsPipelineServer::qoe_snapshot()`].
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct QoeSnapshot {
     /// Total QoE reports received from client.
     pub total_qoe_reports: u64,
@@ -607,6 +611,7 @@ impl FrameTracker {
 
 /// Codec capabilities determined from negotiation
 #[derive(Debug, Clone, Default)]
+#[non_exhaustive]
 pub struct CodecCapabilities {
     /// AVC420 (H.264 4:2:0) is available
     pub avc420: bool,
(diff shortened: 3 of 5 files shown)


crates/ironrdp-egfx/src/server.rs
```diff
@@ -155,6 +156,7 @@ impl Surface {
 ///
 /// Implements the "Offscreen Surfaces ADM element" from MS-RDPEGFX.
 #[derive(Debug, Default)]
+#[non_exhaustive]
```
CBenoit: question: Is it really useful to use #[non_exhaustive] on a struct with private fields?

CBenoit (commented):
Thanks for the thorough write-up. I'm not convinced this should land as-is; I think the blanket application of `#[non_exhaustive]` makes the API meaningfully worse in exchange for fairly thin SemVer payoff, and the `new()` shims actively work against the stated goal.

First, where `#[non_exhaustive]` doesn't earn its cost:

- **Wire PDU structs** — `WireToSurface1Pdu`, `WireToSurface2Pdu`, `SolidFillPdu`, `SurfaceToSurfacePdu`, `SurfaceToCachePdu`, `CacheToSurfacePdu`, `EvictCacheEntryPdu`, `CreateSurfacePdu`, `DeleteSurfacePdu`, `StartFramePdu`, `EndFramePdu`, `FrameAcknowledgePdu`, `QoeFrameAcknowledgePdu`, `DeleteEncodingContextPdu`, the `MapSurfaceTo*` family, `CacheImportOfferPdu`, `CacheImportReplyPdu`, `ResetGraphicsPdu`. Their shape is fully determined by MS-RDPEGFX. You don't get to grow `CreateSurfacePdu` with a new field — if MS extends the wire format, the result is a new spec revision producing a different PDU (or a new variant of `GfxPdu`), not a silently-added struct field. Marking these `#[non_exhaustive]` denies callers struct-literal construction in exchange for forward-compat flexibility the wire format doesn't actually grant us.

- **Spec-defined value primitives** — `Timestamp`, `Point`, `Color`, `QuantQuality`, `Avc420Region`, `CacheEntryMetadata`. Same argument, more starkly: these are immutable spec primitives. `Timestamp::new(hours, minutes, seconds, milliseconds)` is *literally* the four fields the spec defines; they're not going to grow. The cost (no FRU, awkward pattern-matching) is real and the benefit is essentially zero.

- **Enums whose variants encode spec-defined discriminants** — `CapabilitySet`, `Codec1Type`, `Codec2Type`, `PixelFormat`. These currently use the `Unknown(value)` catch-all pattern, which `crates/ironrdp-pdu/README.md` discusses at length and explicitly flags as a forward-compat hazard (silent behavior change when a new variant is later added, `PartialEq` foot-gun, etc.). The README's recommended remedy is the wrapper-struct-around-integer pattern (cf. `ironrdp_graphics::rle::Code`, `ironrdp_cliprdr::pdu::format_list::ClipboardFormatId`), not `#[non_exhaustive]`. Adding `#[non_exhaustive]` here doesn't address that hazard at all — it's a library-internal silent-construction issue, not a consumer exhaustive-match issue — while still degrading match ergonomics for the named variants. If pre-publish stability for these enums is a real concern, the right move is the wrapper-struct refactor the README recommends, not slapping `#[non_exhaustive]` on the existing layout.

Also, the `new()` constructors are the bigger concern. Adding ~12 positional `pub fn new(a, b, c, d, e)` constructors is exactly the anti-pattern `#[non_exhaustive]` exists to prevent: it freezes field order and count into a *new* positional API. The day a field is added, `new()` either breaks (defeating the entire reason for `#[non_exhaustive]`) or sprouts `new_with_X` variants. They also run against STYLE.md's "no single-use helper" guidance — these exist solely to keep mechanical test-site edits compiling.

If a type genuinely is expected to evolve, the right tool is a builder, not a positional `new`. If it isn't expected to evolve, it doesn't need `#[non_exhaustive]` in the first place.

All that said, I'd keep it on the consumer-facing aggregates that genuinely grow with the implementation:

- `Surface` (client + server), `BitmapUpdate`, `DecodedFrame`, `CodecCapabilities` (client + server), `FrameInfo`, `QoeMetrics`, `QoeSnapshot`, `Surfaces`, `FrameTracker`, `GraphicsPipelineClient`, `GraphicsPipelineServer`.
- Error types: `DecoderError`.
- Arguably `GfxPdu` itself (new PDU variants are a real scenario), accepting the `_ =>` cost on consumers.

My suggestion is to narrow the scope to the handler-facing aggregates + errors (+ optionally `GfxPdu`), drop the wire-PDU and spec-primitive markings, remove the corresponding `new()` shims, and address spec-discriminant enum stability via the wrapper-struct pattern documented in `crates/ironrdp-pdu/README.md` (in a separate PR) rather than via `#[non_exhaustive]`.

Generally speaking, pre-v1 we can absolutely make breaking changes, which is not a reason to not use `#[non_exhaustive]` at all, but we can be a bit more conservative and deliberate.

CBenoit:
I also have no idea what is "F16". Is it an audit artifact you are storing locally?

glamberson:
Thanks for the careful read. I misread the common practice on `#[non_exhaustive]` and went a little crazy on the first pass; you're right on all four points. Hopefully this is a little better.

The narrowed shape keeps `#[non_exhaustive]` on the consumer-facing aggregates that genuinely grow with implementation (`Surface` client+server, `Surfaces`, `BitmapUpdate`, `DecodedFrame`, `CodecCapabilities` client+server, `FrameInfo`, `QoeMetrics`, `QoeSnapshot`, `FrameTracker`, `GraphicsPipelineClient`, `GraphicsPipelineServer`), `DecoderError`, and `GfxPdu` (accepting the `_ =>` cost on consumers). Everything else reverts: wire PDUs go back to bare struct-literal construction, spec primitives (`Timestamp`, `Point`, `Color`, `QuantQuality`, `Avc420Region`, `CacheEntryMetadata`) too, and the eleven positional `new()` shims drop along with them. Test sites revert to struct-literal construction; the one surviving constructor (`DecodedFrame::new`) stays because the type is consumer-facing and still `#[non_exhaustive]` from outside the crate.

For the spec-discriminant enums (`CapabilitySet`, `Codec1Type`, `Codec2Type`, `PixelFormat`): I'll file the wrapper-struct refactor separately, following the pattern documented in \`crates/ironrdp-pdu/README.md\` and the existing precedents in \`ironrdp_graphics::rle::Code\` and \`ironrdp_cliprdr::pdu::format_list::ClipboardFormatId\`. That belongs as its own change, not folded in here.

Force-pushing the narrowed branch shortly.

crates/ironrdp-egfx/src/pdu/cmd.rs
```diff
@@ -947,6 +993,15 @@ pub struct Timestamp {
...
 
 impl Timestamp {
+    pub fn new(hours: u16, minutes: u8, seconds: u8, milliseconds: u16) -> Self {
+        Self {
+            milliseconds,
+            seconds,
+            minutes,
+            hours,
+        }
+    }
```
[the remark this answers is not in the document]
glamberson: Moot: `Timestamp::new` is being dropped in the narrowing.

crates/ironrdp-egfx/src/server.rs
glamberson: Right, purely decorative. Dropped here and on `FrameTracker`, `GraphicsPipelineClient`, `GraphicsPipelineServer`.

crates/ironrdp-egfx/src/decode.rs
```diff
@@ -36,6 +37,12 @@ pub struct DecodedFrame {
     pub height: u32,
 }
 
+impl DecodedFrame {
+    pub fn new(data: Vec<u8>, width: u32, height: u32) -> Self {
+        Self { data, width, height }
+    }
```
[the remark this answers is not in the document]
glamberson: Added `debug_assert_eq!` for the RGBA8888 size invariant in `DecodedFrame::new`.

glamberson:
F16 is the egfx pre-publish hardening finding from a workspace API hygiene audit I keep in a private tracking repo. Internal shorthand that slipped into the PR body; won't surface in upstream prose again.

crates/ironrdp-egfx/src/decode.rs
```diff
@@ -36,6 +37,21 @@ pub struct DecodedFrame {
     pub height: u32,
```
CBenoit, line 37: thought: I think there is an invariant here, where `data.len() == width * height *  4`, but since all the fields are public we can’t guarantee the invariant is uphold correctly. The `new()` function is has a `debug_assert_eq!` to verify this property in a loose way.

CBenoit (approved):
LGTM

crates/ironrdp-egfx/src/decode.rs
glamberson: ack: the `debug_assert_eq!` in `new()` only catches the constructor path; `pub` fields leave the invariant open to post-construction mutation. Will file a follow-up making `data`/`width`/`height` private with getters (`data() -> &[u8]`, `width() -> u32`, `height() -> u32`, plus `into_data(self) -> Vec<u8>`) so the invariant is type-enforced. Will cross-link this thread on filing.

PR #1291, opened 2026-05-20
test(fuzz): add pdu_round_trip oracle and target

glamberson:
## Summary

Adds a fuzz oracle and dedicated target for round-trip exercise of PDU
encode/decode pathways. For each PDU type currently covered by
`pdu_decode`, the oracle attempts `decode` -> `encode_vec` -> re-`decode`
on the encoded bytes. Auto-discovers into CI per #1290.

## Property tested

*No internal panic from inside the encoder or decoder when fed a
decoder-accepted input through both directions of the round-trip.*

Asymmetric `Err` returns (decoder accepts something the encoder reports
as `"Encoding not implemented"`, or vice-versa) are NOT in scope here:
those are tolerated incomplete-impl cases tracked separately. The
oracle silently drops `Err` results from any of the three stages.

What this catches:

- `unreachable!()` reached during encoding of a valid decoded state
  (i.e. the encoder's match arms are missing a variant the decoder
  produces).
- Integer overflow / index-out-of-bounds inside the encoder on
  decoder-accepted inputs.
- Panics in the decoder when fed encoder-produced bytes (re-decode
  path).

## Initial type coverage

Mirrors `pdu_decode`'s type list minus two exclusions documented inline
with their reproducers. As new PDU types gain `Encode` impls, they
extend coverage when added to the macro list.

## Excluded pending follow-up

Two types are temporarily excluded because they hit `unreachable!()` in
the encoder during smoke fuzzing, which is an internal-panic bug that
the oracle cannot silently drop:

- `capability_sets::CapabilitySet`: the encoder's match at
  `crates/ironrdp-pdu/src/rdp/capability_sets/mod.rs:447` reaches
  `unreachable!()` on a variant the decoder accepts but the encoder's
  match arms don't cover. Smoke-fuzz reproducer: `[6, 0, 4, 0]`.
- `headers::ShareControlHeader`: transits through `CapabilitySet`'s
  encoder for `DemandActive` / `ConfirmActive` PDU contents and hits
  the same internal panic.

Both inclusions return once the encoder's match arms are completed. I
will file the underlying bug as a separate issue and link it back here.

## Files

- `crates/ironrdp-fuzzing/src/oracles/mod.rs`: new `pdu_round_trip`
  function and `pdu_round_trip_one!` helper macro.
- `fuzz/fuzz_targets/pdu_round_trip.rs`: new fuzz target binary.
- `fuzz/Cargo.toml`: `[[bin]]` entry for the new target.
- `crates/ironrdp-testsuite-core/tests/fuzz_regression.rs`: new
  `check_pdu_round_trip` regression-replay test.
- `crates/ironrdp-testsuite-core/test_data/fuzz_regression/pdu_round_trip/seed-empty.bin`:
  empty seed file for the regression-replay test.

## Verification

- `cargo xtask check tests -v` green (including
  `check_pdu_round_trip`)
- 60-second smoke fuzz: 2,220,481 runs, no crashes on the current
  type set
- New target auto-discovered by `cargo xtask fuzz list` (per #1290)

Refs #1120 (umbrella) and #1124 (oracle infrastructure).

### DIFF
--- a/crates/ironrdp-fuzzing/src/oracles/mod.rs
+++ b/crates/ironrdp-fuzzing/src/oracles/mod.rs
@@ -180,6 +180,127 @@ pub fn pdu_decode(data: &[u8]) {
     let _ = decode::<ironrdp_rdpsnd::pdu::ClientAudioOutputPdu>(data);
 }
 
+/// Helper for [`pdu_round_trip`].
+///
+/// Exercises `decode` → `encode_vec` → re-`decode`, silently dropping `Err`
+/// results from any stage. The oracle's value is in detecting INTERNAL
+/// panics from inside the encoder/decoder (e.g., `unreachable!()` reached
+/// on a valid decoded state), not in asserting Err-result symmetry. Many
+/// `ironrdp-pdu` types have known asymmetric `Encode` impls that return
+/// `"Encoding not implemented"` for variants the decoder still accepts;
+/// those are tracked separately and not in scope for this oracle.
+macro_rules! pdu_round_trip_one {
+    ($data:expr, $ty:ty) => {{
+        if let Ok(pdu) = ironrdp_core::decode::<$ty>($data) {
+            if let Ok(encoded) = ironrdp_core::encode_vec(&pdu) {
+                let _ = ironrdp_core::decode::<$ty>(&encoded);
+            }
+        }
+    }};
+}
+
+/// Round-trip oracle: for each PDU type, exercise the
+/// `decode` → `encode_vec` → re-`decode` pipeline.
+///
+/// The property tested is *no internal panic from inside the encoder or
+/// decoder when fed a decoder-accepted input through both directions of the
+/// round-trip*. Asymmetric `Err` returns (decoder accepts something the
+/// encoder reports as `"Encoding not implemented"`, or vice-versa) are not
+/// in scope: those are tolerated incomplete-impl cases tracked separately.
+///
+/// What this catches:
+///
+/// - `unreachable!()` reached during encoding of a valid decoded state (i.e.
+///   the encoder's match arms are missing a variant the decoder produces).
+/// - Integer overflow / index-out-of-bounds inside the encoder on
+///   decoder-accepted inputs.
+/// - Panics in the decoder when fed encoder-produced bytes (re-decode path).
+///
+/// What this does NOT catch:
+///
+/// - Encode returning `Err`. Many PDU types intentionally return errors for
+///   partially-implemented variants; exercising them is the encoder
+///   developer's responsibility, not this oracle's.
+/// - Re-decode returning `Err`. Surfaces an asymmetry but not a memory-safety
+///   bug; tracked via filed follow-up issues, not this oracle.
+///
+/// Initial type coverage mirrors `pdu_decode` so the same corpus feeds both
+/// oracles. As new PDU types gain `Encode` impls, they auto-extend coverage
+/// here when added to the macro list below.
+pub fn pdu_round_trip(data: &[u8]) {
+    use ironrdp_pdu::mcs::{ConnectInitial, ConnectResponse, McsMessage};
+    use ironrdp_pdu::nego::{ConnectionConfirm, ConnectionRequest};
+    use ironrdp_pdu::rdp::{ClientInfoPdu, server_error_info, server_license, vc};
+    use ironrdp_pdu::x224::X224;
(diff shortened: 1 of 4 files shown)


glamberson:
Filed #1292 for the underlying `CapabilitySet::BitmapCacheV3` encoder gap. Once that lands, this PR's two excluded types (`capability_sets::CapabilitySet` and `headers::ShareControlHeader`) can return to the round-trip coverage list.

crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -180,6 +180,127 @@ pub fn pdu_decode(data: &[u8]) {
...
+/// those are tracked separately and not in scope for this oracle.
+macro_rules! pdu_round_trip_one {
+    ($data:expr, $ty:ty) => {{
+        if let Ok(pdu) = ironrdp_core::decode::<$ty>($data) {
+            if let Ok(encoded) = ironrdp_core::encode_vec(&pdu) {
+                let _ = ironrdp_core::decode::<$ty>(&encoded);
+            }
+        }
+    }};
+}
```
CBenoit, line 200: It looks to me that this does not need to be a macro at all, a generic function could do the same thing

CBenoit (commented):
Thank you!

Small remark: I wonder if this is rendering useless some other harness such as `pdu_decode`, because we are exercising the decode path here too

crates/ironrdp-fuzzing/src/oracles/mod.rs
glamberson: Looked into this; the borrowing-type cohort blocks the migration. Trait shape is `Decode<'de>`, so the natural form is:

```rust
fn pdu_round_trip_one<'de, T: Decode<'de> + Encode>(data: &'de [u8]) {
    if let Ok(pdu) = decode::<T>(data) {
        if let Ok(encoded) = encode_vec(&pdu) {
            let _ = decode::<T>(&encoded);  // 'encoded < 'de
        }
    }
}
```

That fails on the second `decode::<T>(&encoded)`: `T: Decode<'de>` needs input of lifetime `'de` (the original `data`), but `&encoded` is function-local and shorter. HRTB `T: for<'de> Decode<'de>` lifts that constraint, but only for types that don't themselves borrow. 10 of the 27 types in the list do borrow: `X224<McsMessage<'_>>`, `FastPathUpdatePdu<'_>`, the `surface_commands::*<'_>` cluster, `codecs::rfx::Block<'_>`, `cliprdr::pdu::ClipboardPdu<'_>`, `rdpsnd::pdu::ServerAudioOutputPdu<'_>`, plus a few more. Each instantiation is tied to one specific `'de`, so they can't satisfy the HRTB.

The macro works because each `'_` resolves to a fresh inference variable per call site (first binds the data's lifetime, second binds the encoded vec's). Partial migration (generic fn for the 17 owned types, macro for the 10 borrowing) would compile but splits the implementation without semantic gain.

If you see a shape I'm missing, let me know.

glamberson:
There's real overlap. `pdu_round_trip`'s coverage is technically a superset since its first call exercises the same decoder-panic surface `pdu_decode` targets. But `pdu_decode` is still worth keeping for three operational reasons:

1. **Per-iteration cost.** `pdu_decode` does one decode per input; `pdu_round_trip` does up to three (decode + encode + re-decode). In a fixed-time fuzz run, `pdu_decode` covers ~3x more inputs against the same decoder surface, so decoder-only bugs surface faster there.

2. **Corpus separation.** libFuzzer's corpus distillation runs per-target. `pdu_decode`'s corpus minimizes around decoder-panic-inducing inputs; `pdu_round_trip`'s minimizes around round-trip-fragile inputs. Same fuzzer entropy, different distillation outcomes.

3. **Bug isolation.** A `pdu_decode` crash localizes to the decoder by construction. A `pdu_round_trip` crash could be in decoder, encoder, or re-decoder; you have to inspect the backtrace. Separate targets keep the first hop of triage cheap.

The type lists also differ slightly: `pdu_round_trip` excludes `capability_sets::CapabilitySet` and `headers::ShareControlHeader` pending #1292's `BitmapCacheV3` fix, while `pdu_decode` keeps them.

This matches the existing workspace pattern too: `rle_decompress_bitmap`, `bitmap_stream`, and the codec-specific oracles all share some decoder-path coverage with `pdu_decode` but exist as separate targets for the same operational reasons.

CBenoit (approved):
LGTM!

PR #1298, opened 2026-05-22
fix(egfx): tolerate unknown capability versions instead of failing decode

clintcan:
`CapabilitySet::decode` propagated `CapabilityVersion::try_from`'s error via `?`, so a single unrecognized capability version aborted decoding of the entire CapabilitiesAdvertise PDU during EGFX negotiation.

Some clients advertise a capset version outside the enumerated set — notably the macOS "Windows App" / Microsoft Remote Desktop client — which made the EGFX channel fail to negotiate and could prevent the connection from being established at all when the server enables EGFX.

The `CapabilitySet::Unknown(Vec<u8>)` variant already exists and round-trips raw bytes through `encode`/`size`. This routes the unrecognized-version case into it instead of erroring, so negotiation completes and the server can still select a mutually supported capability version.

Verified against a real macOS Windows App client connecting to an EGFX/AVC420 server: without this change the connection aborts during negotiation; with it, negotiation completes and H.264 renders.

### DIFF
--- a/crates/ironrdp-egfx/src/pdu/cmd.rs
+++ b/crates/ironrdp-egfx/src/pdu/cmd.rs
@@ -1593,11 +1593,24 @@ impl<'de> Decode<'de> for CapabilitySet {
     fn decode(src: &mut ReadCursor<'de>) -> DecodeResult<Self> {
         ensure_fixed_part_size!(in: src);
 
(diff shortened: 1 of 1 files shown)


glamberson:
Yeah, an obvious win,

CBenoit (approved):
Thank you!

PR #1301, opened 2026-05-23
feat(cliprdr): advertise Preferred DropEffect alongside FileGroupDescriptorW

clintcan:
`initiate_file_copy` now advertises **both** `FileGroupDescriptorW` and `Preferred DropEffect` (`CFSTR_PREFERREDDROPEFFECT`) in the FormatList, and `handle_format_data_request` short-circuits a request for the latter with `DROPEFFECT_COPY` (0x00000001 LE).

### Changes

- New constant: `ClipboardFormatName::PREFERRED_DROP_EFFECT = "Preferred DropEffect"`.
- New field: `Cliprdr::local_drop_effect_format_id` — mirrors `local_file_list_format_id` in lifecycle (set in `initiate_file_copy`, cleared on `FormatListResponse::Fail` and at the top of `initiate_copy`).
- Distinct private-range format ID (`0xC0FD`) — the wire only cares about the format *name*; the ID just needs to be locally unique so the request can be routed.
- No backend trait changes — the response is fixed (`DROPEFFECT_COPY`, matching the function's name) and served inline, so backends don't have to learn about the format.

### Why

Windows Explorer pairs `CFSTR_PREFERREDDROPEFFECT` with the file formats locally; together they label the operation as a *copy*. When both are present, Explorer engages its shell file-copy machinery on paste (with native progress UI). Without `PreferredDropEffect`, Explorer falls back to a plain synchronous `IStream` read.

### Use case / verification

Used in [macrdp](https://github.com/clintcan/macrdp) (a Rust RDP server for macOS) to deliver the standard "copy a file from the remote into local Windows Explorer" UX. Verified end-to-end on real mstsc.

The companion `FD_PROGRESSUI` flag (#1299) is what actually triggers the native progress dialog; `PreferredDropEffect` provides the operation-semantics label Windows expects to see alongside the file descriptor for a clipboard file paste. Either change is independently useful, but the user-visible "native progress dialog" UX requires both.

### Related

- #1299 — feat(cliprdr): always set FD_PROGRESSUI in FileDescriptor::encode
- #1300 — feat(cliprdr): add CliprdrBackend::on_format_list_response(ok) hook

### DIFF
--- a/crates/ironrdp-cliprdr/src/lib.rs
+++ b/crates/ironrdp-cliprdr/src/lib.rs
@@ -387,6 +387,13 @@ pub struct Cliprdr<R: Role> {
     /// Tracked so we can recognize FormatDataRequest for our file list.
     local_file_list_format_id: Option<ClipboardFormatId>,
 
+    /// Format ID used for the local "Preferred DropEffect" entry in the
+    /// FormatList sent alongside FileGroupDescriptorW from
+    /// [`Cliprdr::initiate_file_copy`]. Tracked so we can recognize a
+    /// FormatDataRequest for it and respond inline with `DROPEFFECT_COPY`
+    /// (0x00000001) — backends don't have to know about the format.
+    local_drop_effect_format_id: Option<ClipboardFormatId>,
+
     /// Stores the remote file list after receiving it via FormatDataResponse.
     /// Used for validating FileContentsRequest.lindex bounds.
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
LGTM! I like the approach. Copilot suggested to add a test, that is fair, but this can also be part of a follow up PR 🙂

Issue #667, opened 2025-02-10
Scancode map is not working for IronRdp Web when connect using Blink


irvingoujAtDevolution:
When user agent is `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)` the scanCode map fails to convert to the correct scanCode.

mattieb:
This may also be a problem with Safari—I've found that with devolutions-gateway standalone 2025.1.4, using Safari, no keypresses are mapped to anything that makes any sense. Same setup with Firefox works fine.

User agent is Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.4 Safari/605.1.15 if it matters.

CBenoit:
@irvingoujAtDevolution @mattieb This may be fixed on master. Do you think you could try again? Thank you!

mattieb:
> This may be fixed on master. Do you think you could try again? Thank you!

Hi @CBenoit—just wanted to let you know I don’t have that setup anymore. Best of luck to y’all!

PR #1304, opened 2026-05-25
fix(web): include Meta keys in WebKit scancode dispatch

eazelis:
## Symptom

On macOS WebKit (Safari and any WKWebView embedder), pressing Cmd or Cmd+letter inside an `iron-remote-desktop` session isn't delivered to the remote. Cmd alone does nothing on the Windows host; Cmd+R selects the first desktop icon starting with "R" instead of opening the Run dialog — `R` arrives without the Win modifier held.

## Root cause

In `web-client/iron-remote-desktop/src/lib/scancodes.ts`, `KeyCodeToScanCode.webkit` only loaded the base `scanCodeToKeyCode` table. Blink and Gecko each extend the base with entries that carry `MetaLeft (0xE05B)` / `MetaRight (0xE05C)`; WebKit was left out. So `scanCode("MetaLeft")` returned `NaN` on WebKit and the keydown was silently dropped in `sendKeyboard`. KeyR arrived alone, hence the desktop-icon-jump behavior.

## Fix

Extend the WebKit branch with the same extras Blink already uses. The entries (`MetaLeft`/`MetaRight`, `AudioVolume*`, `Lang3`/`4`, etc.) are [W3C UI Events Code spec](https://www.w3.org/TR/uievents-code/) values; WebKit emits them per spec, same as Blink.

Verified manually on macOS WebKit:

- `Cmd` alone → Start menu opens
- `Cmd+R` → Run dialog
- `Cmd+E` → File Explorer
- `Cmd+D` → Show Desktop

## Secondary rename

While in the file, renamed two tables to better reflect their semantics:

- `codeToScanCodeBlinkOverride` → `scanCodeToKeyCodeExtras` — the old name was a double misnomer (table direction is `scanCode → keyCode`, not `code → scanCode`; entries are W3C-spec codes shared by Blink and WebKit, not Blink-private).
- `scanCodeToKeyCodeGeckoOverride` → `scanCodeToKeyCodeGeckoExtras` — both tables are purely additive against the base (verified: 0 scancode collisions).

Happy to drop the rename if you'd prefer the PR strictly to the bug fix.

## Related

Likely addresses #667 — that issue reports a Safari user-agent (`AppleWebKit/605.1.15`) but its title says "Blink", which has probably kept it from being triaged.

### DIFF
--- a/web-client/iron-remote-desktop/src/lib/scancodes.ts
+++ b/web-client/iron-remote-desktop/src/lib/scancodes.ts
@@ -157,7 +157,7 @@ const scanCodeToKeyCode = {
     '0xE06D': 'MediaSelect',
 };
 
-const codeToScanCodeBlinkOverride = {
+const scanCodeToKeyCodeExtras = {
     '0x0077': 'Lang4',
     '0x0078': 'Lang3',
     '0xE008': 'Undo',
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM, thank you!

PR #1308, opened 2026-05-25
test(cliprdr): cover Preferred DropEffect advertise + inline DROPEFFECT_COPY response

clintcan:
## Summary

Adds the regression test requested by the [Copilot review on #1301](https://github.com/Devolutions/IronRDP/pull/1301#discussion_r2169280717) (the PR merged without test coverage for the new behavior).

Two cases:

1. `initiate_file_copy` advertises both \`FileGroupDescriptorW\` **and** \`Preferred DropEffect\` in the outgoing \`FormatList\`.
2. A \`FormatDataRequest\` for the drop-effect format id is answered inline by \`Cliprdr\` (not forwarded to the backend) with exactly 4 bytes \`0x01 0x00 0x00 0x00\` (\`DROPEFFECT_COPY\`, u32 LE).

Test #2 keys off the format **name** (\`ClipboardFormatName::PREFERRED_DROP_EFFECT\`) rather than the hardcoded \`0xC0FD\` id — wire-faithful (the remote keys off the name too), and resilient if the internal id constant ever shifts.

Test uses the existing \`init_ready_client()\` helper from \`test_helpers.rs\` (which advertises \`STREAM_FILECLIP_ENABLED\`, required for \`initiate_file_copy\` to succeed).

## Test plan
- [x] Both new tests pass: \`cargo test -p ironrdp-testsuite-core --test integration_tests_core preferred_drop_effect\`
- [x] The wider clipboard suite (175 tests) is unaffected: \`cargo test -p ironrdp-testsuite-core clipboard\` → 175 passed

### DIFF
--- a/crates/ironrdp-testsuite-core/tests/clipboard/preferred_drop_effect.rs
+++ b/crates/ironrdp-testsuite-core/tests/clipboard/preferred_drop_effect.rs
@@ -42,7 +42,11 @@ fn initiate_file_copy_advertises_drop_effect_alongside_file_group_descriptor() {
     ];
     let messages: Vec<SvcMessage> = cliprdr.initiate_file_copy(files).unwrap().into();
 
-    assert_eq!(messages.len(), 1, "initiate_file_copy should send a single FormatList PDU");
+    assert_eq!(
+        messages.len(),
+        1,
+        "initiate_file_copy should send a single FormatList PDU"
+    );
 
     decode_pdu!(&messages[0] => bytes, pdu);
     let ClipboardPdu::FormatList(format_list) = pdu else {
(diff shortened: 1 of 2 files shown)


crates/ironrdp-testsuite-core/tests/clipboard/preferred_drop_effect.rs
```diff
@@ -0,0 +1,141 @@
+//! Regression tests for the `Preferred DropEffect` companion format
+//! advertised alongside `FileGroupDescriptorW` by [`Cliprdr::initiate_file_copy`].
+//!
+//! Follows up on [PR #1301]: the merge landed without test coverage and
+//! the Copilot review flagged this. The two behaviors covered here:
```
CBenoit: This is not information that is relevant to track in the source tree

CBenoit (commented):
Thank you!

clintcan:
Addressed CBenoit's feedback in [\`f1e7e6b\`](https://github.com/Devolutions/IronRDP/commit/f1e7e6b) — module doc-comment no longer mentions PR/Copilot history; just describes what the tests cover.

The two Copilot comments at line 99 about a redundant first \`initiate_file_copy\` call were already addressed earlier in [\`bbdb5ba\`](https://github.com/Devolutions/IronRDP/commit/bbdb5ba) — the test now decodes the \`FormatList\` from a single call instead of calling \`initiate_file_copy\` twice. Looks like Copilot reviewed the older revision.

crates/ironrdp-testsuite-core/tests/clipboard/preferred_drop_effect.rs
clintcan: Removed this particular comment.

CBenoit (approved):
LGTM! Thank you!!

PR #1313, opened 2026-05-26
fix(pdu): cover BitmapCacheV3 in CapabilitySet encoder

### LINKED ISSUE #1292: pdu: CapabilitySet::BitmapCacheV3 encoder reaches unreachable!() (decoded variant not covered in Encode impl)
## Summary

`CapabilitySet::BitmapCacheV3(Vec<u8>)` reaches `unreachable!()` in the encoder when the decoder has accepted a `BitmapCacheV3` capability. The decoder admits the variant, the `size()` impl covers it, but the `Encode` impl's match arms do not, so the catch-all at `crates/ironrdp-pdu/src/rdp/capability_sets/mod.rs:447` fires.

## Reproducer

4-byte input: `[6, 0, 4, 0]`.

This decodes as a single `CapabilitySet` with type `0x0006` (`BitmapCacheV3CodecID`) and length `0x0004`, producing `CapabilitySet::BitmapCacheV3(vec![])`. Re-encoding panics:

```
thread '<unnamed>' panicked at crates/ironrdp-pdu/src/rdp/capability_sets/mod.rs:447:26:
internal error: entered unreachable code
```

Surfaced by the `pdu_round_trip` fuzz oracle being added in #1291 (60-second smoke fuzz, 2.2M iterations).

## Where it is in the code

The `Encode` impl on `CapabilitySet` (around line 280 onward) explicitly handles each "mandatory" and "optional" variant with its own arm, then ends in a catch-all `_ => { ... }` arm at roughly line 432 that handles the "other" `Vec<u8>` carrying variants via a second inner match (lines 433-447). The inner match covers:

- `Control`
- `WindowActivation`
- `Share`
- `Font`
- `BitmapCacheHostSupport`
- `DesktopComposition`
- `ColorCache`
- `DrawNineGridCache`
- `DrawGdiPlus`
- `Rail`
- `WindowList`

then `_ => unreachable!()`.

`CapabilitySet::BitmapCacheV3(Vec<u8>)` is declared in the enum (last variant in the "other" group), is decoded by the `BitmapCacheV3CodecID` arm in the decode dispatch, and is covered in the `size()` impl. It is not in either the outer match's explicit arms or the inner match's explicit arms, so it falls through to `unreachable!()`.

## Suggested fix

Add a `CapabilitySet::BitmapCacheV3(buffer) => (CapabilitySetType::BitmapCacheV3CodecID, buffer),` arm to the inner match in the encoder. The encode shape mirrors the other `Vec<u8>`-carrying variants in that match: write the type as u16, write the length as u16, write the buffer slice.

## Impact

Any consumer that decodes RDP capability-exchange traffic containing a `BitmapCacheV3` capability and re-encodes it (proxy, gateway, replay, fuzz round-trip) hits the panic. `BitmapCacheV3` is rare on the wire but not impossible (some Windows servers advertise it).

`pdu_round_trip` in #1291 temporarily excludes `capability_sets::CapabilitySet` and the higher-level `headers::ShareControlHeader` (which transits through this encoder for `DemandActive` / `ConfirmActive` PDUs) until this is resolved. Once fixed, both can return to the oracle's coverage list.

## References

- Surfaced by: #1291 (`feat(fuzz): add pdu_round_trip oracle and target`)
- Refs umbrella: #1120

glamberson:
`CapabilitySet::BitmapCacheV3(Vec<u8>)` is in the enum, is decoded by
the `BitmapCacheV3CodecID` arm, and is covered in `size()`, but was
missing from both the outer and inner `match` of the `Encode` impl, so
re-encoding a decoded `BitmapCacheV3` capability set hit the catch-all
`_ => unreachable!()` and panicked in release builds.

Add the missing arm to the inner match so the type follows the same
raw-buffer encode path as the other unparsed capability sets.

The `pdu_round_trip` fuzz oracle had previously excluded
`capability_sets::CapabilitySet` and `headers::ShareControlHeader` for
exactly this reason (`ShareControlHeader` transits through
`CapabilitySet`'s encoder via the Active variants). With the encoder
covered, both types re-join the oracle's coverage list and the
4-byte reproducer `[0x06, 0x00, 0x04, 0x00]` is added to the regression
corpus so the fix is locked in by `check_pdu_round_trip` going forward.

Regression test `bitmap_cache_v3_round_trip_does_not_panic` added in
`crates/ironrdp-testsuite-core/tests/pdu/rdp.rs` exercises decode ->
encode round-trip on the same input and asserts byte-equality with the
original. Placed in testsuite-core per the project convention that PDU
crates stay test-light and round-trip coverage lives in
`ironrdp-testsuite-core/tests/<protocol>/`.

Test plan: `cargo xtask check fmt/lints/tests/typos/locks` all pass.

Closes #1292.


CBenoit (approved):
LGTM

PR #1319, opened 2026-05-26
feat(server): handle SuppressOutput / RefreshRectangle and expose state

clintcan:
## Summary

The server currently logs `SuppressOutput` and `RefreshRectangle` as `Unexpected share data pdu` and drops them. That leaves a server streaming high-bitrate EGFX/H.264 frames into a minimized client which accumulates them; on refocus the client must chew through the backlog before it can present the current frame, locking up its input dispatch for seconds.

This PR pattern-matches the two PDUs and flips a shared `Arc<AtomicBool>` so display backends can hold a clone and skip frame emission while the client has signaled it doesn't need updates.

- mstsc sends \`SuppressOutput { desktop_rect: None }\` on minimize.
- mstsc sends \`SuppressOutput { Some(rect) }\` and/or \`RefreshRectangle\` on refocus.

## API

Two new methods on \`RdpServer\`:

- \`display_suppressed_handle() -> Arc<AtomicBool>\` returns a clone of the shared flag so a backend can read it (typical pattern: backend spawns a thread that polls and skips encode/ship while the flag is \`true\`).
- \`set_display_suppressed_handle(handle: Arc<AtomicBool>)\` replaces the internally-created flag with one the caller already shared with the display backend before constructing the server. Necessary because the backend is moved into the builder before the server exists; without a setter there's no way to give the same \`Arc\` to both ends.

## Backward compatibility

- \`RdpServer::new()\` signature unchanged (new field initialized internally).
- Builder unchanged.
- For existing users who don't call the new methods: zero behavior change. Only the log for these two PDUs drops from \`warn\` to \`debug\` (Unexpected → handled-and-logged).
- All existing tests pass.

## Real-world result

Verified on macOS / mstsc with H.264 over EGFX: minimizing mstsc for several minutes used to leave the desktop frozen and unresponsive for several seconds on refocus while it processed the buffered frames. With the flag honored downstream, refocus is instant. Shipped in macrdp v0.5.47.

## Test plan

- [x] \`cargo build -p ironrdp-server --features egfx\`
- [x] \`cargo clippy -p ironrdp-server --features egfx --all-targets -- -D warnings\`
- [x] \`cargo test -p ironrdp-server --features egfx\` (5/5 pass)
- [x] \`rustup run nightly cargo fmt --all -- --check\`
- [x] Real-world verification on mstsc + macrdp (long minimize + refocus, disconnect/reconnect, cargo-build-under-load with YouTube)

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -338,9 +338,15 @@ enum RunState {
 }
 
 impl RdpServer {
-    #[expect(
-        clippy::too_many_arguments,
-        reason = "called via the builder; positional parameters are an internal detail"
+    // The lint only fires with the `egfx` feature on (8 args including
+    // `gfx_factory`); without it the parameter count is 7 and the lint
+    // is satisfied. `cfg_attr` keeps `#[expect]` strict in both modes.
+    #[cfg_attr(
+        feature = "egfx",
+        expect(
+            clippy::too_many_arguments,
+            reason = "called via the builder; positional parameters are an internal detail"
+        )
     )]
     pub fn new(
         opts: RdpServerOptions,
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -1,4 +1,5 @@
 use core::net::SocketAddr;
+use core::sync::atomic::{AtomicBool, Ordering};
 use core::time::Duration;
 use std::rc::Rc;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


crates/ironrdp-server/src/server.rs
```diff
@@ -374,6 +388,45 @@ impl RdpServer {
...
+    ///
+    /// Use this when the display backend needs to read the same flag the
+    /// per-connection PDU handler writes to: create one `Arc<AtomicBool>`
+    /// in the application, hand a clone to the display, then call this to
+    /// swap the server's internal default for that shared instance. Must
+    /// be called before any client connects (the flag is read by every
+    /// per-connection task; replacing it mid-session is racy).
+    pub fn set_display_suppressed_handle(&mut self, handle: Arc<AtomicBool>) {
+        self.display_suppressed = handle;
+    }
```
CBenoit: It’s acknowledged in the function’s docstring, but I don’t think it’s a great API because it’s easy to misuse.

I would suggest to take one of the following approaches:

- Remove this method from RdpServer, and instead add a new method to RdpServerBuilder (it’s impossible for a session to be started before RdpServer is built; my preferred approach)
- Add a runtime check, either with `assert!` or return a `Result::Err` when this method is called while a session is running.

crates/ironrdp-server/src/server.rs
```diff
@@ -338,6 +338,10 @@ enum RunState {
 }
 
 impl RdpServer {
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "called via the builder; positional parameters are an internal detail"
+    )]
     pub fn new(
```
CBenoit, line 345: thought: I agree with the idea here, but `new` is currently `pub`, so we need to change that to `pub(crate)` in a follow up PR for the point to hold

CBenoit (approved):
LGTM! Thank you

PR #1320, opened 2026-05-26
feat(fuzz): add message_decoding_invariants oracle and target

### LINKED ISSUE #1124: Add advanced fuzz targets to verify more invariants (round-trip, framing, state-machine…)

glamberson:
This PR adds an assertion-based oracle that checks the `Encode::size()`
soundness contract on every decoder-accepted input. The asserted
property is `pdu.size() == encode_vec(&pdu).len()`.

The `Encode` trait's `size()` is the contract caller code relies on for
buffer sizing (`ensure_size!`, `cast_length!`, downstream allocations).
A size that lies about its own length produces under-allocation
(truncated encode) or over-allocation (wasted memory and
buffer-overflow risk depending on surrounding context). The workspace
already asserts this property against known-good byte fixtures in
`ironrdp-testsuite-core/tests/pdu/rdp.rs::buffer_length_is_correct_for_*`.
This oracle extends the assertion across the fuzz input surface.

This oracle differs from `pdu_round_trip` in two ways:

- `pdu_round_trip` silently drops `Err` results. It catches panics
  only.
- `message_decoding_invariants` asserts the size contract. A violation
  aborts the fuzz iteration as a libFuzzer crash. The bug-reporting
  mechanism follows the oracle module's documentation.

This oracle catches the following bug classes that `pdu_round_trip`
does not:

- `Encode::size()` returns N but `encode` writes M, where M does not
  equal N. Callers then over-allocate or under-allocate buffers.
- The encoder cannot reconstruct decode-acceptable bytes to the same
  length, which means decode loses framing structure.

This oracle does not catch the following bug classes, which other
oracles cover:

- `pdu_decode` covers decode-time panics and OOM.
- `pdu_round_trip` covers re-decode equality after round-trip via its
  silent-drop pattern. Assertion-based re-decode stays out of scope
  here to keep this oracle's failure mode unambiguous.

Type coverage matches `pdu_round_trip`. The cohort includes
`capability_sets::CapabilitySet` and `headers::ShareControlHeader`.
The latter transits through `CapabilitySet`'s encoder via the Active
variants. The same encoder path exercises both types.

The new target auto-discovers into CI via the `cargo xtask fuzz list`
dynamic fan-out mechanism.

The test plan covers two checks. First, all five
`cargo xtask check fmt/lints/tests/typos/locks` gates pass on the
post-rebase head. Second, the new `check_message_decoding_invariants`
regression-replay test in
`crates/ironrdp-testsuite-core/tests/fuzz_regression.rs` passes
against the seed corpus.

This PR refs the #1120 fuzzing umbrella. The PR also closes #1124.

## Rebase 2026-08-19

Rebased onto current master (was CONFLICTING after the RDPEUDP/RDPEMT series landed). Two conflicts, both purely additive insertions colliding with unrelated new master content at the same point:

- `fuzz/Cargo.toml`: master's 8 new RDPEUDP/RDPEMT `[[bin]]` entries plus this PR's own `message_decoding_invariants` entry, both kept.
- `crates/ironrdp-fuzzing/src/oracles/mod.rs`: master's 8 new oracle functions plus this PR's own, both appended. The automatic merge dropped master's closing brace for its last function (`rdpeudp_ack_vector`'s oracle), the same failure mode already seen twice this session on sibling fuzz PRs rebasing past the same master content. Caught before continuing the rebase by diffing the resolved file against both source commits directly; fixed and re-verified byte-identical to both.

All checks re-run on the new head: `cargo xtask check fmt/lints/tests/typos/locks`, a direct `cargo check --all-targets` in `fuzz/`, and `cargo xtask wasm check`, all green. `fuzz/Cargo.lock` unchanged.

### DIFF
--- a/crates/ironrdp-fuzzing/src/oracles/mod.rs
+++ b/crates/ironrdp-fuzzing/src/oracles/mod.rs
@@ -1139,3 +1139,127 @@ pub fn rdpeudp_ack_vector(data: &[u8]) {
         );
     }
 }
+
+/// Helper for [`message_decoding_invariants`].
+///
+/// On every successful `decode`, asserts that `pdu.size()` accurately reports
+/// the encoded length: `pdu.size() == encode_vec(&pdu).len()`. This is the
+/// `Encode` trait's implicit soundness contract; downstream callers rely on
+/// `size()` for buffer sizing (`ensure_size!`, `cast_length!`, etc.) and a lie
+/// here produces buffer overflows or under-allocations in encode paths.
+///
+/// Distinct from [`pdu_round_trip`]: that oracle silently drops `Err` and
+/// catches panics only. This oracle ASSERTS on the size contract, so a
+/// violation aborts the fuzz iteration as a libFuzzer crash.
+///
+/// What this catches (that `pdu_round_trip` does not):
+///
+/// - `Encode::size()` lies about its own size (returns N but `encode` writes
+///   N+M bytes or fails after writing some bytes), causing buffer
+///   over-allocation or under-allocation in callers.
+/// - Decode-acceptable bytes that the encoder cannot reconstruct to the same
+///   length (lossy decode that loses framing structure).
+///
+/// What this does NOT catch (covered by other oracles):
+///
+/// - Decode-time panics or OOM (covered by `pdu_decode`).
+/// - Re-decode equality after round-trip (covered by `pdu_round_trip`'s
+///   silent-drop pattern; assertion-based re-decode is intentionally out of
+///   scope here to keep the oracle's failure mode unambiguous).
+macro_rules! decode_size_invariant_one {
+    ($data:expr, $ty:ty) => {{
+        use ironrdp_core::Encode as _;
+        let mut cursor = ironrdp_core::ReadCursor::new($data);
+        if let Ok(pdu) = ironrdp_core::decode_cursor::<$ty>(&mut cursor) {
+            if let Ok(encoded) = ironrdp_core::encode_vec(&pdu) {
+                let size_reported = pdu.size();
+                let actual_len = encoded.len();
+                assert!(
+                    size_reported == actual_len,
+                    "{} violates Encode::size() contract: pdu.size() = {}, encode_vec(&pdu).len() = {}",
+                    ::core::any::type_name::<$ty>(),
+                    size_reported,
+                    actual_len
+                );
+            }
+        }
+    }};
+}
+
+/// Message-decoding invariants oracle: for each PDU type, exercise the
+/// `decode -> size() -> encode_vec` pipeline and assert that the type's
+/// reported `size()` matches the actual encoded length.
+///
+/// The property tested is the `Encode` trait soundness contract: on any
+/// decoder-accepted input, the decoded PDU's `size()` method must accurately
+/// report the byte length the encoder will produce. Caller code uses `size()`
+/// for buffer sizing (`ensure_size!`, `cast_length!`), so a violation
+/// produces real downstream bugs (under-allocation -> truncated encode,
+/// over-allocation -> wasted memory or buffer-overflow risk depending on
+/// surrounding context).
+///
+/// The bug class is distinct from `pdu_round_trip`: that oracle catches
+/// encoder panics on decoder-accepted inputs, while this one catches the
+/// strictly weaker but distinct class of size-contract violations.
+///
+/// Type coverage matches `pdu_round_trip`: any input that exercises the
+/// round-trip oracle's decoder set also exercises this oracle's
+/// size-invariant assertion by construction.
+pub fn message_decoding_invariants(data: &[u8]) {
+    use ironrdp_pdu::mcs::{ConnectInitial, ConnectResponse, McsMessage};
+    use ironrdp_pdu::nego::{ConnectionConfirm, ConnectionRequest};
+    use ironrdp_pdu::rdp::capability_sets::CapabilitySet;
+    use ironrdp_pdu::rdp::headers::ShareControlHeader;
+    use ironrdp_pdu::rdp::{ClientInfoPdu, server_error_info, server_license, vc};
+    use ironrdp_pdu::x224::X224;
+    use ironrdp_pdu::{bitmap, codecs, fast_path, gcc, input, pcb, surface_commands};
+
+    // Connection-time PDUs
+    decode_size_invariant_one!(data, X224<ConnectionRequest>);
+    decode_size_invariant_one!(data, X224<ConnectionConfirm>);
+    decode_size_invariant_one!(data, X224<McsMessage<'_>>);
+    decode_size_invariant_one!(data, ConnectInitial);
+    decode_size_invariant_one!(data, ConnectResponse);
+    decode_size_invariant_one!(data, ClientInfoPdu);
+    decode_size_invariant_one!(data, pcb::PreconnectionBlob);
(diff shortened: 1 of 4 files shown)


crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -314,6 +314,133 @@ pub fn pdu_round_trip(data: &[u8]) {
...
+/// over-allocation -> wasted memory or buffer-overflow risk depending on
+/// surrounding context).
+///
+/// Implements #1124 task 2 (`message_decoding_invariants`). Companion to
+/// `pdu_round_trip` (#1124 task 1): the round-trip oracle catches encoder
+/// panics on decoder-accepted inputs; this oracle catches the strictly weaker
+/// but distinct class of size-contract violations.
+///
+/// Initial type coverage mirrors `pdu_round_trip` so the same corpus feeds
+/// both oracles.
```
CBenoit: Can you elaborate on that? Is the intention to reuse the corpus?

CBenoit (commented):
Thank you!

This is looking good to me. My only suggestion here would be to make the patch more "self-contained":

>  Implements #1124 task 2 (`message_decoding_invariants`). Companion to `pdu_round_trip` (#1124 task 1)

Gets much less relevant after the work is merged, and we only care about the final shape of the code

CBenoit:
https://github.com/Devolutions/IronRDP/pull/1313 got merged!

crates/ironrdp-fuzzing/src/oracles/mod.rs
glamberson: You're right that the phrasing was loose. I went back to check what "corpus" actually means in this codebase. The word refers to the per-oracle regression seed directory at `test_data/fuzz_regression/<oracle>/` per the #1120 blueprint. The original sentence was not claiming anything about that directory. It was reaching for the looser idea that any input exercising one oracle's decoder set also exercises the other's by construction, which is true by virtue of the matching type set and independent of any actual corpus sharing. I rewrote the line without the `corpus` framing. For consistency, I also applied the parallel fix at L240 in `pdu_round_trip`'s doc-comment, which carries the same imprecise wording from the original #1291 template. Leaving only one occurrence in master would propagate the drift forward.

glamberson:
You're right that the ticket-numbered framing in the doc-comment doesn't earn its space after the PR lands. I stripped the prefix that pointed at #1124 task 2 and #1291 task 1. I rewrote the bug-class contrast that followed it so it stands on its own. I also rewrote the parallel referential framing in the commit-message body for the same reason.

CBenoit:
The new size assertion is tautological: `encode_vec` allocates a buffer of `pdu.size()` and returns that whole buffer, so its length always matches the reported size after successful encoding; under-reported sizes instead cause an ignored encode error, while over-reported sizes retain zero-filled trailing bytes. Consequently, this oracle cannot detect the claimed `Encode::size()` contract violations.

crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -613,3 +613,127 @@ impl ironrdp_cliprdr::backend::CliprdrBackend for NoopCliprdrFuzzBackend {
...
+                assert!(
+                    size_reported == actual_len,
+                    "{} violates Encode::size() contract: pdu.size() = {}, encode_vec(&pdu).len() = {}",
+                    ::core::any::type_name::<$ty>(),
+                    size_reported,
+                    actual_len
+                );
+            }
+        }
+    }};
```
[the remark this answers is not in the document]
glamberson: Confirmed and fixed, but not by tightening the same assertion. encode_vec allocates its buffer as vec![0; pdu.size()] and returns that buffer unchanged on success, so comparing pdu.size() against encode_vec(&pdu).len() is tautologically true regardless of what encode actually writes, exactly as described here.

Replaced it with a version that encodes into a buffer deliberately larger than pdu.size() via ironrdp_core::encode, which returns the encoder's real byte count instead of one derived from size() itself, and compares that against size() directly. An over-write or under-write is now directly observable instead of masked by a buffer sized from the same value being tested.

I also tried a different redesign first (comparing decode-consumed bytes to re-encoded length) and dropped it: it produced an immediate false positive on bitmap::rdp6::BitmapStream, which legitimately doesn't fully self-delimit via decode when RLE compression is off. The size()-vs-written design doesn't have that problem since it never inspects decode's cursor position. Verified with a 6.2M-execution local fuzz run across all 28 covered types, zero crashes, plus the existing regression-replay test against the seed corpus.

mamoreau-devolutions (approved):

Issue #1077, opened 2026-01-14
RFE: Add MS-RDPEUSB support


elmarco:
Teach client & server about USB device redirection.

I will submit a GSoC proposal through the QEMU community:
https://wiki.qemu.org/Internships/ProjectIdeas/RDPUSB

See linked sub-issues below. The PDU layer (#1136) can be started in parallel with the ironrdp-dvc infrastructure (#1135), since it is pure serialization with no dependency on either. The backend (#1139) can also be started early by defining the trait up front and implementing it against stubs.

uchouT:
I'm designing the `DvcProcessor` for MS-RDPEUSB and want to confirm some points about the channel topology.

My reading of MS-RDPEUSB is that the protocol uses two kinds of DVC instances, both sharing the channel name "URBDRC":

- **One control DVC** — handles capability exchange, `CHANNEL_CREATED`, and
  `ADD_VIRTUAL_CHANNEL`.
- **N device DVCs** (one per redirected USB device) — handle `ADD_DEVICE` and the per-device USB Device Interface messages (IO_CONTROL, TRANSFER_*, etc.).

## Does the device DVC also perform capability exchange?

§3.2.5.5.1 says

> The server MUST send RIM_EXCHANGE_CAPABILITY_REQUEST when the USB redirection virtual channel is connected

which is ambiguous about whether this applies to every URBDRC DVC or just the control one.

FreeRDP's client handles cap requests on any DVC defensively, so it can't disambiguate the server side. Can someone confirm whether Windows server sends `RIM_EXCHANGE_CAPABILITY_REQUEST` on device DVCs?

## Interface manipulation (QI_REQ / QI_RSP / IFACE_RELEASE)

§3.1.5.2 says

> any interface, including the default one, MUST be released with an Interface Release message.

In practice FreeRDP never sends `IFACE_RELEASE` and only uses receipt of one on the channel notification interface as a state trigger; `QI_REQ`/`QI_RSP` don't appear at all.

Is it safe to follow FreeRDP — accept `IFACE_RELEASE` but never send it, and ignore `QI_REQ`/`QI_RSP` entirely?

## Confirming ADD_DEVICE routing

[...]

PR #1321, opened 2026-05-27
refactor(rdpeusb): split PDU and TS_URB enum wrapper

uchouT:
- Add interface manipulation PDUs from [MS-RDPEXPS 2.2.2](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpexps/ebe401f0-f22e-4de4-9cd3-2a55e5493500) that MS-RDPEUSB inherits
- Split `UrbdrcServerPdu` / `UrbdrcClientPdu` by DVC roles, see #1077
- Pack `InterfaceId` and `Mask` into a single `u32` on `SharedMsgHeader`, align with MS-RDPEXPS
- Split `TsUrb` into `TsUrbIn` and `TsUrbOut`
- `TsUrbHeader` contains `ts_urb_size`
- Turn `UrbFunction` and `DeviceTextType` to raw u32

TODO:
- [x] Check TS_URB, refine in this PR if necessary

### DIFF
--- a/crates/ironrdp-rdpeusb/src/pdu/mod.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/mod.rs
@@ -46,7 +46,7 @@ impl Decode<'_> for UrbdrcServerPdu {
         let header = SharedMsgHeader::decode_with_function_id(src)?;
         let f_id = header.function_id.expect("missing function id");
 
-        match unpack(header.interface_id)? {
+        match unpack(header.iface_id)? {
             (InterfaceId::CAPABILITIES, Mask::None) => {
                 if f_id == FunctionId::RIM_EXCHANGE_CAPABILITY_REQUEST {
                     RimExchangeCapabilityRequest::decode(src, header).map(Self::Caps)
@@ -147,7 +147,7 @@ impl Decode<'_> for UrbdrcClientPdu {
         let header = SharedMsgHeader::decode(src)?;
         ensure_size!(in: src, size: 4 /* function id */);
 
-        match unpack(header.interface_id)? {
+        match unpack(header.iface_id)? {
             (InterfaceId::CAPABILITIES, Mask::None) => {
                 RimExchangeCapabilityResponse::decode(src, header).map(Self::Caps)
             }
--- a/crates/ironrdp-rdpeusb/src/pdu/mod.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/mod.rs
@@ -4,11 +4,13 @@
 //!
 //! [1]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/a1004d0e-99e9-4968-894b-0b924ef2f125
 
-use ironrdp_core::{Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteCursor, invalid_field_err};
+use ironrdp_core::{
+    Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteCursor, ensure_size, invalid_field_err,
+};
 
 use crate::pdu::caps::{RimExchangeCapabilityRequest, RimExchangeCapabilityResponse};
 use crate::pdu::completion::{IoControlCompletion, UrbCompletion, UrbCompletionNoData};
-use crate::pdu::header::{FunctionId, InterfaceId, Mask, SharedMsgHeader};
+use crate::pdu::header::{FunctionId, InterfaceId, Mask, SharedMsgHeader, unpack};
 use crate::pdu::notify::ChannelCreated;
 use crate::pdu::sink::{AddDevice, AddVirtualChannel};
 use crate::pdu::usb_dev::{
@@ -19,6 +21,7 @@ use crate::pdu::usb_dev::{
 pub mod caps;
 pub mod completion;
 pub mod header;
+pub mod iface_manipulation;
 pub mod notify;
 pub mod sink;
 pub mod usb_dev;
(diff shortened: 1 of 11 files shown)


crates/ironrdp-rdpeusb/src/pdu/mod.rs
```diff
@@ -117,98 +180,185 @@ macro_rules! fill_server_pdu_arms {
...
-                    Err(invalid_field_err!(
-                        "SHARED_MSG_HEADER",
-                        "invalid CHANNEL_CREATED header"
-                    ))
-                }
+
+        match unpack(header.iface_id)? {
+            (InterfaceId::CAPABILITIES, Mask::None) => {
+                RimExchangeCapabilityResponse::decode(src, header).map(Self::Caps)
             }
```
uchouT, line 267: I skip decode interface manipulation PDU for interfaces that don't expect FunctionId, otherwise I have to peek_u32 then dispatch.

crates/ironrdp-rdpeusb/src/pdu/mod.rs
```diff
@@ -117,98 +180,185 @@ macro_rules! fill_server_pdu_arms {
...
+impl Decode<'_> for UrbdrcClientDevicePdu {
+    fn decode(src: &mut ReadCursor<'_>) -> DecodeResult<Self> {
+        let header = SharedMsgHeader::decode(src)?;
+
+        match unpack(header.iface_id)? {
+            (InterfaceId::DEVICE_SINK, Mask::Proxy) => Self::decode_sink(src, header),
+            (InterfaceId::NOTIFY_SERVER, Mask::Proxy) => Self::decode_notification(src, header),
+            (udev_iface, Mask::Stub) => {
+                QueryDeviceTextRsp::decode(src, header.msg_id, udev_iface).map(Self::DevTextRsp)
+            }
```
uchouT, line 313: I skip decode interface manipulation PDU for interfaces that don't expect FunctionId, otherwise I have to peek_u32 then dispatch.

uchouT:
I originally had a TODO here to finish checking the TS_URB / TS_URB_RESULT encoding and decoding against MS-RDPEUSB, but after digging into it more, I think this needs an explicit boundary decision.

What should the RDPEUSB layer guarantee before a TS_URB is handed to the backend?

My current understanding is that the PDU / processor layer should validate the MS-RDPEUSB wire and protocol invariants: size/count fields, URB Function to TS_URB variant mapping, TransferIn/TransferOut direction rules, OutputBufferSize rules, RequestId/completion matching state, NoAck constraints, etc.

What I’m less sure about is whether this layer should also validate that the URB is fully valid according to the underlying Windows USB/URB semantics. That seems to require checking more than MS-RDPEUSB itself and may involve a lot of the [WDK URB documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/usb/).

Some of the current implementation appears to be influenced by URB/WDK documentation rather than only by the MS-RDPEUSB wire format. That may be intentional, but I’d like to clarify the intended contract:

Should the backend receive a TS_URB that is only guaranteed to be well-formed and valid as an MS-RDPEUSB protocol request, with USB/URB semantic validity checked by the backend/device execution layer?

Or should PDU decoding / processor logic reject URBs that are semantically invalid according to WDK URB rules before they reach the backend?

@elmarco @CBenoit

uchouT:
> Some of the current implementation appears to be influenced by URB/WDK documentation rather than only by the MS-RDPEUSB wire format. That may be intentional, but I’d like to clarify the intended contract:
> 
> Should the backend receive a TS_URB that is only guaranteed to be well-formed and valid as an MS-RDPEUSB protocol request, with USB/URB semantic validity checked by the backend/device execution layer?
> 
> Or should PDU decoding / processor logic reject URBs that are semantically invalid according to WDK URB rules before they reach the backend?

[5210b32](https://github.com/Devolutions/IronRDP/pull/1321/commits/5210b32bd06a403f2d9fbbb0eeef2b518f9e6548) is an example. Does this trend make sense?

CBenoit:
> > Some of the current implementation appears to be influenced by URB/WDK documentation rather than only by the MS-RDPEUSB wire format. That may be intentional, but I’d like to clarify the intended contract:
> > Should the backend receive a TS_URB that is only guaranteed to be well-formed and valid as an MS-RDPEUSB protocol request, with USB/URB semantic validity checked by the backend/device execution layer?
> > Or should PDU decoding / processor logic reject URBs that are semantically invalid according to WDK URB rules before they reach the backend?
> 
> [5210b32](https://github.com/Devolutions/IronRDP/pull/1321/commits/5210b32bd06a403f2d9fbbb0eeef2b518f9e6548) is an example. Does this trend make sense?

Yes, I think we should not bother parsing and validating anything more if we don’t really care about it at the wire-level. The backend will naturally check further as it actually needs to consume that data.

elmarco:
That looks good to me.

Claude pointed out an issue with 
```debug_assert!(header.function_id == Some(FunctionId::RIMCALL_RELEASE));```:

    Client-side PDU decode paths (UrbdrcClientControlPdu, UrbdrcClientDevicePdu)
    use SharedMsgHeader::decode which reads only 8 bytes and sets
    function_id: None. The function_id is read separately in decode_sink /
    decode_notification. When these paths call InterfaceRelease::from_header,
    the debug_assert!(header.function_id == Some(RIMCALL_RELEASE)) fires
    because function_id is None -- even though the caller already matched on
    the correct FunctionId.

    The assert is unnecessary: from_header is only reachable after the caller
    matched FunctionId::RIMCALL_RELEASE, so correctness is guaranteed by
    the call site, not by the header field.

CBenoit:
In general, it’s better to have more debug_assert! that strictly necessary just so it’s easier to catch bugs with tests and fuzzing, but it’s suspicious if function_id is always None.

uchouT:
Cleaned. I left that debug_assert when `SharedMsgHeader` was still attempting to decode `FunctionId`.

CBenoit (commented):
@uchouT LGTM. Is the PR ready to be merged?

uchouT:
> uchouT (@uchouT) LGTM. Is the PR ready to be merged?

Yes. Thanks.

CBenoit (approved):
LGTM!

Issue #1314, opened 2026-05-26
pdu: audit unreachable!() / panic!() sites for reachability via fuzz / spec-conformant inputs


glamberson:
## Background

PR #1313 (closes #1292) fixed a `unreachable!()` panic in `CapabilitySet::BitmapCacheV3`'s encoder that was reachable from any decoder-accepted input carrying the `BitmapCacheV3CodecID` capability set. The bug was surfaced by PR #1291's `pdu_round_trip` fuzz oracle within ~60 seconds of smoke fuzzing on a 4-byte reproducer (`[0x06, 0x00, 0x04, 0x00]`).

The shape of the bug — decoder accepts variant X, `size()` covers variant X, encoder match does not cover variant X, falls through to `_ => unreachable!()` — is unlikely to be unique to `CapabilitySet`. Other `Encode` impls with catch-all `unreachable!()` arms in `crates/ironrdp-pdu/src/` may have the same gap. Beyond encoder-match-missing-arms, internal invariants asserted via `unreachable!()` / `panic!()` / `assert!()` elsewhere in the crate may also be reachable from spec-conformant or attacker-controlled inputs.

This issue tracks a systematic audit pass over `crates/ironrdp-pdu/src/` to classify every panic site.

## Scope

`crates/ironrdp-pdu/src/` only for this issue. Adjacent crates (`ironrdp-egfx`, `ironrdp-graphics`, `ironrdp-bulk`, `ironrdp-rdpdr`, etc.) can be audited under separate sibling issues if this one finds the pattern productive.

## Approach

Walk every `unreachable!()`, `panic!()`, and `assert!()` call site in `crates/ironrdp-pdu/src/`. Classify each as:

- **(a) Provably-unreachable internal invariant.** The condition is genuinely impossible given upstream constraints (e.g., type-system-enforced state). Leave the panic; optionally demote to `debug_assert!()` to remove the release-build cost while preserving development-time enforcement.
- **(b) Reachable via fuzz inputs.** The `pdu_round_trip` or `pdu_decode` oracle can reach this site within bounded fuzzing time. File a fix PR with the offending input as a regression-corpus entry.
- **(c) Reachable via attacker-controlled wire inputs.** Same as (b) but the input is plausibly attacker-shaped (e.g., negotiable on the wire, not requiring an internal helper invocation). File fix PR with priority.

## Relationship to issue #1274

[...]

PR #1328, opened 2026-05-27
refactor(pdu): make the CapabilitySet encoder exhaustive

glamberson:
Part of the `unreachable!()` / `panic!()` audit pass on #1314, scoped to `crates/ironrdp-pdu/src/rdp/capability_sets/`.

**`mod.rs` (`CapabilitySet::Encode::encode` raw-buffer fallback):** the inner `_ => unreachable!()` catch-all is replaced with an explicit enumeration of the 17 structured `CapabilitySet` variants already handled by the outer match. Adding a new variant to `CapabilitySet` is now a compile error here until it is routed in this `Encode` impl. The arm body stays `unreachable!()` (genuinely unreachable under the current outer-match structure). PR #1313 (BitmapCacheV3 encoder `unreachable!()` reached on decoder-accepted input) is why a runtime catch-all is the wrong shape here.

**`bitmap_codecs/mod.rs` (`CodecProperty` decode for `GUID_REMOTEFX | GUID_IMAGE_REMOTEFX`):** keeps its `match guid { ... _ => unreachable!() }` structure per review. `guid` is already validated by the outer arm, so the `_` branch is a redundant correctness check that fires under tests and fuzzing if a future change breaks the invariant. The only change is an inline message documenting that invariant.

Both sites are class (a) (provably unreachable in current code), confirmed by `pdu_round_trip` (#1291) and `message_decoding_invariants` (#1320) exercising `CapabilitySet` and `ShareControlHeader` for millions of iterations post-#1313 with no panics.

Refs #1314.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs/mod.rs
+++ b/crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs/mod.rs
@@ -281,7 +281,12 @@ impl<'de> Decode<'de> for Codec {
                 match guid {
                     GUID_REMOTEFX => CodecProperty::RemoteFx(property),
                     GUID_IMAGE_REMOTEFX => CodecProperty::ImageRemoteFx(property),
-                    _ => unreachable!(),
+                    // `guid` is validated as RemoteFX or ImageRemoteFX by the outer
+                    // match arm, so the `_` branch is genuinely dead. Keep it as a
+                    // redundant correctness check that fires loudly under tests and
+                    // fuzzing if a future change to the outer arm breaks that
+                    // invariant. Not reachable from the wire.
+                    _ => unreachable!("guid validated as RemoteFX or ImageRemoteFX by the outer match"),
                 }
             }
             GUID_IGNORE => CodecProperty::Ignore,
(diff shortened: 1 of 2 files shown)


crates/ironrdp-pdu/src/rdp/capability_sets/bitmap_codecs/mod.rs
```diff
@@ -278,10 +278,13 @@ impl<'de> Decode<'de> for Codec {
...
-                    GUID_IMAGE_REMOTEFX => CodecProperty::ImageRemoteFx(property),
-                    _ => unreachable!(),
+                // `guid` is `GUID_REMOTEFX` or `GUID_IMAGE_REMOTEFX` by the
+                // outer match arm; the previous nested `match` plus
+                // `_ => unreachable!()` carried a dead arm.
+                if guid == GUID_REMOTEFX {
+                    CodecProperty::RemoteFx(property)
+                } else {
+                    CodecProperty::ImageRemoteFx(property)
                 }
```
CBenoit, line 290: thought: I would keep the `unreachable!()` here rather than fold it into an `if/else`. `guid` is already validated as `GUID_REMOTEFX | GUID_IMAGE_REMOTEFX` by the outer arm, so the `_` branch is genuinely dead -- and it's *because* it should be dead that `unreachable!()` is the right tool: it's a redundant correctness check that fires loudly under tests/fuzzing if a future change (e.g. a third GUID added to the outer arm) breaks the invariant. The `else` turns that into a silent mislabel instead. Note this branch isn't reachable from the wire, only from a logic regression, so the usual "no panics in decoders" rule doesn't apply.

Optionally add a message to document the invariant inline: `unreachable!("guid validated as RemoteFX or ImageRemoteFX by the outer match")`.

The other exhaustive-match changes in the PR look good to me.

glamberson: Sure, reverted to the `match` with `unreachable!()` and added your suggested message documenting the invariant. Agreed that for a branch that is only reachable via a future logic regression, the tripwire is more useful than an `else` that would silently mislabel. The exhaustive arm in `mod.rs` is unchanged.

CBenoit (approved):
Thank you!

PR #1331, opened 2026-05-27
fix(egfx)!: make DecodedFrame fields private with getters to enforce size invariant

glamberson:
## Summary

- Picks up the follow-up CBenoit invited via `thought:` on #1284 (`discussion_r3271919807`); acknowledged inline as `discussion_r3273666208` with a commitment to file the follow-up.
- After #1284, `DecodedFrame` in `crates/ironrdp-egfx/src/decode.rs` is `#[non_exhaustive]` with public `data`, `width`, `height` fields. `DecodedFrame::new()` enforces the `data.len() == width * height * 4` RGBA8888 invariant via `debug_assert_eq!`, but the `pub` fields let consumers bypass the guard via post-construction mutation. `#[non_exhaustive]` closes struct-literal construction from outside the crate but does not lock down field-level writes.
- This PR makes the three fields private and exposes them via read-only getters: `data() -> &[u8]`, `width() -> u32`, `height() -> u32`. A consume accessor `into_data(self) -> Vec<u8>` covers callers needing owned ownership transfer without an extra clone.
- `DecodedFrame::new()` remains the sole construction path. The internal `openh264_impl::decode` previously used a struct literal; this PR routes it through `new()` so the same-crate path also benefits from the invariant guard.

## Validation

- Two unit tests in `ironrdp_egfx::decode::tests` cover the new surface: `getters_return_constructor_inputs` verifies the three getters round-trip the constructor inputs, and `into_data_yields_owned_buffer` verifies the consume path returns the original RGBA buffer.
- All 12 internal call sites in `crates/ironrdp-egfx/src/client.rs` and `crates/ironrdp-testsuite-core/tests/egfx/decode.rs` updated to use the getters; a workspace grep confirms no remaining field-style accesses on `DecodedFrame` instances.
- `cargo xtask ci` clean locally on `1.89.0`: fmt, lints, tests, typos, locks all green.

## Notes

- Breaking change for external consumers reading the public fields directly. Migration path is mechanical:
  - `frame.data` becomes `frame.data()` (returns `&[u8]` rather than `&Vec<u8>`; callers needing `Vec` go through `into_data()` or `.to_vec()`)
  - `frame.width` becomes `frame.width()`
  - `frame.height` becomes `frame.height()`
- `!` marker on the conventional commit subject so release-plz computes a breaking-version bump on the next `ironrdp-egfx` cut.
- Resolves the follow-up commitment from `discussion_r3273666208` on the merged #1284 thread; the original `thought:` is at `discussion_r3271919807`.

### DIFF
--- a/crates/ironrdp-egfx/src/client.rs
+++ b/crates/ironrdp-egfx/src/client.rs
@@ -751,10 +751,10 @@ impl GraphicsPipelineClient {
         // Decoded frame must be at least as large as the destination rectangle.
         // Larger is expected (macroblock alignment) and handled by cropping.
         // Smaller means the server sent mismatched dimensions.
-        if frame.width < u32::from(dest_width) || frame.height < u32::from(dest_height) {
+        if frame.width() < u32::from(dest_width) || frame.height() < u32::from(dest_height) {
             warn!(
-                frame_width = frame.width,
-                frame_height = frame.height,
+                frame_width = frame.width(),
+                frame_height = frame.height(),
                 dest_width,
                 dest_height,
                 "decoded frame smaller than destination rectangle"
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
Thank you! LGTM

PR #1332, opened 2026-05-27
feat(nscodec): introduce ironrdp-nscodec crate + opt-in server integration

### LINKED ISSUE #1350: Pilot the per-codec-crate pattern by introducing `ironrdp-nscodec`
NSCodec is small, self-contained, and not yet implemented. Low-risk way to nail down the per-crate template.

## Scope

- New crate `crates/ironrdp-nscodec`.
- Encoder behind `encoder` feature, off by default.
- Decoder reserved for future `decoder` feature.
- No default features.
- Opt-in `ironrdp-server` integration behind a separate `nscodec` feature.

## Acceptance

- Builds with no default features, in workspace.
- Server integration compiles only with `ironrdp-server`'s `nscodec` feature on.

clintcan:
Implements the NSCodec encoder agreed on in discussion #1322 ([link](https://github.com/Devolutions/IronRDP/discussions/1322)). Follows the architecture @CBenoit suggested there: a new `ironrdp-nscodec` peer crate (rather than folding the encoder into `ironrdp-server`), encoder behind a feature flag, no default features, server integration also behind its own opt-in feature.

## Summary

Two commits, intentionally separable:

1. **`feat(nscodec): introduce ironrdp-nscodec crate with MS-RDPNSC encoder`**
   - New `crates/ironrdp-nscodec/` crate. Encoder behind `encoder` feature, off by default. `default-features = false` consumers get an empty shell.
   - Implementation of MS-RDPNSC §3.1.5: RGB→YCoCg conversion, per-plane RLE, 20-byte frame header.
   - Public API: `encoder::encode(data, w, h, stride, format, color_loss_level) -> Vec<u8>` decoupled from `BitmapUpdate` so the server-side glue is thin.
   - Verified byte-identical against a production-tested reference encoder over **1008 fuzzed cases** (8 pixel formats × 6 sizes × 7 CLL values × 3 random seeds). No behavioral drift from the implementation that has been in use against macOS Microsoft Remote Desktop for the last several months.
   - Limitations called out in code: no 4:2:0 subsampling yet (`ChromaSubsamplingLevel = 0` always); encoder-only (the IronRDP client decodes NSCodec via `ironrdp-session` today).

2. **`feat(server): wire NSCodec encoder via ironrdp-nscodec behind a feature gate`**
   - New opt-in `nscodec` feature on `ironrdp-server` pulling in `ironrdp-nscodec` with its `encoder` feature.
   - Replaces the `CodecProperty::NsCodec(_) => ()` no-op with an active arm gated on the new feature; the no-op is preserved when the feature is off, so default builds have zero behavior change.
   - `NsCodecHandler` slots into the existing `BitmapUpdater` selection at lower priority than RemoteFX (priority: qoiz > qoi > remotefx > nscodec). NSCodec exists mainly for clients whose legacy bitmap-codec list advertises only NSCodec — notably the macOS Microsoft Remote Desktop / Windows App, which otherwise falls through to raw/RLE BitmapUpdate at much higher bandwidth.
   - CLL is taken from the client's confirmed `NsCodec` capability so the server encodes at the same shift the client decodes against.

## Test plan

- [x] `cargo build -p ironrdp-nscodec` (no features) — empty shell builds clean
- [x] `cargo build -p ironrdp-nscodec --features encoder` — encoder builds clean
- [x] `cargo test -p ironrdp-nscodec --features encoder` — 7/7 unit tests pass (RLE / YCoCg / header layout)
- [x] Byte-equivalence harness vs production reference: 1008/1008 identical
- [x] `cargo clippy --features encoder --all-targets -- -D warnings` (and no-features variant) — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo build -p ironrdp-server` (default) — clean, no behavior change
- [x] `cargo build -p ironrdp-server --features nscodec` — clean
- [x] `cargo build -p ironrdp-server --no-default-features --features nscodec` — clean
- [x] `cargo build -p ironrdp-server --features nscodec,qoi,qoiz,egfx,helper` — clean
- [x] `cargo clippy -p ironrdp-server --features helper,__bench,nscodec --all-targets --locked -- -D warnings` (mirrors CI) — clean
- [x] `cargo test -p ironrdp-server --features nscodec` — 5/5 doctests pass
- [x] Full `cargo build --workspace` — no impact on other crates

## Notes for review

- The `last-4-bytes-raw` plane convention isn't in the MS-RDPNSC text but is mandatory by the FreeRDP reference implementation that Microsoft NSCodec decoders are written against. Comment in the encoder explains this.
- `debug_assert!(color_loss_level in 1..=7)` because the i8 intermediate storage of Co/Cg overflows at CLL=0. Defensive clamping in release builds keeps the output decodable-shaped rather than wrapping-garbage. Real servers should advertise CLL ≥ 1.
- The decoder side (`decoder` feature) is intentionally *not* pre-declared — happy to add it in a future PR if and when there's a need.

## Discussion

[#1322](https://github.com/Devolutions/IronRDP/discussions/1322) — NSCodec encoder, asked & answered by @elmarco / @CBenoit.


glamberson:
@clintcan, the encoder reads cleanly and the byte-equivalence harness against your production reference over 1008 fuzzed cases is exactly the kind of evidence that earns merge. The implementation work is solid and I am glad to see it offered.

I have posted a placement observation on #1322 about the dedicated-crate shape versus folding into `ironrdp-graphics/src/nscodec/` alongside the other bitmap codecs. That is a question for the architecture call made on the discussion, not a comment on the implementation here. If the placement ends up shifting, the encoder lift is mechanical and your harness carries over unchanged.

crates/ironrdp-server/src/server.rs
```diff
@@ -137,6 +137,14 @@ impl RdpServerOptions {
...
             .any(|codec| matches!(codec.property, CodecProperty::QoiZ))
     }
+
+    #[cfg(feature = "nscodec")]
+    fn has_nscodec(&self) -> bool {
+        self.codecs
+            .0
+            .iter()
+            .any(|codec| matches!(codec.property, CodecProperty::NsCodec(_)))
+    }
```
CBenoit, line 147: question: Would it make sense to have this method defined regardless of the `nscodec` feature flag, but with a blanket implementation returning an unconditional `false` when the codec is not part of the build?

rationale: Typically less conditional compilation churn in consumer code, but I’m not sure how useful that is broadly.

crates/ironrdp-nscodec/README.md
```diff
@@ -0,0 +1,30 @@
+# ironrdp-nscodec
+
+NSCodec ([MS-RDPNSC]) implementation for IronRDP.
+
+NSCodec is a legacy bitmap codec used in the RDP "Surface Bits" command path. It
+predates RemoteFX and QOI/QOIZ but remains the only legacy codec advertised by
```
CBenoit: nit: I think QOI/QOIZ being an IronRDP-specific experimentation I wouldn’t mention it along something like RemoteFX

crates/ironrdp-nscodec/LICENSE-MIT
CBenoit, line 1: issue: Same comment for both LICENSE-APACHE and LICENSE-MIT files. IIRC we are using symlink everywhere else. Copy-paste the symlinks so it’s identical everywhere and we don’t have duplicate files.

CBenoit (commented):
Thank you @clintcan and @glamberson. Implementation is also looking good to me.
I flagged a few places, and I’ll request a Copilot review for sanity, address if relevant.
My intention is to merge this PR once I am done with publishing latest versions for `ironrdp-egfx` and `ironrdp-server`.

crates/ironrdp-server/src/server.rs
clintcan: Good question. I went with the gated method to follow the existing `qoi` / `qoiz` precedent: `has_qoi` and `has_qoiz` are themselves `#[cfg(feature = "qoi")]` / `#[cfg(feature = "qoiz")]` (server.rs:126/134), and the whole chain is gated to match — the `nscodec` field on `UpdateEncoderCodecs`, its `set_nscodec` setter, `has_nscodec`, and the codec-selection match arm.

A blanket `has_nscodec() -> false` would compile, but it wouldn't remove the `#[cfg]` at the one call site: the match arm's body calls `update_codecs.set_nscodec(...)`, and that setter (plus the field it writes) only exists under the feature, so the arm has to stay gated either way. Making just `has_nscodec` unconditional would then leave it inconsistent with `has_qoi`/`has_qoiz` without saving any churn.

If you'd prefer the blanket-`false` style as a broader convention, the cleaner move is to apply it to `qoi`/`qoiz` too in a separate pass so they stay consistent — happy to do that here or as a follow-up, whichever you'd rather. For this PR I'd lean toward keeping it matching the current pattern.

CBenoit: Sounds good to me. You convinced me to not change anything unless we really see a benefit to it.

CBenoit (approved):
LGTM! Thank you

PR #1334, opened 2026-05-27
feat(egfx): cascade Arbitrary derives across ironrdp-egfx public PDU types

glamberson:
## Summary

- Extends the `Arbitrary` cascade from PR #1272 (ironrdp-pdu) and PR #1284 (extension to pdu types in egfx-adjacent modules) to the `ironrdp-egfx` crate's own public type surface.
- Adds the `arbitrary` feature to `ironrdp-egfx/Cargo.toml`, cascading through `ironrdp-pdu/arbitrary` and `bitflags/arbitrary` so variant payload types and bitflag types pick up their `Arbitrary` impls automatically.
- Adds `#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]` to every public PDU type in `crates/ironrdp-egfx/src/pdu/`:
  - `common.rs`: `Point`, `Color`, `PixelFormat`
  - `avc.rs`: `QuantQuality`, `Avc420BitmapStream<'a>`, `Avc444BitmapStream<'a>`, `Avc420Region`, `Encoding` (bitflags)
  - `cmd.rs`: `GfxPdu` plus all 23 variant types, `RawCapabilitySet`, `CapabilitySet`, `CapabilityVersion`, `Codec1Type`, `Codec2Type`, `Timestamp`, `QueueDepth`, `CacheEntryMetadata`, and the six `CapabilitiesV*Flags` bitflag structs.
- Adds the `ironrdp-egfx/arbitrary` case to the xtask feature matrix (`xtask/src/features.rs`), mirroring the existing `ironrdp-pdu/arbitrary` entry so CI verifies the feature builds cleanly.

## Validation

- `cargo xtask ci` clean locally on `1.89.0`: fmt, lints, tests, typos, locks all green.
- `cargo xtask check features --case ironrdp-egfx/arbitrary` clean.
- Existing 12 `ironrdp-egfx` unit tests continue to pass.
- Default build (no `arbitrary` feature) unchanged: no new dependencies pulled, no behavior change.

## Notes

- Prerequisite cascade for the `egfx_multi_frame` fuzz target (target 5 under #1316), which will use `Arbitrary`-derived `Vec<GfxPdu>` to drive a single decoder + surface-cache pair across each fuzz iteration. The target itself follows in a separate PR.
- Pure additive change at the SemVer surface: every derive is gated behind `#[cfg_attr(feature = "arbitrary", ...)]`, and the feature is `optional = true` in Cargo.toml. Default consumers see no change.
[...]

Issue #1316, opened 2026-05-26
egfx: extend fuzz coverage to OpenH264-wrapper, ZGFX, multi-frame state, and encoder round-trip


glamberson:
**Part of #1120.** This issue is an egfx-scoped child of the main fuzzing umbrella, focusing on the codec-pipeline depth specific to MS-RDPEGFX. It sits alongside #1122 (Arbitrary impls) and #1124 (advanced fuzz targets) under the #1120 umbrella, not as a parallel framing.

## Why a separate child issue rather than comments on #1120

#1120 is broad and workspace-wide. The egfx-specific fuzz coverage gaps below are six concrete targets across four oracle classes, scoped to one crate's codec pipeline. That is enough scope to warrant its own tracking surface as a child issue rather than getting buried in #1120's comment thread (where our 2026-05-18 `issuecomment-4483012331` on the ZGFX + memory-budget question is currently sitting).

## Background

PR #1271 (merged 2026-05-25) extended the `pdu_decode` oracle in `crates/ironrdp-fuzzing/src/oracles/mod.rs` to include egfx wire-PDU types: `GfxPdu`, `CacheToSurfacePdu`, `EgfxRawCapabilitySet`, `Avc420BitmapStream<'_>`, `Avc444BitmapStream<'_>`, `QuantQuality`, `Point`, `Color`, `CapabilitiesAdvertisePdu`. Two real bugs surfaced and were fixed in-PR (`Avc444BitmapStream::decode` `split_at` panic on malformed `streamLen`; `Avc420BitmapStream::decode` `Vec::with_capacity(num_regions)` ASan OOM on attacker-controlled u32).

PR #1291 (merged 2026-05-25) added `pdu_round_trip` for `ironrdp-pdu` types, but did not extend into `ironrdp-egfx`'s own encoders.

The result is that egfx is fuzzed at the wire-PDU-decode layer but the pipeline below that is not. The unfuzzed areas are below.

## Coverage gaps

| Surface | Wire-PDU decode | Round-trip | State-machine | Memory-budget |
|---|---|---|---|---|
| GfxPdu top-level | covered (#1271) | gap | gap | gap |
| Avc420BitmapStream wire | covered (#1271, 1 bug fixed) | gap | gap | gap |
| Avc444BitmapStream wire | covered (#1271, 1 bug fixed) | gap | gap | gap |
| Wire-decoded AVC bytes → OpenH264 input wrapper | gap | n/a | gap | gap |
[...]

PR #1336, opened 2026-05-28
feat(fuzz): add egfx_multi_frame oracle and target

glamberson:
## Summary

- Implements target 5 of #1316 (egfx fuzz-coverage umbrella): the `egfx_multi_frame` oracle and target.
- Harness shape (per the design note posted as `issuecomment-4559459227` on #1316): an `Arbitrary`-derived `Vec<GfxPdu>` drives a single `GraphicsPipelineClient` + surface-cache pair across each fuzz iteration, exposing cross-PDU state corruption that single-shot fuzzers cannot reach.
- Each PDU is encoded back to wire bytes, wrapped in a single uncompressed ZGFX segment, and fed through the client's public `DvcProcessor::process` entry. The harness deliberately routes through `process()` rather than the private `handle_pdu` so production traffic's full path is exercised (ZGFX-decompress + GfxPdu-decode + per-variant dispatch + state machine + surface cache).
- The uncompressed ZGFX wrap bypasses the ZGFX decoder layer (covered separately by PR #1333) and concentrates fuzz pressure on dispatch + state machine.

## Validation

- `cargo xtask check fmt/lints/tests/typos/locks` all pass.
- `check_egfx_multi_frame` regression-replay test passes against the seed-empty corpus entry.
- 15-minute rigorous libFuzzer + ASan fuzz on the final SHA: **3,659,157 iterations in 901 seconds (~4,061 exec/s sustained), peak RSS 96 MB, zero panics, zero sanitizer reports, zero OOMs.** Corpus grew to ~4,558 entries via coverage discovery.

## What this catches

- Panics or sanitizer reports along the dispatch and state-machine path under adversarially-ordered PDU sequences.
- Inconsistent surface-cache state under attacker-controlled `CreateSurface` / `DeleteSurface` / `Map*` orderings.
- Corrupted frame-id state from interleaved `StartFrame` / `EndFrame` / `FrameAcknowledge` sequences.
- ZGFX-wrapper integration bugs separate from the standalone ZGFX coverage in `egfx_zgfx_decompress` (PR #1333).

## What this does NOT catch

- Cross-frame H.264 decoder state corruption. The client is constructed with `h264_decoder: None`, so `WireToSurface1` PDUs carrying AVC payloads skip the decoder. The standalone `egfx_avc420_decode` (PR #1326) and `egfx_avc444_decode` (PR #1327) targets cover the H.264 wrapper. Wiring a real or mock H.264 decoder into this harness can be a follow-up if frame-to-frame H.264 state coverage surfaces as a gap.

## Notes

- The first smoke fuzz on the in-progress harness surfaced a `bit_field::set_bits` panic in `Timestamp::encode` where `derive(Arbitrary)` generated `u8`/`u16` values exceeding the encoder's 6-bit and 10-bit pack widths. The fix landed preemptively in PR #1334 as manual `Arbitrary` impls for `Timestamp`, `QuantQuality`, and the `Encoding` bitflag (mapping each field to its wire-allowed range). That follows the `sanitize via mask` shape from #1122's body.
- The oracle uses `as _` imports for `Arbitrary` and `DvcProcessor` per clippy's `unused_trait_names` lint.
- `NoOpHandler` is an empty struct: every `GraphicsPipelineHandler` method has a default impl in the trait.

Refs #1316.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2704,6 +2704,7 @@ dependencies = [
  "ironrdp-cliprdr-format",
  "ironrdp-core 0.2.1",
  "ironrdp-displaycontrol",
+ "ironrdp-dvc",
  "ironrdp-egfx",
  "ironrdp-graphics",
  "ironrdp-pdu",
--- a/crates/ironrdp-fuzzing/Cargo.toml
+++ b/crates/ironrdp-fuzzing/Cargo.toml
@@ -20,7 +20,8 @@ ironrdp-rdpdr.path = "../ironrdp-rdpdr"
 ironrdp-rdpsnd.path = "../ironrdp-rdpsnd"
 ironrdp-cliprdr-format.path = "../ironrdp-cliprdr-format"
 ironrdp-displaycontrol.path = "../ironrdp-displaycontrol"
-ironrdp-egfx.path = "../ironrdp-egfx"
+ironrdp-dvc.path = "../ironrdp-dvc"
+ironrdp-egfx = { path = "../ironrdp-egfx", features = ["arbitrary"] }
 ironrdp-svc.path = "../ironrdp-svc"
 
 [lints]
--- a/crates/ironrdp-fuzzing/src/oracles/mod.rs
+++ b/crates/ironrdp-fuzzing/src/oracles/mod.rs
@@ -421,6 +421,88 @@ pub fn egfx_avc420_decode(data: &[u8]) {
     }
 }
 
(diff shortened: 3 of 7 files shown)


CBenoit:
No concrete blocking, non-blocking, or question findings were missed. The MS-RDPEGFX handoff's coverage-only boundaries are non-defects, and risk:low, ai-reviewed/1, and human-required are appropriate with no additional risk labels.

CBenoit (approved):

PR #1339, opened 2026-05-28
feat(agent): introduce ironrdp-agent crate

CBenoit:
Add a CLI-driven, daemon-backed RDP client designed for programmatic
(e.g. LLM) consumption. A single binary plays two roles: a long-lived
daemon that owns the ironrdp-client engine and one RDP session, and a
short-lived CLI that drives it over a local IPC transport (Unix domain
socket / Windows named pipe).

Highlights:
- Binary, length-delimited IPC protocol using ironrdp-core's
  Encode/Decode traits (no JSON). Connection config travels as a
  binary-encoded PropertySet inside a strictly-typed Request::Connect;
  runtime input/query operations are strictly-typed messages.
- Secrets never reach the IPC reader: ConfigBuilder::build strips every
  ironrdp_cfg::is_secret_key property, and the daemon seeds its live bag
  from the post-build config, so dumps/status/logs cannot leak them.
- Operator overlay: daemon-start --overlay FILE preloads a .rdp file
  applied on top of every connect (overlay wins) to provision any
  setting out of band, credentials in particular. Status reports
  credentials_loaded so a caller knows whether it must supply a password.
- Separate logging: the daemon's own logs go to stderr (default INFO,
  IRONRDP_LOG), while each RDP session's logs are captured into a small
  queryable ring buffer (default DEBUG) via a thread-local subscriber,
  refinable per-connect with --log-directive for troubleshooting.
- --help-agent prints a structured, LLM-friendly operation guide.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2453,6 +2453,7 @@ dependencies = [
  "ironrdp-propertyset",
  "ironrdp-rdpfile",
  "libc",
+ "png",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2440,6 +2440,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ironrdp-agent"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "ironrdp-cfg",
+ "ironrdp-client",
+ "ironrdp-core",
+ "ironrdp-input",
+ "ironrdp-pdu",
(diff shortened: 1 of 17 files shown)


CBenoit:
V1 is functional enough to let the LLM open notepad and write messages into it:

image

I see a few improvements:

- Support for clipboard
- Support for sending batch of inputs (multiple key presses and releases in a single command)
- Support for recording a trail of commands that we could save as-is to create e2e test cases

mamoreau-devolutions (approved):

PR #1343, opened 2026-05-30
docs(rdpsnd): document RdpsndServerHandler::start wFormatNo contract

clintcan:
## What

`RdpsndServerHandler` and its methods are currently undocumented. This PR adds rustdoc to the trait, with the load-bearing detail on `start`'s return value.

## Why

The `Option<u16>` returned by `start` is stamped **directly** onto every `Wave`/`Wave2` PDU as `wFormatNo` (`RdpsndServer::wave`). The non-obvious part — with no doc to warn you — is *which list that index addresses*:

- A compliant client resolves each wave's format as `ClientFormats[wFormatNo]` against the list **it** sent in the Client Audio Formats PDU, and rejects any `wFormatNo >= NumberOfClientFormats`. FreeRDP does exactly this in [`rdpsnd_recv_wave2_pdu`](https://github.com/FreeRDP/FreeRDP/blob/master/channels/rdpsnd/client/rdpsnd_main.c) (and `rdpsnd_recv_wave_info_pdu`) — out-of-range is dropped, so **all audio silently disappears**.
- The client's list is its accepted *subset* of the server's offered formats, so it generally differs from `get_formats()` in both length and order. Returning a `get_formats()` index only works when the chosen format happens to sit at the same position in both lists.

This bites the moment a handler offers more than one format and a client accepts only some of them: indexing `get_formats()` sends the wrong (or out-of-range) `wFormatNo`, and the failure mode is silent — no error, just no sound. We hit this in a downstream server when adding a second (AAC) format alongside PCM: mstsc kept working (it lists the chosen format at the same index in both), but FreeRDP/Thincast clients that accepted only PCM got an out-of-range index and went silent. Picking the format and then returning its position within `client_format.formats` fixes it.

The doc makes the contract explicit so the next implementer doesn't have to rediscover it from a wire capture.

## Notes

- Docs only — no behavior change.
- `cargo check -p ironrdp-rdpsnd` and `RUSTDOCFLAGS="-D warnings" cargo doc -p ironrdp-rdpsnd --no-deps` both pass; no reformatting from `cargo fmt`.
- Possible follow-up (happy to discuss separately): make the contract harder to get wrong by having `start` return the chosen `AudioFormat` (or its `client_format.formats` index) and letting the crate compute/validate `wFormatNo`. That's an API/semver change, so I kept this PR to docs only.

### DIFF
--- a/crates/ironrdp-rdpsnd/src/server.rs
+++ b/crates/ironrdp-rdpsnd/src/server.rs
@@ -28,11 +28,37 @@ pub enum RdpsndServerMessage {
     Error(Box<dyn RdpsndError>),
 }
 
+/// Handler for the server side of the Audio Output Virtual Channel (`RDPSND`).
+///
+/// Implementations supply the list of audio formats the server offers, decide
+/// which format to use once the client replies, and produce the audio waves to
+/// stream (via [`RdpsndServer::wave`]).
 pub trait RdpsndServerHandler: Send + core::fmt::Debug {
+    /// The audio formats the server advertises in the Server Audio Formats and
+    /// Version PDU (MS-RDPEA 2.2.2.1).
     fn get_formats(&self) -> &[pdu::AudioFormat];
 
+    /// Called once the client has replied with the formats it accepts
+    /// (`client_format`, the Client Audio Formats and Version PDU). Returns the
+    /// `wFormatNo` to stamp on every subsequent Wave/Wave2 PDU, or [`None`] if
+    /// no offered format is acceptable (no audio is then streamed).
+    ///
+    /// **The returned index addresses `client_format.formats` — the formats the
+    /// client just echoed back — NOT the server's own [`get_formats`] list.**
+    /// The client resolves each wave's format as `ClientFormats[wFormatNo]`
+    /// against the list *it* sent, and a compliant client rejects any
+    /// `wFormatNo >= client_format.formats.len()`, silently dropping all audio.
+    /// The client's list is its accepted subset of the server's formats, so the
+    /// two lists generally differ in both length and ordering; an index into
+    /// [`get_formats`] only happens to work when the chosen format sits at the
+    /// same position in both. Pick the format you intend to send, then return
+    /// its position within `client_format.formats`.
+    ///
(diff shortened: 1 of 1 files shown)


clintcan:
This is ready for review whenever someone has a moment. It's docs-only — adds rustdoc to `RdpsndServerHandler::start` documenting that the returned `u16` is the wire `wFormatNo`, which the client resolves against **its own** Client Audio Formats list (not the server's `get_formats()`); FreeRDP/Thincast reject `wFormatNo >= NumberOfClientFormats` and silently drop the wave. `cargo check -p ironrdp-rdpsnd` and `RUSTDOCFLAGS=-D warnings cargo doc` both pass. Happy to adjust wording or split out the follow-up API idea floated in the description.

CBenoit (approved):
This is useful documentation, as it’s typical when you had to dive into a rabbit hole because of a an obscure bug. Thank you for contributing it.

CBenoit:
>     * Possible follow-up (happy to discuss separately): make the contract harder to get wrong by having `start` return the chosen `AudioFormat` (or its `client_format.formats` index) and letting the crate compute/validate `wFormatNo`. That's an API/semver change, so I kept this PR to docs only.

I would be more than happy to have a misuse-resistant API instead of the error-prone we have currently (even if it’s now well documented). Generally speaking, I think this would be a superior API on many levels: hard to misuse, self-explanatory types and signature, the crate can also perform additional sanity checks -- e.g.: is the selected format is not part of the common set?

On another note, it also makes sense to me if the function only gets to see the audio formats _in common_ with the client, and we can let the crate handle everything else (if there is nothing in common, we don’t even need to call the method in the first place). I guess it depends how far we want to go in the abstraction. I’m unsure whether it can be genuinely useful for the consumers to see the full client format set unconditionally.

clintcan:
Opened #1356 to track the misuse-resistant `start` API — full proposal and the open question (how far to abstract / index-vs-format return shape) are there so it's easier to follow than a comment on a merged PR. Carrying over my read from above: hand `start` only the negotiated intersection (each entry pre-tagged with its `wFormatNo`), keep `get_formats()` for the server's advertised list, and don't call `start` at all when there's nothing in common.

PR #1344, opened 2026-05-31
fix(graphics): bound ZGFX compressor hash table size

glamberson:
Summary

The ZGFX LZ77 compressor's compact_hash_table trims per-prefix position lists but never reduces the number of distinct prefixes. Incompressible input (an already-encoded H.264 EGFX payload, for example) produces a near-unique 3-byte prefix per byte, so once the table passes MAX_HASH_TABLE_ENTRIES it stays over the cap and add_to_history re-runs compaction on every literal byte, each pass scanning the whole table while freeing nothing. That is O(n * table_size) per frame: compressing 100 KB of incompressible data took about 10 s, and the table grew unbounded with the history. This is reachable from ironrdp-egfx server output when CompressionMode is Auto or Always and the payload is already compressed.

The fix evicts whole least-recently-seen prefixes down to a low watermark when the table exceeds the cap, so compaction runs at most once per (MAX - target) inserted prefixes and the table stays bounded. Match distance is already capped at MAX_MATCH_DISTANCE, so dropping the oldest prefixes does not change reachable matches; compressible input is unaffected.

Validation

cargo xtask check fmt/lints/tests/typos/locks all pass. The new regression test compresses 100 KB of deterministic high-entropy data, asserts the round trip through Decompressor, and asserts the table stays within MAX_HASH_TABLE_ENTRIES (about 10 s to a few ms).

Notes

No public API change; compact_hash_table and the new constant are private. Output remains valid ZGFX (round trip verified); only the worst-case compaction cost and peak table size change.

### DIFF
--- a/crates/ironrdp-graphics/src/zgfx/compressor.rs
+++ b/crates/ironrdp-graphics/src/zgfx/compressor.rs
@@ -26,6 +26,10 @@ const MAX_POSITIONS_PER_PREFIX: usize = 32;
 /// Trigger hash table compaction when entry count exceeds this
 const MAX_HASH_TABLE_ENTRIES: usize = 50_000;
 
+/// Compaction evicts down to this low watermark, so it runs at most once per
+/// `MAX_HASH_TABLE_ENTRIES - COMPACT_TARGET_ENTRIES` inserted prefixes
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM!

Issue #790, opened 2025-05-15
ironrdp-web becomes unresponsive when remote framebuffer update rapidly


gabrielbauman:
If you visit testufo.com in a browser on a remote RDP server using ironrdp-web, it becomes next to impossible to interact with the session using the mouse or keyboard due to lag. Tested this with the version of the web component tagged 0.9.1 in git.

This also happens when trying to play video in the remote session. The larger or more frequently updated the moving area is on screen, the greater the lag. Covering the moving area with another window or bringing up the task manager ends the lag almost immediately and the session becomes responsive again.

The Microsoft native RDP client seems to handle testufo.com and remote video just fine.

I know the native client offloads video surfaces to a separate channel, but I don't think that happens for canvasses within a browser (could be wrong). Just wanted to capture this here.

glamberson:
@CBenoit, @gabrielbauman: circling back as I said I would.

The WASM client is built. ~4,600 LOC, MIT-licensed, working E2E against a Linux RDP server with H.264 EGFX over Worker-thread decode and Canvas2D `drawImage(VideoFrame)` rendering. The three diagnostic conclusions from my March comment held up empirically: worker decode keeps the main thread responsive, WebCodecs plus `drawImage(VideoFrame)` on a Canvas2D context is GPU-fast (no wgpu needed for the baseline), and EGFX channel registration is what unlocks the bandwidth profile. Both RDCleanPath (proxied) and direct-connect modes work for the connector handshake. Builds clean against current IronRDP master (commit `0282d182`).

I want to put this in your hands. Three reasonable submission shapes; the codebase supports any of them.

## Submission options

**Option 1: absorb as `ironrdp-wasm`.** Move the crate into `crates/ironrdp-wasm` in the IronRDP workspace as a sibling to `ironrdp-web`. I rebrand and license-align, file the PR series that incrementally introduces the crate (probably 4-6 PRs: scaffold, EGFX handler, WebCodecs codec, transport, connection sequence, integration). Devolutions owns it from then on. `ironrdp-web` stays as-is for the existing Cloudflare gateway use case; `ironrdp-wasm` serves the direct-connect, EGFX, video-heavy use case.

[...]

PR #1345, opened 2026-05-31
feat(egfx)!: surface total_frames_decoded on the frame-ack callback

glamberson:
Summary

GraphicsPipelineHandler::on_frame_ack already receives frame_id and queue_depth, but handle_frame_acknowledge drops the third field the client sends in FrameAcknowledgePdu (MS-RDPEGFX 2.2.2.13): total_frames_decoded, the client's running decoded-frame count. encoded_count - total_frames_decoded is an instant measure of the client's decode backlog, which a server needs for closed-loop EGFX flow control. Issue #790 (ironrdp-web becomes unresponsive under rapid framebuffer updates) is the user-facing symptom of having no such signal.

This extends on_frame_ack to pass total_frames_decoded through. The field is already decoded; the handler just dropped it before the callback. ironrdp-egfx is unpublished (publish = false), so the breaking signature change has no external SemVer impact; the only in-tree consumers are the trait default and the egfx testsuite handler, both updated here. ironrdp-server inherits the default and is unchanged.

Validation

cargo xtask check fmt/lints/tests/typos/locks all pass.

Notes

Breaking on the GraphicsPipelineHandler trait (hence feat!), scoped to ironrdp-egfx plus its testsuite. This surfaces the field only; it adds no flow-control policy.

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -767,7 +767,10 @@ pub trait GraphicsPipelineHandler: Send {
     fn on_ready(&mut self, negotiated: &CapabilitySet);
 
     /// Called when a frame has been acknowledged by the client
-    fn on_frame_ack(&mut self, _frame_id: u32, _queue_depth: u32) {}
+    ///
+    /// `total_frames_decoded` is the client's running decoded-frame count
+    /// (MS-RDPEGFX 2.2.2.13), for decode-backlog flow control.
+    fn on_frame_ack(&mut self, _frame_id: u32, _queue_depth: u32, _total_frames_decoded: u32) {}
 
     /// Called when QoE metrics are received from client (V10+)
     fn on_qoe_metrics(&mut self, _metrics: QoeMetrics) {}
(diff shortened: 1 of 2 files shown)


crates/ironrdp-egfx/src/server.rs
```diff
@@ -767,7 +767,10 @@ pub trait GraphicsPipelineHandler: Send {
     fn on_ready(&mut self, negotiated: &CapabilitySet);
 
     /// Called when a frame has been acknowledged by the client
-    fn on_frame_ack(&mut self, _frame_id: u32, _queue_depth: u32) {}
+    ///
+    /// `total_frames_decoded` is the client's running decoded-frame count
+    /// (MS-RDPEGFX 2.2.2.13), for decode-backlog flow control.
+    fn on_frame_ack(&mut self, _frame_id: u32, _queue_depth: u32, _total_frames_decoded: u32) {}
```
CBenoit: style: A style nit, but it’s better to not use `_` in traits as IDE will auto-generate using `_` even though the parameters are used.
```suggestion
    fn on_frame_ack(&mut self, frame_id: u32, queue_depth: u32, total_frames_decoded: u32) {
        let _ = frame_id;
        let _ = queue_depth;
        let _ = total_frames_decoded;
    }
```

CBenoit (approved):
LGTM

PR #1348, opened 2026-06-01
fix(pdu)!: send NetworkAutoDetect over the MCS message channel

glamberson:
Stacked on #1347 (acceptor message-channel negotiation). Until that merges,
the diff here also contains its commit; review the top commit
(`fix(pdu)!: send NetworkAutoDetect over the MCS message channel`). I will
rebase this to master-only once #1347 lands.

NetworkAutoDetect was modeled as a slow-path Share Data PDU
(ShareDataPduType::AutoDetect, a fabricated 0x3B value) carried on the I/O
channel. Per MS-RDPBCGR 2.2.14.3 / 2.2.14.4 the auto-detect PDUs are framed by
a Basic Security Header (SEC_AUTODETECT_REQ / SEC_AUTODETECT_RSP) and ride the
MCS message channel, so neither mstsc nor xfreerdp ever answered the requests.

ironrdp-pdu: add AutoDetectReqPdu / AutoDetectRspPdu, wrapping the request and
response data with the security header (mirroring the multitransport PDUs).
Remove the ShareDataPdu::AutoDetectReq / AutoDetectRsp variants and the
ShareDataPduType::AutoDetect discriminant, which did not correspond to any real
PDUTYPE2. This is the breaking change.

ironrdp-server: send auto-detect requests and receive responses on the message
channel surfaced by the acceptor, with the new framing.

ironrdp-connector and ironrdp-session: the client requests the message channel
(Client Message Channel Data), advertises SUPPORT_NET_CHAR_AUTODETECT, joins the
channel, and answers RTT requests on it. The connector handles connect-time
auto-detection (previously a no-op) and the session handles continuous
auto-detection. Resolves the client side of the message-channel TODO (#140).

Tested: the in-process client/server tests in ironrdp-testsuite-core exercise
the full negotiation and continuous auto-detect; new ironrdp-pdu round-trip
tests cover the security-header framing; the session auto-detect tests were
migrated to the message channel.

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -23,6 +23,8 @@ use crate::{
 pub struct ConnectionResult {
     pub io_channel_id: u16,
     pub user_channel_id: u16,
+    /// MCS channel ID of the message channel, when one was negotiated.
+    pub message_channel_id: Option<u16>,
     pub share_id: u32,
     pub static_channels: StaticChannelSet,
     pub desktop_size: DesktopSize,
@@ -134,6 +136,8 @@ pub struct ClientConnector {
     /// The client address to be used in the Client Info PDU.
     pub client_addr: SocketAddr,
     pub static_channels: StaticChannelSet,
+    /// MCS message channel ID assigned by the server, once negotiated.
+    pub message_channel_id: Option<u16>,
 }
 
 impl ClientConnector {
@@ -143,6 +147,7 @@ impl ClientConnector {
             state: ClientConnectorState::ConnectionInitiationSendRequest,
             client_addr,
             static_channels: StaticChannelSet::new(),
+            message_channel_id: None,
         }
     }
 
@@ -208,6 +213,31 @@ impl ClientConnector {
     }
 }
 
+fn advance_licensing_exchange(
+    mut license_exchange: LicenseExchangeSequence,
+    io_channel_id: u16,
+    user_channel_id: u16,
+    input: &[u8],
+    output: &mut WriteBuf,
+) -> ConnectorResult<(Written, ClientConnectorState)> {
+    let written = license_exchange.step(input, output)?;
+
+    let next_state = if license_exchange.state.is_terminal() {
+        ClientConnectorState::MultitransportBootstrapping {
+            io_channel_id,
+            user_channel_id,
+        }
+    } else {
+        ClientConnectorState::LicensingExchange {
+            io_channel_id,
+            user_channel_id,
+            license_exchange,
+        }
+    };
+
+    Ok((written, next_state))
+}
+
 impl Sequence for ClientConnector {
     fn next_pdu_hint(&self) -> Option<&dyn PduHint> {
         match &self.state {
@@ -220,7 +250,20 @@ impl Sequence for ClientConnector {
             ClientConnectorState::BasicSettingsExchangeWaitResponse { .. } => Some(&ironrdp_pdu::X224_HINT),
             ClientConnectorState::ChannelConnection { channel_connection, .. } => channel_connection.next_pdu_hint(),
             ClientConnectorState::SecureSettingsExchange { .. } => None,
-            ClientConnectorState::ConnectTimeAutoDetection { .. } => None,
+            ClientConnectorState::ConnectTimeAutoDetection { .. } => {
+                // Wait for input only when a message channel was negotiated, so
+                // we can receive connect-time auto-detect requests there. With a
+                // message channel the server always sends a PDU next in this phase
+                // (a connect-time Auto-Detect Request on the message channel, or
+                // the first licensing PDU on the I/O channel), so waiting here
+                // cannot stall. Without one, this state reads nothing and
+                // transitions straight to licensing.
+                if self.message_channel_id.is_some() {
+                    Some(&ironrdp_pdu::X224_HINT)
+                } else {
+                    None
+                }
+            }
             ClientConnectorState::LicensingExchange { license_exchange, .. } => license_exchange.next_pdu_hint(),
             ClientConnectorState::MultitransportBootstrapping { .. } => None,
             ClientConnectorState::CapabilitiesExchange {
@@ -390,9 +433,10 @@ impl Sequence for ClientConnector {
                     return Err(general_err!("can't satisfy server security settings"));
                 }
 
-                if server_gcc_blocks.message_channel.is_some() {
-                    warn!("Unexpected ServerMessageChannelData GCC block (not supported)");
-                }
+                self.message_channel_id = server_gcc_blocks
+                    .message_channel
+                    .as_ref()
+                    .map(|data| data.mcs_message_channel_id);
 
                 if server_gcc_blocks.multi_transport_channel.is_some() {
                     warn!("Unexpected MultiTransportChannelData GCC block (not supported)");
@@ -426,7 +470,9 @@ impl Sequence for ClientConnector {
                         channel_connection: if skip_channel_join {
                             ChannelConnectionSequence::skip_channel_join()
                         } else {
-                            ChannelConnectionSequence::new(io_channel_id, static_channel_ids)
+                            let mut join_channel_ids = static_channel_ids;
+                            join_channel_ids.extend(self.message_channel_id);
+                            ChannelConnectionSequence::new(io_channel_id, join_channel_ids)
                         },
                     },
                 )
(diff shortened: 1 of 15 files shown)


crates/ironrdp-connector/src/connection.rs
```diff
@@ -485,12 +501,37 @@ impl Sequence for ClientConnector {
...
+                    let written =
+                        respond_to_connect_time_autodetect(pdu.request, message_channel_id, user_channel_id, output)?;
+                    (
+                        written,
+                        ClientConnectorState::ConnectTimeAutoDetection {
+                            io_channel_id,
+                            user_channel_id,
+                        },
+                    )
+                } else {
```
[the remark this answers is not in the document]
glamberson: Reworked this to demux by MCS channel rather than by whether the payload parses as auto-detect. A PDU on the message channel is never a licensing PDU, so a message-channel PDU that is not an auto-detect request is now ignored (we stay in `ConnectTimeAutoDetection`) instead of being forwarded to the licensing sequence. In the canonical sequence the PDUs you cite (multitransport bootstrap, heartbeat) arrive after licensing, so in practice this hardens the phase-exit invariant to be channel-based rather than fixing a reachable abort, but the channel is the correct discriminator and matches how the channels are separated in the spec.

crates/ironrdp-server/src/server.rs
```diff
@@ -1391,26 +1426,18 @@ impl RdpServer {
 /// `deactivate_all()`. In practice, RDP clients do not validate `share_id`
 /// on server-initiated PDUs, but a future refactor could thread the
```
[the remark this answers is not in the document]
glamberson: Rewrote the doc comment to match the function: it now describes encoding the Auto-Detect Request for the MCS message channel with the SEC_AUTODETECT_REQ Basic Security Header per [MS-RDPBCGR] 2.2.14.3, and dropped the stale `share_id` paragraph, which no longer applies on this path.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -499,9 +560,39 @@ impl Sequence for ClientConnector {
...
-                },
-            ),
+                    );
+                    // If a PDU was read (message channel present) it is the first
+                    // licensing PDU; process it here and advance the same way the
+                    // LicensingExchange state would, so it is not stepped twice.
+                    // Otherwise nothing was read and the licensing sequence runs
+                    // from its first step as before.
+                    if self.message_channel_id.is_some() {
+                        let written = license_exchange.step(input, output)?;
```
CBenoit: suggestion: This inlines the first licensing step and re-implements the terminal-state transition (is_terminal() -> MultitransportBootstrapping else LicensingExchange) that already lives in the LicensingExchange arm. Two copies of the same control flow will drift the next time we touch the licensing sequence. Can we route this first PDU through the normal LicensingExchange entry instead of stepping it here, so the transition logic stays in one place?

crates/ironrdp-connector/src/connection.rs
```diff
@@ -485,12 +501,57 @@ impl Sequence for ClientConnector {
...
+                            decode::<rdp::autodetect::AutoDetectReqPdu>(&data.user_data).ok()
+                        }
+                        _ => None,
+                    }
+                });
+                let on_message_channel = self.message_channel_id.is_some_and(|message_channel_id| {
+                    decode::<X224<mcs::McsMessage<'_>>>(input).is_ok_and(|mcs| {
+                        matches!(mcs.0, mcs::McsMessage::SendDataIndication(data) if data.channel_id == message_channel_id)
+                    })
+                });
```
CBenoit: nitpick: We decode input as X224<McsMessage> twice here: once to build `autodetect`, once for the `on_message_channel` guard. Decode once, match on the SendDataIndication, and branch on whether it parses as AutoDetectReqPdu.

crates/ironrdp-connector/src/connection.rs
```diff
@@ -212,7 +217,15 @@ impl Sequence for ClientConnector {
...
-            ClientConnectorState::ConnectTimeAutoDetection { .. } => None,
+            ClientConnectorState::ConnectTimeAutoDetection { .. } => {
+                // Wait for input only when a message channel was negotiated, so
+                // we can receive connect-time auto-detect requests there.
+                if self.message_channel_id.is_some() {
+                    Some(&ironrdp_pdu::X224_HINT)
+                } else {
+                    None
+                }
+            }
```
CBenoit, line 266: question (non-blocking): Returning X224_HINT only when message_channel_id is Some relies on the server always sending something next in this phase (an auto-detect req, or the first licensing PDU) and never stalling. I believe that's true per the canonical sequence: could you add a one-line comment asserting it?

crates/ironrdp-connector/src/connection.rs
```diff
@@ -639,6 +731,27 @@ pub fn encode_send_data_request<T: Encode>(
...
+        AutoDetectRequest::RttRequest { sequence_number, .. } => {
+            let response = AutoDetectRspPdu::new(AutoDetectResponse::RttResponse { sequence_number });
+            let written = encode_send_data_request(user_channel_id, message_channel_id, &response, output)?;
+            Written::from_size(written)
+        }
+        // The Network Characteristics Result is informational, and bandwidth
+        // measurement is driven implicitly by the connect-time payload exchange.
+        // Neither requires an immediate reply.
+        _ => Ok(Written::Nothing),
+    }
```
CBenoit: question (non-blocking): We only reply to RttRequest at connect time and treat bandwidth-measure / NetworkCharacteristicsResult as no-ops. I’m not really sure here, I asked Claude and it said that some connect-time sequences do expect a Bandwidth Measure Results response. Have you confirmed mstsc/a real server doesn't block waiting for one here? If it's intentional, the comment is enough; just want it to be a decision, not an omission.

CBenoit (commented):
Thanks for chasing this down, the protocol diagnosis is right: 0x3b was never a real PDUTYPE2, and auto-detect belongs on the message channel behind a Basic Security Header. My one real concern is the connector connect-time state machine. Also I think the migrated tests cover the continuous (session) path well, but I don't see coverage for the connector connect-time branch specifically (the inlined-licensing path and the "ignore a non-autodetect message-channel PDU and keep listening" path). Those are the two riskiest branches in the PR. Could you add a test for that?

crates/ironrdp-connector/src/connection.rs
glamberson: Good catch. I pulled the licensing step plus the terminal-state transition into a single helper, advance_licensing_exchange, and both the connect-time auto-detect branch and the LicensingExchange arm now call it, so the is_terminal -> MultitransportBootstrapping / LicensingExchange transition lives in exactly one place.

I kept it as a shared helper rather than routing the PDU back through the LicensingExchange state because, when the message channel is negotiated, ConnectTimeAutoDetection's next_pdu_hint has already caused this first licensing PDU to be read before this arm runs, so it has to be consumed in the same step or it is lost. The no-message-channel path still transitions to LicensingExchange with nothing written and lets the normal arm step the next PDU. The helper gives the single source of truth without changing the one-PDU-per-step contract.

crates/ironrdp-connector/src/connection.rs
glamberson: Done. There is now a single decode into message_channel_pdu, which is Some((message_channel_id, data)) only when the PDU is a SendDataIndication on the negotiated message channel. We then branch on whether data.user_data parses as an AutoDetectReqPdu (answer it and keep listening) or not (a message-channel PDU we ignore in this phase); None means it is the first licensing PDU on the I/O channel. The second X224<McsMessage> decode is gone.

crates/ironrdp-connector/src/connection.rs
glamberson: Added. The comment on this arm now states the invariant: with a message channel negotiated, the server always sends a PDU next in this phase (a connect-time Auto-Detect Request on the message channel, or the first licensing PDU on the I/O channel), so the wait here cannot stall; without a message channel the state reads nothing and transitions straight to licensing. Your read of the canonical sequence matches mine.

crates/ironrdp-connector/src/connection.rs
glamberson: It is intentional, and you are right that the spec defines a reply: a connect-time Bandwidth Measure Stop ([MS-RDPBCGR] 2.2.14.1.4) is meant to be answered with a Bandwidth Measure Results PDU, so no-op'ing it is a real conformance gap, not a non-event.

What makes it safe to skip here is that connect-time auto-detect is informational: the server advances to licensing whether or not it receives bandwidth results. The strongest evidence is IronRDP's own history. Before this change the connector answered nothing at all at connect time, not even RTT (the old path was a dead no-op on the wrong channel), and it still connected to Windows and mstsc-class servers. If a server gated the sequence on auto-detect completion, that never would have worked. I have not run a fresh mstsc capture for this specific case, so I am leaning on that shipping history plus the informational model rather than a new packet trace.

I have made it an explicit decision in the code: the comment now states that Bandwidth Measure Stop is spec-defined to warrant a Results reply, that we deliberately skip it because the sequence does not block on it, and that full connect-time bandwidth measurement (replying to Bandwidth Measure Stop with Bandwidth Measure Results) is left as a follow-up rather than part of this framing fix.

glamberson:
Added connector connect-time tests in a new tests/connector/autodetect.rs, covering the ConnectTimeAutoDetection demux you flagged: an Auto-Detect Request on the message channel is answered and the phase continues; an unrelated message-channel PDU is ignored and the phase continues; and a first licensing PDU on the I/O channel advances the connector out of auto-detection into the licensing path. That exercises the two branches you called out as riskiest plus the auto-detect path itself.

The four inline points are addressed in their threads: the inlined licensing transition is now a shared helper so the terminal-state transition lives in one place; the message-channel input is decoded once; the next_pdu_hint gate carries a comment stating why it cannot stall; and the connect-time bandwidth no-op is now an explicit documented decision, with the Bandwidth Measure Results reply noted as a follow-up.

I have also rebased the branch onto current master.

CBenoit (approved):
LGTM, thank you!

CBenoit:
CI is broken because of the problem you already addressed in https://github.com/Devolutions/IronRDP/pull/1390
I couldn’t rebase on master and force-push your branch even though edits are supposed to be allowed to maintainers. I’m sorry for the trouble, I’ll make sure to merge once it’s rebased again.

glamberson:
Yeah, no biggie. Rebased again, thanks.

glamberson:
@CBenoit I'm just waiting on your action regarding merge order on this and #1435 to see how these sort out before looking at your refactoring work in Issue #1422 etc. further. Thanks.

CBenoit:
@glamberson Thank you! As for https://github.com/Devolutions/IronRDP/issues/1422, I went ahead with phase 1 which was the most straightforward and easy win, but I don’t have time to go ahead with the other phases immediately. You can expect a significant cool-down period, during which I’m open to suggestions if anything looks off to you.

And oops, I kind of lost track again, this PR was ready to be merged, going ahead now. Fortunately, it touches a strictly different set of API, so both changes are compatible. Just in time for the next release batch.

PR #1358, opened 2026-06-02
fix(dvc)!: add channel_id parameter to DvcChannelListener::create

uchouT:
Add `channel_id` args in `DynamicChannelListener` trait's create method.

`channel_id` is required when implementing `UrbdrcListener` in #1137


CBenoit (approved):
LGTM! I changed the commit from `refactor!` to `fix!` because if it’s not just changing an implementation detail, it’s changing the public API, and that is something consumers need to care about 🙂

PR #1359, opened 2026-06-02
feat(rdpsnd)!: misuse-resistant format negotiation for RdpsndServerHandler

### LINKED ISSUE #1356: rdpsnd: make RdpsndServerHandler::start misuse-resistant (hand it the common formats, let the crate compute wFormatNo)
Following up on the discussion in #1343 (where the `wFormatNo` contract was documented) — @CBenoit suggested turning that documented-but-error-prone contract into a misuse-resistant API, and I'd like to track it here.

### Problem

`RdpsndServerHandler::start` is currently:

```rust
fn start(&mut self, client_format: &ClientAudioFormatPdu) -> Option<u16>;
```

The returned `u16` is stamped directly onto every Wave/Wave2 PDU as `wFormatNo`, which the client resolves as an index into **its own** Client Audio Formats list. That puts two identical, error-prone steps on every implementer:

1. Intersect the server's advertised formats (`get_formats()`) with the client's list.
2. Translate the chosen format back into its index *in the client's list*.

Step 2 is the silent footgun: returning the server-list index instead works only while the lists coincide, and otherwise yields `wFormatNo >= NumberOfClientFormats`, which FreeRDP/Thincast hard-reject (`rdpsnd_recv_wave2_pdu`) → every wave silently dropped, no audio. #1343 documents this; this issue is about making it unrepresentable.

### Proposal

Hand `start` only the formats common to both sides, each pre-tagged with its wire `wFormatNo`, and let the crate own the index math + validation:

```rust
/// Formats supported by BOTH sides, in the server's advertised preference
/// order. Each entry carries the wire `wFormatNo` (its index in the client's
/// own list), so the consumer never computes it. Guaranteed non-empty: if
/// there is no common format, `start` is not called at all.
pub struct NegotiatedFormat {
    pub format: AudioFormat,
    wformat_no: u16, // crate-owned; stamped onto every Wave/Wave2
}

fn start(&mut self, common: &[NegotiatedFormat]) -> Option<usize>; // index into `common`, or None to decline
```

Because the consumer can only return something *from* `common`, "is the selected format part of the common set?" is true by construction, and the crate can hard-validate the returned index.

### Open question (from @CBenoit)

How far to take the abstraction — in particular, whether consumers ever benefit from seeing the full client format set. My read: they don't. The only uses beyond the two steps above are logging the set length and `version`, and `version`/`flags` are already reachable via `RdpsndServer::version()`/`flags()`. So I'd keep `get_formats()` for what the server advertises, hand `start` only the negotiated intersection, and skip calling it when there's nothing in common.

One shape decision: `start` returning an index into `common` (above) vs. returning the chosen `&AudioFormat` / a small `FormatId` newtype. Happy to implement whichever is preferred.

This is an API/semver change to `ironrdp-rdpsnd`; I can put up the PR once the shape is settled.

clintcan:
Implements the misuse-resistant `RdpsndServerHandler` format-negotiation API discussed in #1356, in the two-method shape you proposed there.

### Problem

The old `start(&ClientAudioFormatPdu) -> Option<u16>` made every handler compute `wFormatNo` itself — an index into the *client's* format list. Returning the wrong index (e.g. a server-list index) yields `wFormatNo >= NumberOfClientFormats`, which a compliant client rejects, silently dropping all audio. Every handler also re-implemented the same server∩client intersection. (Documented in #1343.)

### Change

Move the negotiation into the crate and split selection from lifecycle:

```rust
fn choose_format<'a>(&mut self, common: &'a [NegotiatedFormat]) -> Option<&'a NegotiatedFormat>;
fn start(&mut self, format: &NegotiatedFormat);
```

- The crate computes `common` — the formats from `get_formats()` the client also advertised, in the server's preference order, each tagged with its client-list `wFormatNo` — and calls `choose_format`, then `start` with the chosen format.
- `NegotiatedFormat` has a private `wformat_no` and no public constructor, and `choose_format` returns a *borrow* of an element of `common`, so a handler cannot pick a format the client didn't accept nor emit an out-of-range `wFormatNo`.
- `choose_format` is not called when nothing is in common.
- `choose_format` (pure selection) and `start` (encoder init / producer startup) are kept distinct, per your note that `start` is also where producers initialize.

Resolves the footgun in #1343. Closes #1356.

### Two decisions worth a look (easy to change)

1. **`NegotiatedFormat.format` is private with a `format()` accessor** rather than `pub format` — `clippy::partial_pub_fields` rejects mixing a `pub` field with the private `wformat_no`, and this matches the direction of #1331. Happy to switch to `pub` + `#[allow]` if you prefer.
2. **Format matching uses WAVEFORMATEX identity** (`audio_format_eq`: wave-format tag, channels, sample rate, bit depth), deliberately ignoring derived fields (`n_avg_bytes_per_sec`, `n_block_align`) and the codec `data` blob, which a client may not echo byte-for-byte. This is looser than the type's derived `==`. I went with identity matching because strict equality can miss formats a client genuinely supports (→ silent no-audio, the very bug this targets), but if you'd rather the crate use full `==`, it's a one-line change.

### Notes

- **Breaking change** (pre-v1): replaces `RdpsndServerHandler::start`. Known downstream implementors that will need a small update: `lamco-rdp-server`, `hypr-rdp`, `cosmic-ext-rdp-server`, `ARISU`, `macrdp`. (Per the #1356 discussion, the straight replacement was preferred over an additive/deprecated path.)
- The bundled example server is updated to the new shape.

### Testing

- New unit tests for `negotiate_formats` / `audio_format_eq`: client-index mapping, the PCM-only/AAC-first regression, empty overlap, and identity equality. (Enabled the lib test harness in `Cargo.toml`.)
- `cargo clippy -p ironrdp-rdpsnd --all-targets -- -D warnings` and the example (`--features cliprdr,connector,rdpsnd,server`) are clean; `cargo test -p ironrdp-rdpsnd` passes.

### DIFF
--- a/crates/ironrdp-rdpsnd/Cargo.toml
+++ b/crates/ironrdp-rdpsnd/Cargo.toml
@@ -14,7 +14,7 @@ categories.workspace = true
 
 [lib]
 doctest = false
-test = false
+# test = false
 
 [features]
 default = []
--- a/crates/ironrdp-rdpsnd/Cargo.toml
+++ b/crates/ironrdp-rdpsnd/Cargo.toml
@@ -14,11 +14,16 @@ categories.workspace = true
 
 [lib]
 doctest = false
-# test = false
+test = false
 
 [features]
 default = []
 std = []
+# Internal (PRIVATE!) feature used to aid testing.
+# Don't rely on this whatsoever. It may disappear at any time.
+# It uses `visibility` to expose otherwise-private negotiation helpers to the
+# integration testsuite (the lib has no inline test harness — `test = false`).
+__test = ["dep:visibility"]
 
 [dependencies]
 bitflags = "2.11"
@@ -26,6 +31,7 @@ tracing = { version = "0.1", features = ["log"] }
 ironrdp-svc = { path = "../ironrdp-svc", version = "0.7" } # public
 ironrdp-core = { path = "../ironrdp-core", version = "0.2", features = ["alloc"] }
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.8", features = ["alloc"] } # public
+visibility = { version = "0.1", optional = true }
 
 [lints]
 workspace = true
--- a/crates/ironrdp-rdpsnd/src/server.rs
+++ b/crates/ironrdp-rdpsnd/src/server.rs
@@ -58,6 +58,14 @@ impl NegotiatedFormat {
     pub fn format(&self) -> &pdu::AudioFormat {
         &self.format
     }
+
+    /// Test-only accessor for the crate-private `wformat_no`, exposed for the
+    /// integration testsuite behind the private `__test` feature. Not a stable API.
+    #[cfg(feature = "__test")]
+    #[doc(hidden)]
+    pub fn wformat_no(&self) -> u16 {
+        self.wformat_no
+    }
 }
 
 /// Handler for the server side of the Audio Output Virtual Channel (`RDPSND`).
@@ -97,8 +105,14 @@ pub trait RdpsndServerHandler: Send + core::fmt::Debug {
     /// [`choose_format`]. This is the lifecycle hook: initialize encoder state,
     /// spawn the producer, etc. Waves are then emitted via [`RdpsndServer::wave`].
     ///
+    /// Return `Err` if initialization fails (e.g. the encoder can't be created).
+    /// The crate then **declines the negotiated format** — exactly as if
+    /// [`choose_format`] had returned [`None`] — rather than leaving the channel
+    /// "negotiated" but silently producing no audio. The error is logged by the
+    /// crate.
+    ///
     /// [`choose_format`]: RdpsndServerHandler::choose_format
-    fn start(&mut self, format: &NegotiatedFormat);
+    fn start(&mut self, format: &NegotiatedFormat) -> Result<(), Box<dyn RdpsndError>>;
 
     /// Called when the audio stream is torn down (e.g. the client closed the
     /// channel or the session ended).
@@ -222,6 +236,7 @@ impl RdpsndServer {
 /// client resolves waves against (MS-RDPEA). The result mirrors the server's
 /// ordering so the handler can express preference simply by `get_formats`
 /// order, while the `wFormatNo` always points into the client list.
+#[cfg_attr(feature = "__test", visibility::make(pub))]
 fn negotiate_formats(
     server_formats: &[pdu::AudioFormat],
     client_formats: &[pdu::AudioFormat],
@@ -241,16 +256,24 @@ fn negotiate_formats(
         .collect()
 }
 
-/// Compare two audio formats by their WAVEFORMATEX identity — wave format tag,
-/// channel count, sample rate, and bit depth. Derived fields
-/// (`n_avg_bytes_per_sec`, `n_block_align`) and the codec-specific `data` blob
-/// are deliberately ignored: a client echoes back a format it accepts but is
-/// not guaranteed to reproduce those byte-for-byte.
+/// Compare two audio formats for negotiation. The WAVEFORMATEX identity fields
+/// — wave format tag, channel count, sample rate, bit depth — must match, and so
+/// must the codec-specific extra-data blob (`data`).
+///
+/// The two derived fields (`n_avg_bytes_per_sec`, `n_block_align`) are
+/// deliberately ignored: they are computable from the others and a client may
+/// legitimately not echo them back byte-for-byte. The `data` blob is a different
+/// category, though — for codecs whose extra-format bytes carry real
+/// configuration (AAC's HEAACWAVEINFO extra data is the clear case, MS-RDPEA
+/// 2.2.2.1.1's `cbSize` + extra data), ignoring it could match two genuinely
+/// incompatible formats, so it IS compared.
+#[cfg_attr(feature = "__test", visibility::make(pub))]
 fn audio_format_eq(a: &pdu::AudioFormat, b: &pdu::AudioFormat) -> bool {
     a.format == b.format
         && a.n_channels == b.n_channels
         && a.n_samples_per_sec == b.n_samples_per_sec
         && a.bits_per_sample == b.bits_per_sample
+        && a.data == b.data
 }
 
 impl_as_any!(RdpsndServer);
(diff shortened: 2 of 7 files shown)


clintcan:
Friendly ping on this when you have a moment 🙂 — it's been a few weeks and I wanted to make sure it didn't slip off the radar.

This implements the exact two-method shape we settled on in #1356: `choose_format(&mut self, common: &[NegotiatedFormat]) -> Option<&NegotiatedFormat>` for pure selection + `start(&mut self, format: &NegotiatedFormat)` as the lifecycle hook, with the crate owning the `wFormatNo` computation/validation. CI is green and it's currently mergeable (no conflicts).

Two small in-PR points I'd flagged for your call are still open whenever you review: (1) the `format()` accessor vs a `pub format` field (clippy `partial_pub_fields` / #1331), and (2) `audio_format_eq` doing a WAVEFORMATEX-identity match vs a full `==`. Happy to adjust either, and glad to rebase if it drifts before you get to it. No rush — just flagging it's ready.

CBenoit:
Thank you! I added this PR to my review queue! I’m requesting a Copilot review in the meantime.

crates/ironrdp/examples/server.rs
```diff
@@ -290,30 +279,32 @@ impl RdpsndServerHandler for SndHandler {
...
+        let fmt = format.format().clone();
 
         let mut opus_enc = if fmt.format == WaveFormat::OPUS {
             let n_channels: opus2::Channels = match fmt.n_channels {
                 1 => opus2::Channels::Mono,
                 2 => opus2::Channels::Stereo,
                 n => {
                     warn!("Invalid OPUS channels: {}", n);
-                    return Some(0);
+                    return;
```
CBenoit: issue: start() now returns (). So format_no is committed before start runs, and start has no way to say "I couldn't initialize." We can see in this example where the encoder creation fails and we do a bare return, after which the library still sets format_no = Some(wformat_no) and considers the stream live. The problem is that at this point no producer task was spawned, and we have a silent "negotiated, no audio" state, which is a failure class I wish we could kill in this PR. I think we need to either return Result<(), E> or a bool so the library can roll back to None.

crates/ironrdp-rdpsnd/Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories.workspace = true
 
 [lib]
 doctest = false
-test = false
+# test = false
```
CBenoit: issue: `test = false` is deliberate policy here, not incidental, let's not re-enable it.

Keeping the lib free of an inline test harness is a constraint we hold across the workspace, and a commented-out manifest key will just rot. The behavior is well worth testing though. Here is a route that keeps the policy intact:

- add an `__internal` feature (doc-hidden, excluded from `default`),
- behind it, expose thin `#[doc(hidden)] pub` shims over the private items: `negotiate_formats`, `audio_format_eq`, and a `wformat_no()` accessor on `NegotiatedFormat` (the current tests read the private `wformat_no` field directly, which only works because they live in-module, so the shim has to surface that too),
- move the tests into the appropriate testsuite crate

`[lib] test = false` stays, the internal surface stays out of the public API and docs, and the assertions you already wrote port over almost verbatim.

crates/ironrdp-rdpsnd/src/server.rs
```diff
@@ -173,6 +216,43 @@ impl RdpsndServer {
...
+/// channel count, sample rate, and bit depth. Derived fields
+/// (`n_avg_bytes_per_sec`, `n_block_align`) and the codec-specific `data` blob
+/// are deliberately ignored: a client echoes back a format it accepts but is
+/// not guaranteed to reproduce those byte-for-byte.
+fn audio_format_eq(a: &pdu::AudioFormat, b: &pdu::AudioFormat) -> bool {
+    a.format == b.format
+        && a.n_channels == b.n_channels
+        && a.n_samples_per_sec == b.n_samples_per_sec
+        && a.bits_per_sample == b.bits_per_sample
+}
```
CBenoit: issue: ignoring the codec `data` blob is unsafe for some formats we actually advertise.

Agreed on the derived fields (`n_avg_bytes_per_sec` / `n_block_align` are computable and a client may legitimately not echo them), but the `data` blob is a different category: for codecs whose extra-format bytes carry real configuration (AAC is the clear case per MS-RDPEA 2.2.2.1.1's cbSize/extra data, worth confirming per codec in our `get_formats`), blanket-ignoring it can match two genuinely incompatible variants.

Concretely, with the current looseness plus `format: server_format.clone()` in `negotiate_formats`: we stream the *server's* definition while `wFormatNo` points the client at *its* entry. If they differ only in `data`, the server encodes one config and the client decodes another, garbled/no audio. Two distinct server formats differing only in `data` also collapse onto the same client index, yielding two `NegotiatedFormat`s sharing one `wFormatNo`.

PCM is fine (no extra data); the codecs that aren't need either codec-aware matching or `data` included in the comparison for those. Note `equality_uses_waveformatex_identity_only` currently encodes the too-loose decision, so it'd flip with this.

crates/ironrdp-rdpsnd/src/server.rs
```diff
@@ -220,8 +300,27 @@ impl SvcProcessor for RdpsndServer {
...
+                    // `chosen` borrows `common` (not `self`), so the encoder
+                    // is read off it and the handler is free to borrow again
+                    // for the `start` lifecycle hook.
+                    let wformat_no = chosen.wformat_no;
+                    self.handler.start(chosen);
+                    Some(wformat_no)
+                } else {
+                    debug!("Handler declined every common audio format; audio disabled");
+                    None
+                };
```
CBenoit: thought: I would value a test on the `SvcProcessor` wiring over the isolated helper tests; but fine as a follow-up, or I may pick it up myself.

The helpers are well covered; the part that's new and easy to get wrong is the state-machine path in `process`: `choose_format` skipped when `common` is empty, decline -> `None`, the captured `wformat_no` matching the chosen format, `start` called exactly once. A black-box test: drive the processor with a Client Audio Formats PDU through a fake handler, emit a wave, assert the `wFormatNo` on the outgoing PDU; it would exercises all of that through the public surface, so it would live as a normal `tests/` integration test and wouldn't even need the `__internal` shim from the Cargo.toml thread.

Not blocking: happy to take it as a follow-up PR, or leave it 🙂

CBenoit (commented):
Finally went through it! Thank you, I think this is the right direction, and  I love the new API.

clintcan:
Thanks for the thorough review @CBenoit! Pushed a follow-up addressing all of it:

**`start()` can now fail (your `examples/server.rs:299` + Copilot's `server.rs:318`).** `start` returns `Result<(), Box<dyn RdpsndError>>`. The crate commits `format_no` *before* calling `start` (so a producer that emits a wave during `start` can't race an unset index), and on `Err` rolls it back to `None` — declining the format exactly as `choose_format` returning `None` would, instead of the silent "negotiated, no audio" state. The example's two encoder-init failures now `return Err(...)`.

**`test = false` kept (your `Cargo.toml:17`).** Restored `test = false`, added a private feature that exposes `negotiate_formats` / `audio_format_eq` and a `wformat_no()` accessor on `NegotiatedFormat`, and moved the tests into `ironrdp-testsuite-core` (`tests/rdpsnd/server.rs`). One deviation from your suggestion: I named the feature **`__test`** and used **`visibility::make(pub)`** rather than `__internal` + hand-written shims — that's the existing workspace idiom (`ironrdp-cliprdr` does exactly this, `ironrdp-testsuite-core` already pulls crates in via `__test`, and `__test` is in xtask's `EXCLUDE_FEATURES` so the feature-matrix skips it; `__internal` wouldn't be excluded). Same outcome — doc-hidden, out of the public API/docs. Happy to rename if you'd prefer `__internal`.

**`data` is now compared (your `server.rs:254`).** `audio_format_eq` compares the extra-data blob; the two derived fields stay ignored. `cbSize=0` decodes to `None` on both sides (`pdu/mod.rs` decode), so PCM/AAC are unaffected, while AAC variants that differ only in HEAACWAVEINFO no longer collapse. The old `equality_uses_waveformatex_identity_only` test flipped as you predicted (now `equality_ignores_derived_fields_but_not_extra_data`), plus a new `extra_data_must_match_for_otherwise_identical_formats`.

**`SvcProcessor` wiring test (your `server.rs:323`).** Went ahead and added it — black-box through the public surface (no `__test` shim), driving a fake handler: a shared format streams (`start` called exactly once, `format_no` committed), no common format skips `choose_format`, and a failing `start` declines (no wave streamed) — which also covers the new roll-back path.

Green locally: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --features helper,__bench --locked -- -D warnings`, and `cargo test --workspace --locked` all pass (also verified rdpsnd builds + clippies clean both with and without `__test`).

CBenoit (approved):
Thank you! Very nice improvement, I'm sure it will be useful to prevent subtle bugs =)

PR #1199, opened 2026-04-01
test(graphics): add progressive RFX round-trip tests and documentation

glamberson:
Final PR in the multi-codec EGFX implementation described in #1158 (Section 4). Its two prerequisites have since merged, so this now applies directly to master.

## Summary

Add encode/decode round-trip tests verifying bounded-tolerance reconstruction through the Progressive RFX pipeline. Coverage:

- **First-pass round-trip** (`first_pass_encode_decode_round_trip_lossless` + `_reduce_extrapolate`): forward DWT → progressive quantization → RLGR1 entropy code → RLGR1 decode → progressive dequant → inverse DWT. With LOSSLESS quants the only residual error is LeGall 5/3 DWT integer truncation; assertions `max_err ≤ 4` (standard) / `≤ 6` (reduce-extrapolate).
- **First-pass with quantization** (`first_pass_encode_decode_with_quantization`): the same pipeline at realistic non-lossless quants; assertion `mean_err < 200` for the flat-tile input.
- **Upgrade-pass round-trip** (`upgrade_pass_encode_decode_round_trip`): `encode_upgrade_pass` + `decode_upgrade_pass` on the SRL + raw-bit wire format, with synthetic post-DWT post-quant coefficients. Asserts the monotonic-refinement contract (MS-RDPRFX 3.1.8.1.7.2): post-decode L1 distance to the refined target cannot exceed pre-decode distance.
- **Color conversion round-trip** (`rgba_ycbcr_reconstruct_round_trip`): RGB → YCbCr (forward BT.601) → RGB (inverse BT.601 matching `TileState::reconstruct_to_rgba`) on a gradient input; assertion `max_err ≤ 2` per channel.
- **Quantization round-trip** (`quantize_dequantize_ccq_round_trip`): `quantize_component_ccq` + `dequantize_component_ccq` on varied input.

Module-level documentation expanded with the decode/encode pipeline layout, state-management types, and progressive-quantization notes.

A libFuzzer-driven round-trip oracle and full EGFX-PDU integration round-trip remain follow-ups; the unit-level coverage above is the algorithmic foundation those build on.

## Stack ordering

No longer stacked. This was step 3 of 3 behind #1197 (merged 2026-05-21) and #1198 (merged 2026-05-25); with both upstream it has been rebased onto master and carries only its own two commits.

## Test plan

[...]

PR #1370, opened 2026-06-10
fix(graphics): mirror decoder GR-mode adaptation in RLGR1 encoder

rdeangel:
## Problem

In `rlgr::encode`, the RLGR1 GR-mode zero branch updates `kp` with `UP_GR` (+4), while the decoder's corresponding branch (`compute_rlgr1_magnitude`) uses `UQ_GR` (+3), which is the value MS-RDPRFX 3.1.8.1.7.2 and FreeRDP's `rfx_rlgr.c` prescribe for GR mode. Because RLGR adaptation is implicit, encoder and decoder must update `k` identically from the coded symbols alone; with this mismatch the two sides' `k` diverges as soon as a zero value is coded in GR mode, and the remainder of the bitstream is misinterpreted (mode selection and run lengths depend on `k`).

## Evidence

A randomized encode→decode round-trip (deterministic LCG, mostly-zero inputs with small magnitudes, i.e. the shape of quantized wavelet coefficients) failed on **1877 of 2000** inputs before this change. Example:

```
input:   [0, 2, -1, 7, 0, 0,  4,  -6, 7, 0, 1, 0, -4,  0,  0,  4, 1]
decoded: [0, 2, -1, 7, 0, 0,  2, -16, 5, 0, 2, 1,  4, -1, -1, -2, 0]
```

The divergence starts right after the first GR-mode zero. With the one-line fix, all 2000 round-trips pass.

## Changes

- Use `UQ_GR` in the GR-mode zero branch of the RLGR1 encoder, mirroring the decoder.
- Add a `rlgr1_round_trip_randomized` regression test (`rlgr.rs` previously had no tests).

Found while debugging RemoteFX Progressive (EGFX) interop against xrdp in an IronRDP-based client.

### DIFF
--- a/crates/ironrdp-graphics/src/rlgr.rs
+++ b/crates/ironrdp-graphics/src/rlgr.rs
@@ -122,7 +122,11 @@ pub fn encode(mode: EntropyAlgorithm, input: &[i16], tile: &mut [u8]) -> Result<
                         let two_ms = get_2magsign(input_first);
                         code_gr(&mut bits, &mut krp, two_ms);
                         if two_ms == 0 {
-                            kp = min(kp + UP_GR, KP_MAX);
+                            // Must mirror the decoder's GR-mode adaptation
+                            // (`compute_rlgr1_magnitude`): `UQ_GR` on a zero,
+                            // `DQ_GR` on a non-zero. Using `UP_GR` here makes the
+                            // adaptive `k` diverge between encoder and decoder.
+                            kp = min(kp + UQ_GR, KP_MAX);
                         } else {
                             kp = kp.saturating_sub(DQ_GR);
                         }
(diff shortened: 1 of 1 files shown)


glamberson:
Nice catch. The `UQ_GR` fix is correct: it brings the GR-mode `kp` update in line with the decoder's `compute_rlgr1_magnitude`.

One note for merge ordering: elmarco's draft #1179 fixes this same `UP_GR`/`UQ_GR` line too, alongside two other RLGR encoder bugs it found, so there's a one-line overlap between the two PRs. No opinion from me on sequencing, just flagging it so the fix doesn't get duplicated. Whatever's easiest: this lands as a focused fix and #1179 rebases without that hunk, or it waits on #1179.

Good to see this one fixed regardless. It's on the RemoteFX Progressive encode path that #1199's round-trip tests cover.

rdeangel:
Thanks for flagging the overlap with #1179. If this lands first, elmarco can drop the overlapping hunk on rebase, otherwise happy to wait, whatever you prefer.

CBenoit:
Reviewed the RLGR1 encoder and decoder adaptation paths: the zero GR-code branch now applies UQ_GR, matching compute_rlgr1_magnitude and the stated MS-RDPRFX behavior, while non-zero handling remains DQ_GR; the included deterministic round-trip regression covers the divergence scenario. No contradictory issue or API impact was found.

glamberson:
Confirmed correct, and worth recording why the original looked plausible.

The encode pseudocode in MS-RDPRFX 3.1.8.1.7.3.2 is unambiguous on the direction:

```
if (!twoMs) { k = UpdateParam(&kp, UqGR); }
else        { k = UpdateParam(&kp, -DqGR); }
```

A zero increases `kp` by `UqGR`, which is what this PR does and what `compute_rlgr1_magnitude` has always done on the decode side.

The trap is that the `#define` block a few lines above it in the same section labels the constant the other way round: `#define UQ_GR (3) // increase in kp after nonzero symbol in GR mode`. FreeRDP calls that out directly in `rfx_rlgr.c`:

```c
/* NOTE: as of Aug 2011, the algorithm is still wrongly documented
   and the update direction is reversed */
if (twoMs) { k = UpdateParam(&kp, -DQ_GR); }
else       { k = UpdateParam(&kp, UQ_GR); }
```

So the pseudocode, FreeRDP's encoder and IronRDP's own decoder all agree with this change, and only the constant's comment disagrees. Worth keeping the code comment you added, since anyone who checks the fix against the `#define` alone will conclude it's backwards.

One suggestion, take it or leave it. `crates/ironrdp-testsuite-core/tests/graphics/rlgr.rs` already exists and already does `use ironrdp_graphics::rlgr::*`, while `rlgr.rs` doesn't carry an inline test module today. Since `encode` and `decode` are public API, the round-trip test might sit better there.

On ordering against #1179: this should go first. It's one line where that's three, it carries the regression test, and #1179 has been a draft since March. Landing this reduces #1179 to its other two fixes, the RL-mode trailing value on input exhaustion and the RLGR3 `two_ms2` default.

Unrelated, but since it's the same file: I'm looking at the encoder's `BitStream` writes, which are unbounded and panic when RLGR output exceeds the tile buffer (Devolutions/IronRDP#1557). That work is confined to the `BitStream` impl rather than the body of `encode`, so it shouldn't conflict with this.

Issue #139, opened 2023-05-16
Support Server Redirection PDUs


CBenoit:
- [TODO comment in code](https://github.com/Devolutions/IronRDP/blob/f5a6ac7751dce6125f8faada03b288ab600035bd/crates/ironrdp-connector/src/connection.rs#L867)

> REDIRECTION_SUPPORTED
> 0x00000001

> The client can receive server session redirection packets. If this flag
> is set, the ServerSessionRedirectionVersionMask MUST contain the
> server session redirection version that the client supports.

t184256:
This renders "login mode" gnome-remote-desktop 50.2 and 51.0 unusable with Haven in a curious way: the connection goes through, but the main process eats 100% CPU. The remote assistance mode is not affected, since it doesn't do redirection.

PR #1371, opened 2026-06-10
fix(connector): stay in CapabilitiesExchange when activation handles DeactivateAll

### LINKED ISSUE #1362: Invalid state error using gnome-remote-desktop
Compiled into wasm using https://github.com/electerm/ironrdp-wasm after the release of #1254 

Windows rdp is working in browser, and Ubuntu 24.04.4 rdp works using mstsc but errors in the browser

relevant log bits:
```
2026-06-05T18:13:04.539Z DEBUG connect_finalize: ironrdp_connector::connection_activation: Skipping Server Deactivate All PDU received during Capabilities Exchange, awaiting Server Demand Active

console.js:38 The console connection failed:  [invalid state (this is a bug) @ /home/mahcsig/.cargo/git/checkouts/ironrdp-98b53f60bfdbb711/f21470c/crates/ironrdp-connector/src/lib.rs:405] general error
```

I'm not sure what would help get to the fix, but here are the pcaps and debug console. If anything else could help, please let me know.

[rdp-pcap.zip](https://github.com/user-attachments/files/28650609/rdp-pcap.zip)

[console.log](https://github.com/user-attachments/files/28650552/console.log)

rdeangel:
Fixes #1362

## Problem

#1254 taught the inner `ConnectionActivationSequence` to skip a Server Deactivate All PDU received before Server Demand Active (sent by e.g. Windows Server and gnome-remote-desktop as part of a Deactivation-Reactivation Sequence, MS-RDPBCGR 1.3.1.3): it returns `Written::Nothing` and remains in `CapabilitiesExchange`.

However, the outer `ClientConnector::step()` only accepts `ConnectionFinalization` as the resulting inner state after calling `connection_activation.step()`, so the same scenario still fails with `invalid state (this is a bug)` when it happens during the initial connection sequence — exactly the symptom reported in #1362.

## Changes

- Add the missing match arm in `ClientConnector::step()`: when the inner sequence stays in `CapabilitiesExchange`, the outer connector stays in `CapabilitiesExchange` too and waits for the next input.
- Add a regression test mirroring the existing #1254 tests, driving the outer `ClientConnector` instead of the inner sequence (fails with `invalid state` before the fix).

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -558,6 +558,14 @@ impl Sequence for ClientConnector {
                         written,
                         ClientConnectorState::ConnectionFinalization { connection_activation },
                     ),
+                    // The inner sequence stays in CapabilitiesExchange when it receives a
+                    // Server Deactivate All PDU before the Server Demand Active PDU (sent
+                    // by e.g. Windows Server and gnome-remote-desktop); mirror it here and
+                    // wait for the next input.
+                    ConnectionActivationState::CapabilitiesExchange { .. } => (
+                        written,
+                        ClientConnectorState::CapabilitiesExchange { connection_activation },
+                    ),
                     _ => return Err(general_err!("invalid state (this is a bug)")),
                 }
             }
--- a/crates/ironrdp-testsuite-core/tests/session/connection_activation.rs
+++ b/crates/ironrdp-testsuite-core/tests/session/connection_activation.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 
 use ironrdp_connector::connection_activation::{ConnectionActivationSequence, ConnectionActivationState};
-use ironrdp_connector::{Credentials, DesktopSize, Sequence as _, Written};
+use ironrdp_connector::{ClientConnector, ClientConnectorState, Credentials, DesktopSize, Sequence as _, Written};
 use ironrdp_core::{WriteBuf, encode_vec};
 use ironrdp_pdu::gcc;
 use ironrdp_pdu::mcs::{McsMessage, SendDataIndication};
@@ -92,6 +92,26 @@ fn deactivate_all_during_capabilities_exchange_stays_in_same_state() {
             ConnectionActivationState::CapabilitiesExchange { .. }
         ),
         "state should remain CapabilitiesExchange after DeactivateAll"
+    );
+}
+
+#[test]
+fn client_connector_stays_in_capabilities_exchange_on_deactivate_all() {
+    let config = test_config();
+    let mut connector = ClientConnector::new(config.clone(), "127.0.0.1:3389".parse().unwrap());
+    connector.state = ClientConnectorState::CapabilitiesExchange {
+        connection_activation: ConnectionActivationSequence::new(config, IO_CHANNEL_ID, USER_CHANNEL_ID),
+    };
+
+    let frame = encode_server_share_control(ShareControlPdu::ServerDeactivateAll(ServerDeactivateAll));
+    let mut output = WriteBuf::new();
+
+    let written = connector.step(&frame, &mut output).unwrap();
+
+    assert_eq!(written, Written::Nothing);
+    assert!(
+        matches!(connector.state, ClientConnectorState::CapabilitiesExchange { .. }),
+        "outer connector state should remain CapabilitiesExchange after DeactivateAll"
     );
 }
 


mahcsig:
@rdeangel thank you for looking into this! 

I've tried compiling again from your branch and it does get past the `ServerDectivateAll`, but it now errors out saying `[decode_send_data_indication @ /code/src/github.com/rdeangel/IronRDP/crates/ironrdp-connector/src/lib.rs:409] reason: received disconnect provider ultimatum: UserRequested`

rdeangel:
Thanks for testing again, good to see it's getting past ServerDeactivateAll now.
  
From the error location, the connection is getting quite far: it's past NLA and licensing, into capabilities exchange. It handles the ServerDeactivateAll and waits for a ServerDemandActive, but the server sends a  disconnectProviderUltimatum and drops instead. So I don't think this PR is causing the disconnect, it just lets the connection get far enough to reach it, where before it stopped at the "invalid state" error, but I'd rather not guess at why the server is dropping without more to go on.

A couple of things that would really help narrow it down:

  - What server is this (Windows version / GNOME RD / xrdp / other)?
  - Does the same server connect OK with another client like FreeRDP or mstsc? That'd tell us whether it's something specific to IronRDP or a server-side decision.
  - If you can grab a trace log (or Wireshark capture) of the PDUs right after the ServerDeactivateAll, that'd show whether the server sends anything before the disconnect.

One small note: the UserRequested name is a little misleading, it's just the generic MCS code for a deliberate close, not necessarily anything a user did.

Separately, this did make me wonder if it'd be worth surfacing DisconnectProviderUltimatum as a typed error instead of a string, so it's easier to handle distinctly but that's independent of this PR.

mahcsig:
That makes sense, to help narrow it down:
- the server is Ubuntu 24.0.4 LTS, using the gnome-remote-desktop service
  - win11 connects properly, so it may be something specific gnome is doing?
- mstsc is able to connect to the same server
- I'm not sure how to start the capture after the ServerDeactivateAll, but here is the updated pcap of the connection attempt after the fix: [rdp-ironrdp.zip](https://github.com/user-attachments/files/28898384/rdp-ironrdp.zip)

rdeangel:
Here's what I see in it: there's a single TCP connection that completes the TLS handshake and NLA, gets through the capabilities exchange, and then the server closes it (it sends the FIN). The client never reconnects. That pattern  authenticate, server drops, and mstsc working against the same box  is what you'd expect from gnome-remote-desktop's "Remote Login" handover, where the server authenticates you on one endpoint and then redirects the client to the actual session. mstsc follows that redirect automatically; right now IronRDP doesn't, so it just sees the disconnect.

I couldn't confirm it from the pcap alone, though, because everything after the handshake is inside TLS, so the redirect PDU (if that's what it is) isn't readable.

One thing worth flagging: the branch you built from only has the capabilities-exchange fix. It's missing two capability advertisements that my working setup uses  EGFX dynamic-channel support, and redirection support in the cluster data (the old TODO(#139)). The redirection one matters here, because some servers only perform the handover redirect when the client advertises that it supports it.

I've put together a test branch that adds those two, plus some temporary logging that prints each PDU during the activation sequence (look for [redir-diag] lines). Could you rebuild from it and share the output? The most useful part is the last [redir-diag] line before the disconnect, and whether any line shows looks_like_redirect=true.

Branch: https://github.com/rdeangel/IronRDP/tree/test/gnome-handover-diag

If that confirms a redirect, the next step is bigger: IronRDP would need actual server-redirection handling (parse the redirect, then reconnect with the routing token and the session credentials it hands back). That's not in the library yet  TODO(#139) only covers advertising support, not following the redirect so it'd be a separate piece of work, but at least we'd know that's the right direction.

mahcsig:
I've compiled the new branch, and there are no `looks_like_redirect=true`, it looks like it's getting a data pdu instead of the ServerDemandActive pdu it expects

[console.log](https://github.com/user-attachments/files/28966862/console.log)

rdeangel:
Right, Looking at the log, the connection reaches this point:

2026-06-15T16:37:40.326Z  INFO ironrdp_web::session: Connected!

The canvas initializes and the session starts. What happens next is a separate issue:

2026-06-15T16:37:40.402Z DEBUG ironrdp_web::session: Received Server Deactivate All PDU, executing Deactivation-Reactivation Sequence

Session error: [General] [ConnectionActivation::CapabilitiesExchange @ /code/src/github.com/rdeangel/IronRDP/crates/ironrdp-connector/src/lib.rs:409] reason: unexpected Share Control PDU during capabilities exchange: got Data PDU (expected Server Demand Active PDU)

Gnome sends a Server Deactivate All immediately after connecting, which triggers a Deactivation-Reactivation Sequence.
Per MS-RDPBCGR 1.3.1.3
(https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/dfc234ce-481a-4674-9a5d-2a7bafb14432), after a Deactivate All the Capability Exchange phase should be re-executed. Instead, gnome sends a Data PDU, which IronRDP rejects.

This is a different bug and would need either gnome-side investigation or a follow-up IronRDP fix to tolerate Data PDUs during reactivation the same way the DeactivateAll case is already handled.

mahcsig:
With mstsc working to connect to the same box, it should probably be caught in IronRDP, no?

f1ori:
I’m interested in getting gnome remote desktop to work with IronRDP too.
First note that there are two usage scenarios for gnome remote desktop. One is to allow somebody to join a normal graphical session (share it) and the other is a headless scenario, where the user logs in and creates a new session. The server redirect is only needed for the headless scenario. In the headless case the user at first is connected to a login session where he can provide username and password and only then he is redirected to the actual session. At least this is my understanding.

This means we can focus on the sharing type of session first.

When connecting with IronRDP git master, I get this error in the system journal:

`gnome-remote-desktop-daemon: [RDP] Client did not advertise support for the Graphics Pipeline, closing connection
`

I used AI to enable EGFX graphics pipeline support in ironrdp-viewer. And now I can connect to grd in the sharing case and see an image. The current state is in https://github.com/f1ori/IronRDP/tree/test/gnome-remote-desktop The is only a vibecoded branch. It does not work properly for Windows connections yet.

Are there plans to integrate EGFX in the viewer? Are there still blockers?

CBenoit (approved):
LGTM, thank you for looking into this!

CBenoit:
If you noticed other issues when connecting to gnome remote desktop, it would help if you could open dedicated issues so the context is not forgotten in a merged PR.

PR #1372, opened 2026-06-10
fix(web): recover file-upload state when a paste is interrupted or never pulled

ymarcus93:
## Problem

`RdpFileTransferProvider.uploadFiles()` sets `uploadState` and refuses to start another upload while
it is set (`throw "Upload already in progress"`). `uploadState` was cleared only when the remote
pulled the advertised files **to completion**, or on `dispose()`. So any paste that is advertised but
doesn't run to completion wedges every later upload until the page is reloaded. That happens whenever:

- **The remote never pulls** the files — the paste lands in a non-file target, the clipboard monitor
  overwrites our file FormatList before the paste is pulled, or the remote rejects the FormatList with
  `CB_RESPONSE_FAIL` so it never requests the files.
- **The remote replaces the clipboard mid-paste** — it copies its own content, which per MS-RDPECLIP
  makes it the new clipboard owner and supersedes our advertise, so our files are never pulled.
- **The remote starts pulling, then stops** — the transfer stalls partway, so completion (the only
  thing that cleared `uploadState`) never arrives.

Backends also had no way to know whether an advertise was accepted or rejected.

## Fix

Release `uploadState` (reject `completion`, emit an upload `error`, resume clipboard monitoring) on
every path where an advertised paste won't run to completion — not just the happy-path completion:

1. **Paste-ack watchdog (primary, peer-agnostic).** Armed on advertise, disarmed on the first
   `FileContentsRequest`. On timeout (60s, matching `Cliprdr`'s `lock_inactivity_timeout`) it fails the
   pending upload regardless of how the peer behaves. Covers "advertised but never pulled."
2. **`on_format_list_response` (explicit reject).** Newly surfaced from the cliprdr hook (#1300)
   through `ironrdp-web` as a `formatListResponseCallback` / `format-list-response` event; on
   `CB_RESPONSE_FAIL` the provider fails the pending upload at once. The event also lets a frontend
   drive paste (inject on accept, retry on reject).
3. **`handleFilesAvailable` (remote replaced the clipboard with files).** When the remote copies its
   own files mid-paste the web client auto-requests the list (no delayed rendering) and surfaces it
   here — i.e. the remote took clipboard ownership, so our advertise is superseded and we fail the
   pending upload. The symmetric counterpart of (2): that recovers when the remote *rejects* our
   advertise, this when it *replaces* it.
4. **Upload inactivity watchdog.** Armed and reset on every `FileContentsRequest` once pulling has
   begun — picking up where the paste-ack watchdog (which only covers the never-pulled case) leaves
   off. If requests then stop for 60s the paste stalled — e.g. a text/image copy took the clipboard
   (which arrives as a FormatList **without** `FileGroupDescriptorW`, so it never reaches
   `handleFilesAvailable`), or the transfer simply died — and the upload is failed. A
   slow-but-progressing transfer keeps resetting it, so it is never killed.

Together (1) and (4) give continuous coverage across the paste's lifetime (before the first pull, then
after it), while (2) and (3) short-circuit the two cases the remote signals explicitly.

Monitoring stays suppressed until the first `FileContentsRequest` (not just the wire send), closing the
window where the clipboard monitor could clobber our file FormatList.

`handleUnlock` is a deliberate **no-op**: `Unlock` is not a reliable failure signal — the peer, and our own cliprdr on the lock-inactivity timeout, emit it routinely (often before any `FileContentsRequest`) — so recovery is
left to the watchdogs.

## Scope

- `ironrdp-web`: `FormatListResponse` backend message + `on_format_list_response` wiring; the cliprdr
  trait method has a no-op default (#1300), so it is non-breaking for other backends.
- `iron-remote-desktop-rdp` (web): provider recovery paths (paste-ack + inactivity watchdogs,
  `handleFilesAvailable` release, reject handling) and a `formatListResponseCallback` factory. No
  `iron-remote-desktop` change.

## Tests

`RdpFileTransferProvider.test.ts`:

- Monitoring is deferred past the wire send and resumed on the first `FileContentsRequest`.
- The paste-ack watchdog fails a never-pulled upload and resumes monitoring (and the next upload then
  succeeds).
- The inactivity watchdog fails a pulled-then-idle upload, releasing the wedge, but does **not** fail a
  pulled upload that keeps making progress.
- `handleFilesAvailable` releases an in-flight upload when the remote replaces the clipboard.
- An `Unlock` does not fail an in-flight upload (it is not a paste-failure signal).
- `on_format_list_response`: extension exposure, event emission, reject-on-fail, and a reject with no
  upload in progress is a no-op.

### DIFF
--- a/crates/ironrdp-web/src/clipboard.rs
+++ b/crates/ironrdp-web/src/clipboard.rs
@@ -173,6 +173,16 @@ pub(crate) enum WasmClipboardBackendMessage {
     LocksExpired {
         clip_data_ids: Vec<u32>,
     },
+    /// [MS-RDPECLIP] 2.2.3.2 Remote's response to one of our outbound Format Lists.
+    ///
+    /// `ok` is `true` when the remote accepted the advertised formats
+    /// (`CB_RESPONSE_OK`) and `false` when it rejected them (`CB_RESPONSE_FAIL`).
+    /// A rejected advertise is silently discarded by the remote, so a file paste
+    /// cannot proceed; backends use this to detect and recover from a refused
+    /// paste instead of stalling until a lock timeout.
+    FormatListResponse {
+        ok: bool,
+    },
 
     // JS-initiated file transfer operations
     /// JS requests file contents from remote (download).
@@ -235,6 +245,7 @@ pub(crate) struct JsClipboardCallbacks {
     pub(crate) on_lock: Option<js_sys::Function>,
     pub(crate) on_unlock: Option<js_sys::Function>,
     pub(crate) on_locks_expired: Option<js_sys::Function>,
+    pub(crate) on_format_list_response: Option<js_sys::Function>,
 }
 
 impl WasmClipboard {
@@ -841,6 +852,16 @@ impl WasmClipboard {
                     );
                 }
             }
+            WasmClipboardBackendMessage::FormatListResponse { ok } => {
+                if let Some(callback) = self.js_callbacks.on_format_list_response.as_ref() {
+                    if let Err(e) = callback.call1(&JsValue::NULL, &JsValue::from_bool(ok)) {
+                        error!(error = ?e, ok, "Failed to call JS format list response callback");
+                        return Ok(());
+                    }
+                } else {
+                    trace!(ok, "Format list response received but no JS callback registered");
+                }
+            }
             // The following variants are handled directly in the event loop and should never reach here
             WasmClipboardBackendMessage::FileContentsRequestSend { .. }
             | WasmClipboardBackendMessage::FileContentsResponseSend { .. }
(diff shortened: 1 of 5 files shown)


ymarcus93:
Hey @CBenoit, do you mind taking a look at this PR? It fixes an issue where file upload no longer works (with no way to recover except a hard refresh) if we get to this broken state

ymarcus93:
@CBenoit Ok pushed amended commit addressing Copilot changes. I have 1 more follow-up commit that build off this PR that fix some other upload issues I've noticed during testing. I'll wait for this PR to get merged before submitting another PR to avoid churn and conflicts

CBenoit (approved):
LGTM! Thank you for the high quality fix. Looking forward to the next PR 🙂

ymarcus93:
> LGTM! Thank you for the high quality fix. Looking forward to the next PR 🙂

@CBenoit Thank you! Here is the next one https://github.com/Devolutions/IronRDP/pull/1381 ready for review, building off of this PR

PR #1374, opened 2026-06-12
perf(web): replace softbuffer with direct put_image_data canvas present

irvingoujAtDevolution:
## Summary

The web client presented frames through `softbuffer`, whose web backend repacks
the **whole surface** (RGBA → u32 → RGBA into a fresh buffer) on every present.
This replaces it with a direct `put_image_data` that uploads only the dirty
region, and drops the `softbuffer` dependency.

Same idea as the IronVNC change.

## What changed

- Remove the `softbuffer` dependency; present each dirty region with
  `put_image_data` at its origin.
- No full-surface buffer and no per-region scratch. `extract_partial_image` fills
  a single `WriteBuf` reused across frames, so steady-state draws don't allocate.
- Force opaque alpha before upload (kept — see Correctness).
- Add `WriteBuf::filled_mut` to `ironrdp-core` (mutable counterpart of `filled`).
- `web-sys`: add `CanvasRenderingContext2d` + `ImageData`, drop the softbuffer-only
  features.

## Performance

Draw-stage time on a 1080p replay (595 frames / 110 dirty regions), headless
Chromium, 8 measured passes × 3 runs, median. Both rows are reproducible branches
off the replay-bench harness; the only difference is the render path.

| Render path | draw (ms) | vs softbuffer | branch |
|---|--:|--:|---|
| softbuffer `present_with_damage` | ~1031 | — | `bench/draw-softbuffer` |
| this PR (direct upload, reused `WriteBuf`) | ~97 | **~10.6×** | `bench/draw-zerocopy` |

- The win is structural: upload the dirty region instead of repacking the whole
  surface every present.
- Reusing one `WriteBuf` (vs a per-frame allocation) keeps the steady-state draw
  allocation-free; the remaining cost is the unavoidable `ImageData` JS copy.
- Output is **byte-identical**: framebuffer CRC32 `2d8e1b79` matches the recorded
  ground truth and the rendered-canvas FNV-1a is unchanged.
- Absolute ms carry ~±15% noise from machine load (decode drifted 1.5–1.9 s); the
  ratio held across runs.

Reproduce:
```sh
git checkout bench/draw-softbuffer   # or bench/draw-zerocopy
cd crates/ironrdp-web && wasm-pack build --target web --release -- --features bench
cd bench-harness && node run.mjs --capture /bench-corpus/<your>.irdprec --passes 8
```

## Correctness

`put_image_data` stores alpha verbatim, and the decoded framebuffer isn't
guaranteed opaque — it's zero-initialised, a widened whole-rows region can cover
not-yet-painted columns (alpha 0), and the QOI-RGBA path copies source alpha. So
we force alpha opaque before upload. A scan-then-conditionally-force was tried and
is *slower* than just forcing (the check touches the same bytes), so the
unconditional force stays.

## Follow-up (separate PR)

Guarantee framebuffer opacity upstream in `ironrdp-session` (init alpha to `0xff` +
clamp `apply_rgba32`); after that the web side can drop the alpha force entirely.

### DIFF
--- a/crates/ironrdp-web/src/canvas.rs
+++ b/crates/ironrdp-web/src/canvas.rs
@@ -11,9 +11,9 @@ use wasm_bindgen::{Clamped, JsCast as _};
 use web_sys::ImageData;
 use web_sys::{CanvasRenderingContext2d, HtmlCanvasElement};
 
-/// Web render surface. Owns the canvas's 2D context and a reusable RGBA scratch buffer: each dirty
-/// region's pixels are copied once into the scratch (alpha forced opaque), then blitted with
-/// `put_image_data` at the region's origin.
+/// Web render surface. Owns the canvas's 2D context; each dirty region is blitted directly with
+/// `put_image_data` at the region's origin, after forcing its alpha opaque in place. The region
+/// buffer is the caller's throwaway copy, so no scratch buffer is kept here.
 ///
 /// This replaced a softbuffer-backed path that converted RGBA -> u32 `0RGB` (our pass) and then let
 /// softbuffer repack u32 -> RGBA per frame into a freshly allocated buffer — two pixel passes over
@@ -23,7 +23,6 @@ use web_sys::{CanvasRenderingContext2d, HtmlCanvasElement};
 pub(crate) struct Canvas {
     canvas: HtmlCanvasElement,
     ctx: CanvasRenderingContext2d,
-    rgba: Vec<u8>,
 }
 
 impl Canvas {
@@ -35,7 +34,6 @@ impl Canvas {
         Ok(Self {
             canvas: render_canvas,
             ctx,
-            rgba: Vec::new(),
         })
     }
 
@@ -47,23 +45,17 @@ impl Canvas {
         self.canvas.set_height(height.get());
     }
 
-    /// `buffer` is the region's RGBA sub-image (as produced by `extract_partial_image`).
-    pub(crate) fn draw(&mut self, buffer: &[u8], region: InclusiveRectangle) -> anyhow::Result<()> {
-        // Refill the reusable scratch from `buffer` in a single copy. `clear` keeps the existing
-        // capacity, so steady-state frames reuse the allocation and `extend_from_slice` writes the
-        // pixels straight into spare capacity with no zero-fill; only a larger-than-seen region
-        // reallocates.
-        self.rgba.clear();
-        self.rgba.extend_from_slice(buffer);
-        let dst = self.rgba.as_mut_slice();
-
-        // Force opaque alpha: most decode paths already write 0xFF, but the QOI path copies source
-        // alpha, and `put_image_data` stores alpha verbatim into the canvas.
-        for pixel in dst.chunks_exact_mut(4) {
+    /// Blits one dirty region. `buffer` is the region's RGBA sub-image — a throwaway copy produced
+    /// by `extract_partial_image` — mutated in place to force opaque alpha before upload.
+    pub(crate) fn draw(&mut self, buffer: &mut [u8], region: InclusiveRectangle) -> anyhow::Result<()> {
+        // Force opaque alpha in place. The decoded framebuffer leaves the alpha channel as
+        // don't-care (it starts at 0 and the decode loop skips it), but `put_image_data` stores
+        // alpha verbatim, so the region would otherwise render transparent on the canvas.
+        for pixel in buffer.chunks_exact_mut(4) {
             pixel[3] = 0xFF;
         }
 
-        blit(&self.ctx, dst, &region)
+        blit(&self.ctx, buffer, &region)
     }
 }
 
--- a/crates/ironrdp-web/src/canvas.rs
+++ b/crates/ironrdp-web/src/canvas.rs
@@ -49,7 +49,19 @@ impl Canvas {
     pub(crate) fn draw(&mut self, buffer: &[u8], region: InclusiveRectangle) -> anyhow::Result<()> {
         let len = buffer.len();
         if self.rgba.len() < len {
-            self.rgba.resize(len, 0);
+            let initialized_len = self.rgba.len();
+            self.rgba.reserve(len - initialized_len);
+
+            let (initialized, uninitialized) = buffer.split_at(initialized_len);
+            self.rgba.copy_from_slice(initialized);
+            let spare = &mut self.rgba.spare_capacity_mut()[..uninitialized.len()];
+            for (dst, src) in spare.iter_mut().zip(uninitialized.iter().copied()) {
+                dst.write(src);
+            }
+
+            // SAFETY: The prefix was already initialized, and `spare` above initializes exactly
+            // the additional bytes needed to make `self.rgba.len() == len`.
+            unsafe { self.rgba.set_len(len) };
         }
         let dst = &mut self.rgba[..len];
         dst.copy_from_slice(buffer);
--- a/crates/ironrdp-web/src/canvas.rs
+++ b/crates/ironrdp-web/src/canvas.rs
@@ -11,15 +11,8 @@ use wasm_bindgen::{Clamped, JsCast as _};
 use web_sys::ImageData;
 use web_sys::{CanvasRenderingContext2d, HtmlCanvasElement};
 
-/// Web render surface. Owns the canvas's 2D context; each dirty region is blitted directly with
-/// `put_image_data` at the region's origin, after forcing its alpha opaque in place. The region
-/// buffer is the caller's throwaway copy, so no scratch buffer is kept here.
-///
-/// This replaced a softbuffer-backed path that converted RGBA -> u32 `0RGB` (our pass) and then let
-/// softbuffer repack u32 -> RGBA per frame into a freshly allocated buffer — two pixel passes over
-/// the whole surface plus a per-frame allocation. The direct path drops the u32 round-trip and the
-/// per-frame allocation, measuring an order of magnitude faster present at 4K with byte-identical
-/// canvas output. Mirrors the same fix in IronVNC.
+/// Web render surface: owns the canvas's 2D context and blits each dirty region's RGBA sub-image
+/// directly with `put_image_data` at the region's origin, after forcing alpha opaque (see [`Canvas::draw`]).
 pub(crate) struct Canvas {
     canvas: HtmlCanvasElement,
     ctx: CanvasRenderingContext2d,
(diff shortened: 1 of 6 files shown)


irvingoujAtDevolution:
Same as VNC, remove softbuffer

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,104 @@
...
-                .skip(region_top_usize)
-                .take(region_height_usize)
-            {
-                let src_row = src.by_ref().take(region_width_usize);
+        let len = buffer.len();
+        if self.rgba.len() < len {
+            self.rgba.resize(len, 0);
+        }
+        let dst = &mut self.rgba[..len];
+        dst.copy_from_slice(buffer);
```
RRRadicalEdward: We can do it faster without filling the buffer with zeros prior to copying, since we are initializing it with data from `buffer` after all.

```rust
if self.rgba.len() < len {
     self.rgba.reserve(len);
     // SAFETY: We later initialize the exact `len` of `self.rgba` by copying `len` bytes from `buffer`. 
     unsafe { self.rgba.set_len(len) };
}

let dst = &mut self.rgba[..len];
dst.copy_from_slice(buffer);
```
But, I wonder how much of speed up it gives. 

Also, we could even make it branchless:
```rust
let target_len = max(self.rgba.len(), len);
// In, case `target_len` == `self.rgba.len()`,  `reserve` does nothing, and `self.rgba.set_len` sets a length, which it already contains. 
// In, case ``target_len` == `len`, it's the same case as above. 

self.rgba.reserve(target_len);
// SAFETY: We later initialize the exact `len` of `self.rgba` by copying `len` bytes from `buffer`. 
unsafe { self.rgba.set_len(target_len ) };

let dst = &mut self.rgba[..target_len];
dst.copy_from_slice(buffer);
```
But, we need to measure both versions. The first one should definitely be faster, than filling with zero. 
As for the second version, I'm not sure - we don't have `len` comparing branch, which is great, but it means that a branch inside `self.rgba.reserve(...);` executes more often.

irvingoujAtDevolution: Hi Alex, yes I benched it with [this branch](https://github.com/Devolutions/IronRDP/tree/bench/record-replay)
There is improvement, but is minimal, around 5-7 ms per draw, which is around 5%, the downside is to introduce unsafe blocks, I personally believe it's not worth it, but let me know

RRRadicalEdward: 5% is nice to have. I don't think that unsafe is that bad, because we work with it correctly

CBenoit: Agree the zero-fill is avoidable. One option to make the intent a bit clearer is `spare_capacity_mut()`: it hands back a `&mut [MaybeUninit<u8>]` over the reserved capacity, so we write through `MaybeUninit` instead of forming a `&mut [u8]` to not-yet-written bytes (a little friendlier under Miri). It doesn't remove the `unsafe` though. We still need a `set_len` afterward to make the bytes live, so it's really a readability/correctness-framing improvement rather than a different category:

```rust
if self.rgba.capacity() < len {
    self.rgba.reserve(len - self.rgba.len());
}
let spare = &mut self.rgba.spare_capacity_mut()[..len - self.rgba.len()];
// ... or just keep reserve + set_len; same shape
```

Honestly the two real choices are still: keep `resize(len, 0)` (safe, eats the zero-fill) or `reserve` + `unsafe { set_len }` (drops it, costs one SAFETY-commented block). For a `Vec<u8>` scratch the unsafe is legitimate and not UB as long as we only read what we copy.

On significance: the ~5% is a repeatable win, but worth keeping in perspective. The structural change here (killing the double pixel-pass + per-frame alloc) is what got us the 20-50×, and the draw stage isn't the bottleneck anymore at 83ms/4K. So this is mostly a preference call rather than something load-bearing.

That said, I lean slightly toward taking the perf, a single well-scoped unsafe with a clear SAFETY comment is cheap to maintain, and free performance is free performance. Happy either way! 🙂

RRRadicalEdward: Instead of 
```rust
if self.rgba.capacity() < len {
    self.rgba.reserve(len - self.rgba.len());
}
```
We could just do:
```rust
self.rgba.reserve(len - self.rgba.len());
```
`Vec::reserve` anyway doesn't grow if the current capacity already can handle the additional amount of data, so the call is a no-op in this case.

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,104 @@
...
+/// Blits `rgba` (a `region`-sized RGBA buffer) onto the canvas at the region's origin.
+fn blit(ctx: &CanvasRenderingContext2d, rgba: &[u8], region: &InclusiveRectangle) -> anyhow::Result<()> {
+    #[cfg(target_arch = "wasm32")]
+    {
+        let image = ImageData::new_with_u8_clamped_array_and_sh(
+            Clamped(rgba),
+            u32::from(region.width()),
+            u32::from(region.height()),
+        )
+        .map_err(|err| anyhow!("ImageData::new failed: {err:?}"))?;
```
CBenoit: question: Can’t this be `.context("ImageData::new failed")` instead?

EDIT: I guess the error is a JsValue, so not something implementing `std::error::Error`

CBenoit (commented):
Nice!

irvingoujAtDevolution:
@RRRadicalEdward @CBenoit after a closer look, I realized that this rgba buffer might not be needed to begin with, so removed it, no allocation/initialization issue anymore. It's even 20% faster than what we have in the first draft!

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,99 @@
...
-
-                dst_row
-                    .iter_mut()
-                    .skip(region_left_usize)
-                    .take(region_width_usize)
-                    .zip(src_row)
-                    .for_each(|(dst, src)| *dst = src);
-            }
+    /// Blits one dirty region. `buffer` is the region's RGBA sub-image — a throwaway copy produced
+    /// by `extract_partial_image` — mutated in place to force opaque alpha before upload.
```
CBenoit: Documentation assumes the caller is using `extract_partial_image`; the LLM is right, but in terms of documentation that’s bad. If we want this type of API, we should document more like this:

```suggestion
    /// Blits one dirty region.
    ///
    /// `buffer` is the region's RGBA sub-image, it is modified in place to force opaque alpha before upload,
    /// and thus is typically a throwaway buffer.
```

The caller can decide how the sub-image is obtained, and whether it’s okay to let `draw` modifies it, optionally performing the copy upfront if that’s necessary. (So far, it’s not.)

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,99 @@
...
+
+/// Web render surface. Owns the canvas's 2D context; each dirty region is blitted directly with
+/// `put_image_data` at the region's origin, after forcing its alpha opaque in place. The region
+/// buffer is the caller's throwaway copy, so no scratch buffer is kept here.
+///
+/// This replaced a softbuffer-backed path that converted RGBA -> u32 `0RGB` (our pass) and then let
+/// softbuffer repack u32 -> RGBA per frame into a freshly allocated buffer — two pixel passes over
+/// the whole surface plus a per-frame allocation. The direct path drops the u32 round-trip and the
+/// per-frame allocation, measuring an order of magnitude faster present at 4K with byte-identical
+/// canvas output. Mirrors the same fix in IronVNC.
```
CBenoit: Keeping some historical documentation on how we ended up converging may be okay in order to justify the current design, but it currently sounds like a "prompt artifact": "Mirrors the same fix in IronVNC" is completely irrelevant outside of this PR.

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,99 @@
...
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen::{Clamped, JsCast as _};
+#[cfg(target_arch = "wasm32")]
+use web_sys::ImageData;
+use web_sys::{CanvasRenderingContext2d, HtmlCanvasElement};
+
+/// Web render surface. Owns the canvas's 2D context; each dirty region is blitted directly with
+/// `put_image_data` at the region's origin, after forcing its alpha opaque in place. The region
+/// buffer is the caller's throwaway copy, so no scratch buffer is kept here.
+///
```
CBenoit: Documentation is describing too much the shape of the caller’s code. Documentation should be about the type, function or module it is attached to. It’s of course very helpful to also document things such as preconditions and invariants, that need to be taken into account in caller code, and to some extent design rationale; but it should not be the other way around. Describing the exact caller code shape is like overfitting and that kind of documentation will become stale as soon as the caller code is changed (at a distance).

crates/ironrdp-web/src/canvas.rs
```diff
@@ -1,93 +1,99 @@
...
-            }
+    /// Blits one dirty region. `buffer` is the region's RGBA sub-image — a throwaway copy produced
+    /// by `extract_partial_image` — mutated in place to force opaque alpha before upload.
+    pub(crate) fn draw(&mut self, buffer: &mut [u8], region: InclusiveRectangle) -> anyhow::Result<()> {
+        // Force opaque alpha in place. The decoded framebuffer leaves the alpha channel as
+        // don't-care (it starts at 0 and the decode loop skips it), but `put_image_data` stores
+        // alpha verbatim, so the region would otherwise render transparent on the canvas.
+        for pixel in buffer.chunks_exact_mut(4) {
+            pixel[3] = 0xFF;
         }
```
CBenoit, line 44: question: Just challenging the established pattern here: do we really need to force opaque alpha? I assume that if we don’t do that we end up with visual artifacts? My understanding is that we are just extracting a sub-image from an otherwise fully rendered image, and that we don’t really need any extra cleaning step.

crates/ironrdp-web/src/session.rs
```diff
@@ -868,8 +868,8 @@ impl iron_remote_desktop::Session for Session {
                     }
                     ActiveStageOutput::GraphicsUpdate(region) => {
                         // PERF: some copies and conversion could be optimized
-                        let (region, buffer) = extract_partial_image(&image, region);
-                        gui.draw(&buffer, region).context("draw updated region")?;
+                        let (region, mut buffer) = extract_partial_image(&image, region);
+                        gui.draw(&mut buffer, region).context("draw updated region")?;
```
CBenoit: This is hinted by the PERF comment, but any chance we could optimize further by reusing the `buffer` instead of returning a fresh one each time?

We already have a solid primitive for such patterns: the `WriteBuf` type. It prevents ever growing buffers by shrinking to a good default, but can otherwise be reused by resetting it. The idea is simple: keep a `WriteBuf` outside the loop, pass it as `&mut` to `extract_partial_image`, call `unfilled_to(dst_buf_size)` internally, then use a `filled_mut` function (akin to `filled`) to retrieve the throwaway buffer. Once we are done drawing, call `clear` to reset the `filled` cursor.

The vec allocation + 0-filling found inside `extract_partial_image` should be amortized. Can you check if we get some extra speed?

CBenoit (commented):
We’re getting at a good place!

RRRadicalEdward (approved):
Good job 👍. 
Benoit already pointed out good things. I have nothing to add

RRRadicalEdward:
> Alex Yusiuk (@RRRadicalEdward) Benoît Cortier (@CBenoit) after a closer look, I realized that this rgba buffer might not be needed to begin with, so removed it, no allocation/initialization issue anymore. It's even 20% faster than what we have in the first draft!

Could you apply this method to IronVNC as well?

crates/ironrdp-web/src/canvas.rs
irvingoujAtDevolution: @CBenoit yeah, works for most cases but it's not guaranteed. Framebuffer starts zero-filled ([here](https://github.com/Devolutions/IronRDP/blob/8fe59c19e253aa106a0a94c72dd0f962b881794c/crates/ironrdp-session/src/image.rs#L151)) and the QOI-RGBA path keeps source alpha ([here](https://github.com/Devolutions/IronRDP/blob/8fe59c19e253aa106a0a94c72dd0f962b881794c/crates/ironrdp-session/src/image.rs#L848)) — plus a tall update gets widened to full width, so early / post-resize frames can upload not-yet-painted columns as transparent.

Best case we'd guarantee opaque upstream (init the framebuffer to `0xff` + clamp `apply_rgba32`), but for the scope of this PR I think it's better to keep the force and fix it in a follow-up. Sound good?

crates/ironrdp-web/src/canvas.rs
irvingoujAtDevolution: updated, see below

crates/ironrdp-web/src/canvas.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/canvas.rs
irvingoujAtDevolution: updated

crates/ironrdp-web/src/session.rs
irvingoujAtDevolution: updated, now it's exactly zero per-draw allocation.

irvingoujAtDevolution:
Updated, we now use WriteBuf to achive effectively `O(1)` allocation per draw on the Rust side rendering path.

crates/ironrdp-web/src/canvas.rs
CBenoit: Sounds good to me! My only suggestion then is to track this with an issue for visibility

crates/ironrdp-web/src/image.rs
```diff
@@ -2,18 +2,29 @@
...
+use ironrdp_core::WriteBuf;
+
+/// Copies the dirty `region` into `buffer` from its current cursor (clear it between regions).
+/// The returned rect may be wider than `region`: the whole-rows path widens to full image width.
+pub(crate) fn extract_partial_image(
+    image: &DecodedImage,
+    region: InclusiveRectangle,
+    buffer: &mut WriteBuf,
+) -> InclusiveRectangle {
     // PERF: needs actual benchmark to find a better heuristic
```
CBenoit, line 14: Thought: Since you have the tools for benchmarking, I think it may be worth to look into this comment as well. As a follow up 🙂

CBenoit (approved):
LGTM! Good job! Before we merge, let's measure the final gains and clean up the final commit body so it matches the result we ended up having 🙂

PR #1375, opened 2026-06-15
fix(cliprdr): release outgoing locks before initiating a file copy

ymarcus93:
## Problem

When the client downloads files from the remote, it sends a **Lock** PDU (`CB_LOCK_CLIPDATA`) so the
Shared Clipboard Owner retains the File Stream data while we pull it chunk-by-chunk — even if the
clipboard changes mid-transfer ([[MS-RDPECLIP] 2.2.4.1](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeclip/150bac72-bc7f-42e5-9e8e-cb5a0ddc7dbc)). `Cliprdr` tracks these in `outgoing_locks`.

These locks are deliberately **not** released the instant a *remote* clipboard change arrives: they
transition to `Expired` and are cleaned up lazily by the inactivity sweep in `drive_timeouts`
(`lock_inactivity_timeout`, default 60s), so a download from the previous clipboard can keep running
after the remote copies something new. That is correct for the remote-driven case.

But there is **no path that releases them when the _local_ side takes clipboard ownership itself** —
i.e. when it calls `initiate_file_copy` to advertise its own files. In that case the outgoing locks
point at data we are replacing: we send a fresh `FormatList` while still holding `Lock`s on the
previous owner's File Stream data. The server is left tracking a lock for a transfer that will never
complete, overlapping with our new ownership advertise.

Some servers handle that overlap badly — notably Windows `rdpclip.exe`: the new copy can fail to be
pulled (a local→remote paste hangs), and the server's clipboard state can desync. In practice a user
who interrupts an in-flight download by starting a file copy hits exactly this.

## Fix

Release the outgoing locks **before** advertising the new clipboard, from the one path the lazy sweep
never covers — the local side taking ownership:

- New `Cliprdr::release_outgoing_locks()` sends an `Unlock` PDU (`CB_UNLOCK_CLIPDATA`) for, and drops,
  **every** entry in `outgoing_locks` — `Active` as well as `Expired` — clears `current_lock_id`, and
  notifies the backend via `on_outgoing_locks_cleared`. This aborts any still-in-flight download that
  was holding one of those locks; that's intended, since taking ownership replaces the very data the
  download was pulling, so it could not complete anyway. The backend uses the callback to tear those transfers
down cleanly.
- `initiate_file_copy()` calls it and places the `Unlock` PDUs **before** the `FormatList` on the
  wire, so the server releases the stale locks first and then processes the new ownership advertise.

Scoped to `initiate_file_copy` (a file copy) only — **not** `initiate_copy` (text/image) — so the
existing behavior where a download survives a non-file clipboard change is preserved. The lazy
inactivity sweep is unchanged for the remote-driven case; this only adds an explicit release for the
case the timer never reached: the local side replacing the clipboard with its own files. No-op when no
outgoing locks are held.

## Scope

- `ironrdp-cliprdr` only. No `ironrdp-web` / binding changes — `initiate_file_copy`'s return already
  flows through existing wiring and the extra `Unlock` PDUs ride the same channel. Non-breaking for
  other backends and roles; standalone (no dependency on other in-flight clipboard work).

## Tests

`ironrdp-testsuite-core` (`clipboard/lock_lifecycle.rs`):

- With two outgoing download locks held, `initiate_file_copy` emits an `Unlock` for **each** lock
  **before** the `FormatList` (verified by decoding the returned PDUs in order) and clears
  `outgoing_locks` + `current_lock_id`.
- With no locks held, `initiate_file_copy` sends only the `FormatList` (no spurious `Unlock`).

### DIFF
--- a/crates/ironrdp-cliprdr/src/lib.rs
+++ b/crates/ironrdp-cliprdr/src/lib.rs
@@ -1027,6 +1027,47 @@ impl<R: Role> Cliprdr<R> {
         // Backend notification deferred until actual cleanup
     }
 
+    /// Immediately sends `Unlock` PDUs for — and drops — every outgoing clipboard lock.
+    ///
+    /// Outgoing locks are created when we download files from the remote: each asks the
+    /// Shared Clipboard Owner to retain File Stream data so we can keep pulling it even
+    /// after the clipboard changes ([MS-RDPECLIP] 2.2.4.1). They are normally released
+    /// lazily by the inactivity sweep in [`Self::drive_timeouts`], so concurrent downloads
+    /// that outlive a *remote* clipboard change aren't aborted.
+    ///
+    /// When the **local** side takes clipboard ownership itself (initiating a file copy),
+    /// those locks point at data we are replacing. Leaving them held while we advertise a
+    /// fresh `FormatList` makes the server track a lock for a download that will never
+    /// finish; some servers (notably Windows `rdpclip.exe`) react badly to that overlap.
+    /// Releasing them up front keeps the lock/ownership state consistent.
+    ///
+    /// Returns the `Unlock` PDUs to send (these must precede the new `FormatList` on the
+    /// wire); empty when no locks are held.
+    fn release_outgoing_locks(&mut self) -> Vec<SvcMessage> {
+        if self.outgoing_locks.is_empty() {
+            return Vec::new();
+        }
+
+        let cleared: Vec<u32> = self.outgoing_locks.keys().copied().collect();
+        self.outgoing_locks.clear();
+        self.current_lock_id = None;
+
+        info!(
+            count = cleared.len(),
+            "Releasing outgoing locks before taking clipboard ownership"
+        );
+
+        let messages = cleared
+            .iter()
+            .map(|id| into_cliprdr_message(ClipboardPdu::UnlockData(LockDataId(*id))))
+            .collect();
+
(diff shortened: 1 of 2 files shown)


crates/ironrdp-cliprdr/src/lib.rs
```diff
@@ -1027,6 +1027,47 @@ impl<R: Role> Cliprdr<R> {
...
+        }
+
+        let cleared: Vec<u32> = self.outgoing_locks.keys().copied().collect();
+        self.outgoing_locks.clear();
+        self.current_lock_id = None;
+
+        info!(
+            count = cleared.len(),
+            "Releasing outgoing locks before taking clipboard ownership"
+        );
```
CBenoit, line 1058: question: Should we drop this to `debug!` or `trace!`? I’m not sure it’s something we need to show by default 🤔

CBenoit:
Thank you! The fix is looking good to me, I just have one minor concern about the log level of new trace.

CBenoit (approved):
@ymarcus93 For velocity, I’ll merge and send a follow up PR to lower the log level

CBenoit:
It occurred to me that the log level remark was true for many other call sites, I updated that in this PR: https://github.com/Devolutions/IronRDP/pull/1379

ymarcus93:
> Yuval Marcus (@ymarcus93) For velocity, I’ll merge and send a follow up PR to lower the log level

Thanks! I thought I had pushed that fix already

PR #1382, opened 2026-06-24
fix(pdu): set COMPRESSION_USED on the FastPath update header when compressed

glamberson:
`FastPathUpdatePdu::encode` writes the update header byte before setting the `COMPRESSION_USED` bit, so a PDU encoded with `compression_flags` set emits the trailing compression flags byte but never marks it in the header.

`FastPathUpdatePdu::decode` keys on that bit (bits 6..7 of the update header) to decide whether to consume the compression flags byte. With the bit clear, it skips that byte and reads the data length from the wrong offset, so a compressed FastPath update does not round-trip through encode/decode.

This is currently latent: nothing in the workspace encodes a compressed fast-path output PDU (the server only decompresses, on the client side), so the path is never exercised. It surfaces as soon as anything does (it turned up while writing a bulk-decompression round-trip test).

The fix sets the bit on the header before it is written, and adds a round-trip test (`compressed_update_round_trips`) covering the compressed case.


CBenoit (approved):
Good catch! Thank you!

PR #1383, opened 2026-06-25
perf(graphics): portable SIMD inverse DWT (wide + SWAR)

irvingoujAtDevolution:
## Summary

On the WASM web client, frame **decode** dominates (~93% of frame time on a 1080p RemoteFX replay), and within decode the **RFX inverse DWT was ~48%** (the YCbCr→RGBA convert is already SIMD via `yuv`; the entropy/RLE stages are inherently sequential). This vectorizes the inverse DWT with the portable [`wide`](https://crates.io/crates/wide) crate (`i16x8`), so the same code lowers to **wasm `simd128`, x86 SSE/AVX, and ARM NEON** — desktop and browser both benefit.

The encode path is unchanged.

## How it stays bit-exact (no `unsafe`, no `cfg` split)

The lifting steps need i32 intermediates only for the averages. Overflow-free SWAR identities let the whole kernel stay in `i16` lanes (no widen/narrow):

- `ceil_avg(a,b)  = (a|b) - ((a^b)>>1)`  ≡ `(a + b + 1) >> 1`
- `floor_avg(a,b) = (a&b) + ((a^b)>>1)`  ≡ `(a + b) >> 1`

and `(2x+1)>>1 == x` / `(x+x)>>1 == x` simplify the first/last rows. Every other op is wrapping `i16` arithmetic, identical to the old `i32`-intermediate-then-`as i16` truncation.

## Performance

1080p RemoteFX replay, headless Chromium, wasm release `+simd128`, 8-pass median:

| inverse DWT | decode (ms) |
|---|--:|
| scalar (baseline) | ~1598 |
| **portable `wide` SIMD** | **~985** |

→ inverse DWT ~2×, **~39% off the decode stage**. (Absolute ms carry ~±15% machine-load noise; the ratio is stable. Per-frame this is a throughput win — decode was already within real-time budget.)

## Correctness

Verified bit-exact three ways:
- the replay-bench **framebuffer CRC32** is unchanged,
- the existing **native DWT tests** pass (so it's exact on x86 too, not just wasm),
- an **exhaustive** check of the SWAR identities over all `i16 × i16` pairs (0 mismatches).

## Notes

- `wide` is a single-user dep in `ironrdp-graphics`; chosen over `std::simd` (still nightly-only) and over per-arch intrinsics (one portable kernel vs three).
- Reproducible bench branches: `bench/draw-*` (renderer) and the DWT measurements were taken on the replay-bench harness branch (the capture corpus is gitignored).

### DIFF
--- a/crates/ironrdp-graphics/Cargo.toml
+++ b/crates/ironrdp-graphics/Cargo.toml
@@ -26,6 +26,7 @@ byteorder = "1.5" # TODO: remove
 num-derive.workspace = true # TODO: remove
 num-traits.workspace = true # TODO: remove
 yuv = { version = "0.8", features = ["rdp"] }
+wide = "0.7" # portable SIMD for the inverse DWT (wasm/x86/ARM); `std::simd` is still nightly-only
 
 [dev-dependencies]
 bmp = "0.5"
--- a/crates/ironrdp-graphics/src/dwt.rs
+++ b/crates/ironrdp-graphics/src/dwt.rs
@@ -1,4 +1,33 @@
 use ironrdp_pdu::utils::SplitTo as _;
+use wide::i16x8;
+
+/// Max RFX sub-band width. The 8-wide tiling and the `[_; MAX_SUBBAND_WIDTH + 1]` scratch in the
+/// inverse passes rely on `subband_width` being one of {8, 16, 32}.
+const MAX_SUBBAND_WIDTH: usize = 32;
+
+/// Loads 8 contiguous `i16` from `s[off..]` into a vector. The caller guarantees `off + 8 <= s.len()`.
+#[inline]
+fn vld(s: &[i16], off: usize) -> i16x8 {
+    i16x8::from(<[i16; 8]>::try_from(&s[off..off + 8]).expect("off + 8 within bounds"))
+}
+
+/// Stores a vector into `s[off..]`.
+#[inline]
+fn vst(s: &mut [i16], off: usize, v: i16x8) {
+    s[off..off + 8].copy_from_slice(&v.to_array());
+}
+
+/// Ceil average `(a + b + 1) >> 1`, overflow-free (SWAR; arithmetic shift).
+#[inline]
+fn ceil_avg(a: i16x8, b: i16x8) -> i16x8 {
+    (a | b) - ((a ^ b) >> 1)
+}
+
+/// Floor average `(a + b) >> 1`, overflow-free (SWAR; arithmetic shift).
+#[inline]
+fn floor_avg(a: i16x8, b: i16x8) -> i16x8 {
+    (a & b) + ((a ^ b) >> 1)
+}
 
 pub fn encode(buffer: &mut [i16], temp_buffer: &mut [i16]) {
     encode_block::<32>(&mut *buffer, temp_buffer);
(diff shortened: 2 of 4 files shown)


crates/ironrdp-graphics/Cargo.toml
```diff
@@ -26,6 +26,7 @@ byteorder = "1.5" # TODO: remove
 num-derive.workspace = true # TODO: remove
 num-traits.workspace = true # TODO: remove
 yuv = { version = "0.8", features = ["rdp"] }
+wide = "0.7" # portable SIMD for the inverse DWT (wasm/x86/ARM); `std::simd` is still nightly-only
```
CBenoit, line 29: praise: Thank you for documenting the rationale. At some point we may migrate to `std::simd`, when it’s stabilized.

CBenoit (commented):
This is looking good to me, do you have any concern to address before removing the draft status?

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -1,4 +1,33 @@
...
+use wide::i16x8;
+
+/// Max RFX sub-band width. The 8-wide tiling and the `[_; MAX_SUBBAND_WIDTH + 1]` scratch in the
+/// inverse passes rely on `subband_width` being one of {8, 16, 32}.
+const MAX_SUBBAND_WIDTH: usize = 32;
+
+/// Loads 8 contiguous `i16` from `s[off..]` into a vector. The caller guarantees `off + 8 <= s.len()`.
+#[inline]
+fn vld(s: &[i16], off: usize) -> i16x8 {
+    i16x8::from(<[i16; 8]>::try_from(&s[off..off + 8]).expect("off + 8 within bounds"))
```
RRRadicalEdward: _wide_ has an API for creating a SIMD from a slice. Use `i16x8::from_slice_unaligned`

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -1,4 +1,33 @@
...
+}
+
+/// Stores a vector into `s[off..]`.
+#[inline]
+fn vst(s: &mut [i16], off: usize, v: i16x8) {
+    s[off..off + 8].copy_from_slice(&v.to_array());
+}
```
RRRadicalEdward, line 18:
```suggestion
fn vst(s: &mut [i16], off: usize, v: i16x8) {
    s[off..off + 8].copy_from_slice(v.as_array());
}
```
No need to make an array. You can get a reference to inner data directly.

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -1,4 +1,33 @@
...
+
+/// Ceil average `(a + b + 1) >> 1`, overflow-free (SWAR; arithmetic shift).
+#[inline]
+fn ceil_avg(a: i16x8, b: i16x8) -> i16x8 {
+    (a | b) - ((a ^ b) >> 1)
```
RRRadicalEdward, line 23: It would be beneficial to contribute to `wide` to add this operation directly there. There is already a specialized SIMD for that - _mm_avg_epu16, but wide doesn't have this API. So we need to use other SIMD operations to achieve that. If the API was present is `wide`, we could do it at the cost of only a single SIMD operation, which is faster. 

I used to contribute to `wide` when optimizing IronVNC with SIMDs. The author of the library was open to contributions.

crates/ironrdp-graphics/src/dwt.rs
RRRadicalEdward: Also, you can get subslice without repeating off: `s[off..][..8]`. IMHO, it's cleaner.
You can make use of this pattern in many places in this file.

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -1,4 +1,33 @@
...
+
+/// Max RFX sub-band width. The 8-wide tiling and the `[_; MAX_SUBBAND_WIDTH + 1]` scratch in the
+/// inverse passes rely on `subband_width` being one of {8, 16, 32}.
+const MAX_SUBBAND_WIDTH: usize = 32;
+
+/// Loads 8 contiguous `i16` from `s[off..]` into a vector. The caller guarantees `off + 8 <= s.len()`.
+#[inline]
+fn vld(s: &[i16], off: usize) -> i16x8 {
+    i16x8::from(<[i16; 8]>::try_from(&s[off..off + 8]).expect("off + 8 within bounds"))
+}
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -119,94 +148,119 @@ fn decode_block(buffer: &mut [i16], temp_buffer: &mut [i16], subband_width: usiz
...
-    let total_width = subband_width * 2;
+// One band of the inverse horizontal pass, vectorized along `n` (`wide`): `low`/`high` are the two
+// source subband rows (len `sw`), `dst` the reconstructed row (len `2*sw`, even/odd interleaved).
+// Bit-exact with the former scalar code (SWAR averages + wrapping `i16` arithmetic). `sw` is a
+// multiple of 8 and ≤ 32.
+fn horizontal_band(low: &[i16], high: &[i16], dst: &mut [i16], sw: usize) {
+    debug_assert!(
+        sw % 8 == 0 && sw <= MAX_SUBBAND_WIDTH,
+        "sw must be a multiple of 8 and <= MAX_SUBBAND_WIDTH"
+    );
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -119,94 +148,119 @@ fn decode_block(buffer: &mut [i16], temp_buffer: &mut [i16], subband_width: usiz
...
+// first/last rows, the averages use the overflow-free SWAR `ceil_avg`/`floor_avg`, and every other
+// op is wrapping `i16` arithmetic (identical to i32-intermediate-then-truncate).
+// Precondition: `subband_width` is a multiple of 8 and <= MAX_SUBBAND_WIDTH (for the 8-wide tiling).
+fn inverse_vertical(buffer: &mut [i16], temp_buffer: &[i16], subband_width: usize) {
+    let sw = subband_width;
+    let tw = sw * 2;
+    debug_assert!(
+        sw % 8 == 0 && sw <= MAX_SUBBAND_WIDTH,
+        "sw must be a multiple of 8 and <= MAX_SUBBAND_WIDTH"
+    );
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/ironrdp-graphics/src/dwt.rs
irvingoujAtDevolution: updated

crates/ironrdp-graphics/src/dwt.rs
irvingoujAtDevolution: updated

irvingoujAtDevolution:
Hi @CBenoit and @RRRadicalEdward — sorry for the delay. Honestly, I wasn't confident in this change before, since I had no experience in the graphics area. I spent about a day of focused work learning DWT and SIMD, and re-derived the transform by hand as an ugly-but-mathematically-equivalent version to make sure I actually understood it (verified bit-exact against the existing tests). I'm now comfortable marking this PR ready for review since I understand what's going on. Thanks for the patience!

crates/ironrdp-graphics/src/dwt.rs
irvingoujAtDevolution: Will do as follow up

irvingoujAtDevolution:
@CBenoit @RRRadicalEdward Ready for review

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -109,104 +138,124 @@ fn dwt_horizontal<const SUBBAND_WIDTH: usize>(mut buffer: &mut [i16], dwt: &[i16
...
+// LL(3); the L band reconstructs from LL+HL, the H band from LH+HH. Each row is reconstructed by
+// `horizontal_band`.
+fn inverse_horizontal<const SUBBAND_WIDTH: usize>(buffer: &[i16], temp_buffer: &mut [i16]) {
+    let sw = SUBBAND_WIDTH;
+    let tw = sw * 2;
+    let ssw = sw * sw;
+    let hl = &buffer[0..ssw];
+    let lh = &buffer[ssw..2 * ssw];
+    let hh = &buffer[2 * ssw..3 * ssw];
+    let ll = &buffer[3 * ssw..4 * ssw];
```
RRRadicalEdward: Maybe, do it this way, so you don't duplicate the `ssw` variable?
```suggestion
    let mut bands = buffer.chunks(ssw);
    let hl = bands.next()?;
    let lh = bands.next()?;
    let hh = bands.next()?;
    let ll = bands.next()?;
```

crates/ironrdp-graphics/src/dwt.rs
```diff
@@ -109,104 +138,124 @@ fn dwt_horizontal<const SUBBAND_WIDTH: usize>(mut buffer: &mut [i16], dwt: &[i16
...
+    let (l_dst, h_dst) = temp_buffer.split_at_mut(ssw * 2);
+
+    for r in 0..sw {
+        let row = r * sw;
+        horizontal_band::<SUBBAND_WIDTH>(&ll[row..row + sw], &hl[row..row + sw], &mut l_dst[r * tw..r * tw + tw]);
```
RRRadicalEdward:
```suggestion
        horizontal_band::<SUBBAND_WIDTH>(&ll[row..][..sw], &hl[row..][..sw], &mut l_dst[r * tw..][..tw]]);
```
The same for the line below

RRRadicalEdward (approved):
LGTM. I have only a few minor comments

PR #1384, opened 2026-06-26
feat(tls): expose negotiated TLS version and cipher suite

glamberson:
## What

Add `NegotiatedTls` and a backend-neutral `negotiated()` accessor to
`ironrdp-tls`, reporting the TLS protocol version and cipher suite
negotiated for an established `TlsStream`.

- rustls: reports both, read from the live `ClientConnection`.
- native-tls: returns `None` for both (the wrapper only exposes the peer
  certificate, ALPN, and the channel-binding token).
- stub: returns `None`.

`upgrade()` is unchanged, so this is purely additive.

## Why

Downstream RDP diagnostic tooling wants to report a server's negotiated
TLS parameters alongside the certificate `upgrade()` already returns;
today that information is discarded even though rustls exposes it. This
follows the crate's existing pattern of normalizing backend-specific
output into a neutral type, as `upgrade()` already does for the
certificate via `x509-cert`.

## Notes

Version and cipher are returned as their standard string names (e.g.
`TLSv1_3`, `TLS13_AES_256_GCM_SHA384`). Happy to reshape into a typed
form if you prefer.

### DIFF
--- a/crates/ironrdp-tls/src/lib.rs
+++ b/crates/ironrdp-tls/src/lib.rs
@@ -23,7 +23,21 @@ compile_error!("a TLS backend must be selected by enabling a single feature out 
 
 // The whole public API of this crate.
 #[cfg(any(feature = "stub", feature = "native-tls", feature = "rustls"))]
-pub use impl_::{TlsStream, upgrade};
+pub use impl_::{TlsStream, negotiated, upgrade};
+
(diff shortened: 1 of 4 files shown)


CBenoit (approved):
LGTM! Thank you, nice addition

PR #1065, opened 2025-12-23
feat(cliprdr): add request_file_contents method

glamberson:
Per [MS-RDPECLIP section 2.2.5.3][file-contents-spec], the Local Clipboard Owner
sends File Contents Request PDU to retrieve file data from the Shared Clipboard
Owner during paste operations.

This enables server implementations to request file contents from clients,
completing the bidirectional file transfer capability.

## Changes

- `SendFileContentsRequest` message variant in `ClipboardMessage`
- `request_file_contents()` method on `Cliprdr<R>`
- Server handler wiring

## Dependencies

Depends on #1064 (lock/unlock methods).

[file-contents-spec]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeclip/cbc851d3-4e68-45f4-9292-26872a9209f2

PR #1388, opened 2026-06-29
feat(cliprdr): dispatch initiate_file_copy via ClipboardMessage

glamberson:
## Summary

- `Cliprdr::initiate_file_copy` (added in #1166) records a local file list so the peer's `FileContentsRequest` can be serviced. #1166 wired it through the web client and FFI but not through the `ClipboardMessage` backend API.
- A server or client built on that API can receive files but cannot offer them: `SendInitiateCopy` calls `initiate_copy`, leaving `local_file_list` empty, so an incoming `FileContentsRequest` is rejected with "no file list available" (MS-RDPECLIP 3.1.5.4.6) before the backend is consulted.
- Adds `ClipboardMessage::SendInitiateFileCopy(Vec<FileDescriptor>)` and dispatches it to `Cliprdr::initiate_file_copy` in every `ClipboardMessage` consumer: `ironrdp-server`, `ironrdp-client`, `ironrdp-web`, and the `ffi` message-type mapping.

## Validation

- `cargo xtask check fmt/lints/tests/typos/locks` pass.
- `ironrdp-web` additionally checked on `wasm32-unknown-unknown`.

## Notes

- Non-breaking, following the precedent of #194, #1065, and #1166, which added `ClipboardMessage` variants as `feat` while updating every in-tree consumer (done here). The receive-side path (`request_file_contents` / `submit_file_contents`) is unchanged.
- The `ffi` binding categorizes the new message (`ClipboardMessageType::SendInitiateFileCopy`); surfacing the `FileDescriptor` payload through FFI accessors would add a new opaque type surface and is left as a follow-up.

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -640,6 +640,10 @@ async fn active_session(
                                     Some(cliprdr.initiate_copy(&formats)
                                         .map_err(|e| session::custom_err!("CLIPRDR", e))?)
                                 }
+                                ClipboardMessage::SendInitiateFileCopy(files) => {
+                                    Some(cliprdr.initiate_file_copy(files)
+                                        .map_err(|e| session::custom_err!("CLIPRDR", e))?)
+                                }
                                 ClipboardMessage::SendFormatData(response) => {
                                     Some(cliprdr.submit_format_data(response)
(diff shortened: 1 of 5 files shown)


CBenoit (approved):
LGTM. Thank you!

PR #1394, opened 2026-06-30
feat(rdpeusb): implement urbdrc server processors

uchouT:
Part of #1138 

- Implement control and per-device urbdrc server processors, similar to the urbdrc client, maintaining a request_id -> pending_io at the state machine layer.
- introduce backend-facing `rdpeusb::io` request and completion types shared by client and server implementations.
- Refactor TsUrb variants, the rationale is documented in the 0d3d5539 commit message.
- As a follow-up to #1365, handle no-ack OUT transfers through a dedicated client backend method.

> [!NOTE]
> Types in `rdpeusb::io` define a system boundary, worth good design.

cc @elmarco


elmarco:
lgtm, it could use more doc comments and tests, but I suppose this will come next.

How to trigger copilot checks? @CBenoit

CBenoit (commented):
I’m sorry, I need to trigger it manually for now, but I’m discussing internally about having them automatically

CBenoit (approved):

PR #1198, opened 2026-04-01
feat(graphics,egfx): add progressive RFX server encode and mixed-codec frames

glamberson:
Depends on #1197.
Depended on by #1199 (round-trip tests). Part of the multi-codec EGFX implementation described in #1158 (Section 4).

## Summary

Add server-side progressive encode pipeline and multi-codec frame support.

**ironrdp-graphics**: Progressive RFX encoder. First-pass tiles produced via forward DWT + progressive quantization + RLGR1 entropy coding. Upgrade-pass tiles produced via SRL encoding for zero-DAS positions plus raw-bit encoding for non-zero-DAS positions, with monotonic-refinement semantics per MS-RDPRFX 3.1.8.1.7.2.

**ironrdp-egfx**: `GraphicsPipelineServer::send_progressive_frame()` for single-codec progressive frames, plus `send_mixed_frame()` for `MixedTilePayload` batches that combine ClearCodec, Progressive RFX, and AVC420 tiles in a single `StartFrame`/`EndFrame` pair. `MixedTilePayload` is `#[non_exhaustive]` so future codec additions land without a SemVer break.

## Stack ordering

This is step 2 of 3. The full chain (must merge in PR-number order):

| Step | PR | Scope |
|------|-----|-------|
| 1 | #1197 | Progressive RFX decode pipeline + EGFX client/server integration |
| **2** | **this PR** | Progressive RFX server-side encode + mixed-codec frames |
| 3 | #1199 | Progressive RFX encode/decode round-trip tests |

Branch is stacked on `feat/progressive-rfx-decode` (#1197).

## Note (post-rebase fix)

During the 2026-05-20 rebase onto current master, fixed a missing `InclusiveRectangle` import that was masked in the original branch base (`MixedTilePayload::ClearCodec.destination` was declared as `InclusiveRectangle` but pushed into `WireToSurface1Pdu.destination_rectangle` which expects `ExclusiveRectangle`). Corrected the enum variant to use `ExclusiveRectangle` directly per MS-RDPEGFX 2.2.1.4.1 wire-format conventions.

## Test plan

`cargo xtask check fmt/lints/tests/typos/locks` all pass. 10 new unit tests cover `rgba_to_ycbcr`, `quantize_component_ccq`, `RawBitWriter`, `encode_first_pass`, and `encode_upgrade_pass`.

PR #1395, opened 2026-06-30
fix(graphics): don't require CONTEXT block on every progressive frame

rdeangel:
When a server sends a WireToSurface2 PDU for a codec context that's already been established, the CONTEXT block is often omitted. MS-RDPEGFX 2.2.4.2 specifies SYNC + CONTEXT as the establishment sequence for a context (keyed by codec_context_id); frames after that point reference the existing context and don't carry a fresh CONTEXT.

The current decoder treats CONTEXT as mandatory on every decode_bitmap call. Servers that follow the spec and omit it on subsequent frames get rejected with MissingBlock(\"CONTEXT\"), and the image is stuck on the coarse first pass for the rest of the session.

Confirmed against real servers: xrdp and GNOME Remote Desktop both send the CONTEXT block on the first frame only. Reproducing it requires a multi-frame progressive stream from one of those servers, so there's no test in this PR. The fix has been verified end to end against both xrdp and GNOME Remote Desktop.

The change is small. If the current frame has a CONTEXT block, use it (unchanged). If not, fall back to the use_reduce_extrapolate value stored when the context was first created. Only when neither is available, i.e. the very first frame for a context arrived without a CONTEXT block, do we still error out.

Relates to the progressive codec introduced in #1197 / #1198. Open to adding a regression test if there's a preferred way to fixture a multi-frame stream here, or if maintainers want me to generate one from FreeRDP's vectors.

### DIFF
--- a/crates/ironrdp-graphics/src/progressive.rs
+++ b/crates/ironrdp-graphics/src/progressive.rs
@@ -1076,16 +1076,30 @@ impl ProgressiveDecoder {
 
         let blocks = decode_progressive_stream(bitmap_data)?;
 
-        // Extract context flags from the CONTEXT block. Per MS-RDPEGFX 2.2.4.2
-        // a Progressive stream MUST begin with SYNC + CONTEXT; treat absence as
-        // a malformed stream rather than silently defaulting band layout.
-        let use_reduce_extrapolate = blocks
-            .iter()
-            .find_map(|block| match block {
-                ProgressiveBlock::Context(ctx) => Some(ctx.uses_reduce_extrapolate()),
-                _ => None,
-            })
-            .ok_or(ProgressiveDecodeError::MissingBlock("CONTEXT"))?;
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM, thank you!

CBenoit:
question: I just noticed: on the surface-resize branch, a CONTEXT-less resize frame now reallocates using the fallback flag and discards accumulated tile state. Pre-existing, but worth confirming that's acceptable for the resize + omitted-CONTEXT combination.

> Open to adding a regression test if there's a preferred way to fixture a multi-frame stream here, or if maintainers want me to generate one from FreeRDP's vectors.

I won't block this PR on this, but can we get the two-frame regression fixture you offered? Frame 1 with CONTEXT establishing a `codec_context_id`, frame 2 without, asserting frame 2 decodes via the stored flag rather than erroring. A static byte-array fixture avoids needing a live xrdp/GRD server and pins the exact loosening this PR introduces.

PR #1396, opened 2026-06-30
feat(connector): send ClientClusterData with REDIRECTION_SUPPORTED

### LINKED ISSUE #139: Support Server Redirection PDUs

rdeangel:
The cluster field in create_gcc_blocks has been left at \`None\` with a TODO(#139) note for as long as the file has existed. This fills it in.

Sends the values mstsc sends on a fresh connection: \`REDIRECTION_SUPPORTED\`, \`RedirectionVersion::V4\`, \`redirected_session_id\` of 0. Session brokers and terminal-server farms rely on this block (MS-RDPBCGR 2.2.1.3.6) to route the connection; without it the client can be misrouted or dropped on brokered setups.

The field was already optional, so existing servers that don't care about cluster data keep working.

Closes #139.

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -646,9 +646,10 @@ fn create_gcc_blocks<'a>(
     static_channels: impl Iterator<Item = &'a StaticVirtualChannel>,
 ) -> ConnectorResult<gcc::ClientGccBlocks> {
     use ironrdp_pdu::gcc::{
-        ClientCoreData, ClientCoreOptionalData, ClientEarlyCapabilityFlags, ClientGccBlocks, ClientNetworkData,
-        ClientSecurityData, ColorDepth, ConnectionType, EncryptionMethod, HighColorDepth, MonitorOrientation,
-        RdpVersion, SecureAccessSequence, SupportedColorDepths,
+        ClientClusterData, ClientCoreData, ClientCoreOptionalData, ClientEarlyCapabilityFlags, ClientGccBlocks,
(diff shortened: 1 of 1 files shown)


CBenoit:
Thank you, but I’m not sure it’s enough. This sets `REDIRECTION_SUPPORTED` in `ClientClusterData`, which tells the server we can handle an in-session Server Redirection PDU (MS-RDPBCGR 2.2.13) if it sends one. But `ShareControlPduType::ServerRedirect` has no decoder and falls into the generic "unexpected share control PDU type" error path in `ShareControlPdu::from_type`. Advertising support we don't have risks turning a previously-working connection to a broker/farm into a hard failure. #139 was also about implementing the receive-side decode + a connector state to handle it before we flip this flag.

rdeangel:
Thanks for the explanation, that’s a fair point, I missed that `ShareControlPduType::ServerRedirect` has no decoder and would fall into the error path. Flipping this flag without the receive-side handling could indeed break working broker/farm connections.

I’ll close this PR. I stumbled across this flag while testing other things and didn’t realize there was more to it on the receive side. The real fix is the larger piece from #139 (decode + connector state), and I currently have no way to directly test that against a broker/farm that issues Server Redirection PDUs, so I’m not in a position to land it properly. Apologies for the noise.

PR #1397, opened 2026-07-01
feat(acceptor): expose the client's keyboard layout on AcceptorResult

clintcan:
## What

The client announces its keyboard layout identifier (KLID) in the GCC Client Core Data (MS-RDPBCGR 2.2.1.3.2, `keyboardLayout`). `ironrdp-acceptor` already decodes that block in `BasicSettingsWaitInitial` but keeps only the early-capability flags, so a server built on the acceptor has no way to learn the client's layout.

This adds `AcceptorResult::keyboard_layout: u32` (the announced KLID; `0` when the client didn't send one), captured during `BasicSettingsWaitInitial` and stored on `Acceptor` so it survives a deactivation-reactivation (same as `desktop_size`).

## Why

A server can use it to select a matching **server-side** keyboard layout without touching any local input state — e.g. translating positional scancodes through the client's announced layout for non-US clients. Today the KLID is decoded and then thrown away.

This mirrors the existing pattern of surfacing already-parsed client data on `AcceptorResult`:
- #1155 — `credentials`
- #1347 — `message_channel_id`

The motivating downstream is a macOS RDP server (macrdp) that auto-detects the client's keyboard layout; this is the last piece it vendors a fork of the acceptor for on this front.

## Notes

- **Additive** field on a result struct that only `get_result` constructs internally — labeled `feat(acceptor):` to match #1155/#1347 (same shape).
- No new tests: the acceptor lib sets `test = false` (`Cargo.toml`) and there's no acceptor harness in `ironrdp-testsuite-core`, so there's no in-tree place to assert it (the value is a straight passthrough of the already-decoded `gcc_blocks.core.keyboard_layout`). Happy to add coverage wherever you'd prefer it to live.
- `cargo build` / `cargo fmt --check` / `cargo clippy -p ironrdp-acceptor --all-targets -- -D warnings` clean (only the pre-existing workspace MSRV-config notice).

### DIFF
--- a/crates/ironrdp-acceptor/src/connection.rs
+++ b/crates/ironrdp-acceptor/src/connection.rs
@@ -31,6 +31,7 @@ pub struct Acceptor {
     user_channel_id: u16,
     message_channel_id: Option<u16>,
     desktop_size: DesktopSize,
+    keyboard_layout: u32,
     server_capabilities: Vec<CapabilitySet>,
     static_channels: StaticChannelSet,
     saved_for_reactivation: AcceptorState,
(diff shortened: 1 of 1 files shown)


clintcan:
Flagging this small one for your queue 🙂 — CI green and `mergeable`. It's a companion to #1373: just surfaces the client's already-decoded KLID on `AcceptorResult` (same additive shape as #1155 `credentials` and #1347 `message_channel_id`), so a server can pick a matching layout without a vendored acceptor fork.

Only open point is testing: the acceptor lib is `test = false` and there's no acceptor harness in `ironrdp-testsuite-core`, so I left it untested (it's a straight passthrough of `gcc_blocks.core.keyboard_layout`). Happy to add coverage wherever you'd prefer it.

CBenoit (approved):
LGTM! Simple enough 🙂

PR #1400, opened 2026-07-01
fix(graphics): apply YCbCr->RGB fixed-point descale in progressive decoder/encoder

rdeangel:
## Problem

Against some real-world EGFX servers (xrdp, GNOME Remote Desktop), the progressive (RemoteFX) decoder produces posterised output. Smooth colour gradients collapse to a handful of saturated primaries and luma loses its mid-tones. Spatial detail (text, icons) stays crisp, which points away from a scaling issue and toward colour conversion.

## Root cause

`reconstruct_to_rgba()` passes the dequantized Y/Cb/Cr values into the BT.601 YCbCr->RGB conversion, but the fixed-point headroom that `dequantize_component_ccq` intentionally retains isn't removed before conversion.

`dequantize_component_ccq` applies the spec-defined `<< (quant_value - 1)` shift. For the wire quant table this works out to +5 bits of headroom, which matches FreeRDP's reference decoder (see the `/* -6 + 5 = -1 */` comment in `libfreerdp/primitives/prim_colors.c`). The colour-conversion stage is the natural place to remove that headroom (`R = ((CrR + Y) >> divisor) >> 5`). This PR adds the missing `>> 5` and the corresponding `128 << 5 = 4096` luma offset. Without them, each channel exceeds the legal ±128 range and `clamp_u8` pins it to 0 or 255.

Worth noting: this isn't the integer inverse DWT. The DWT is unity-gain, verified separately below.

## Verification

1. **DWT unity-gain check.** A flat 64×64 tile of value 100 forward-transforms with no coefficient exceeding ~100 and round-trips back to exactly 100, so the `×2^5` scale doesn't originate there. Added `tests/dwt_gain.rs` to lock this in.
2. **Live server quant trace.** Against an xrdp stream, the per-band `<< (q-1)` factors land at +5 bits of headroom (max coefficient scaled ~32×). The +5 bits come from the dequant table.
3. **FreeRDP cross-check.** `libfreerdp/codec/progressive.c` and `libfreerdp/primitives/prim_colors.c` apply the same `<< (q-1)` shift and the matching `>> 5` removal at the colour stage, so this patch's math lines up with the reference.
4. **Visual confirmation.** With the `>> 5` descale applied, xrdp and GNOME RD streams render with correct colour gradients; without it, the same streams are posterised.

## Scope of this PR

Kept symmetric so the crate's internal encode<->decode round-trip test stays identity:
- **Decoder:** `>> 5` plus `+128` luma offset in `reconstruct_to_rgba` (matching FreeRDP).
- **Encoder:** matching `<< 5` plus offset on the RGB->YCbCr path in `rgba_to_ycbcr`, so the round-trip doesn't regress.

## Files
- `crates/ironrdp-graphics/src/progressive.rs` — decoder descale and encoder scale-up
- `crates/ironrdp-graphics/tests/dwt_gain.rs` — regression test confirming the DWT is unity-gain

### DIFF
--- a/crates/ironrdp-graphics/src/progressive.rs
+++ b/crates/ironrdp-graphics/src/progressive.rs
@@ -449,9 +449,14 @@ pub fn rgba_to_ycbcr(pixels: &[u8], y_out: &mut [i16], cb_out: &mut [i16], cr_ou
         let cb = (-11059 * r - 21709 * g + 32768 * b + 32768) >> 16;
         let cr = (32768 * r - 27439 * g - 5329 * b + 32768) >> 16;
 
-        y_out[i] = clamp_i16(y);
-        cb_out[i] = clamp_i16(cb);
-        cr_out[i] = clamp_i16(cr);
+        // Add the 5-bit fixed-point headroom the decoder removes with `>> 5` in
+        // `reconstruct_to_rgba`. The base dequant retains +5 bits (matching
+        // FreeRDP's `/* -6 + 5 = -1 */`); scaling up here keeps the
+        // encode->decode round-trip an identity. Range after `<< 5` is
+        // [-4096, 4064], well within i16.
+        y_out[i] = clamp_i16(y << 5);
+        cb_out[i] = clamp_i16(cb << 5);
+        cr_out[i] = clamp_i16(cr << 5);
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/ironrdp-graphics/src/progressive.rs
```diff
@@ -830,9 +835,17 @@ impl TileState {
...
-            let cr = i32::from(cr_buf[i]);
+            // The base dequantization (`<< (quant - 1)`) retains +5 bits of
+            // fixed-point headroom by design (the DWT itself is unity-gain --
+            // see tests/dwt_gain.rs), so the spatial coefficients arrive scaled
+            // up by 2^5. FreeRDP folds this descale into its colour primitive
+            // (yCbCrToRGB_16s8u_P3AC4R shifts by `divisor + 5`); we apply the
+            // matching `>> 5` here. Without it every value overflows the ±128
+            // range and clamps to 0/255 -- luma loses mid-tones and chroma
+            // collapses to pure primaries (the posterised image).
+            let y = (i32::from(y_buf[i]) >> 5) + 128;
```
mamoreau-devolutions, line 846: MS-RDPEGFX §3.3.8.2.3 delegates this conversion to MS-RDPRFX §3.1.8.2.5, which only level-shifts Y by 128. The `>> 5` is compensating for `dequantize_component_ccq` using `quant - 1`, but MS-RDPRFX §3.1.8.1.5 Figure 8 defines the base scale as `1 << (quantization factor - 6)`. For the usual factor 6, spatial components already have the specified `[-128, 127]` scale. Please correct base quantization/dequantization instead of baking its constant error into color conversion; otherwise intermediate DWT values can truncate before this descale.

crates/ironrdp-graphics/tests/dwt_gain.rs
```diff
@@ -0,0 +1,49 @@
...
+    // Round-trip must be identity.
+    let mut rt = [p; N];
+    let mut tmp2 = [0i16; N];
+    dwt::encode(&mut rt, &mut tmp2);
+    dwt::decode(&mut rt, &mut tmp2);
+    let (rmin, rmax) = rt
+        .iter()
+        .fold((i16::MAX, i16::MIN), |(lo, hi), &v| (lo.min(v), hi.max(v)));
+
+    // Guard: unity-gain. No forward coefficient exceeds the input magnitude P
```
mamoreau-devolutions, line 40: This only checks a flat DWT input and never calls either changed function, `rgba_to_ycbcr` or `TileState::reconstruct_to_rgba`, so it will pass if the new `<< 5` or `>> 5` is removed or incorrect. Please add a known progressive/RFX tile vector with a factor-6 base quant table and assert reconstructed RGB values, plus a complementary assertion on the encoder's wire coefficients.

mamoreau-devolutions:
fixed by https://github.com/Devolutions/IronRDP/pull/1499

PR #1403, opened 2026-07-02
refactor(rdpeusb): seperate raw and validated `InternalIoControl`

uchouT:
Keep INTERNAL_IO_CONTROL as a lossless wire representation so unknown IOCTL codes and buffer fields remain decodable.

Convert raw PDUs into supported backend requests at the state-machine boundary. This centralizes semantic validation and makes invalid outbound requests unrepresentable.


CBenoit (approved):
LGTM! Thank you!

PR #1404, opened 2026-07-02
feat(acceptor)!: clamp honored client desktop size to an operator maximum

clintcan:
Follow-up to #1373 (the resource-hardening angle you flagged in review — thanks for the go-ahead 🙂).

## Problem

`#1373` gated honor-client-desktop-size behind a bare `bool`. With it on, the acceptor adopts the client-requested desktop size bounded only by the protocol range `[200, 8192]`. But the desktop size is a client-controlled `u16`, and the server still builds its framebuffer/encoder from the negotiated size — so a client could request e.g. `8192x8192` and drive the server's allocation off an untrusted number (~256 MiB per frame buffer). Mild, and only on an opt-in default-off path, but it's a resource-exhaustion vector driven purely by a number the client picks.

Your review comment: *"[200, 8192] is a protocol ceiling, not a resource guard … tracked the 'clamp/range policy rather than a bare bool' idea as a future follow-up (an operator-set max size)."* This is that PR.

## Change

Replace the `bool` with `Option<DesktopSize>` carrying an **operator-set maximum**:

- `None` (default) — disabled; always enforce the server-provided size (unchanged behavior).
- `Some(max)` — honor the client's request, **clamped per dimension to `max`**. The client can ask for a smaller desktop, never a larger one.

The acceptor clamps the requested `width`/`height` to `max` *before* the existing `validate_desktop_size` protocol-range check, so the negotiated size can never exceed what the operator is willing to render — set `max` to the host display's native resolution (or whatever ceiling the server can afford).

Touches:
- `ironrdp-acceptor`: field + `set_honor_client_desktop_size(Option<DesktopSize>)` + clamp-then-validate in the Client Core Data adopt path.
- `ironrdp-server`: `RdpServerOptions::honor_client_desktop_size: Option<DesktopSize>` + `RdpServerBuilder::with_honor_client_desktop_size(Option<DesktopSize>)`.

## Breaking (pre-v1)

`RdpServerOptions::honor_client_desktop_size` is now `Option<DesktopSize>`; the acceptor setter and the builder method take `Option<DesktopSize>` instead of `bool`. Callers migrate `true` → `Some(max)` (choosing a ceiling) and `false` → `None`. Same set of downstream server impls as before (lamco-rdp-server, hypr-rdp, cosmic-ext-rdp-server, ARISU, macrdp).

## Design points for review

1. **Clamp vs reject.** I clamp per dimension (`min(requested, max)`), so a client over the cap gets the largest allowed size rather than falling back to the server default. Per-dimension clamping can shift the aspect ratio (e.g. client `1600x1200`, max `1920x1080` → `1600x1080`); the display handler + client-side scaling absorb that, and it keeps honoring useful right up to the ceiling. Happy to switch to reject-if-over-max (keep server size) if you'd prefer no aspect change.
2. **No inline test**, same as #1373 — the acceptor lib is `test = false` and there's no acceptor harness in `ironrdp-testsuite-core`. The clamp is a two-line `min` ahead of the existing `validate_desktop_size`; I'm glad to add coverage wherever you'd want acceptor tests to live (a pure `clamp_desktop_size` helper + a testsuite-core case, if useful).

Built + `clippy --features egfx -D warnings` clean on `ironrdp-acceptor` + `ironrdp-server`.

### DIFF
--- a/crates/ironrdp-acceptor/src/connection.rs
+++ b/crates/ironrdp-acceptor/src/connection.rs
@@ -150,10 +150,11 @@ impl Acceptor {
     /// later in the client's Confirm Active is, per [MS-RDPBCGR] 2.2.1.13.2,
     /// the value the client copied from the *server's* Demand Active, so it
     /// cannot be used to discover what the client originally asked for. When
-    /// this is enabled and the client's request is within the protocol-legal
-    /// range, the acceptor negotiates that size from the start (it is written
-    /// into the server's Bitmap capability set before Demand Active is sent),
-    /// avoiding a Deactivation-Reactivation resize round trip.
+    /// this is enabled, the client's request is first clamped per dimension to
+    /// the operator maximum and then validated against the protocol-legal range;
+    /// if the clamped size is legal the acceptor negotiates it from the start (it
+    /// is written into the server's Bitmap capability set before Demand Active is
+    /// sent), avoiding a Deactivation-Reactivation resize round trip.
     ///
     /// Pass `Some(max)` to honor the client's request, clamped per dimension to
     /// `max`: the client can ask for a *smaller* desktop than `max`, but never a
@@ -567,9 +568,10 @@ impl Sequence for Acceptor {
                         }
                     } else {
                         debug!(
-                            width = requested_width,
-                            height = requested_height,
-                            "Client requested an out-of-range desktop size; keeping the server-provided size"
+                            requested = ?DesktopSize { width: requested_width, height: requested_height },
+                            clamped = ?DesktopSize { width: clamped_width, height: clamped_height },
+                            max = ?max,
+                            "Client-requested desktop size is out of protocol range after clamping to the operator maximum; keeping the server-provided size"
                         );
                     }
                 }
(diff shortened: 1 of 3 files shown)


clintcan:
Addressed the Copilot review in `f32886a5`:
- Reworded both doc blocks (acceptor setter + builder `with_honor_client_desktop_size`) to describe the actual **clamp-then-validate** behavior — an oversized client request is honored *clamped down*, not rejected.
- The out-of-range `debug!` now logs `requested` + `clamped` + `max` and says "out of protocol range after clamping," so a misconfigured `max` (e.g. < 200) reads correctly instead of blaming the client.

clintcan:
Rebased onto the latest `master` — the only conflict was a trivial union in `builder.rs` (the merged auto-reconnect-cookie field sat next to this PR's `honor_client_desktop_size` type change), so it's `MERGEABLE` again and CI is green.

This is the honor-size resource-hardening follow-up you green-lit in the #1373 approval (replacing the bare `bool` with an operator-set `Option<DesktopSize>` cap, so an authenticated client can't drive an 8192×8192 framebuffer allocation). It's had no human review since it was opened, so flagging it as ready whenever you have a moment. Happy to adjust anything.

mamoreau-devolutions (approved):

PR #1405, opened 2026-07-02
feat(server): send the Server Auto-Reconnect Cookie during logon

clintcan:
## What

Adds an optional **Server Auto-Reconnect Cookie** (MS-RDPBCGR 2.2.4.3 `ARC_SC_PRIVATE_PACKET`) to `ironrdp-server`. When set, `RdpServer` sends a Save Session Info PDU (`LogonExtended` + `AUTO_RECONNECT_COOKIE`) carrying it once per connection, right after activation (Confirm Active processed, encoder built), on the IO channel.

## Why

A client only enters its **automatic reconnection sequence** (MS-RDPBCGR 1.3.1.5) on an *ungraceful* disconnect if it was handed this cookie during logon. Without it, a dropped connection just reports as disconnected — **mstsc in particular won't auto-reconnect at all**. Today `ironrdp-server` never sends the cookie, so there's no way for a server to opt into that behavior.

The concrete use case: a server that *intentionally* drops a connection and expects the client to come straight back — e.g. a recovery path that cycles the session — currently forces the user to reconnect by hand. With the cookie provisioned, the client re-establishes on its own (re-authenticating via NLA/CredSSP from cached credentials). It's also just standard behavior a real RDP server provides.

## API

Mirrors the existing `credential_validator` pattern exactly — a builder method plus a runtime setter:

```rust
// build time
RdpServer::builder()
    .with_auto_reconnect_cookie(Some(ServerAutoReconnect { logon_id, random_bits }))
    // ...

// or dynamically
server.set_auto_reconnect_cookie(Some(cookie));
```

`ServerAutoReconnect` (already public in `ironrdp-pdu::rdp::session_info`) carries a `logon_id` and a 16-byte `random_bits` (generate from a CSPRNG). A per-connection guard (`auto_reconnect_sent`, reset in `run_connection_with`) sends it exactly once — not again on a Deactivation-Reactivation.

## Scope / additive

- **Additive, non-breaking.** Default is `None` (send nothing); existing servers are byte-for-byte unaffected.
- All PDU types already exist in `ironrdp-pdu` (`rdp::session_info::{SaveSessionInfoPdu, LogonInfoExtended, LogonExFlags, ServerAutoReconnect, InfoType, InfoData}`) — this is purely wiring the server-side send.
- Reuses the existing `encode_share_data_pdu` helper.

## Design point for review — the returning cookie

This PR implements the **send** side only: it *enables* the client's automatic reconnection. It does **not** validate the `ARC_CS_PRIVATE_PACKET` the client sends back on reconnect (MS-RDPBCGR 2.2.4.4). For a server that re-authenticates every connection by other means (NLA/CredSSP) that's sufficient and safe, and it's documented as such on the setter. If you'd prefer the crate to also offer *validation* of the returning cookie (so it can be an authentication factor — the server would store issued `(logon_id, random_bits)` and verify the client's `SecurityData`/`ARC_CS` on the next connect), I'm happy to do that as a follow-up, or fold it in here — it's a larger, stateful feature so I kept this PR to the send path. Let me know which you'd prefer.

Built + `clippy --features egfx -D warnings` clean; tests compile.

### DIFF
--- a/crates/ironrdp-server/src/builder.rs
+++ b/crates/ironrdp-server/src/builder.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use anyhow::Result;
 use ironrdp_pdu::rdp::capability_sets::{BitmapCodecs, server_codecs_capabilities};
+use ironrdp_pdu::rdp::session_info::ServerAutoReconnect;
 use tokio_rustls::TlsAcceptor;
 
 use super::clipboard::CliprdrServerFactory;
@@ -43,6 +44,7 @@ pub struct BuilderDone {
     display_suppressed: Option<Arc<AtomicBool>>,
     autodetect_rtt: Option<Arc<AtomicU32>>,
     honor_client_desktop_size: bool,
+    auto_reconnect_cookie: Option<ServerAutoReconnect>,
 }
 
 pub struct RdpServerBuilder<State> {
@@ -144,6 +146,7 @@ impl RdpServerBuilder<WantsDisplay> {
                 display_suppressed: None,
                 autodetect_rtt: None,
                 honor_client_desktop_size: false,
+                auto_reconnect_cookie: None,
             },
         }
     }
@@ -166,6 +169,7 @@ impl RdpServerBuilder<WantsDisplay> {
                 display_suppressed: None,
                 autodetect_rtt: None,
                 honor_client_desktop_size: false,
+                auto_reconnect_cookie: None,
             },
         }
     }
@@ -289,6 +293,23 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
+    /// Provision the Server Auto-Reconnect Cookie (MS-RDPBCGR 2.2.4.3
+    /// `ARC_SC_PRIVATE_PACKET`) handed to the client during logon.
+    ///
+    /// When set to `Some`, the server sends a Save Session Info PDU carrying the
+    /// cookie once per connection, right after activation, which is what lets a
+    /// client automatically re-establish the session after an *ungraceful*
+    /// disconnect (MS-RDPBCGR 1.3.1.5, "Automatic Reconnection") rather than
+    /// reporting the connection as simply lost. Generate the cookie's 16-byte
+    /// `random_bits` from a CSPRNG. `None` (the default) sends no cookie.
+    ///
+    /// See [`RdpServer::set_auto_reconnect_cookie`] for the runtime equivalent
+    /// and a note on the (unvalidated) returning `ARC_CS_PRIVATE_PACKET`.
+    pub fn with_auto_reconnect_cookie(mut self, cookie: Option<ServerAutoReconnect>) -> Self {
+        self.state.auto_reconnect_cookie = cookie;
+        self
+    }
+
     pub fn build(self) -> RdpServer {
         let mut server = RdpServer::new(
             RdpServerOptions {
@@ -309,6 +330,7 @@ impl RdpServerBuilder<BuilderDone> {
             self.state.autodetect_rtt,
         );
         server.set_credential_validator(self.state.credential_validator);
+        server.set_auto_reconnect_cookie(self.state.auto_reconnect_cookie);
         server
     }
 }
--- a/crates/ironrdp-server/src/builder.rs
+++ b/crates/ironrdp-server/src/builder.rs
@@ -293,7 +293,7 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
-    /// Provision the Server Auto-Reconnect Cookie (MS-RDPBCGR 2.2.4.3
+    /// Provision the Server Auto-Reconnect Cookie (MS-RDPBCGR 2.2.4.2
     /// `ARC_SC_PRIVATE_PACKET`) handed to the client during logon.
     ///
     /// When set to `Some`, the server sends a Save Session Info PDU carrying the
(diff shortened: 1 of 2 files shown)


clintcan:
Addressed the Copilot review in `c1cf3186`: the doc links now point at the reachable `RdpServer::builder` (and name `with_auto_reconnect_cookie` in code formatting) instead of `crate::RdpServerBuilder::…`, which doesn't resolve.

Root cause aside, in case it's of interest: `RdpServerBuilder` isn't re-exported from the crate root (`lib.rs` exports `RdpServer`/`RdpServerOptions` but not the builder type), so *every* `crate::RdpServerBuilder::…` intra-doc link in `server.rs` is currently broken. A one-line `pub use builder::RdpServerBuilder;` would fix them all at once, but that's a public-API surface change I left out of this PR — happy to send it separately if you'd want it.

crates/ironrdp-server/src/server.rs
```diff
@@ -1474,6 +1528,29 @@ impl RdpServer {
...
+        // once per connection. This Save Session Info PDU is what lets a client
+        // automatically re-establish the session after an ungraceful disconnect
+        // (MS-RDPBCGR 1.3.1.5). Not re-sent on a Deactivation-Reactivation
+        // sequence (`auto_reconnect_sent` guard, reset per connection).
+        if !self.auto_reconnect_sent {
+            if let Some(cookie) = self.auto_reconnect_cookie.clone() {
+                let pdu = rdp::headers::ShareDataPdu::SaveSessionInfo(rdp::session_info::SaveSessionInfoPdu {
+                    info_type: rdp::session_info::InfoType::LogonExtended,
+                    info_data: rdp::session_info::InfoData::LogonExtended(rdp::session_info::LogonInfoExtended {
+                        present_fields_flags: rdp::session_info::LogonExFlags::AUTO_RECONNECT_COOKIE,
```
mamoreau-devolutions, line 1551: **P1 — validate the returning cookie before advertising this feature.** Setting LOGON_EX_AUTORECONNECTCOOKIE causes clients to reconnect without resending credentials, but the acceptor decodes ClientInfoPdu and retains only credentials; it drops the session ID and reconnect verifier. No code computes or compares the HMAC verifier, or binds the new connection to the prior session. MS-RDPBCGR §5.5 requires the server to look up the session random, calculate the verifier, and grant access only on a match. Please expose the reconnect fields through the acceptor and validate them before accepting the resumed connection.

crates/ironrdp-server/src/server.rs
```diff
@@ -1474,6 +1528,29 @@ impl RdpServer {
...
+                        auto_reconnect: Some(cookie),
+                        errors_info: None,
+                    }),
+                });
+                let data = encode_share_data_pdu(pdu, result.io_channel_id, result.user_channel_id)?;
+                writer.write_all(&data).await.context("send auto-reconnect cookie")?;
+                self.auto_reconnect_sent = true;
```
mamoreau-devolutions, line 1558: **P1 — this prevents the required cookie rotation.** After the first send, uto_reconnect_sent remains true, so set_auto_reconnect_cookie cannot send an updated cookie to the active client. The same configured cookie is also reused by each later 
un_connection. MS-RDPBCGR §5.5 requires the cookie to be regenerated whenever a client connects to the session or it is reset, and §3.3.6.2 requires hourly updates. Add a session-bound cookie lifecycle and a path to emit replacements while the connection remains active.

crates/ironrdp-server/src/builder.rs
```diff
@@ -289,6 +293,23 @@ impl RdpServerBuilder<BuilderDone> {
         self
     }
 
+    /// Provision the Server Auto-Reconnect Cookie (MS-RDPBCGR 2.2.4.3
```
mamoreau-devolutions: **P2 — correct the section reference.** ARC_SC_PRIVATE_PACKET is MS-RDPBCGR §2.2.4.2; §2.2.4.3 defines the client-side ARC_CS_PRIVATE_PACKET. The same incorrect server-packet reference appears in the RdpServer docs.

mamoreau-devolutions (changes requested):
The initial Save Session Info encoding uses the correct extended-logon form, but the implementation advertises automatic reconnect without implementing the required verifier validation or cookie lifecycle.

clintcan:
Thanks for the thorough review — the section references are fixed in `7bc1b76` (and I caught a sibling: the returning `ARC_CS_PRIVATE_PACKET` was cited as 2.2.4.4, now 2.2.4.3).

On the two P1s — you're right that this PR only *sends* the cookie; it doesn't validate the returning `ARC_CS_PRIVATE_PACKET` or rotate. I'd like to propose landing it in two phases, and get your read on whether that's acceptable:

**This PR (send-side, opt-in, default off).** It's already useful and safe for the common IronRDP-server case where CredSSP/NLA gates every connection: there, the reconnect still completes CredSSP from the client's cached credentials before the RDP-layer auto-reconnect, so the cookie is *not* a credential bypass — it only spares the user a manual reconnect. The doc already states plainly that the returning cookie is not validated and that a server treating the cookie as an authentication factor must verify it.

**Follow-up (the ARC lifecycle you're describing).** Verifier validation per §5.5 (expose the reconnect fields through the acceptor, look up the session random, compute + compare the HMAC, bind the resumed connection) plus the cookie lifecycle (per-connection regeneration, hourly refresh per §3.3.6.2, and clearing the send latch so `set_auto_reconnect_cookie` can push a replacement to a live client). That's a meaningful chunk of acceptor + server work and reviews better on its own.

If you'd rather not advertise `LOGON_EX_AUTORECONNECTCOOKIE` at all until validation exists in-tree, I understand — I'm happy to build the full lifecycle here instead. Just let me know which you'd prefer and I'll proceed. I can open a tracking issue for the validation/rotation piece either way.

mamoreau-devolutions (approved):

PR #1409, opened 2026-07-05
Credential hand-over to sub-servers (Wrdp/canonical stack)

rcarmo:
Hello! 

I am creating a “wrdp” server that follows the xrdp-sesman model, and broke off this set of commits (along with a little note you can discard) that allows IronRDP to effectively own credential handling while still enabling session hand-off to a process that will then handle the actual user session setup - without that process needing to reinvent the wheel.


rcarmo:

PR #1410, opened 2026-07-05
Expose CredSSP delegated credentials to servers

rcarmo:
This is a counterpart to #1409 - I wanted to make it really clear where the credential capture was made.


rcarmo:

PR #1411, opened 2026-07-05
Add post-auth connection binder hook

rcarmo:
This is another portion of auth binding glue following #1409 and #1410


rcarmo:

PR #1412, opened 2026-07-05
Wrdp/reactivation credential cache

rcarmo:
This is the final piece in the #1409 #1410 #1411 series - it is effectively an in-session cache to cope with Windows App’s foibles when negotiating credentials, and only holds on to the credentials during session setup.


rcarmo:

PR #1415, opened 2026-07-06
fix(pdu): decode MousePdu wheel rotation as two's complement, matching encode

### LINKED ISSUE #1414: MousePdu::decode mis-decodes negative wheel rotation (sign-magnitude instead of two's complement)
## Summary

`ironrdp-pdu`'s `MousePdu::decode` (crates/ironrdp-pdu/src/input/mouse.rs) computes
`number_of_wheel_rotation_units` incorrectly for negative values, and the bug means
the codec doesn't even round-trip its own output.

## Current decode

```rust
let wheel_rotations_bits = flags_raw as u8;
let number_of_wheel_rotation_units = if flags.contains(PointerFlags::WHEEL_NEGATIVE) {
    -i16::from(wheel_rotations_bits)
} else {
    i16::from(wheel_rotations_bits)
};
```

## The problem

Per MS-RDPBCGR 2.2.8.1.1.3.1.1.3, the wheel-rotation field (`WheelRotationMask`,
0x01FF) is a **9-bit two's-complement** value — the `WHEEL_NEGATIVE` (0x0100) bit
doubles as the sign bit of that 9-bit field, not an independent "negate this
magnitude" flag. The correct decode (matching this crate's own `encode`, and
matching FreeRDP's `update_read_pointer_event`) is:

```rust
if negative { value = -(0x100 - byte) } else { value = byte }
```

i.e. for a byte of `0xFF` (255) with `WHEEL_NEGATIVE` set, the true value is `-1`,
not `-255`.

**This crate's own `encode` side already does this correctly** —
`MousePdu::encode` computes `self.number_of_wheel_rotation_units as u8` (a
truncating cast, i.e. proper two's complement), so a delta of `-1` is encoded as
byte `0xFF` + the negative flag. `decode` then reads that same byte back as
`-255`, not `-1`. Concretely: `decode(encode(MousePdu { number_of_wheel_rotation_units: -1, .. }))`
produces `-255`, not `-1` — **the codec does not round-trip.**

The encode-side truncating cast was introduced in #71 (2023); the decode side
appears to have never been updated to match, so this has likely been present
since then.

## Real-world impact

Found while debugging a scroll-runaway bug in a third-party RDP server
(clintcan/macrdp#113): a client sending fine-grained wheel deltas of `-1`/`-2`/`-3`
per tick (a Windows App / macOS trackpad-driven scroll) had every downward tick
decoded as `-255`/`-254`/`-253` — a max-magnitude scroll on every gentle tick,
in one direction only (positive deltas decode correctly, since no sign bit is
involved). Clients that only ever send ±120-multiple notches never notice, because
the decode error (`-136` instead of the true `-120`) doesn't change the outcome
after typical downstream scaling — which is likely why this has gone unnoticed for
~3 years.

## Suggested fix

```rust
let number_of_wheel_rotation_units = if flags.contains(PointerFlags::WHEEL_NEGATIVE) {
    i16::from(wheel_rotations_bits) - 0x100
} else {
    i16::from(wheel_rotations_bits)
};
```

(`byte - 256` for the negative case, equivalent to `-(256 - byte)`.) I'm happy to
open a PR with this fix plus a round-trip test if that's useful — let me know.

antonmos:
Fixes #1414.

## Summary

`MousePdu::decode` computed a negative `number_of_wheel_rotation_units` as
plain sign-magnitude (`-i16::from(byte)`), but per MS-RDPBCGR
2.2.8.1.1.3.1.1.3 the `WheelRotationMask` (0x01FF) field is 9-bit
**two's complement** — `WHEEL_NEGATIVE` (0x0100) is the sign bit of that
9-bit value, not an independent "negate this magnitude" flag.

`MousePdu::encode` already does this correctly (a truncating cast,
`self.number_of_wheel_rotation_units as u8`, is proper two's complement),
so the two sides disagreed and the codec didn't round-trip:
`decode(encode(-1))` produced `-255`, not `-1`.

## Fix

```rust
let number_of_wheel_rotation_units = if flags.contains(PointerFlags::WHEEL_NEGATIVE) {
    i16::from(wheel_rotations_bits) - 0x100
} else {
    i16::from(wheel_rotations_bits)
};
```

Added:
- `wheel_rotation_units_round_trip_through_encode_decode` — round-trips
  every `i8` value through `encode`/`decode` (this fails on `master` for
  any negative value before this fix).
- `small_negative_wheel_rotation_decodes_correctly` — decodes a
  hand-built wire buffer (`WHEEL_NEGATIVE` + byte `0xFF`) and asserts the
  result is `-1`, not `-255`.

## Test plan
- [x] `cargo test -p ironrdp-pdu --lib` — 369 passed (2 new)
- [x] `cargo clippy -p ironrdp-pdu --lib -- -D warnings` — clean on the
      changed file (two pre-existing, unrelated errors elsewhere in the
      crate reproduce identically on `master` without this change —
      confirmed via `git stash`)
- [x] `rustfmt --check` clean

### DIFF
--- a/crates/ironrdp-pdu/src/input/mouse.rs
+++ b/crates/ironrdp-pdu/src/input/mouse.rs
@@ -68,8 +68,15 @@ impl<'de> Decode<'de> for MousePdu {
         )]
         let wheel_rotations_bits = flags_raw as u8;
 
+        // Per MS-RDPBCGR 2.2.8.1.1.3.1.1.3, WheelRotationMask (0x01FF) is a 9-bit
+        // TWO'S-COMPLEMENT field: WHEEL_NEGATIVE (0x0100) is the sign bit of that
+        // 9-bit value, not an independent "negate this magnitude" flag. So a byte
+        // of 0xFF with WHEEL_NEGATIVE set means -1, not -255. This must mirror
+        // `encode` below, which already produces a proper two's-complement byte
+        // via a truncating cast (`self.number_of_wheel_rotation_units as u8`) —
+        // without this, `decode(encode(x))` does not round-trip for x < 0.
         let number_of_wheel_rotation_units = if flags.contains(PointerFlags::WHEEL_NEGATIVE) {
-            -i16::from(wheel_rotations_bits)
+            i16::from(wheel_rotations_bits) - 0x100
         } else {
             i16::from(wheel_rotations_bits)
         };
@@ -99,5 +106,50 @@ bitflags! {
         const DOWN = 0x8000;
 
         const _ = !0;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use ironrdp_core::{decode, encode_vec};
+
+    use super::*;
+
+    fn mouse_pdu(number_of_wheel_rotation_units: i16) -> MousePdu {
+        MousePdu {
+            flags: PointerFlags::VERTICAL_WHEEL,
+            number_of_wheel_rotation_units,
+            x_position: 0,
+            y_position: 0,
+        }
+    }
+
+    #[test]
+    fn wheel_rotation_units_round_trip_through_encode_decode() {
+        // Every representable value must survive an encode/decode round trip.
+        // This previously failed for small negative values: encode(-1) produced
+        // byte 0xFF + WHEEL_NEGATIVE, which decode incorrectly read back as -255
+        // (sign-magnitude) instead of -1 (two's complement, matching encode).
+        for value in i8::MIN..=i8::MAX {
+            let value = i16::from(value);
+            let pdu = mouse_pdu(value);
+            let buffer = encode_vec(&pdu).unwrap();
+            let decoded: MousePdu = decode(buffer.as_slice()).unwrap();
+            assert_eq!(
+                decoded.number_of_wheel_rotation_units, value,
+                "round trip failed for {value}"
+            );
+        }
+    }
+
+    #[test]
+    fn small_negative_wheel_rotation_decodes_correctly() {
+        // WHEEL_NEGATIVE set, byte = 0xFF -> true value is -1 (two's complement:
+        // byte - 0x100), NOT -255 (sign-magnitude: -byte).
+        let flags = (PointerFlags::VERTICAL_WHEEL | PointerFlags::WHEEL_NEGATIVE).bits() | 0x00FF;
+        let mut buffer = [0u8; 6];
+        buffer[0..2].copy_from_slice(&flags.to_le_bytes());
+        let pdu: MousePdu = decode(buffer.as_slice()).unwrap();
+        assert_eq!(pdu.number_of_wheel_rotation_units, -1);
     }
 }
--- a/crates/ironrdp-pdu/src/input/mouse.rs
+++ b/crates/ironrdp-pdu/src/input/mouse.rs
@@ -26,6 +26,10 @@ impl Encode for MousePdu {
             PointerFlags::empty().bits()
         };
 
+        // The wire field is 9-bit two's complement, representable range
+        // [-256, 255] — narrower than i16. A value outside that range
+        // truncates silently here (predates this fix; not this PR's job to
+        // widen the field, but worth flagging for the next reader).
         #[expect(
             clippy::as_conversions,
             clippy::cast_sign_loss,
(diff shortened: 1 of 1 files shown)


crates/ironrdp-pdu/src/input/mouse.rs
```diff
@@ -101,3 +108,48 @@ bitflags! {
...
+    #[test]
+    fn wheel_rotation_units_round_trip_through_encode_decode() {
+        // Every representable value must survive an encode/decode round trip.
+        // This previously failed for small negative values: encode(-1) produced
+        // byte 0xFF + WHEEL_NEGATIVE, which decode incorrectly read back as -255
+        // (sign-magnitude) instead of -1 (two's complement, matching encode).
+        for value in i8::MIN..=i8::MAX {
+            let value = i16::from(value);
+            let pdu = mouse_pdu(value);
+            let buffer = encode_vec(&pdu).unwrap();
```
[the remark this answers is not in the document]
CBenoit: Indeed

crates/ironrdp-pdu/src/input/mouse.rs
```diff
@@ -68,8 +68,15 @@ impl<'de> Decode<'de> for MousePdu {
         )]
         let wheel_rotations_bits = flags_raw as u8;
 
+        // Per MS-RDPBCGR 2.2.8.1.1.3.1.1.3, WheelRotationMask (0x01FF) is a 9-bit
+        // TWO'S-COMPLEMENT field: WHEEL_NEGATIVE (0x0100) is the sign bit of that
+        // 9-bit value, not an independent "negate this magnitude" flag. So a byte
+        // of 0xFF with WHEEL_NEGATIVE set means -1, not -255. This must mirror
+        // `encode` below, which already produces a proper two's-complement byte
```
CBenoit: nitpick: Comment says "encode below" but encode is above decode. Otherwise excellent comment.

crates/ironrdp-pdu/src/input/mouse.rs
CBenoit, line 8: thought: The struct field is wider than the wire field, and encode truncates silently. `number_of_wheel_rotation_units: i16` can hold values the 9-bit wire field can't (e.g. 300), and `encode` silently truncates them via `as u8` (300 would serialize as a small positive). This predates the PR and isn't its job to fix, but the new round-trip test now implicitly documents the contract only for the tested subrange. If you widen the test per the suggestion above, worth a one-line comment noting the representable domain is [-256, 255], or a `debug_assert!` in encode, otherwise the next person may assume the full i16 round-trips.

antonmos: Will do!

crates/ironrdp-pdu/src/input/mouse.rs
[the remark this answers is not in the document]
antonmos: Widened the loop to the actual 9-bit two's-complement domain, `-256i16..=255i16`. Pushed in a0b2d414.

crates/ironrdp-pdu/src/input/mouse.rs
antonmos: Fixed — pushed in a0b2d414.

crates/ironrdp-pdu/src/input/mouse.rs
antonmos: Added a comment at the encode truncating cast noting the representable domain is [-256, 255] (i16 is wider than the 9-bit wire field). Pushed in a0b2d414.

antonmos: Replaced the comment with a `debug_assert!` checking `(-256..=255).contains(&self.number_of_wheel_rotation_units)` in encode. Confirmed it can't fire against the existing fuzz corpus: `pdu_round_trip_one!` only ever re-encodes a value that just came out of `decode`, and decode's two's-complement math can't produce anything outside that range. Pushed in 9fa725f9.

crates/ironrdp-pdu/src/input/mouse.rs
```diff
@@ -26,6 +26,14 @@ impl Encode for MousePdu {
             PointerFlags::empty().bits()
         };
 
+        // The wire field is 9-bit two's complement: representable range is
+        // [-256, 255], narrower than i16.
+        debug_assert!(
+            (-256..=255).contains(&self.number_of_wheel_rotation_units),
+            "number_of_wheel_rotation_units out of the 9-bit two's-complement range [-256, 255]: {}",
+            self.number_of_wheel_rotation_units
+        );
```
CBenoit, line 35: thought: Out of scope for this PR, but it may be worth to consider a new type pattern for the number of wheel rotation units (`WheelRotationUnits(i64)`) with the invariant that its value is between -256 and 255, so that it’s guaranteed at the type-level.

CBenoit (approved):
Thank you!

PR #1416, opened 2026-07-07
feat(dvc): create channels with assigned id

uchouT:
## Summary

- Add `create_channel_with` for creating dynamic channel with their assigned dynamic channel id.

## Motivation

Some dynamic channel processors need their assigned channel ID during construction, for example when creating a backend handle that routes messages to that channel. The existing `create_channel()` API only exposes the ID after the processor has already been constructed.

cc @elmarco

### DIFF
--- a/crates/ironrdp-dvc/src/server.rs
+++ b/crates/ironrdp-dvc/src/server.rs
@@ -1,8 +1,10 @@
 use alloc::boxed::Box;
 use alloc::collections::BTreeMap;
+use alloc::collections::btree_map::Entry;
 use alloc::vec::Vec;
 use core::any::TypeId;
 use core::fmt;
+use core::sync::atomic::AtomicUsize;
 
 use ironrdp_core::{Decode as _, DecodeResult, ReadCursor, impl_as_any, invalid_field_err};
 use ironrdp_pdu::{self as pdu, decode_err, encode_err, pdu_other_err};
@@ -42,6 +44,54 @@ impl Drop for DynamicChannel {
     }
 }
 
+/// A reserved dynamic channel ID awaiting a channel processor.
+///
+/// The ID is provisional until [`Self::create`] is called. Reserved IDs increase
+/// monotonically and are never reused for the lifetime of the server. Dropping a
+/// reservation without creating the channel permanently skips its ID.
+#[derive(Debug)]
+pub struct DynamicChannelReservation {
+    server_id: usize,
+    channel_id: u32,
+}
+
+impl DynamicChannelReservation {
+    /// Returns the reserved dynamic channel ID.
+    pub fn channel_id(&self) -> u32 {
+        self.channel_id
+    }
+
+    /// Registers the channel and returns its DVC Create Request message.
+    ///
+    /// # Panics
+    ///
+    /// Panics if reservation finalized on a different DrdynvcServer than the one that minted it.
+    pub fn create<T>(self, server: &mut DrdynvcServer, channel: T) -> PduResult<SvcMessage>
+    where
+        T: DvcServerProcessor + 'static,
+    {
+        assert_eq!(
+            self.server_id, server.id,
+            "reservation finalized on a different DrdynvcServer than the one that minted it"
+        );
+        let channel_name = channel.channel_name().into();
+        let channel_id = self.channel_id();
+        server
+            .dynamic_channels
+            .insert_reserved(channel_id, channel, ChannelState::Creation);
+
+        // TODO: Align the TypeId lookup semantics of DrdynvcServer::create_channel and
+        // DrdynvcServer::with_dynamic_channel in a separate public-contract PR, then make
+        // DrdynvcServer::create_channel use the reservation path.
+        server
+            .type_id_to_channel_id
+            .entry(TypeId::of::<T>())
+            .or_insert(channel_id);
+        let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
+        as_svc_msg_with_flag(req)
+    }
+}
+
 struct DynamicChannelAllocator {
     dynamic_channels: BTreeMap<u32, DynamicChannel>,
     next_channel_id: u32,
@@ -77,13 +127,27 @@ impl DynamicChannelAllocator {
     where
         T: DvcServerProcessor + 'static,
     {
+        let channel_id = self.reserve_channel();
+        self.insert_reserved(channel_id, processor, state);
+        channel_id
+    }
+
+    fn insert_reserved<T>(&mut self, channel_id: u32, processor: T, state: ChannelState)
+    where
+        T: DvcServerProcessor + 'static,
+    {
+        let Entry::Vacant(entry) = self.dynamic_channels.entry(channel_id) else {
+            unreachable!("reserved dynamic channel ID must be vacant")
+        };
+        entry.insert(DynamicChannel::new(processor, channel_id, state));
+    }
+
+    fn reserve_channel(&mut self) -> u32 {
         let channel_id = self.next_channel_id;
-        self.dynamic_channels
-            .insert(channel_id, DynamicChannel::new(processor, channel_id, state));
         self.next_channel_id = self
             .next_channel_id
             .checked_add(1)
-            .expect("dynamic channels reaches `u32::MAX`");
+            .expect("dynamic channels reached `u32::MAX`");
         channel_id
     }
 
@@ -117,10 +181,14 @@ impl DynamicChannel {
         self.processor.as_any().type_id()
     }
 }
+
+static NEXT_SERVER_ID: AtomicUsize = AtomicUsize::new(0);
+
 /// DRDYNVC Static Virtual Channel (the Remote Desktop Protocol: Dynamic Virtual Channel Extension)
 ///
 /// It adds support for dynamic virtual channels (DVC).
 pub struct DrdynvcServer {
+    id: usize,
     dynamic_channels: DynamicChannelAllocator,
     type_id_to_channel_id: BTreeMap<TypeId, u32>,
 }
@@ -145,6 +213,7 @@ impl DrdynvcServer {
 
     pub fn new() -> Self {
         Self {
+            id: NEXT_SERVER_ID.fetch_add(1, core::sync::atomic::Ordering::Relaxed),
             dynamic_channels: DynamicChannelAllocator::new(),
             type_id_to_channel_id: BTreeMap::new(),
         }
@@ -169,7 +238,7 @@ impl DrdynvcServer {
     ///
     /// # Panics
     ///
-    /// Panics if the number of registered dynamic channels reaches `u32::MAX`.
+    /// Panics if the number of registered dynamic channels reached `u32::MAX`.
     #[must_use]
     pub fn with_dynamic_channel<T>(mut self, channel: T) -> Self
     where
@@ -221,9 +290,28 @@ impl DrdynvcServer {
     {
         let channel_name = channel.channel_name().into();
 
+        // TODO: Reuse self.reserve_channel() after the TypeId lookup contract is aligned
+        // in a separate PR.
         let channel_id = self.dynamic_channels.insert_channel(channel, ChannelState::Creation);
         let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
         as_svc_msg_with_flag(req)
+    }
+
+    /// Reserves the next dynamic channel ID without creating the channel.
+    ///
+    /// Reserved IDs increase monotonically and are never reused. Dropping the returned
+    /// reservation without creating the channel skips its ID, preventing ABA issues.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the dynamic channel ID space is exhausted.
+    #[must_use]
+    pub fn reserve_channel(&mut self) -> DynamicChannelReservation {
+        let channel_id = self.dynamic_channels.reserve_channel();
+        DynamicChannelReservation {
+            channel_id,
+            server_id: self.id,
+        }
     }
 
     fn remove_by_channel_id(&mut self, id: u32) -> Option<DynamicChannel> {
@@ -231,7 +319,7 @@ impl DrdynvcServer {
             let type_id = dvc.processor_type_id();
 
             // Only matters for pre-registered channels
-            if let alloc::collections::btree_map::Entry::Occupied(entry) = self.type_id_to_channel_id.entry(type_id)
+            if let Entry::Occupied(entry) = self.type_id_to_channel_id.entry(type_id)
                 && entry.get() == &id
             {
                 entry.remove();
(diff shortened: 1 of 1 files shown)


elmarco (commented):
## Review

Thanks for the PR — the motivation is clear and the implementation is clean. A few things to address:

### 1. `type_id_to_channel_id` not updated on `create()`

`with_dynamic_channel` registers the processor type in `type_id_to_channel_id` (so `get_channel_id_by_type::<T>()` works), but `DynamicChannelReservation::create()` doesn't — it only has access to the `VacantEntry`, not the `DrdynvcServer`.

This means `get_channel_id_by_type::<T>()` will silently return `None` for channels created via the reservation path. Meanwhile `remove_by_channel_id` will try to clean up a mapping that was never inserted (harmless but asymmetric).

If this is intentional, it should be documented on `reserve_channel()` (e.g. "channels created through a reservation are not discoverable via `get_channel_id_by_type`"). If not, `create()` needs access to the type-id map — which likely means the reservation should borrow `&mut DrdynvcServer` instead of just the `VacantEntry`, or `create()` should return something that lets the caller register the type afterward.

### 2. Leaked channel IDs on drop

The doc correctly says dropping leaves the ID unused, but `next_channel_id` has already been incremented, so the ID slot is permanently lost. With `u32::MAX` IDs this is fine in practice, but if reservations are created and dropped in a retry loop, the ID space erodes silently.

Worth adding a note that IDs are not reclaimed. Alternatively, deferring the increment to `create()` would avoid this, though the borrowing gets trickier.

### 3. Test coverage

There are no tests for the new API. A minimal test exercising the reserve → `channel_id()` → `create()` → verify-channel-exists flow would catch regressions.

### 4. Minor nit

The panic message `"dynamic channels reaches u32::MAX"` matches the existing one in `insert_channel`, but both read better as "reached" (past tense). Not a blocker.

elmarco:
Claude decided to post the review directly :)

crates/ironrdp-dvc/src/server.rs
```diff
@@ -110,21 +117,20 @@ impl DynamicChannelAllocator {
...
 
-    fn reserve_channel(&mut self) -> DynamicChannelReservation<'_> {
+    fn entry_channel(&mut self) -> (&mut u32, VacantEntry<'_, u32, DynamicChannel>) {
         let channel_id = self.next_channel_id;
-        self.next_channel_id = self
-            .next_channel_id
+        self.next_channel_id
             .checked_add(1)
-            .expect("dynamic channels reaches `u32::MAX`");
+            .expect("dynamic channels reached `u32::MAX`");
```
elmarco: why not do the overflow check in create() ?

uchouT: The idea is to fail fast.

Because DynamicChannelReservation<'_> takes a &mut to next_channel_id, concurrent channel creation is already impossible. Catching the overflow early at entry() guarantees we panic immediately, which cleanly avoids any wasted work or state changes before reaching create().

uchouT: Or we can make `reserve_channel()` return Option<DynamicChannelReservation<'_>>, `None` for overflow.

elmarco: yes, better avoid any reachable expect()/unwrap()

elmarco:
lgtm, @CBenoit ?

crates/ironrdp-dvc/src/server.rs
```diff
@@ -169,14 +216,16 @@ impl DrdynvcServer {
...
     #[must_use]
     pub fn with_dynamic_channel<T>(mut self, channel: T) -> Self
     where
         T: DvcServerProcessor + 'static,
     {
         let channel_id = self.dynamic_channels.insert_channel(channel, ChannelState::Pending);
-        self.type_id_to_channel_id.insert(TypeId::of::<T>(), channel_id);
+        self.type_id_to_channel_id
+            .entry(TypeId::of::<T>())
+            .or_insert(channel_id);
```
CBenoit: issue: type_id_to_channel_id changes are unrelated to the reservation API and change a public contract. Same in `create_channel`. Those are two separate public-contract shifts riding along with the feature:
- with_dynamic_channel flips duplicate-type resolution from last-wins to first-wins.
- create_channel-created channels become visible to get_channel_id_by_type, where before
    they returned None.

This is arguably a consistency fix, but they belong in their own PR that can be reviewed and reverted independently, especially since the CHANGELOG is generated from commits and this change would otherwise be invisible to downstream consumers.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -42,6 +43,39 @@ impl Drop for DynamicChannel {
...
+}
+
+impl DynamicChannelReservation<'_> {
+    /// Returns the reserved dynamic channel ID.
+    pub fn channel_id(&self) -> u32 {
+        *self.entry.key()
+    }
+
+    /// Registers the channel and returns its DVC Create Request message.
+    pub fn create<T>(self, channel: T) -> PduResult<SvcMessage>
```
CBenoit: praise: The move-only token is the right shape: single-use finalization at compile time for free.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -42,6 +43,39 @@ impl Drop for DynamicChannel {
...
 }
 
+/// A reserved dynamic channel ID awaiting a channel processor.
+///
+/// Dropping this value without calling [`Self::create`] does not consume the reserved ID.
+pub struct DynamicChannelReservation<'a> {
+    next_channel_id: &'a mut u32,
+    entry: VacantEntry<'a, u32, DynamicChannel>,
+    type_id_to_channel_id: &'a mut BTreeMap<TypeId, u32>,
+}
```
CBenoit: suggestion: Drop the borrow, and carry a plain channel id + a server instance id instead.

rationale: The reservation holding `&'a mut` into the server means the whole DrdynvcServer is exclusively borrowed while a reservation is live. I think that fights the use case: if constructing the processor (or its backend handle) needs to touch the server, you're stuck, and you can't hold more than one reservation.
By switching the handle to a plain value: { server_id: usize, channel_id: u32 }. It holds no borrow, so multiple reservations can be outstanding and the server stays mutable between reserve and create. `create` then takes `&mut DrdynvcServer`.

The residual risk of dropping the borrow is finalizing a handle on a different DrdynvcServer than minted it, which silently succeeds without bumping that server's counter and collides later. I do think it’s anecdotal, because we typically don’t deal with multiple DrdynvcServer at once. That being said, we can close this gap at runtime: stamp each server with a unique id from a global `AtomicUsize` (you can use `Relaxed` read-write operation; carried as a fiel), copy it into the handle, and assert equality in `create` using `assert_eq!`. This converts silent corruption into a loud, deterministic panic and keeps the handle Send/Sync/Debug/storable.

For the idea:

```rust
use core::sync::atomic::{AtomicUsize, Ordering};

static NEXT_SERVER_ID: AtomicUsize = AtomicUsize::new(0);

pub struct DrdynvcServer {
    server_id: usize, // <- new field
    /* … */
}

impl DrdynvcServer {
    pub fn new() -> Self {
        Self { server_id: NEXT_SERVER_ID.fetch_add(1, Ordering::Relaxed), /* … */ } // <- get a unique runtime ID
    }
    pub fn reserve_channel(&mut self) -> DynamicChannelReservation {
        DynamicChannelReservation { server_id: self.server_id, channel_id: self.dynamic_channels.reserve() } // <- pass on the server ID to the handle, so they are tied together.
    }
}

pub struct DynamicChannelReservation {
    server_id: usize,
    channel_id: u32,
}

impl DynamicChannelReservation {
    pub fn channel_id(&self) -> u32 { self.channel_id }

    pub fn create<T>(self, server: &mut DrdynvcServer, channel: T) -> PduResult<SvcMessage>
    where T: DvcServerProcessor + 'static {
        assert_eq!(
            self.server_id, server.server_id,
            "reservation finalized on a different DrdynvcServer than the one that minted it",
        );
        // … insert_reserved(self.channel_id, channel) as before …
    }
}
```

CBenoit: suggestion: Eager allocation is fine.

rationale: Multiple outstanding handles requires assigning distinct ids at reserve time, i.e. bumping next_channel_id eagerly, i.e. a dropped-uncreated reservation skips an id. That's fine: the space is u32 and no session approaches it. Put that rationale in the doc: monotonic, never reused, dropping skips (not reuses) the id, panic on u32::MAX exhaustion. Keep the existing note that channel_id() is provisional until create(). We also don’t really need to deal with exhaustion in practice, and it’s fine to panic by default, the API should reflect that.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -222,15 +271,30 @@ impl DrdynvcServer {
         let channel_name = channel.channel_name().into();
 
         let channel_id = self.dynamic_channels.insert_channel(channel, ChannelState::Creation);
+        self.type_id_to_channel_id
+            .entry(TypeId::of::<T>())
+            .or_insert(channel_id);
         let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
         as_svc_msg_with_flag(req)
     }
```
CBenoit, line 238: suggestion: Rewrite create_channel in terms of the reserve API.

With the borrow gone, this composes cleanly and kills the current duplication between create_channel and the reservation's create():

```rust
    pub fn create_channel<T>(&mut self, channel: T) -> PduResult<SvcMessage>
    where T: DvcServerProcessor + 'static {
        self.reserve_channel().create(self, channel)
    }
```

crates/ironrdp-dvc/src/server.rs
```diff
@@ -42,6 +43,39 @@ impl Drop for DynamicChannel {
...
+        self.type_id_to_channel_id
+            .entry(TypeId::of::<T>())
+            .or_insert(channel_id);
+        let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
+        *self.next_channel_id += 1;
+        as_svc_msg_with_flag(req)
+    }
+}
+
 struct DynamicChannelAllocator {
```
CBenoit, line 45: suggestion: Encapsulate the counter bump behind the DynamicChannelAllocator. Expose `reserve() -> u32` (bump + panic-on-max) and `insert_reserved(id, channel)` (vacant-or-panic), so create() doesn't reach through two layers of fields. The checked_add-discard oddity in `entry_channel()` disappears too.

CBenoit (commented):
This is shaping up well, and the core idea is sound.

I mainly have two requests:
- I would like to bundle unrelated public-contract changes in a separate PR landing first on its own.
- About the ergonomics: the reservation mechanism should switch to an owned handle with a runtime ID so we can tie together the `DrdynvcServer` with the handle without borrowing, which is just enough. Code is otherwise functionally sound.

CBenoit:
> ### 2. Leaked channel IDs on drop
> 
> The doc correctly says dropping leaves the ID unused, but `next_channel_id` has already been incremented, so the ID slot is permanently lost. With `u32::MAX` IDs this is fine in practice, but if reservations are created and dropped in a retry loop, the ID space erodes silently.
> 
> Worth adding a note that IDs are not reclaimed. Alternatively, deferring the increment to `create()` would avoid this, though the borrowing gets trickier.

This comment was against a diff I didn’t review, but I think my redesign suggestion is reintroducing something similar: don’t reclaim skipped IDs, because monotonic, never-reused IDs are a deliberate safety property to avoid ABA holes, and the `u32` space is effectively inexhaustible for a per-session allocator: exhausting the space needs ~4.3 billion reserve-drop cycles. Even a pathological no-backoff retry loop burning thousands of failed reservations a second would need hours-to-days of continuous churn in a single session to bite, which is arguably a caller bug. The realistic version of "retry loop" is tamer than it sounds: the fallible step is processor construction before create (build the processor from channel_id(), then finalize), and retry loops are should be bounded _and_ backoff exponentially. I don’t expect more than a handful of IDs to be consumed at worse before the retry loop is stopped.

crates/ironrdp-dvc/src/server.rs
CBenoit: I’m handling that part.

CBenoit:
Review stages found no PR-specific correctness issue. The protocol concern about a capabilities guard matches pre-existing create-channel behavior and is not a regression introduced here; the change remains low risk with technical-debt classification.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -42,6 +44,54 @@ impl Drop for DynamicChannel {
...
+        // DrdynvcServer::with_dynamic_channel in a separate public-contract PR, then make
+        // DrdynvcServer::create_channel use the reservation path.
+        server
+            .type_id_to_channel_id
+            .entry(TypeId::of::<T>())
+            .or_insert(channel_id);
+        let req = DrdynvcServerPdu::Create(CreateRequestPdu::new(channel_id, channel_name));
+        as_svc_msg_with_flag(req)
+    }
+}
```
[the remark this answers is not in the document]
uchouT: I did consider `create_channel_with<T>(&mut self, f: impl FnOnce(u32) -> PduResult<T>)`.

I initially rejected this API shape because processor construction and DVC channel creation have two independent error domains. Making the closure generic over `E` is not sufficient by itself, because the operation may still produce a `PduError` after the closure succeeds. The API would therefore need either a nested result such as:

```rust
Result<PduResult<SvcMessage>, E>
```
or an additional public error type combining `E` and `PduError`.

The reservation API keeps these two phases and their errors separate, and additionally supports deferred channel construction. The trade-off is that it introduces a larger and more misuse-prone public API.

@CBenoit @elmarco Would like to hear your insight.

crates/ironrdp-dvc/src/server.rs
```diff
@@ -221,11 +233,39 @@ impl DrdynvcServer {
...
+
+    /// Creates a new DVC using a processor built with its assigned channel ID.
+    ///
+    /// The next channel ID is reserved and passed to `build`, allowing the
+    /// processor or one of its dependencies to use the ID during construction.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of registered dynamic channels reaches `u32::MAX`.
+    pub fn create_channel_with<T, E, F>(&mut self, build: F) -> Result<SvcMessage, E>
```
uchouT, line 248: I'm considering to add default type for error generic type E: `E = PduError`

crates/ironrdp-dvc/src/server.rs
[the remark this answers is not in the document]
CBenoit: Context: we discussed on Matrix.
Outcome: I think you are right here, let’s go with this API.

crates/ironrdp-dvc/src/server.rs
CBenoit: Discussed on Matrix: this does not fix ordinary inference. Keeping `<T, E, F>` without a default is probably good for now. Open to a follow up PR if you see fit!

CBenoit (approved):
LGTM!

PR #1417, opened 2026-07-07
feat(server): add rdpeusb server integration

uchouT:
Related discussion: #1516.

*Currently blocked by:*
- #1682 — `ironrdp-usb`, the protocol-independent USB layer
- #1683 — `ironrdp-rdpeusb::usb`, the RDPEUSB adapter, which is also blocked by #1682

> **Reviewing this PR:** because #1682 and #1683 have not merged yet, the *Files changed* tab shows the cumulative diff. The incremental diff for this PR alone is here:
> https://github.com/uchouT/IronRDP/compare/ironrdp-rdpeusb-usb...ironrdp-server

This PR carries the design rationale for all three, since it is the first place where the whole stack is visible at once.

[qemu-rdp](https://gitlab.com/uchouT/qemu-display/-/tree/feat/usbredir) now redirects USB mass storage over RDP end to end, verified against a FreeRDP client built from source. These three PRs split the reusable part out into IronRDP, so that any RDP server can offer USB redirection instead of reimplementing the channel.

## The layering

#1516 proposed splitting this into three layers. The implementation followed that split; this section records what each layer ended up owning.

```text
ironrdp-usb        USB semantics: descriptors, control requests, transfers.
      ↑            Knows nothing about any transport.
ironrdp-rdpeusb    RDPEUSB adapter: TS_URB forms, USBD conventions,
      ↑            capability negotiation, channel state.
ironrdp-server     Server facade: request lifetime, cancellation, routing.
      ↑            Speaks ironrdp-usb types outward.
application        A usbredir bridge, like qemu-rdp, macrdp, or any other consumer.
```

The rule used to place a concept is: **it may live in a layer only if its contract can be explained using that layer's vocabulary alone** — USB, or MS-RDPEUSB, or a generic server facade. A concept that can only be justified by what one downstream consumer happens to need stays in that consumer.

That rule is not free, so two worked examples:

- `CompletionData` (#1683) and `InterfaceAlloc` (this PR) were both added because the server layer needed them, and both stayed in `ironrdp-rdpeusb` because both are explainable purely in RDPEUSB terms: one is the union of the four completion result types the channel defines, the other allocates the interface IDs RDPEUSB requires per redirected device. Being needed downstream was neither sufficient on its own nor disqualifying.
- Going the other way: usbredir packet types and IDs, its synchronous/asynchronous ordering rules, its inflight scheduling, and its stream policy all stayed out of these crates, because none of them can be stated without naming usbredir.

## What the facade exposes

Behind an optional `usb` feature:

- **`DeviceFactory`** creates a per-device backend when the client opens a redirection channel; **`UsbRedirDevice`** receives the device announcement, device text, and channel close.
- **`UsbDeviceHandle`** drives the device in `ironrdp-usb` terms: `get_descriptor`, `get_configuration`, `select_configuration`, `get_interface`, `select_interface`, `control_transfer`, `bulk_transfer`, `interrupt_transfer`, `isochronous_transfer`, `clear_halt`, `reset_device`, and `current_frame_number`, plus `query_device_text_request` and `retract_request`.
- **`PendingRequest`**, **`PendingHandle`**, **`CompletionFut`**, and **`RawPending`** own the request lifetime: a caller can await a completion, hold a cancellation handle, or drop the request. Dropping a submitted request emits `CANCEL_REQUEST`, so an abandoned transfer does not stay in flight on the client.
- **`RdpServerBuilder::with_usb_factory`** opts a server in. The channel is created only when the client negotiates the capability.

No `TS_URB`, RDPEUSB processor, or `ServerEvent` wiring appears in any of those signatures.

## What stays in this layer, and why

Request identifiers, the pending-request map, cancellation, device closure, and completion routing all live here rather than in `ironrdp-rdpeusb`. Keeping them out of the protocol crate is what lets #1683 stay sans-I/O and independently testable, and it means a second consumer of the facade gets the lifetime handling for free without inheriting RDPEUSB details.

## Also in this PR

`refactor(rdpeusb): expose InterfaceAlloc at the crate root` — `InterfaceAlloc` was a private helper of `UrbdrcListener`, so a server driving the channel had no way to allocate the interface IDs it must assign to redirected devices. It is now public at the crate root, with the initial value expressed through `Default`. This is a pure relocation: the allocation range and the exhaustion behaviour are unchanged.

## TODOs

Stated up front rather than left to be discovered:

1. **The `usb` feature is not linted in CI.**. Locally, `cargo clippy -p ironrdp-server --features usb --all-targets -- -D warnings` is clean. Whether to add the feature to the CI matrix is a maintainer call, so I have not touched CI configuration here.
2. **RDPEUSB types still reaches the consumer.** `RdpUsbDeviceAnnounceInfo` carries the raw `DeviceAnnounce`, so a consumer that wants the device speed reads `DeviceSpeed`, an RDPEUSB PDU type. The goal is for a downstream bridge to need no `ironrdp-rdpeusb` dependency at all. I would rather agree on the facade shape than guess at it, so it is left as a follow-up.


uchouT:
cc @clintcan, macrdp (another server built on ironrdp-server) maintainer. This facade is meant to cover that case too.

clintcan:
Thanks for the cc, @uchouT — and yes, macrdp is squarely the downstream this is for. We've carried a vendored server-direction URBDRC implementation for a while (our "divergence 16", ~1400 lines), and I'd genuinely like to delete it in favor of this. I diffed your stack (`ironrdp-usb` + `ironrdp-rdpeusb::usb` + this server PR) against ours — here's what I can offer from the field.

**The hard-won mstsc-interop core is all here, and nicely factored.** The URB-level fixes we had to discover the painful way against real mstsc are all present in `ironrdp-rdpeusb::usb`: **typed** control URBs (`CLASS_INTERFACE` / `GET_DESCRIPTOR_FROM_INTERFACE` / …) with a generic fallback; `USBD_SHORT_TRANSFER_OK` on bulk/interrupt IN; `SelectConfiguration` emitting one interface-info **per interface number** plus the **full** configuration descriptor; and `SET_FEATURE`/`CLEAR_FEATURE` routed as `TRANSFER_IN`. Those are the exact things real Windows rejects a header-only or generic form of — having them in a protocol-independent layer is better than where we keep them. The `RequestId` correlation with PDU-layer completion-shape validation is also cleaner than ours.

**Two things I'd flag before an mstsc device actually streams — both FreeRDP-lenient / mstsc-strict traps we already hit:**

**1 — The per-device channel needs its own capability exchange, or mstsc closes it.** Your per-device processor's `start()` (`UrbdrcDeviceServer`) returns `CHANNEL_CREATED` directly, with **no preceding `RIM_EXCHANGE_CAPABILITY_REQUEST`** — your main `UrbdrcControlServer::start()` sends one, the per-device one doesn't. That's fine on FreeRDP (its readiness is global to the main channel), but we found live (mstsc, mid-2026) that **mstsc closes a per-device channel that gets `CHANNEL_CREATED` / `RIMCALL_RELEASE` without a preceding capability exchange** — it keeps the channel open-but-silent waiting for the caps request, so the device never emits `ADD_DEVICE`. Our per-device processor therefore runs the *full* caps → `CHANNEL_CREATED` handshake on each per-device channel, same as the main one. Since this PR is FreeRDP-verified, the concrete question is: **does an mstsc RemoteFX-USB device ever reach `ADD_DEVICE` on this branch?** If mstsc is hard for you to test, I'm happy to — it's the single highest-value thing to verify here. (Note mstsc gates RemoteFX USB behind a Group Policy + reboot and excludes mass storage, so the test device is a HID/gamepad or camera, not a flash drive.)

**2 — The processors fail-hard on a decode/state error; we deliberately don't.** Both `UrbdrcControlServer` / `UrbdrcDeviceServer` `?`-propagate on a bad decode / `"invalid state"` / `"completion mismatch"`. On our dispatch seam that tears down the whole RDP session — so for an *opt-in* channel we swallow decode errors (`Ok(vec![])`) rather than let a malformed URBDRC PDU kill a working session. (It's also where cargo-fuzz found a few real panics in the PDU decoders, since patched.) Might be worth settling a policy — whether DRDYNVC dispatch isolates a per-processor error to its own channel, or the processor tolerates decode errors — since a redirected-USB channel shouldn't be able to drop the session.

**Smaller notes:**
- The two-phase `submit → PendingRequest.wait()` + `EndpointAddress` addressing + facade-owned pipe binding is a nice model (auto-`CANCEL_REQUEST` on drop; a `CompletionFut` dropping into a `FuturesUnordered` suits our bulk read-ahead engine well). It's a real but worthwhile port for us — we currently address by raw `pipe_handle` and single-await.
- Re your TODO about `RdpUsbDeviceAnnounceInfo` still exposing the RDPEUSB `DeviceAnnounce`: we read `VID:PID:bcdDevice` off the announce/descriptor to dedup a device the client double-announces (one controller per physical device). A transport-independent announce that surfaces those cleanly would let that policy stay downstream where it belongs.

Net: this is a strong replacement for our div-16 — if 1–2 land (or 1 is proven unnecessary on mstsc), we'd adopt it and drop ~1400 lines of vendored code, keeping only our presenting side. Happy to validate the mstsc path (gamepad + camera) and review the split PRs from a real-downstream perspective — point me at whichever piece you'd like eyes on first.

uchouT:
@clintcan: Thanks for your strong feedback. I can't run mstsc at hand so
your additional information about the mstsc behavior helps a lot.

On Mon Aug 17, 2026 at 11:08 PM CST, clintcan wrote:
> clintcan left a comment (Devolutions/IronRDP#1417)
>
> Thanks for the cc, @uchouT — and yes, macrdp is squarely the downstream this is for. We've carried a vendored server-direction URBDRC implementation for a while (our "divergence 16", ~1400 lines), and I'd genuinely like to delete it in favor of this. I diffed your stack (`ironrdp-usb` + `ironrdp-rdpeusb::usb` + this server PR) against ours — here's what I can offer from the field.
>
> **The hard-won mstsc-interop core is all here, and nicely factored.** The URB-level fixes we had to discover the painful way against real mstsc are all present in `ironrdp-rdpeusb::usb`: **typed** control URBs (`CLASS_INTERFACE` / `GET_DESCRIPTOR_FROM_INTERFACE` / …) with a generic fallback; `USBD_SHORT_TRANSFER_OK` on bulk/interrupt IN; `SelectConfiguration` emitting one interface-info **per interface number** plus the **full** configuration descriptor; and `SET_FEATURE`/`CLEAR_FEATURE` routed as `TRANSFER_IN`. Those are the exact things real Windows rejects a header-only or generic form of — having them in a protocol-independent layer is better than where we keep them. The `RequestId` correlation with PDU-layer completion-shape validation is also cleaner than ours.
>
> **Two things I'd flag before an mstsc device actually streams — both FreeRDP-lenient / mstsc-strict traps we already hit:**
>
> **1 — The per-device channel needs its own capability exchange, or mstsc closes it.** Your per-device processor's `start()` (`UrbdrcDeviceServer`) returns `CHANNEL_CREATED` directly, with **no preceding `RIM_EXCHANGE_CAPABILITY_REQUEST`** — your main `UrbdrcControlServer::start()` sends one, the per-device one doesn't. That's fine on FreeRDP (its readiness is global to the main channel), but we found live (mstsc, mid-2026) that **mstsc closes a per-device channel that gets `CHANNEL_CREATED` / `RIMCALL_RELEASE` without a preceding capability exchange** — it keeps the channel open-but-silent waiting for the caps request, so the device never emits `ADD_DEVICE`. Our per-device processor therefore runs the *full* caps → `CHANNEL_CREATED` handshake on each per-device channel, same as the main one. Since this PR is FreeRDP-verified, the concrete question is: **does an mstsc RemoteFX-USB device ever reach `ADD_DEVICE` on this branch?** If mstsc is hard for you to test, I'm happy to — it's the single highest-value thing to verify here. (Note mstsc gates RemoteFX USB behind a Group Policy + reboot and excludes mass storage, so the test device is a HID/gamepad or camera, not a flash drive.)

Good catch, I've discussed with @glamberson in #1077 about this topic.
This doesn't conflict with the 3 PRs, I'm happy to open another PR
handling this, but it takes some time because I'm currently preparing
for my GSoC final submission. And yes, please do validate the mstsc path
if you can; that's the most useful thing right now, since I only have
FreeRDP.

>
> **2 — The processors fail-hard on a decode/state error; we deliberately don't.** Both `UrbdrcControlServer` / `UrbdrcDeviceServer` `?`-propagate on a bad decode / `"invalid state"` / `"completion mismatch"`. On our dispatch seam that tears down the whole RDP session — so for an *opt-in* channel we swallow decode errors (`Ok(vec![])`) rather than let a malformed URBDRC PDU kill a working session. (It's also where cargo-fuzz found a few real panics in the PDU decoders, since patched.) Might be worth settling a policy — whether DRDYNVC dispatch isolates a per-processor error to its own channel, or the processor tolerates decode errors — since a redirected-USB channel shouldn't be able to drop the session.
>

I agree the problem isn't urbdrc-specific — `ironrdp-server`'s own
`AInputHandler`, displaycontrol... propagate the same way, and on the
server path a processor `Err` reaches `run_connection` and tears down
the whole connection.

But I'd rather not fix it by swallowing inside each processor: every DVC
would reimplement the same policy, and `Ok(vec![])` also discards the
error, so the dispatch layer can neither log it nor close just that
channel. cc @elmarco @CBenoit

> **Smaller notes:**
> - The two-phase `submit → PendingRequest.wait()` + `EndpointAddress` addressing + facade-owned pipe binding is a nice model (auto-`CANCEL_REQUEST` on drop; a `CompletionFut` dropping into a `FuturesUnordered` suits our bulk read-ahead engine well). It's a real but worthwhile port for us — we currently address by raw `pipe_handle` and single-await.
> - Re your TODO about `RdpUsbDeviceAnnounceInfo` still exposing the RDPEUSB `DeviceAnnounce`: we read `VID:PID:bcdDevice` off the announce/descriptor to dedup a device the client double-announces (one controller per physical device). A transport-independent announce that surfaces those cleanly would let that policy stay downstream where it belongs.
>
> Net: this is a strong replacement for our div-16 — if 1–2 land (or 1 is proven unnecessary on mstsc), we'd adopt it and drop ~1400 lines of vendored code, keeping only our presenting side. Happy to validate the mstsc path (gamepad + camera) and review the split PRs from a real-downstream perspective — point me at whichever piece you'd like eyes on first.

On where to look first: **`ironrdp-usb`**. Its API is the hardest to
change once released, and you're the strong reviewer who can tell
whether those types are genuinely protocol-independent rather than
quietly shaped by usbredir — which is the boundary I care most about in
this split.

Thanks for your help.

mamoreau-devolutions (approved):

PR #1418, opened 2026-07-08
feat(rdpeusb)!: tolerate unrecognized device-reported USB capability values

clintcan:
## Problem

`UsbDeviceCaps::decode` (MS-RDPEUSB `USB_DEVICE_CAPABILITIES`, 2.2.11) rejects any
value it doesn't explicitly name for the four **device-reported** capability
fields — `UsbBusInterfaceVersion`, `USBDI_Version`, `SupportedUsbVersion`,
`DeviceIsHighSpeed`. In particular `SupportedUsbVersion` only knew USB 1.0/1.1/2.0
(`0x100`/`0x110`/`0x200`).

A real **USB 3.2** device reports `SupportedUsbVersion = 0x320`, which isn't named,
so decoding its `ADD_DEVICE` returns `unsupported_value_err!` — and that error
propagates out of the DVC processor, tearing down the URBDRC channel. A modern
USB-3 device therefore can't be redirected at all.

## Fix

Make the four device-reported enums data-carrying: the named values plus an
`Other(u32)` fallback (with `from_u32`/`to_u32` helpers replacing the
`#[repr(u32)]` + `as u32` encode), so an unrecognized value is preserved verbatim
instead of failing the decode. `SupportedUsbVer` additionally names the USB
3.0/3.1/3.2 versions (`0x300`/`0x310`/`0x320`) for readable output.

The **framing constants** (`CbSize`, `HcdCapabilities`) stay strict — they
validate the PDU layout, not device data, so an invalid value there remains a
genuine decode error.

Encoding named values is byte-identical (the `to_u32` results equal the previous
`#[repr(u32)]` discriminants), so this is transparent on the wire for existing
devices.

## Verification

- Round-trip tested in `ironrdp-testsuite-core` (new `tests/rdpeusb/sink.rs`): a
  USB 3.2 capability set decodes and round-trips; an unrecognized value decodes as
  `Other`.
- Verified end-to-end against a real USB-3.2 flash drive redirected over a
  FreeRDP-with-`urbdrc` client: `ADD_DEVICE` now fully decodes and the device
  enumerates, where it previously errored out.

## Breaking change

`SupportedUsbVer` / `UsbdiVer` / `UsbBusIfaceVer` / `DeviceSpeed` are no longer
`#[repr(u32)]` fieldless enums — each gains an `Other(u32)` variant (and
`SupportedUsbVer` gains `Usb30`/`Usb31`/`Usb32`). Downstream code matching these
exhaustively will need a wildcard arm. No other crate in this workspace is
affected.

## Deliberately out of scope (happy to fold in if you'd prefer)

- The client-side builder `to_supported_usb_version` still maps `bcdUSB` to at
  most `Usb20`; it could emit the new USB-3 versions so an announced device
  reports its true `bcdUSB`. Kept separate since this PR targets the decode path.
- `NoAckIsochWriteJitterBufferSizeInMs` decode stays strict (`0` or `10..=512` per
  spec) — only the version/speed enums are loosened here.

### DIFF
--- a/crates/ironrdp-rdpeusb/src/pdu/sink.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/sink.rs
@@ -219,7 +219,9 @@ impl UsbDeviceCaps {
         usb_bus_iface_ver: UsbBusIfaceVer,
         device_speed: DeviceSpeed,
     ) -> Result<(), &'static str> {
-        if matches!(usb_bus_iface_ver, UsbBusIfaceVer::V0) && matches!(device_speed, DeviceSpeed::HighSpeed) {
+        if usb_bus_iface_ver.to_u32() == UsbBusIfaceVer::V0.to_u32()
+            && device_speed.to_u32() == DeviceSpeed::HIGH_SPEED.to_u32()
+        {
             Err("must be 0x00000000 when UsbBusInterfaceVersion is 0x00000000")
         } else {
             Ok(())
@@ -236,17 +238,13 @@ impl Encode for UsbDeviceCaps {
 
         dst.write_u32(Self::CB_SIZE);
 
-        #[expect(clippy::as_conversions)]
-        {
-            dst.write_u32(self.usb_bus_iface_ver as u32);
-            dst.write_u32(self.usbdi_ver as u32);
-            dst.write_u32(self.supported_usb_ver as u32);
-        }
+        dst.write_u32(self.usb_bus_iface_ver.to_u32());
+        dst.write_u32(self.usbdi_ver.to_u32());
+        dst.write_u32(self.supported_usb_ver.to_u32());
 
         dst.write_u32(Self::HCD_CAPS);
 
-        #[expect(clippy::as_conversions)]
-        dst.write_u32(self.device_speed as u32);
+        dst.write_u32(self.device_speed.to_u32());
 
         dst.write_u32(self.no_ack_isoch_write_jitter_buf_size.0);
 
@@ -270,32 +268,22 @@ impl Decode<'_> for UsbDeviceCaps {
         if cb_size != Self::CB_SIZE {
             return Err(unsupported_value_err!("CbSize", format!("{cb_size}")));
         }
-        let usb_bus_iface_ver = match src.read_u32() {
-            0x0 => UsbBusIfaceVer::V0,
-            0x1 => UsbBusIfaceVer::V1,
-            0x2 => UsbBusIfaceVer::V2,
-            value => return Err(unsupported_value_err!("UsbBusInterfaceVersion", format!("{value}"))),
-        };
-        let usbdi_ver = match src.read_u32() {
-            0x500 => UsbdiVer::V0x500,
-            0x600 => UsbdiVer::V0x600,
-            value => return Err(unsupported_value_err!("USBDI_Version", format!("{value}"))),
-        };
-        let supported_usb_ver = match src.read_u32() {
-            0x100 => SupportedUsbVer::Usb10,
-            0x110 => SupportedUsbVer::Usb11,
-            0x200 => SupportedUsbVer::Usb20,
-            value => return Err(unsupported_value_err!("SupportedUsbVersion", format!("{value}"))),
-        };
+        // These four fields are device-reported capability values that grow over
+        // time (USB 3.x, SuperSpeed, newer bus-interface revisions). Each is a
+        // newtype over the raw value, so an unrecognized value is preserved
+        // verbatim instead of failing the decode: rejecting it tears down the
+        // URBDRC channel for an otherwise-usable device (a real USB 3.2 device
+        // reports SupportedUsbVersion 0x320, which the named constants did not
+        // cover). The framing constants (`CbSize`, `HcdCapabilities`) stay strict
+        // — they validate the PDU layout, not device data.
+        let usb_bus_iface_ver = UsbBusIfaceVer::from_u32(src.read_u32());
+        let usbdi_ver = UsbdiVer::from_u32(src.read_u32());
+        let supported_usb_ver = SupportedUsbVer::from_u32(src.read_u32());
         let hcd_caps = src.read_u32();
         if hcd_caps != Self::HCD_CAPS {
             return Err(unsupported_value_err!("HcdCapabilities", format!("{hcd_caps}")));
         }
-        let device_speed = match src.read_u32() {
-            0x0 => DeviceSpeed::FullSpeed,
-            0x1 => DeviceSpeed::HighSpeed,
-            value => return Err(unsupported_value_err!("DeviceIsHighSpeed", format!("{value}"))),
-        };
+        let device_speed = DeviceSpeed::from_u32(src.read_u32());
         Self::check_device_speed(usb_bus_iface_ver, device_speed)
             .map_err(|reason| invalid_field_err!("USB_DEVICE_CAPABILITIES::DeviceIsHighSpeed", reason))?;
         let no_ack_isoch_write_jitter_buf_size = match src.read_u32() {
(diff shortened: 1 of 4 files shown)


uchouT:
I think there is a spec/codecs-layer question here.

[MS-RDPEUSB 2.2.11](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/98d4650e-b6d8-47e5-b71b-4d320ab542ee) defines `Supported_USB_Version` as: “The value MUST be one of the following:” USB 1.0 `0x100`, USB 1.1 `0x110`, USB 2.0 `0x200`. So, strictly speaking, `0x300`/`0x320` are outside the documented PDU syntax.

But FreeRDP currently writes the raw USB device descriptor `bcdUSB` into this field in `urbdrc_send_add_device()`, so a USB 3.x device can indeed produce `0x300`, `0x310`, `0x320`, etc. I’m not yet sure whether this is merely a FreeRDP bug or whether Microsoft’s server also accepts these values in practice. I’ll test this and follow up.

Relevant FreeRDP evidence:

- The libusb backend returns the raw device descriptor `bcdUSB` value:
  https://github.com/FreeRDP/FreeRDP/blob/f3b4347105114fe7453828736bea069999af319f/channels/urbdrc/client/libusb/libusb_udevice.c#L1002-L1003

- The `ADD_DEVICE` path reads `BCD_USB` into `bcdUSB`:
  https://github.com/FreeRDP/FreeRDP/blob/f3b4347105114fe7453828736bea069999af319f/channels/urbdrc/client/urbdrc_main.c#L416-L419

- It then writes `bcdUSB` directly into `USB_DEVICE_CAPABILITIES::Supported_USB_Version`:
  https://github.com/FreeRDP/FreeRDP/blob/f3b4347105114fe7453828736bea069999af319f/channels/urbdrc/client/urbdrc_main.c#L320-L324

If the test confirms that Microsoft accepts higher `bcdUSB` values here, then I agree we should relax this in the PDU codec for interoperability.

Fwiw, per `ironrdp-pdu/README.md`, for resilient parsing the preferred shape is to “consider providing a newtype struct wrapping the appropriate integer type”, rather than adding `Other(u32)` to an enum.

clintcan:
Thanks for digging into this — your FreeRDP trace matches exactly what we captured live, and it's the observation that motivated this PR: a real USB 3.2 flash drive (`bcdUSB = 0x0320`) redirected through FreeRDP's urbdrc/libusb client announces `Supported_USB_Version = 0x320` on the wire. Before this change, the decode error propagated out of the channel processing and tore down the whole session for that device — a metadata field killing the channel.

On the spec text: agreed, 2.2.11 says "MUST be one of" 0x100/0x110/0x200 — but the structure reads as authored in the USB-2 era and never updated (`DeviceIsHighSpeed` models only full/high speed, no SuperSpeed either). Meanwhile the deployed client population, per your trace, writes raw `bcdUSB` into the field. So whichever way your Microsoft-server test lands, a decoder that hard-errors here rejects traffic real clients demonstrably send — resilient parsing (with the named values kept for logs) seems right for interop either way. That said, your test would settle whether Windows' own server is lenient or FreeRDP is emitting values Microsoft never accepts, and I'd be glad to fold the result into the PR description. (For completeness: our capture is FreeRDP-as-client only — the devices we've exercised over mstsc's RemoteFX USB redirection were all USB 2.0-era, so we have no mstsc datapoint for 0x300+.)

On the shape: fair point, and the README's dual-representation critique (`Other(0x200)` vs `Usb20` representing the same wire value) applies to this PR as written. Happy to reshape the four fields into newtype structs with associated consts (the `http::StatusCode` pattern the README recommends) if the maintainers prefer — the diff is about the same size and round-trip encoding is preserved.

crates/ironrdp-testsuite-core/tests/rdpeusb/sink.rs
```diff
@@ -0,0 +1,37 @@
...
+/// An unrecognized device-reported capability value is preserved verbatim in
+/// `Other` rather than failing the decode.
+#[test]
+fn unknown_capability_value_decodes_as_other() {
+    let original = caps(SupportedUsbVer::Other(0x9999));
+    let encoded = encode_vec(&original).expect("encode should succeed");
+    let decoded: UsbDeviceCaps = decode(&encoded).expect("unknown version should decode as Other");
+    assert_eq!(decoded.supported_usb_ver, SupportedUsbVer::Other(0x9999));
+    assert_eq!(original, decoded);
+}
```
CBenoit: nitpick: These tests assert round-trip, not the wire contract.
Both tests build a UsbDeviceCaps in Rust, encode it with the crate, then decode; so they lean on the crate's own encode to produce the bytes.
The reported bug is a *decode* failure of bytes a real device sends; a compensating encode bug would pass these. A decode-from-raw-&[u8] test (28 bytes with 0x320 at the SupportedUsbVersion offset) would reproduce the actual failure and also pin the field order/offset independently of encode.

suggestion: Also it may be possible to use rstest for fixture-based testing and deduplicate code. It would be easy to also include some of the other variants although we don’t need to be exhaustive at this point.

CBenoit (commented):
Jumping in to express my preference clearly and give a small review.

First, I appreciate the spec-grounded and empirical investigation. Indeed, the spec does not always describe the reality; actually that’s common enough for RDP. FreeVNC typically does not deviate from the spec by mistake (although it may happen), I like to think of it as an accumulation of targeted workarounds based on what is observed in the real world and/or from reverse engineering to some extent. Always good to question, of course, but in this case, it makes sense: we’ve got USB V3 and variants for a while now.

Moving away from the variant `Other` fallback which can alias a named value, breaking round-trip identity and Eq, is a requirement on my side before we merge this PR (we are hunting down this kind of pattern across the database).

That’s pretty much the biggest issue here, otherwise everything is pretty neat!

Out of scope here, but it may also be worth evaluate things like the hard-fail in `UsbDeviceCaps::check_device_speed` (cc @uchouT).

It also sounds like the other out of scope actions mentioned in the PR body are worth considering and implementing in follow up PRs.

uchouT:
> Out of scope here, but it may also be worth evaluate things like the hard-fail in `UsbDeviceCaps::check_device_speed` (cc uchouT (@uchouT)).
> 
> It also sounds like the other out of scope actions mentioned in the PR body are worth considering and implementing in follow up PRs.

Exactly. I haven’t fully reviewed the TS_URB codec yet, so more similar issues may surface once RDPEUSB gets exercised by real integrations (macrdp, qemu-rdp, etc).

I’m currently working on the ironrdp-server RDPEUSB integration, and I expect qemu-rdp to eventually run on top of it. So I’m glad @clintcan is uncovering these issues before we run into them deeper in qemu-rdp development :)

clintcan:
Reshaped in `80c7c8b9`. Both points addressed:

- **Dropped the `Other` fallback.** `SupportedUsbVer`/`UsbdiVer`/`UsbBusIfaceVer`/`DeviceSpeed` are now newtype structs over `u32` with named associated consts (`SupportedUsbVer::USB_20`, `DeviceSpeed::HIGH_SPEED`, …) — the `http::StatusCode` shape the README recommends. A newtype can't alias a named value, so round-trip identity and `Eq` are exact by construction; an unnamed device value is just the wrapped `u32`. `from_u32`/`to_u32` construct/read the raw value.
- **Added a decode-from-raw-bytes test.** `usb3_capabilities_decode_from_raw_bytes` decodes a hand-authored 28-byte `USB_DEVICE_CAPABILITIES` with `0x320` at the `Supported_USB_Version` offset (12) — reproducing the actual decode failure independent of this crate's encode, and pinning the field offset. The round-trip test is now an `rstest` fixture over every named version plus an unnamed value.

On the spec datapoint: still glad to fold in your Microsoft-server result whenever you get to it, but as discussed it's decoupled from the merge now — the decoder tolerates whatever the device reports either way.

The out-of-scope items (the `check_device_speed` hard-fail, other TS_URB codec edges) I'm happy to take as follow-up PRs — and agreed they'll keep surfacing as macrdp / qemu-rdp exercise the server direction.

crates/ironrdp-rdpeusb/src/pdu/sink.rs
```diff
@@ -219,7 +219,9 @@ impl UsbDeviceCaps {
         usb_bus_iface_ver: UsbBusIfaceVer,
         device_speed: DeviceSpeed,
     ) -> Result<(), &'static str> {
-        if matches!(usb_bus_iface_ver, UsbBusIfaceVer::V0) && matches!(device_speed, DeviceSpeed::HighSpeed) {
+        if usb_bus_iface_ver.to_u32() == UsbBusIfaceVer::V0.to_u32()
+            && device_speed.to_u32() == DeviceSpeed::HIGH_SPEED.to_u32()
```
uchouT: Per [MS-RDPEUSB] 2.2.11:
> If `UsbBusInterfaceVersion` is 0x00000000, `DeviceIsHighSpeed` MUST be 0x00000000.

Since `DeviceSpeed` is changed to new type, which can contain any u32 value, we should modify the check logic:
```suggestion
        if usb_bus_iface_ver.to_u32() == UsbBusIfaceVer::V0.to_u32()
            && device_speed.to_u32() != DeviceSpeed::FULL_SPEED.to_u32()
```

uchouT (changes requested):

clintcan:
Thank you - good catch. Pushed in `04c76e7a`.

Changed the check to compare against `FULL_SPEED` instead of `HIGH_SPEED`:

```rust
if usb_bus_iface_ver.to_u32() == UsbBusIfaceVer::V0.to_u32()
    && device_speed.to_u32() != DeviceSpeed::FULL_SPEED.to_u32()
```

Since `DeviceSpeed` is now a lenient newtype that can hold any `u32`, the old `== HIGH_SPEED` only rejected the named value `0x1` and would have accepted e.g. `0x2` at bus-interface version 0. `!= FULL_SPEED` (i.e. non-zero) matches the MS-RDPEUSB 2.2.11 wording you quoted.

Also added a raw-decode test (`device_speed_must_be_zero_at_bus_iface_version_0`) covering `V0 + 0x0` (valid), `V0 + 0x1` / `V0 + 0x2` (rejected — the `0x2` case is the one the old check missed), and `V2 + 0x1` (constraint doesn't apply).

uchouT:
@CBenoit can we just merge this?

glamberson:
Aside, since you're on the capability-tolerance theme: the EGFX capset side already went tolerant in #1305. `RawCapabilitySet` keeps the raw version and bytes for versions this build doesn't recognize (`m == encode(decode(m))`), and the server parses known caps and skips unknown ones instead of failing the advertise decode. So if macrdp is still carrying a vendored EGFX capset-version workaround, you can probably drop it and lean on upstream now.

clintcan:
Thanks for the flag — already done, as it happens. We dropped the vendored `ironrdp-egfx` back when the `CapabilitySet::decode` tolerance fix went upstream in #1298, and our current pin sits past #1305 too, so `capabilities_advertise` already takes `Vec<RawCapabilitySet>` and does `filter_map(|raw| raw.parsed().ok().flatten())` — known caps parsed, unrecognized versions skipped. So we're fully on upstream tolerance now, no vendored capset workaround left. Appreciate you looking out for it.

CBenoit (approved):
LGTM!

PR #1419, opened 2026-07-08
fix(session)!: reduce dependency on ironrdp-connector

zmb3:
Decode-only consumers previously had to compile the entire stack pulled in by ironrdp-connector, including all of the sspi/NLA dependencies.

Move some of the framing code from ironrdp-connector to ironrdp-pdu. This doesn't eliminate the transitive dependency on sspi-rs, that will require some additional reshaping.


crates/ironrdp-session/src/lib.rs
```diff
@@ -6,9 +6,9 @@ mod macros;
 
 pub mod fast_path;
 pub mod image;
-pub mod legacy;
 pub mod pointer;
 pub mod rfx; // FIXME: maybe this module should not be in this crate
+
```
CBenoit:
```suggestion
```

CBenoit (approved):
LGTM, thank you!

PR #1420, opened 2026-07-08
feat(rdpeusb)!: carry the full configuration descriptor in UsbConfigDesc

clintcan:
## What

`TS_URB_SELECT_CONFIGURATION` sets `ConfigurationDescriptorIsValid` and then encodes a `UsbConfigDesc`, but only the 9-byte `USB_CONFIGURATION_DESCRIPTOR` **header** was modelled — `wTotalLength` would claim e.g. 759 bytes with only 9 following.

Per [MS-RDPEUSB] 2.2.9.2 the field must carry the **full** configuration descriptor (all interface / endpoint / class-specific bytes) when the flag is set. Real Windows (mstsc) walks `wTotalLength` and rejects a header-only descriptor with `0x80070057` (E_INVALIDARG), so `URB_SELECT_CONFIGURATION` fails against a real Windows peer. FreeRDP uses the interface array and ignores the descriptor bytes, which masks this.

## Change

Adds a `trailing: Vec<u8>` field holding bytes `9..wTotalLength`:
- `encode` writes it after the 9-byte header.
- `decode` reads `wTotalLength - 9` bytes, clamped to what is present — so a header-only descriptor still decodes (with `trailing` empty), and a spec-conformant full descriptor's trailing bytes are now **consumed** from the cursor instead of left dangling.

`BREAKING CHANGE`: `UsbConfigDesc` gains the public field `trailing: Vec<u8>` (all fields are public, so struct-literal constructors need it).

## Tests

Adds `tests/rdpeusb/ts_urb.rs` (wired into the rdpeusb testsuite): `full_config_descriptor_round_trips`, `header_only_descriptor_still_decodes`, and `select_configuration_carries_full_descriptor`.

## Context

Found while building a server-direction (client-redirected-USB) consumer of this crate, and verified live: mstsc `SelectConfiguration` succeeds (pipe handles returned) for a redirected camera + audio/HID device; FreeRDP mass storage is unaffected. Independent of #1418 (a different file/type in the same crate) — this branch is rebased on current `master`, not stacked on it.

Gates run locally: `cargo fmt --all -- --check`, `cargo clippy -p ironrdp-rdpeusb --all-targets --locked -D warnings`, and the rdpeusb testsuite (all green).

### DIFF
--- a/crates/ironrdp-rdpeusb/src/pdu/usb_dev/ts_urb/utils.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/usb_dev/ts_urb/utils.rs
@@ -617,6 +617,15 @@ pub struct UsbConfigDesc {
     pub configuration: u8,
     pub attributes: u8,
     pub max_power: u8,
+    /// The remainder of the configuration descriptor **after** the 9-byte header —
+    /// all interface / endpoint / class-specific descriptors, i.e. bytes
+    /// `9..total_length`.
+    ///
+    /// [MS-RDPEUSB] 2.2.9.2 requires the *full* configuration descriptor when
+    /// `ConfigurationDescriptorIsValid` is set; real Windows (mstsc) walks
+    /// `wTotalLength` bytes and rejects a header-only descriptor with `0x80070057`
+    /// (`E_INVALIDARG`). Empty means header-only.
+    pub trailing: Vec<u8>,
 }
 
 impl UsbConfigDesc {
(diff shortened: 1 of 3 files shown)


clintcan:
Addressed the review point in `2cde264f`: `UsbConfigDesc::encode` now validates the header against the payload before writing — `bLength` must be the 9-byte header length and `wTotalLength` must equal the header plus the trailing bytes (which also rejects a `trailing` too large for `wTotalLength` to address). So a caller can no longer emit a descriptor whose `wTotalLength` disagrees with the bytes that follow — the exact shape a real Windows peer rejects with `0x80070057`. Added `inconsistent_header_fails_to_encode` covering both a mismatched `wTotalLength` and a wrong `bLength`.

mamoreau-devolutions:
@clintcan this PR just needs a rebase

CBenoit (approved):
LGTM and approving; just needs a rebase before we can merge 🙂

mamoreau-devolutions (approved):

PR #1436, opened 2026-07-09
Correctly crop bitmap updates when TS_BITMAP_DATA.width is padded

Eugeny:
Downstream issue: https://github.com/warp-tech/warpgate/issues/2173

TL;DR: IronRDP decodes the bitmap incorrectly when `update.width` and `update_rectangle.width()` differ

AI disclosure: AI was used to investigate and fix this.

----
`process_bitmap_update` decodes each bitmap at `update.width`/`update.height`, but `DecodedImage::apply_*_bitmap` re-chunk the buffer at `update_rectangle.width()`. When a server pads the header width beyond the rectangle (spec-legal; xrdp always pads to a multiple of 4 pixels, `e = (4 - width) & 3`), each row is offset by the padding → diagonal shear. Flat fills look fine; text/images/the xrdp login dialog hatch. Affects 0.9.0 and `master`, all depths, both compressed and raw paths (`apply_rgb16/15_bitmap`, `apply_bgr24_bitmap`, `apply_rgb32_bitmap`, `apply_rgb8_with_palette`, `apply_rgb24`).

**Fix** — treat the header width as source stride; crop each row to `rectangle.width()` and the row count to `rectangle.height()` before applying (as FreeRDP/mstsc do). Patch leaves `apply_*` untouched — feeding them a tightly-packed rectangle-sized buffer makes their existing `chunks_exact(rectangle_width * …)` correct.

### DIFF
--- a/crates/ironrdp-session/src/fast_path.rs
+++ b/crates/ironrdp-session/src/fast_path.rs
@@ -21,6 +21,32 @@ use crate::palette::Palette;
 use crate::pointer::PointerCache;
 use crate::{SessionError, SessionErrorExt as _, SessionResult, custom_err, reason_err, rfx};
 
+/// Re-pack bitmap pixel data so each row holds exactly `dst_row_bytes`
+/// and there are `rows` of them, dropping any right/bottom padding the server added.
+///
+/// RDP servers may pad `TS_BITMAP_DATA` beyond the destination rectangle: the width up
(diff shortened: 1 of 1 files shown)


mamoreau-devolutions:
superseded by https://github.com/Devolutions/IronRDP/pull/1486

PR #1448, opened 2026-07-13
test(extra): cover proxy pipe cross-process access

mamoreau-devolutions:
## Summary
- verify a `DvcNamedPipeProxy` Windows pipe is visible to a separately spawned process
- require the child process to forward pipe data into the parent DVC callback

## Tests
- `cargo test -p ironrdp-testsuite-extra --test integration_tests_extra dvc_pipe_proxy`
- `cargo clippy -p ironrdp-testsuite-extra --test integration_tests_extra --no-deps -- -D warnings`


crates/ironrdp-dvc-pipe-proxy/src/proxy.rs
```diff
@@ -131,6 +131,11 @@ impl Drop for DvcNamedPipeProxy {
     fn drop(&mut self) {
         if let Some(ctx) = &self.worker {
             // Signal the worker thread to abort.
+            info!(
+                channel_name = %self.channel_name,
+                pipe_name = %self.named_pipe_name,
+                "Dropping DVC named pipe proxy"
+            );
```
CBenoit: issue: 
- `info!` is reserved for rare lifecycle milestones that a consumer wants at default verbosity (e.g. connection/session lifecycle transitions), and it should be uncommon in a library, and never used for anything that repeats during normal operation.
- A proxy being dropped is a routine teardown event that happens once per proxy instance / per channel.

suggestion: Downgrade this to `debug!` (a significant one-off teardown event) or `trace!` if you consider it fine-grained detail.

crates/ironrdp-dvc-pipe-proxy/src/proxy.rs
```diff
@@ -131,6 +131,11 @@ impl Drop for DvcNamedPipeProxy {
     fn drop(&mut self) {
         if let Some(ctx) = &self.worker {
             // Signal the worker thread to abort.
```
CBenoit: nitpick: stale comment placement, "// Signal the worker thread to abort." now sits above the info! block
rather than immediately above the notify_one() it describes. Move it back down to the notify call, or drop it.

crates/ironrdp-testsuite-extra/tests/dvc_pipe_proxy.rs
```diff
@@ -37,6 +32,71 @@ async fn connects_and_forwards_windows_pipe_data() {
...
+    let pipe_path = format!(r"\\.\pipe\{name}");
+    let mut client = (0..200)
+        .find_map(|_| match ClientOptions::new().open(&pipe_path) {
+            Ok(client) => Some(client),
+            Err(_) => {
+                std::thread::sleep(Duration::from_millis(10));
+                None
+            }
+        })
+        .expect("DVC pipe proxy must be visible to a separate process within two seconds");
```
CBenoit: suggestion: Duplicate code that is useful to refactor out. (Duplicated a prior test body above.)

```rust
#[cfg(windows)]
fn open_client_pipe(name: &str, unavailable_msg: &str) -> NamedPipeClient {
    let pipe_path = format!(r"\\.\pipe\{name}");
    (0..200)
        .find_map(|_| ClientOptions::new().open(&pipe_path).ok().or_else(|| {
            std::thread::sleep(Duration::from_millis(10));
            None
        }))
        .expect(unavailable_msg)
}
```

CBenoit (commented):
Thank you!

Some review items from Claude that I think are worth addressing:

> **The hardcoded child-test path is stale-prone with a misleading failure mode**
> 
> dvc_pipe_proxy.rs:48 — "dvc_pipe_proxy::child_process_connects_and_forwards_windows_pipe_data"
> must track both the module name in main.rs (`mod dvc_pipe_proxy;`) and the fn name. If
> either is renamed, `--exact` matches zero tests; libtest then exits 0, so
> `assert!(status.success())` (line 54) passes, and the test instead fails 1s later at
> assert_forwarded_data with "must forward pipe data to its callback" — pointing at a
> forwarding bug that doesn't exist. This is the first use of the current_exe re-exec
> pattern in the repo, so there's no established guard. Consider asserting the child
> actually ran a test (e.g. parse `--list` or check the summary), or at minimum a comment
> binding the string to the fn so a rename is caught in review.
> 
> **child_process_… is a real test that silently no-ops in normal runs**
> 
> dvc_pipe_proxy.rs:62-65 — as a #[tokio::test] it's discovered in every normal run,
> where the env var is absent so it returns Ok early and shows up green. Its name reads
> like a standalone assertion ("connects_and_forwards…") but in the normal suite it
> asserts nothing. A name that signals its helper role (e.g. child_process_helper_…)
> would make the test output honest without changing the --exact filter meaningfully.

crates/ironrdp-testsuite-extra/tests/dvc_pipe_proxy.rs
mamoreau-devolutions: Addressed in 7b095b09: extracted `open_client_pipe`, renamed the no-op test to identify it as a child-process helper, and assert the child harness ran exactly one test so a stale `--exact` selector fails directly.

crates/ironrdp-dvc-pipe-proxy/src/proxy.rs
mamoreau-devolutions: Moved the comment directly above `notify_one()` in 7b095b09.

crates/ironrdp-dvc-pipe-proxy/src/proxy.rs
mamoreau-devolutions: Changed to `debug!` in 7b095b09, keeping the event available for teardown diagnosis without default-verbosity noise.

vnikonov-devolutions (approved):

vnikonov-devolutions (approved):

PR #1452, opened 2026-07-15
fix(session): handle bitmap updates wider than the destination rectangle

lixmal:
## Summary

- Trim each bitmap row to the destination rectangle width before applying it, in both the compressed (RDP 6.0 / RLE) and uncompressed paths in `process_bitmap_update()`

Some servers, like xrdp, round the bitmap width in a Bitmap Update up to a multiple of 4, so the width can be bigger than the destination rectangle. The apply functions use the rectangle width as the row stride, so the extra padding columns push every following row over and the image comes out sheared.

## Relates to

- #314 (Cannot connect to XRDP): this is the "bitmap row padding" gap called out in #1132.

## Validation

- `cargo xtask check fmt -v`
- Tested against XRDP 0.10.1 (Debian 13) via the web client: the login screen renders sheared on master and correct with this change.

### DIFF
--- a/crates/ironrdp-session/src/fast_path.rs
+++ b/crates/ironrdp-session/src/fast_path.rs
@@ -192,6 +192,11 @@ impl Processor {
             trace!("{update:?}");
             buf.clear();
 
+            // The apply functions use the rectangle width as the source row
+            // stride, so a bitmap wider than the rectangle must be trimmed.
+            let bmp_width = usize::from(update.width);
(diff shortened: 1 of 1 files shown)


lixmal:
Closing as a duplicate of #1436, which fixes the same padded-bitmap shear and also crops the row count, not just the width. #1252 covers the same area plus the out-of-bounds writes. Thanks!

PR #1454, opened 2026-07-18
feat(egfx): add Planar frame sender

digyear:
## Summary

Add a server-side `GraphicsPipelineServer::send_planar_frame` API for transmitting pre-encoded RDP6 Planar bitmap streams through EGFX `WireToSurface1`.

The method follows the existing AVC and uncompressed frame APIs:

- waits for EGFX capability negotiation to complete
- respects frame-acknowledgement backpressure and records it in QoE metrics
- validates the target surface
- clips the destination size to the surface dimensions
- queues the required `StartFrame` / `WireToSurface1` / `EndFrame` sequence
- uses `Codec1Type::Planar` (`0x0a`)

## Why

Some EGFX clients, including Microsoft Remote Desktop on Android, advertise `AVC_DISABLED`. They can still consume compressed desktop graphics through the RDP6 Planar codec, but IronRDP currently exposes no server-side API for sending a Planar payload.

Without this path, downstream servers must either fall back to uncompressed EGFX frames, leave the managed EGFX pipeline, or construct protocol PDUs outside `GraphicsPipelineServer`. Exposing Planar alongside the existing AVC, uncompressed, and progressive APIs keeps frame sequencing, acknowledgement-based flow control, surface validation, and QoE accounting inside the library.

This is also a small step toward the multi-codec server API discussed in #1158.

## Validation

- Added an integration test that negotiates an AVC-disabled EGFX session, sends a Planar frame, decompresses the ZGFX output, decodes the PDU, and verifies:
  - `WireToSurface1` is used
  - codec ID is Planar
  - the payload is preserved
  - the destination rectangle is correct
- `cargo test -p ironrdp-testsuite-core --test integration_tests_core planar_frame_uses_planar_codec`
- Tested downstream with Microsoft Remote Desktop for Android through `lamco-rdp-server`

## Notes

This PR only exposes transmission of an already encoded Planar stream. It intentionally does not add a Planar encoder or downstream-specific codec-selection policy.

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -1323,6 +1323,53 @@ impl GraphicsPipelineServer {
         Some(frame_id)
     }
 
+    /// Queue a pre-encoded RDP6 Planar bitmap stream for transmission via EGFX.
+    ///
+    /// Planar is used by clients such as Microsoft Remote Desktop for Android
+    /// when they negotiate EGFX with AVC disabled.
+    ///
+    /// Returns `Some(frame_id)` if queued, `None` if the server is not ready,
+    /// the surface does not exist, or backpressure is active.
+    pub fn send_planar_frame(
+        &mut self,
+        surface_id: u16,
+        planar_data: &[u8],
+        dest_width: u16,
+        dest_height: u16,
+        timestamp_ms: u32,
+    ) -> Option<u32> {
+        if !self.is_ready() {
+            return None;
+        }
+        if self.should_backpressure() {
+            self.qoe.record_backpressure();
+            return None;
+        }
+
+        let surface = self.surfaces.get(surface_id)?;
+        let timestamp = Self::make_timestamp(timestamp_ms);
+        let frame_id = self.frames.begin_frame(timestamp);
+        let dest_rect = ExclusiveRectangle {
+            left: 0,
+            top: 0,
+            right: dest_width.min(surface.width),
+            bottom: dest_height.min(surface.height),
+        };
+
+        self.output_queue
+            .push_back(GfxPdu::StartFrame(StartFramePdu { timestamp, frame_id }));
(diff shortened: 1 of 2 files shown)


glamberson:
This is a valuable but unimplemented feature as indicated for secondary server (android) functionality. 

@CBenoit I didn't fully review the code as I'm walking out the door right now, but I hope you'll prioritize this as a no brainer.

crates/ironrdp-egfx/src/server.rs
```diff
@@ -1323,6 +1323,53 @@ impl GraphicsPipelineServer {
...
+            return None;
+        }
+
+        let surface = self.surfaces.get(surface_id)?;
+        let timestamp = Self::make_timestamp(timestamp_ms);
+        let frame_id = self.frames.begin_frame(timestamp);
+        let dest_rect = ExclusiveRectangle {
+            left: 0,
+            top: 0,
+            right: dest_width.min(surface.width),
```
mamoreau-devolutions, line 1355: `destRect` specifies both the target point and the dimensions of the decoded bitmap (MS-RDPEGFX §2.2.2.1). Clamping this rectangle to the surface while forwarding the same pre-encoded RDP6 Planar stream makes the declared dimensions disagree with the encoded pixels whenever a caller supplies dimensions larger than the surface. That can make the client reject or misdecode the frame and withhold its acknowledgment, eventually stalling this server's flow control.

Reject dimensions that exceed `surface.width` or `surface.height` before `begin_frame` (return `None`, like the other preconditions) rather than clipping them. Please add a test that verifies this path queues no frame PDUs and does not increase frames in flight.

mamoreau-devolutions (changes requested):
The Planar sender’s frame sequencing and codec selection are correct, but the destination rectangle must not be clipped independently of the encoded bitmap data.

mamoreau-devolutions:
fixed by https://github.com/Devolutions/IronRDP/pull/1498

PR #1456, opened 2026-07-18
feat(rdpeusb)!: type URB completion payloads

uchouT:
## Why

URB result types are determined by their originating requests. Keeping results as raw bytes leaked wire details into the backend and allowed invalid request/result combinations.

## Impact

Pending requests now record the expected result type, and raw wire payloads are converted into typed results at the server boundary. Header-only results use `None`, while isochronous completions use `TsUrbIsochTransferResult` directly.

This is a breaking API change for RDPEUSB completion types.

cc @elmarco

### DIFF
--- a/crates/ironrdp-rdpeusb/src/pdu/completion/ts_urb_result.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/completion/ts_urb_result.rs
@@ -17,7 +17,9 @@ use crate::pdu::{
     header::SharedMsgHeader,
     usb_dev::{
         InternalIoControl, IoControl, RegisterRequestCallback, TransferInRequest, TransferOutRequest,
-        ts_urb::{TsUrb, TsUrbGetCurrFrameNum, TsUrbIsochTransfer, TsUrbSelectConfig, TsUrbSelectInterface},
+        ts_urb::{
+            TsUrbGetCurrFrameNum, TsUrbIn, TsUrbIsochTransfer, TsUrbOut, TsUrbSelectConfig, TsUrbSelectInterface,
+        },
     },
 };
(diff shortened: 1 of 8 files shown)


crates/ironrdp-rdpeusb/src/pdu/completion/ts_urb_result.rs
```diff
@@ -45,25 +135,64 @@ impl Decode<'_> for TsUrbResult {
...
 
     fn name(&self) -> &'static str {
         "TS_URB_RESULT"
     }
 
     fn size(&self) -> usize {
-        size_of::<u16>(/* TS_URB_RESULT_HEADER::Size */) + self.header.size() + self.payload.size()
+        size_of::<u16>(/* TS_URB_RESULT_HEADER::Size */)
+            + self.header.size()
+            + if let Some(p) = &self.payload { p.size() } else { 0 }
```
elmarco: self.payload.as_ref().map_or(0, Encode::size)

elmarco (approved):
lgtm

CBenoit:
Review identified two medium-risk questions: malformed or mismatched completions remove pending correlation state before typed-result validation, preventing a later valid completion, and the new generic public completion/PDU types are source-incompatible for downstream callers. A maintainer should confirm the intended recovery behavior and whether this breaking API change is acceptable for the release.

CBenoit:
Human here: breaking change is acceptable. 

> malformed or mismatched completions remove pending correlation state before typed-result validation, preventing a later valid completion

@uchouT Please verify, but I’m not blocking the PR on that

CBenoit (approved):

PR #1458, opened 2026-07-19
fix(ironrdp-pdu): tolerate unknown security header flags in BasicSecurityHeader

### LINKED ISSUE #1457: LicenseHeader decode rejects valid security header flags from Windows Server 2019 with RDS licensing (UpgradeLicense)
IronRDP fails to connect to Windows Server 2019 hosts that use Remote Desktop Licensing (RDS CAL) with an RD Connection Broker. The connection succeeds through NLA/CredSSP but fails immediately afterwards when decoding the Server License PDU - `LicenseHeader::decode() rejects the securityHeaderFlags as invalid.`

FreeRDP connects to the same server without issues.

Related: #269 (same server_license module, different failure mode - #269 is about LICENSE_ERROR_MESSAGE control flow; this is about UpgradeLicense rejecting valid security header flag combinations).

**Environment**
IronRDP version: 0.17 (via ironrdp crate with connector, session, graphics, pdu, core, svc features)
Server: Windows Server 2019, Active Directory authentication, RDS licensing configured, RD Connection Broker deployed
Authentication: NTLM via CredSSP (NLA) - succeeds
Failure point: License exchange, LicenseExchangeState::UpgradeLicense

**Error**

IronRDP connect_finalize failed:
```
  Error {
    context: "decode during SERVER_NEW_LICENSE/LicenseExchangeState::UpgradeLicense",
    kind: Decode(Error {
      context: "<ironrdp_pdu::rdp::server_license::LicenseHeader as ironrdp_core::decode::Decode<'_>>::decode",
      kind: InvalidField {
        field: "securityHeaderFlags",
        reason: "invalid security header flags"
      },
      source: None
    }),
    source: None
  }
```
Sequence of events (from debug logs)
X.224 negotiation - success
TLS upgrade - success
NLA (CredSSP/NTLM) - success: CredSSp finished NLA stage, tscredentials has been written
License exchange - FAILS at SERVER_NEW_LICENSE/UpgradeLicense with "invalid security header flags"

**Working comparison**
A nearly identical server (same OS image, same AD domain, same user/password) connects successfully via IronRDP. The difference is that the failing server has RDS licensing + Connection Broker configured, which causes it to send a SERVER_UPGRADE_LICENSE PDU with different security header flags.

FreeRDP handles both servers without issues, suggesting the flags are valid per the spec but not covered by the current LicenseHeader parser.

**Likely cause**
Per MS-RDPBCGR §2.2.8.1.1.2.1 and MS-RDPELE §2.2.2.7, the license PDU security header can carry flags such as:

SEC_LICENSE_PKT (0x0080) - always present in license PDUs
SEC_LICENSE_ENCRYPT_CS (0x0200) - server can accept encrypted licensing packets
SEC_ENCRYPT (0x0008) - data following the header is encrypted
It appears LicenseHeader::decode() validates the flags with a strict allowlist and rejects certain valid combinations that Windows Server 2019 with RDS licensing sends during UpgradeLicense. FreeRDP's equivalent parser accepts these flags and continues the handshake successfully.

**Impact**
Any Windows Server with RDS licensing that sends SERVER_UPGRADE_LICENSE (instead of a simple STATUS_VALID_CLIENT license-OK response) will fail to connect via IronRDP. This affects enterprise environments with RD Connection Broker / RDS farms - a common deployment pattern.

**Workaround**
In my application we fall back to an external FreeRDP process when IronRDP fails at this stage. This works but means those servers never benefit from the embedded IronRDP client.

**References**
MS-RDPBCGR §2.2.8.1.1.2.1 - Security Header structure
MS-RDPELE §2.2.2.7 - Server Upgrade License (SERVER_UPGRADE_LICENSE)
MS-RDPBCGR §3.2.5.7 - Processing License Error PDU - Valid Client

**Original issue**
https://github.com/totoshko88/RustConn/issues/218

totoshko88:
## Summary

Use `from_bits_truncate()` instead of `from_bits()` when decoding `BasicSecurityHeader` flags. Some servers (e.g., Windows Server 2019 with RDS licensing / RD Connection Broker) send security header flag combinations that include bits not defined in the current bitflags enum. The strict `from_bits()` rejected these as invalid, causing connection failure during the `UpgradeLicense` license exchange phase.

This matches FreeRDP behavior which masks for known flags without rejecting the PDU when unrecognized bits are present.

## What was tested

- Existing unit tests pass (`cargo xtask check tests -v`)
- Lints pass (`cargo xtask check lints -v`)

Closes #1457

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -68,8 +68,10 @@ impl<'de> Decode<'de> for BasicSecurityHeader {
     fn decode(src: &mut ReadCursor<'de>) -> DecodeResult<Self> {
         ensure_fixed_part_size!(in: src);
 
-        let flags = BasicSecurityHeaderFlags::from_bits(src.read_u16())
-            .ok_or_else(|| invalid_field_err!("securityHeader", "invalid basic security header"))?;
+        // Use from_bits_truncate to tolerate unknown flag bits that some servers
+        // (e.g., Windows Server 2019 with RDS licensing) may set.
+        // This matches FreeRDP behavior which masks for known flags without rejecting the PDU.
+        let flags = BasicSecurityHeaderFlags::from_bits_truncate(src.read_u16());
         let _flags_hi = src.read_u16(); // unused
 
         Ok(Self { flags })


glamberson:
Thanks for chasing this down, @totoshko88, and for the detailed repro writeup. I traced the decode path against current master and I think the fix is landing in the wrong spot, so let me lay out what I found before it goes further.

The PR's premise is that the server sets flag bits not defined in `BasicSecurityHeaderFlags`. That turns out not to be possible: the bitflags cover all 16 bits of the `u16` (`EXCHANGE_PKT` 0x0001 through `FLAGSHI_VALID` 0x8000, defined since 2019), so `from_bits()` can never return `None` for any `u16`. With every bit defined, `from_bits_truncate()` is identical to `from_bits()`, so this change is a no-op for decode behavior. (Also worth flagging: #1144 deliberately moved the crate off `from_bits_truncate` to `from_bits_retain`, so reintroducing truncate would cut against that.)

So `BasicSecurityHeader::decode` isn't what's rejecting your server. The error you posted:

```
field: "securityHeaderFlags", reason: "invalid security header flags"
```

comes from `LicenseHeader::decode`, one line below the header decode, at the `LICENSE_PKT` presence check:

```rust
let security_header = BasicSecurityHeader::decode(src)?;   // succeeds
if !security_header.flags.contains(BasicSecurityHeaderFlags::LICENSE_PKT) {
    return Err(invalid_field_err!("securityHeaderFlags", "invalid security header flags"));
}
```

And `LicensePdu::decode` runs every license PDU through `LicenseHeader::decode` first, so this gate fires before any message-type routing. The real signal is that your Win2019 + RDS/Broker server's license-stage PDU doesn't have `LICENSE_PKT` (0x0080) set in the bytes we read as the header, which fits the "different flags with RDS licensing" you observed.

Could you grab a hex dump of the failing PDU, the first ~8 bytes at the point `LicenseHeader::decode` runs (the basic security header is `flags: u16` + `flagsHi: u16`)? That settles in one shot whether:

- `LICENSE_PKT` is genuinely absent, in which case the fix is to relax that hard requirement in `LicenseHeader::decode` to match FreeRDP's leniency and let the existing `LicensingErrorMessage` / `StatusValidClient` path handle it, or
- the header is offset and we're reading the wrong two bytes, a framing fix similar in shape to what you found in #1442.

Happy to help land whichever it turns out to be. Enterprise RDS/Broker interop is exactly the kind of gap worth closing.

mamoreau-devolutions (approved):

PR #1459, opened 2026-07-20
feat(connector): implement multitransport bootstrapping handshake

glamberson:
> Supersedes #1098, rebased onto current master. Same change; opening fresh from the organization fork.

## Summary

Makes the MultitransportBootstrapping state functional instead of a no-op
pass-through. After licensing the server may send 0, 1, or 2 Initiate
Multitransport Request PDUs before capabilities exchange. The connector
reads those, pauses in a new MultitransportPending state, and surfaces
the requests for the application to establish UDP transport (RDPEUDP2 +
TLS + RDPEMT) or decline.

## API

Mirrors the existing `should_perform_X()` pause-point pattern used by TLS
upgrade and CredSSP, but uses `complete_X()` / `skip_X()` rather than
`mark_X_as_done()` because completion carries result data:

- `should_perform_multitransport()`: true when requests have been collected
- `multitransport_requests()`: access the server's request PDUs
- `complete_multitransport(results, output)`: send response PDUs, advance
- `skip_multitransport(output)`: decline, advance

`complete_multitransport` accepts `&[MultitransportResult]` (a `Success` /
`Failure(hresult)` enum) rather than caller-built response PDUs. The
connector builds the response PDUs internally using the stored request
IDs and validates that the result count matches the request count.

## Approach

**Detection.** The state machine try-decodes the incoming PDU as
`MultitransportRequestPdu`; a clean decode means multitransport, a decode
error means Demand Active. For that to be sound the request decoder must
reject a Demand Active, so this PR also tightens `MultitransportRequestPdu`
/ `MultitransportResponsePdu` decoding to require the `BasicSecurityHeader`
flags to equal `SEC_TRANSPORT_REQ` / `SEC_TRANSPORT_RSP` exactly rather than
merely contain the bit (`fix(pdu)` commit). The previous `contains` check
false-positived: a Demand Active's leading `ShareControlHeader` `totalLength`
aliases onto the security header flags field, and a length such as `0x016F`
is a valid flag combination that carries the `TRANSPORT_REQ` bit, so the
Demand Active decoded as a bogus request. The exact check is what
MS-RDPBCGR 2.2.15.1 / 2.2.15.2 specify (each PDU carries that single flag).
No flag-peeking on `BasicSecurityHeader` in the connector.

**Demand Active buffering.** The connector buffers the Demand Active PDU
internally in `MultitransportPending` and replays it through
`ConnectionActivationSequence::step()` when `complete_multitransport()`
or `skip_multitransport()` is called.

**Driver wiring.** Both `ironrdp-async` and `ironrdp-blocking`
`connect_finalize` loops now auto-skip multitransport: when the connector
reports `should_perform_multitransport()`, the driver calls
`skip_multitransport()` on the caller's behalf so the connection succeeds
TCP-only. Applications that want to participate in multitransport setup
must drive the connector directly rather than calling `connect_finalize`.

**Request count.** Per MS-RDPBCGR 2.2.15.1 the server may send 0, 1, or 2
requests (one per transport protocol). The connector rejects subsequent
requests with a `ConnectorError`.

**Wire behavior.** Per the doc-comment on `MultitransportPending`, TCP and
UDP negotiation happen in parallel. The UDP transport is established
alongside the ongoing TCP handshake; its completion is a signal to the
dynamic-channel layer that subsequent channels may migrate to UDP. The
connector's API yield point here is a Rust affordance, not a
spec-mandated TCP pause. Thanks to @hardening for the correction.

## Tests

Connector state-machine tests in `ironrdp-testsuite-core` drive the public
`MultitransportPending` API with the shared `SERVER_DEMAND_ACTIVE` fixture:

- `complete_multitransport()` writes the response PDUs and replays the
  buffered Demand Active, advancing the connector past bootstrapping;
- `skip_multitransport()` replays the Demand Active without responses;
- a `Failure` result is carried through;
- a result/request count mismatch is rejected;
- `complete` / `skip` outside `MultitransportPending` error;
- a Demand Active's user data does not decode as a `MultitransportRequestPdu`
  (regression test for the detection fix above).

Fuzzing: `MultitransportRequestPdu` and `MultitransportResponsePdu` are added
to the `pdu_decode` and `pdu_round_trip` oracles, so the post-licensing
try-decode surface is fuzzed against arbitrary input (auto-discovered into the
fuzz CI matrix). Both targets smoke-run clean.

Full end-to-end coverage against a live server still awaits server-side
`MultitransportRequestPdu` emission, which `ironrdp::server::RdpServer` does
not implement yet.

## Verification

## References

Builds on: #1091
Related: #140

### DIFF
--- a/crates/ironrdp-async/src/connector.rs
+++ b/crates/ironrdp-async/src/connector.rs
@@ -78,6 +78,24 @@ where
     }
 
     let result = loop {
+        if connector.should_perform_multitransport() {
+            // Auto-skip multitransport bootstrapping: this driver does not own
+            // UDP transport setup, so it declines on the application's behalf
+            // and the connection continues TCP-only. Applications that want to
+            // participate in multitransport must drive the connector directly
+            // using `ClientConnector::complete_multitransport()` instead of
+            // calling `connect_finalize`.
+            buf.clear();
+            let written = connector.skip_multitransport(&mut buf)?;
+            if written.size().is_some() {
+                framed
+                    .write_all(buf.filled())
+                    .await
+                    .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
+            }
+            continue;
+        }
+
         single_sequence_step(framed, &mut connector, &mut buf).await?;
 
         if let ClientConnectorState::Connected { result } = connector.state {
--- a/crates/ironrdp-blocking/src/connector.rs
+++ b/crates/ironrdp-blocking/src/connector.rs
@@ -82,6 +82,23 @@ where
     debug!("Remaining of connection sequence");
 
     let result = loop {
+        if connector.should_perform_multitransport() {
+            // Auto-skip multitransport bootstrapping: this driver does not own
+            // UDP transport setup, so it declines on the application's behalf
+            // and the connection continues TCP-only. Applications that want to
+            // participate in multitransport must drive the connector directly
+            // using `ClientConnector::complete_multitransport()` instead of
+            // calling `connect_finalize`.
+            buf.clear();
+            let written = connector.skip_multitransport(&mut buf)?;
+            if written.size().is_some() {
+                framed
+                    .write_all(buf.filled())
+                    .map_err(|e| ironrdp_connector::custom_err!("write all", e))?;
+            }
+            continue;
+        }
+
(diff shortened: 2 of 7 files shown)


glamberson:
Opened in error. The change continues on #1098; closing this.

Issue #1446, opened 2026-07-13
Client can't connect to GNOME Remote Desktop: no Graphics Pipeline (EGFX) advertised — the blocker after #1371


strausmann:
**Summary:** With #1254 and #1371 merged, the connection to GNOME Remote Desktop now gets *past* the Server-Deactivate-All handshake and reaches capabilities exchange — where grd immediately closes it because the IronRDP client advertises **no Graphics Pipeline (MS-RDPEGFX)**. Recent GNOME Remote Desktop dropped its legacy RDP graphics paths and now *requires* EGFX (reproduced here on `gnome-remote-desktop 50`), so this is currently a hard wall for both `ironrdp-web` and the native `ironrdp-client` against grd.

### What grd does

GNOME Remote Desktop's RDP backend rejects any client that doesn't advertise the Graphics Pipeline. Reproduced against `gnome-remote-desktop 50` (system mode / remote-login, Ubuntu), driving it with the real `ironrdp-web` client (bundled in Pangolin) — correct credentials, connection reaches `DEMAND_ACTIVE`, then:

```
[RDP] Client did not advertise support for the Graphics Pipeline, closing connection
[rdp_peer_handle_state_demand_active]: [CONNECTION_STATE_CAPABILITIES_EXCHANGE_DEMAND_ACTIVE] freerdp_peer::Capabilities() callback failed
[RDP] Network or intentional disconnect, stopping session
```

On the client this surfaces as `received disconnect provider ultimatum` (grd's clean MCS disconnect after the rejection). This is not client-credentials or network related. That grd removed the legacy paths and mandates EGFX is stated by GNOME and reproduced by other clients too:
- https://discourse.gnome.org/t/rdp-client-did-not-advertise-support-for-the-graphics-pipeline-closing-connection/27169
- Remmina hit the same: https://gitlab.com/Remmina/Remmina/-/issues/3302

### Why the IronRDP client falls over here

The client stack doesn't wire EGFX into the connection:
- `ironrdp-web/src/session.rs` builds its `connector::Config` with `client_codecs_capabilities` + `ironrdp-graphics` (legacy codecs) and never attaches an EGFX processor.
- `ironrdp-egfx` is currently depended on **only by `ironrdp-server`** — it's at 0× in `ironrdp-connector`, `ironrdp-session`, `ironrdp-client`, and `ironrdp-web`.
[...]

PR #1465, opened 2026-07-21
fix(connector): answer connect-time Bandwidth Measure to unblock FreeRDP servers

glamberson:
## Summary

- The connector answers only the RTT auto-detect request at connect time and returns nothing for the Bandwidth Measure Stop, on the assumption that skipping it does not stall the sequence.
- That assumption fails for FreeRDP-based servers: GNOME Remote Desktop blocks in its `AWAIT_BW_RESULT` state until it receives a Bandwidth Measure Results reply and never proceeds to licensing, so the connection hangs right after the Client Info PDU. Windows servers tolerate the omission, which hid it.
- Reply to a connect-time `BandwidthMeasureStop` with a `BandwidthMeasureResults` PDU carrying the payload size the server handed us, over a nominal interval.

## Scope

This is the unblock alone. The reported interval is nominal (`time_delta_ms: 1`) because the sans-I/O layer has no time source: nothing reaches the connector that says when the bytes arrived. The figure is an informational QoS hint and the server proceeds on receipt, which is what unsticks the connection.

Measuring it properly needs an arrival time threaded down from the I/O driver, which is a breaking change to `Sequence::step` and does not belong in a fix aimed at getting FreeRDP servers connecting. It is #1530, stacked on this branch: it introduces `MonotonicInstant`, has `Framed` record when each read completed, and replaces the nominal figure here with the real Start-to-Stop interval and accumulated byte count.

Split out at @CBenoit's suggestion in review.

## Validation

- `cargo xtask check fmt/typos/lints/tests/locks` all pass on the pinned toolchain.
- Regression test: a connect-time Bandwidth Measure Stop produces a response frame and the auto-detect phase continues.
- Reproduced and fixed live against gnome-remote-desktop 49: before, the connector stalled after Client Info with grd in `AWAIT_BW_RESULT`; after, grd proceeds through licensing and DEMAND_ACTIVE to an active session.

## Notes

- Found while bringing up the client-side Graphics Pipeline against grd (#1446), but it is an independent connector bug affecting any connection to a FreeRDP-based server, not specific to EGFX.
- Previously depended on #1511 for a zero-`payloadLength` regression test. #1511 has merged, and that test moved out with the rest of the measurement work, so there is no dependency left here.

### DIFF
--- a/crates/ironrdp-connector/src/connection.rs
+++ b/crates/ironrdp-connector/src/connection.rs
@@ -1252,7 +1252,9 @@ fn respond_to_connect_time_autodetect(
     user_channel_id: u16,
     output: &mut WriteBuf,
 ) -> ConnectorResult<Written> {
-    use ironrdp_pdu::rdp::autodetect::{AutoDetectRequest, AutoDetectResponse, AutoDetectRspPdu};
+    use ironrdp_pdu::rdp::autodetect::{
+        AutoDetectRequest, AutoDetectResponse, AutoDetectRspPdu, BW_RESULTS_CONNECT_TIME,
+    };
 
     match request {
         AutoDetectRequest::RttRequest { sequence_number, .. } => {
@@ -1260,14 +1262,34 @@ fn respond_to_connect_time_autodetect(
             let written = encode_send_data_request(user_channel_id, message_channel_id, &response, output)?;
             Written::from_size(written)
         }
-        // Only RTT is answered at connect time. A connect-time Bandwidth Measure
-        // Stop ([MS-RDPBCGR] 2.2.14.1.4) is defined to warrant a Bandwidth Measure
-        // Results reply, and the Network Characteristics Result is informational.
-        // We deliberately send neither: connect-time auto-detect is informational
-        // and the server proceeds to licensing whether or not it receives them, so
-        // skipping them does not stall the sequence. Full connect-time bandwidth
-        // measurement (replying to Bandwidth Measure Stop with Bandwidth Measure
-        // Results) is left for a follow-up.
+        // A connect-time Bandwidth Measure Stop ([MS-RDPBCGR] 2.2.14.1.4) warrants a
+        // Bandwidth Measure Results reply ([MS-RDPBCGR] 2.2.14.2.2). This reply must
+        // be sent: FreeRDP-based servers (for example GNOME Remote Desktop) block in
+        // their AWAIT_BW_RESULT state until they receive it and never proceed to
+        // licensing without it, so omitting it stalls the whole connection. We do not
+        // run a stateful connect-time measurement, so we report the payload the
+        // server handed us over a nominal interval; the figure is an informational
+        // QoS hint and the server proceeds on receipt. A precise measurement (timing
+        // the Start/Payload/Stop window) can refine the reported bandwidth later.
+        AutoDetectRequest::BandwidthMeasureStop {
+            sequence_number,
+            payload,
+            ..
+        } => {
+            let byte_count = payload
+                .as_ref()
+                .map_or(0, |p| u32::try_from(p.len()).unwrap_or(u32::MAX));
+            let response = AutoDetectRspPdu::new(AutoDetectResponse::BandwidthMeasureResults {
+                sequence_number,
+                response_type: BW_RESULTS_CONNECT_TIME,
+                time_delta_ms: 1,
+                byte_count,
+            });
+            let written = encode_send_data_request(user_channel_id, message_channel_id, &response, output)?;
+            Written::from_size(written)
+        }
+        // Bandwidth Measure Start and Payload carry no client reply, and the Network
+        // Characteristics Result is informational; nothing to send for those.
         _ => Ok(Written::Nothing),
     }
 }
--- a/crates/ironrdp-testsuite-core/tests/connector/autodetect.rs
+++ b/crates/ironrdp-testsuite-core/tests/connector/autodetect.rs
@@ -171,3 +171,31 @@ fn first_licensing_pdu_leaves_autodetect_for_the_licensing_path() {
         "a completed licensing exchange advances the connector out of auto-detection"
     );
 }
+
+#[test]
+fn connect_time_bandwidth_measure_stop_is_answered_and_phase_continues() {
+    let mut connector = connect_time_autodetect_connector();
+
+    // A connect-time Bandwidth Measure Stop ([MS-RDPBCGR] 2.2.14.1.4) must be
+    // answered with a Bandwidth Measure Results reply. FreeRDP-based servers (for
+    // example GNOME Remote Desktop) block in their AWAIT_BW_RESULT state until they
+    // receive it, so no response stalls the whole connection.
+    let user_data = encode_vec(&AutoDetectReqPdu::new(AutoDetectRequest::bw_stop_connect_time(
+        0x5678,
+        vec![0u8; 1024],
+    )))
+    .unwrap();
+    let frame = server_send_data_indication(MESSAGE_CHANNEL_ID, user_data);
+
+    let mut output = WriteBuf::new();
+    let written = connector.step(&frame, &mut output).unwrap();
+
+    assert!(
+        written.size().is_some(),
+        "a connect-time Bandwidth Measure Stop must produce a Bandwidth Measure Results response frame"
+    );
+    assert!(
+        matches!(connector.state, ClientConnectorState::ConnectTimeAutoDetection { .. }),
+        "the connector keeps listening after answering the bandwidth measurement"
+    );
+}


mamoreau-devolutions:
Review finding: the new Bandwidth Measure Results reply reports a fabricated 	imeDelta (1 ms) and only the stop payload in yteCount. Per MS-RDPBCGR 3.2.5.14, connect-time detection must measure from BW_START, count all BW_PAYLOAD data and the stop PDU (including the 8-byte headers), then return those values. Since the server derives bandwidth from these fields, the current reply can materially overstate bandwidth and affect QoS/codec decisions.\n\nI prepared a follow-up commit that tracks the connect-time measurement and adds response-field coverage: https://github.com/Devolutions/IronRDP/tree/copilot/fix-connect-time-bandwidth-measurement (commit 46b488d).

glamberson:
Finding accepted, and thanks for writing the patch rather than just flagging it. The byte accounting's right, including the 8-byte headers per 3.2.5.14, and narrowing to `request_type: BW_STOP_CONNECT_TIME` fixes a real bug of mine: I was answering a continuous stop with a connect-time response type.

Two things before I take it.

`std::time::Instant::now()` panics on `wasm32-unknown-unknown`, and `ironrdp-connector` compiles for that target: `ironrdp-web` depends on `ironrdp` with the `connector` feature, and `xtask/src/wasm.rs` builds it. So the patch would panic during connect-time auto-detect, which is the scenario this PR exists for.

I'd also drop the `payload: Some(payload)` narrowing. A connect-time stop without a payload then falls to the catch-all and the server blocks again, which is the stall this PR fixes. I'd rather stay lenient on that one field.

For the clock, I'm going to have the connector take the time rather than read it: a caller-supplied monotonic `u64` of milliseconds, epoch arbitrary. Three reasons beyond the wasm target:

- `ironrdp-connector` doesn't read a clock anywhere today, and both drivers hand it bytes and nothing else.
- `Framed::read_exact` keeps leftover bytes in its internal buffer, so a clock read inside the state machine measures how fast we drained an already-filled buffer rather than how long the payload took to arrive. The driver stamping at the read measures the interval the spec is actually asking about.
- There's precedent on both sides. `ironrdp-web` already has `now_ms() -> u64` using `performance.now()` with a `Date::now()` fallback, and `quinn-proto`, already in the tree, takes the time as a parameter instead of reading it internally.

On the `Sequence` side that's Core Tier public API, but it can go in as a defaulted trait method so existing implementors compile untouched and only the drivers opt in.

If you'd rather keep this PR small, say so and I'll just swap `std::time::Instant` for `web_time::Instant` in your commit instead. That's a one-line change and `web-time` is already in `Cargo.lock` via `rustls-pki-types`, so it costs nothing in build time.

I'm applying the same treatment on the server side, since `AutoDetectManager` had the identical internal-clock shape from #1177. Taking the timestamps from the caller turned its RTT tests from "average is under 100 ms" into exact assertions. That's #1487.

glamberson:
Implemented. Your byte accounting is in as written, including `bandwidth_measure_pdu_size` with the 8-byte header, and the narrowing to `BW_STOP_CONNECT_TIME`.

The clock is the one thing I changed. `Sequence` gained a defaulted `set_now_ms(&mut self, _now_ms: u64)`, so existing implementors compile untouched, and the async and blocking drivers stamp it immediately after `read_by_hint` and before `step`. `ironrdp-connector` now reads no clock at all: the only mention of `Instant::now` left in the crate is a doc comment explaining why it isn't called. The reported interval floors at 1, so a driver that supplies no clock at all hands the server a useless denominator rather than a zero one.

One correction to something I said earlier, because I was wrong about the reason.

I said dropping your `payload: Some(payload)` narrowing mattered because a connect-time Stop without a payload would fall through to the catch-all and leave the server blocked. That can't actually happen. Writing a test for it, I found that `BandwidthMeasureStop { request_type: BW_STOP_CONNECT_TIME, payload: None }` encodes to 10 bytes which the decoder then rejects with `expected 2 bytes`: the encoder omits the payload-length field when the payload is `None`, but the decoder reads it unconditionally for a connect-time stop. So from the wire `payload` is always `Some`, possibly empty, and your narrowing would not have dropped anything real.

I've kept the arm lenient anyway, since matching on a field the reply never uses buys nothing, but the justification is defence-in-depth rather than the live bug I claimed. That encode/decode asymmetry looks like a genuine `ironrdp-pdu` bug in its own right, and I haven't filed it anywhere yet.

Tests cover the measurement end to end: a 250 ms window over three PDUs totalling 374 bytes with headers, an empty-payload stop, a stop with no preceding start, and the no-clock floor.

The same treatment is on the server side in #1487, where `AutoDetectManager` had the identical internal-clock shape from #1177. The two touch no files in common, so they can land in either order.

CBenoit:
Automated review found no correctness findings at 181d28f2e6519174cec444616aedd4361ea6a929. The connect-time auto-detection implementation and focused tests are consistent with the reviewed evidence. This remains human-review-ready; independently, the public ClientConnectorState field addition is a suspected breaking API surface requiring API-maintainer assessment.

glamberson:
@cbenoit or @mamoreau-devolutions this fix should be ok now I think. thanks.

crates/ironrdp-async/src/framed.rs
```diff
@@ -261,6 +274,10 @@ where
 
         trace!(length = pdu.len(), "PDU received");
 
+        // Stamped at the read, not inside the sequence: this is the moment the bytes
```
CBenoit: `read_by_hint` can return a PDU retained from an earlier socket read: `Framed` keeps surplus bytes after extracting the previous frame. In that case this timestamp is drain time, not receipt time. If Start/Payload/Stop arrive in one stream read, the Stop is typically reported at the same instant as Start (then floored to 1 ms), which materially inflates the bandwidth result. Capture receive metadata when bytes enter `Framed`, or propagate it with buffered frames, and cover a coalesced Start/Payload/Stop sequence.

crates/ironrdp-blocking/src/connector.rs
```diff
@@ -230,6 +243,10 @@ where
 
         trace!(length = pdu.len(), "PDU received");
 
+        // Stamped at the read, not inside the sequence: this is the moment the bytes
```
CBenoit: The blocking `Framed` has the same leftover-buffer behavior, so this stamp also records extraction time rather than receipt time for coalesced bandwidth-measurement PDUs. Please apply the same receive-time propagation here.

CBenoit (changes requested):
Blocking: the new timing hook records frame-extraction time rather than socket-receipt time when `Framed` has buffered surplus bytes. This can collapse connect-time bandwidth intervals under normal TCP coalescing and substantially overstate the result.

CBenoit (commented):
Hi! Human here.

This fixes the immediate issue, but adding externally supplied time to `Sequence` changes the abstraction’s contract substantially. The default method makes it source-compatible, but integrations that do not call it silently produce fabricated measurements, and the timing requirement now applies to all sequence drivers. Moreover, stamping after `read by_hint` may timestamp extraction from an existing `Framed` buffer rather than actual network arrival.

The general idea of keeping the clock outside of the sans-IO sequence is good though.

Could we split the required Bandwith Measure Results response from the timing measurement API and design the latter deliberately in a separate PR? I suggest to use stacked PRs in this case.

The design I would suggest here:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub struct MonotonicInstant(u64);

impl MonotonicInstant {
    pub const ZERO: Self = Self(0);

    pub fn from_millis(milliseconds: u64) -> Self {
        Self(milliseconds)
    }

    pub fn duration_since(self, earlier: Self) -> core::time::Duration {
        core::time::Duration::from_millis(self.0.saturating_sub(earlier.0))
    }
}

pub trait Sequence: Send {
    fn next_pdu_hint(&self) -> Option<&dyn PduHint>;

    fn state(&self) -> &dyn State;

    fn step(
        &mut self,
        input: &'a [u8],
        received_at: MonotonicInstant,
        output: &mut WriteBuf,
    ) -> ConnectorResult<Written>;

    fn step_no_input(&mut self, output: &mut WriteBuf) -> ConnectorResult<Written>;
}
```

This is a significant breaking change, but in this case I think it is the right call.

crates/ironrdp-async/src/framed.rs
glamberson: Both halves of this are right, and the second one bit harder than the first.

The stamp moved to where bytes enter `Framed`: `read()` records the instant the socket read completed, and `read_by_hint` hands that down unchanged. A PDU extracted from surplus bytes now reports the read that filled the buffer rather than the moment it was drained.

That alone does not fix the coalesced case you described, because when Start, Payload and Stop arrive in one read they share that instant exactly, so the interval is 0. Reporting 0 is worse than the 1 ms floor you saw, not better: a server computing `byteCount * 8 / timeDelta` divides by it. So the connector now treats a zero-length window as unmeasurable rather than instantaneous and reports the floor, with the byte count still exact. Same for a Stop with no preceding Start.

That is a judgment call and it is the weakest part of the change, so I would rather flag it than bury it: at read granularity a coalesced window genuinely cannot be timed, and reporting the floor understates a fast link. If you would rather signal "no measurement" some other way, say so and I will change it.

Covered by `connect_time_bandwidth_coalesced_into_one_read_reports_the_floor`, which drives all three PDUs at one instant.

All of this is in #1530, stacked on this branch, since this PR is now the unblock alone.

crates/ironrdp-blocking/src/connector.rs
glamberson: Done, in #1530. The blocking `Framed` stamps `last_read_at` inside `read()` alongside the `extend_from_slice`, and exposes it the same way the async one does, so both drivers report receipt time rather than extraction time.

glamberson:
Split as you suggested. This PR is now the unblock alone: it answers the Bandwidth Measure Stop with the payload size over a nominal interval, which is what unsticks grd's `AWAIT_BW_RESULT`. Body rewritten, since the old one described the timing work and carried a #1511 dependency that has since merged.

The timing API is #1530, stacked on this branch, and it takes your design essentially as written. One deliberate deviation: your sketch declares `step_no_input` without a default body. I kept the default that forwards to `step(&[], MonotonicInstant::ZERO, output)`, on the grounds that every implementor would otherwise write the same line and some would pick a different instant. Happy to make it required if you would rather each one state it.

On integrations silently producing fabricated measurements: making `received_at` a required parameter rather than a separate setter is what addresses that. There is no call to forget, so a driver that does not measure has to pass `ZERO` explicitly, and the connector treats intervals derived from it as unmeasurable rather than as zero.

Your inline point about `read_by_hint` was the sharper one and is answered in that thread.

CBenoit (approved):
LGTM!

PR #1471, opened 2026-07-23
feat(server): measure bandwidth and report it in NetworkCharacteristicsResult

glamberson:
Stacked on #1470, which is itself stacked on #1487, so the merge order is #1487, then #1470, then this. This branch carries both ancestors' commits, so it builds and tests green on its own. The breaking signature changes visible against master come from #1487; baselined against it, this PR reports no semver update required.

Extends the server-side auto-detect added by #1470. The RTT-only Network Characteristics Result becomes the full baseRTT plus bandwidth plus averageRTT report by adding a continuous Bandwidth Measure over the MCS message channel.

Change: once per N auto-detect ticks, AutoDetectManager emits a Bandwidth Measure transaction (Start, Payload, Stop; continuous request type 0x0014 / 0x0429) back to back on the message channel. The client counts the payload window and replies with a Bandwidth Measure Results PDU (timeDelta plus byteCount); the manager turns that into a kbps figure and upgrades the next Network Characteristics Result from the RTT-only form to the all-fields form (NETCHAR_RESULT_ALL). A new measurement is never started while one is outstanding. It reuses the message-channel framing from #1348 and adds no new PDU types.

Testing: unit and integration tests cover pacing, the Start/Payload/Stop transaction sharing one sequence number, bandwidth-results handling, and the RTT-only to all-fields upgrade. `cargo xtask check fmt/lints/tests/typos/locks` all pass.

## Rebased

Rebased onto `master` on 2026-08-02 as one `--update-refs` operation with the rest of the stack, so the checks run against the current tree rather than the state before that day's merges. No conflicts, no content change. All five gates green on this head independently, not only at the stack tip.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/autodetect.rs
+++ b/crates/ironrdp-pdu/src/rdp/autodetect.rs
@@ -270,6 +270,20 @@ impl AutoDetectRequest {
         }
     }
 
+    /// Construct a Network Characteristics Result carrying RTT only (no bandwidth).
+    ///
+    /// Emits the `NETCHAR_RESULT_RTT` form (baseRTT + averageRTT) for servers
+    /// that measure round-trip time but do not run a bandwidth measurement.
+    pub fn netchar_result_rtt(sequence_number: u16, base_rtt_ms: u32, average_rtt_ms: u32) -> Self {
+        Self::NetworkCharacteristicsResult {
+            sequence_number,
+            request_type: NETCHAR_RESULT_RTT,
+            base_rtt_ms: Some(base_rtt_ms),
+            bandwidth_kbps: None,
+            average_rtt_ms,
+        }
+    }
+
     /// Get the sequence number of this request.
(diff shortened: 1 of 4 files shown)


CBenoit:
Review found blocking issues: continuous bandwidth detection incorrectly uses connect-time-only BW_PAYLOAD handling, and NetworkCharacteristicsSync state is ignored. Additional findings include sequence-wrap correlation risk, oversized malformed public PDU encodings, and stale MCS transport documentation; the security-header concern is not applicable.

glamberson:
Folded into #1470 and closing.

The review round on #1470 gated the Network Characteristics Result on having both
RTT and bandwidth, because a result carrying RTT alone reports part of the
picture as though it were the whole. With that gate, #1470 on its own could only
ever emit the form it now declines to send, so the bandwidth measurement here is
not a follow-up to it: the two are one change.

Everything in this branch is in #1470, which is now titled and described for the
combined scope.

PR #1477, opened 2026-07-27
build: bump Rust toolchain to 1.97.1

irvingoujAtDevolution:
Moves the pinned toolchain from 1.89.0 to 1.97.1 (latest stable) and syncs `rust-version` on all 40 published crates.

**This breaks our own MSRV rule, on purpose.** [ARCHITECTURE.md](https://github.com/Devolutions/IronRDP/blob/master/ARCHITECTURE.md#msrv-policy) says the MSRV is the latest stable that's at least 6 months old, and 1.97.1 shipped on 2026-07-14. If we'd rather stay compliant the ceiling today is ~1.93 — easy to redo, just say the word.

Two other bits while I was in there:

- `clippy.toml` msrv was stale at 1.87, so clippy kept warning it disagreed with `Cargo.toml`. Now it matches the pin.
- added `rust-analyzer` to the pinned components.

The rest is clippy fallout from eight releases. Mostly nested `if`s becoming let-chains, plus `is_multiple_of`, `Option::zip`, `checked_div`, byte-string literals, two merged duplicate inherent impls, and four `capability_sets` modules moved to `mod.rs` form.

One thing to know: `clippy::std_instead_of_core` now tells you to import `core::io`, but that whole module is [still unstable](https://github.com/rust-lang/rust/issues/154046) — I checked, it doesn't compile. Those sites get an `#[expect]` with that reason instead.

### Testing

On 1.97.1: `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, and the CI clippy line (`cargo clippy --workspace --all-targets --features helper,__bench --locked -- -D warnings`) are all clean.

### Not fixed here

`crates/ironrdp-connector/src/connection.rs:264` (`front: Front`) trips `clippy::partial_pub_fields`. That one is **already failing on `feat/vmconnect`** before this PR — I built a clean worktree of the base on 1.89 to confirm. It belongs down in #1475, not here, so CI on this branch stays red until that's sorted.


glamberson:
This is untenable. What is your motivation in trying to do this?

CBenoit:
I’m noting this is against our MSRV policy and I don’t see a rationale for that, so I would be inclined to close this PR. [The toolchain was updated to a version following our policy earlier today](https://github.com/Devolutions/IronRDP/pull/1495)

PR #1480, opened 2026-07-28
fix(error): make source locations opt-in

CBenoit:
Default error display omits locations; alternate formatting and reports with explicit location opt-in preserve diagnostic context.


vnikonov-devolutions (approved):
LGTM

PR #1482, opened 2026-07-29
fix(agent): restrict Windows named pipe to the current user

mamoreau-devolutions:
## Summary

The `ironrdp-agent` Windows named-pipe listener inherited the pipe namespace's default ACL, so **any local user** could connect to a running daemon and drive the session — inject input, capture screenshots, read logs, and trigger NOW remote execution. The Unix listener already locks its socket to `0o600` (owner-only); this mirrors that stance on Windows.

## The fix

Build a **protected DACL** (`D:P(A;;GA;;;{user-sid})`) granting `GENERIC_ALL` only to the current user's SID, and apply it to **every** `CreateNamedPipeW` instance (both the first in `bind` and the replacement minted in `accept`) via tokio's `create_with_security_attributes_raw`. The descriptor is cached on the `Listener` and reused across instances (the kernel copies it on each `CreateNamedPipeW`, so it only needs to outlive each synchronous call).

Implementation follows the established windows-FFI style used by sibling crates (`ironrdp-cliprdr-native`): the `windows` crate (0.62, matching siblings), RAII guards (`OwnedTokenHandle`, `OwnedSecurityDescriptor`) for `CloseHandle`/`LocalFree` cleanup, one unsafe op per block with `// SAFETY:` comments, `.cast::<>()` + `try_from` instead of `as` casts, and `tracing::warn!` on cleanup failure.

## Why this matters

On shared Windows hosts the `\\.\pipe\` namespace is reachable by every local user by default. This is the last line of defense against a different local user taking over a running agent session. It closes the asymmetry with the Unix path, which already documents the `/tmp` fallback as world-writable and explicitly sets `0o600`.

## Test coverage

`transport.rs` previously had **zero tests**. Added a Windows-gated `#[cfg(test)]` smoke test (`security_descriptor_for_current_user_is_non_null`) that exercises the full token → SID → SDDL → security-descriptor pipeline for the current process and asserts a non-null descriptor is produced — the happy path `Listener::bind` depends on. (Asserting the ACL denies a *different* user requires a second token context and is out of scope.)

## Verification

All `cargo xtask` CI-equivalent checks pass on a clean tree:

- `cargo xtask check fmt -v` → All good
- `cargo xtask check lints -v` (workspace clippy `--all-targets -D warnings`) → All good (ironrdp-agent produces zero warnings)
- `cargo test -p ironrdp-agent` → 12 passed; 0 failed (incl. the new test)
- `cargo xtask check locks -v` → All good

## Files

- `crates/ironrdp-agent/Cargo.toml` — add `windows = "0.62"` (cfg(windows), Win32_Foundation/Security/Security_Authorization/System_Threading)
- `crates/ironrdp-agent/src/transport.rs` — `OwnedTokenHandle`/`OwnedSecurityDescriptor` RAII guards, `create_server_instance` helper, `Listener` caches + reuses the SD; new smoke test
- `Cargo.lock` — +1 line registering `windows` as an ironrdp-agent dep (no new versions; crates already present via siblings)

## Release

This is a `fix(agent):` conventional commit, picked up by release-plz into the next release cycle (alongside the existing `feat(agent)` NOW-integration commit) to propose the `0.2.0` bump once merged to `master`. CHANGELOG is auto-generated — no manual edit needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2434,6 +2434,7 @@ dependencies = [
  "tracing-subscriber",
  "uuid",
  "whoami",
+ "windows",
 ]
 
 [[package]]
--- a/crates/ironrdp-agent/Cargo.toml
+++ b/crates/ironrdp-agent/Cargo.toml
@@ -65,5 +65,13 @@ now-proto-pdu = { version = "0.4", features = ["std"] }
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
 
+[target.'cfg(windows)'.dependencies]
+windows = { version = "0.62", features = [
+    "Win32_Foundation",
+    "Win32_Security",
+    "Win32_Security_Authorization",
+    "Win32_System_Threading",
+] }
+
 [lints]
 workspace = true
--- a/crates/ironrdp-agent/src/transport.rs
+++ b/crates/ironrdp-agent/src/transport.rs
@@ -5,7 +5,7 @@
 //! - **Unix**: a [`tokio::net::UnixListener`]/[`tokio::net::UnixStream`] at
 //!   `$XDG_RUNTIME_DIR/ironrdp-agent-<uid>.sock`, falling back to `/tmp/ironrdp-agent-<uid>.sock`
 //!   when `XDG_RUNTIME_DIR` is unset.
-//! - **Windows**: a named pipe at `\\.\pipe\ironrdp-agent-<user>`.
+//! - **Windows**: a named pipe at `\\.\pipe\ironrdp-agent-<user>`, restricted to the current user.
 //!
 //! Framing is identical on both: a little-endian `u32` byte-count prefix followed by the `Encode`d
 //! message body.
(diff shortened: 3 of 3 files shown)


mamoreau-devolutions:
Replying to the Copilot review comment on `transport.rs` (the smoke test only checked the descriptor was non-null):

Done — the test now exercises the actual security property. It creates a real pipe instance via `create_server_instance`, then queries the live pipe's DACL with `GetNamedSecurityInfoW` and asserts:

- the DACL is **protected** (`SE_DACL_PROTECTED` set, so inherited ACEs from the pipe namespace can't widen access),
- exactly **one ACE** is present (`GetAclInformation` → `AceCount == 1`),
- that ACE is an **allow ACE** (`ACCESS_ALLOWED_ACE_TYPE`) granting **full access** (`GENERIC_ALL` in the SDDL, which the kernel maps to `FILE_ALL_ACCESS` on the file-type pipe object), and
- its SID **equals the current process token SID** (`EqualSid`).

So it now fails if the SDDL ever grants `Everyone`, drops the `P` flag, or `create_server_instance` stops applying the descriptor. To share the token/SID logic between production and test, the token-query code was extracted into two helpers (`current_user_token_buffer`, `sid_from_token_buffer`) that both `OwnedSecurityDescriptor::for_current_user` and the test call. The queried descriptor is freed via the existing `OwnedSecurityDescriptor` RAII guard.

All `cargo xtask check fmt/lints/tests/locks` pass (12/12 ironrdp-agent tests, including the new `pipe_dacl_grants_generic_all_only_to_current_user`).

vnikonov-devolutions (approved):

PR #1408, opened 2026-07-04
Fix Win7 RDP PDU and bitmap stride handling

cxxzhang:

PR #1486, opened 2026-07-30
fix(session): preserve bitmap source stride

mamoreau-devolutions:
## Summary

- Preserve `TS_BITMAP_DATA` source stride independently of destination bounds for raw, Interleaved RLE, and RDP6 bitmap updates.
- Remove raw 4-byte scanline padding without collapsing padded source columns into following rows.
- Crop only the source extent beyond the destination and reject empty dimensions explicitly; do not add framebuffer bounds suppression.
- Render decoded RDP6 RGB data top-down and retain bottom-up rendering for raw and RLE data.

## Why this supersedes the overlapping proposals

- #1252 identifies both dimensions but applies a clipped stride, which still misaddresses padded source rows; its broad framebuffer-clipping changes are intentionally excluded.
- #1398 correctly identifies the RDP6 stride/orientation issue, but leaves raw and RLE paths unresolved.
- #1408 includes the related stride issue but also mixes unrelated Win7 activation and web changes; its allocation/inference normalization is unnecessary once source stride is explicit.
- #1436 correctly crops padded rows but repacks each path into temporary buffers; this PR preserves the source stride directly and covers all bitmap codecs.

## Specification basis

- MS-RDPBCGR 2.2.9.1.1.3.1.2.2 (`TS_BITMAP_DATA`): separate destination bounds, dimensions, and bottom-up raw rows with 4-byte row padding.
- MS-RDPBCGR 2.2.9.1.1.3.1.2.3 and 2.2.9.1.1.3.1.2.4: compressed bitmap header and Interleaved RLE stream.
- MS-RDPEGDI 2.2.2.5.1: RDP 6.0 bitmap stream.

## Validation

- `cargo test -p ironrdp-session --lib`
- `cargo check -p ironrdp-session --all-features`
- `cargo clippy -p ironrdp-session --all-targets -- -D warnings`
- `cargo xtask check fmt -v`

### DIFF
--- a/crates/ironrdp-session/src/fast_path.rs
+++ b/crates/ironrdp-session/src/fast_path.rs
@@ -21,6 +21,64 @@ use crate::palette::Palette;
 use crate::pointer::PointerCache;
 use crate::{SessionError, SessionErrorExt as _, SessionResult, custom_err, reason_err, rfx};
 
+#[cfg(test)]
+mod tests {
+    use ironrdp_pdu::bitmap::{BitmapData, Compression};
+
+    use super::*;
+
+    #[test]
+    fn raw_bitmap_removes_byte_padding_without_changing_source_stride() {
+        let mut processor = ProcessorBuilder {
+            io_channel_id: 0,
+            user_channel_id: 0,
+            share_id: 0,
+            enable_server_pointer: false,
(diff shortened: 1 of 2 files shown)


irvingoujAtDevolution (approved):
LGTM

PR #1491, opened 2026-07-31
fix(pdu): key auto-detect optional fields off requestType, not the Option

glamberson:
## Summary

MS-RDPBCGR decides which optional fields an auto-detect message carries by its `requestType`. Two of these message types encoded them by inspecting which `Option`s happened to be set instead, so the encoder and the decoder disagreed about the wire.

This started as a fix for `BandwidthMeasureStop` alone. Review found the connect-time fallback was still non-compliant, and checking whether the same shape appeared elsewhere in the file turned up `NetworkCharacteristicsResult` with the identical defect and a worse consequence, so both are fixed here rather than one now and one later.

## The two failure modes

**Connect-time stop with no payload: encodes to bytes the decoder rejects.**

```rust
AutoDetectRequest::BandwidthMeasureStop {
    sequence_number: 7,
    request_type: BW_STOP_CONNECT_TIME,
    payload: None,
}
```

encodes to ten bytes with `headerLength` 0x06 and no length field. Decoding those bytes
fails with `not enough bytes provided to decode: received 0 bytes, expected 2 bytes`,
because the decoder reads `payloadLength` back for every `BW_STOP_CONNECT_TIME`.

**UDP stop with a payload: silently loses it.**

```rust
AutoDetectRequest::BandwidthMeasureStop {
    sequence_number: 3,
    request_type: BW_STOP_RELIABLE_UDP,
    payload: Some(vec![0xAA; 16]),
}
```

encodes the length and the sixteen bytes, which the decoder never reads back for
`BW_STOP_RELIABLE_UDP` or `BW_STOP_LOSSY_UDP`. `AutoDetectReqPdu::decode` does not check
for trailing bytes, so the payload is dropped in transit without an error rather than
refused.

The second is the worse of the two: a round trip that loses data and reports success.

## Network Characteristics Result (2.2.14.1.5)

The same defect, found by sweeping the file rather than reported.

`0x0840` carries baseRTT and averageRTT, `0x0880` carries bandwidth and averageRTT, `0x08C0` carries all three, and the decoder already read them on that basis. Encoding from the `Option`s meant:

- a `0x0840` result with no `base_rtt_ms` wrote a body the decoder cannot read;
- a `0x0840` result carrying `bandwidth_kbps` instead wrote that bandwidth into the slot the decoder reads as baseRTT, so the value came back **silently corrupted** rather than rejected;
- `headerLength` was always derived from `requestType`, so it could contradict the body it described.

Encode and `size()` now consult `requestType` through a shared helper, a value the type does not carry is dropped rather than written, and a missing one that the type requires is an error.

## Fix

`Encode` and `size()` now key off `requestType`, matching the decoder. That makes the
wire form canonical:

- an absent payload on a connect-time stop encodes as a zero length and reads back as an
  empty one;
- a payload on a UDP stop never reaches the wire.

Decoding and re-encoding reproduces the same bytes in every case. The types can still
express states the protocol cannot, so value-level identity is not achievable, but byte
stability is, and that is what a decoder consuming real traffic depends on.

No public signatures change; this is a behaviour fix inside the existing `Encode` impl.

## Tests

New `pdu/autodetect.rs` in `ironrdp-testsuite-core`:

- a connect-time stop with no payload round-trips, and `size()` agrees with `encode()`;
- a UDP stop carrying a payload encodes header-only and comes back with `payload: None`,
  for both the reliable and lossy request types;
- a connect-time stop preserves a real payload;
- decode-then-encode reproduces identical bytes for both shapes.

Three of the four fail against the current encoder and pass with the fix. The fourth is
a control that passed before and after.

Note the existing `pdu_round_trip` fuzz oracle would not have caught this even with
auto-detect added to it, since it discards the result of the re-decode (`let _ =`). That
is deliberate in the oracle's design and out of scope here, but it is why this went
unnoticed.

## How this was found

Writing a test for the connect-time bandwidth measurement in #1465, which needs to answer
a connect-time stop. The `None` case was constructed to check the reply path stayed
lenient and turned out not to be constructible on the wire at all.

## Verification

[MS-RDPBCGR]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/515150db-4e7a-4c9b-88d8-63f9fe79981f

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/autodetect.rs
+++ b/crates/ironrdp-pdu/src/rdp/autodetect.rs
@@ -305,7 +305,14 @@ impl Encode for AutoDetectRequest {
                 request_type,
                 payload,
             } => {
-                if let Some(data) = payload {
+                // Whether the payload fields appear on the wire is decided by
+                // `requestType`, not by whether a payload was supplied: MS-RDPBCGR
+                // 2.2.14.1.4 puts `payloadLength` and `payload` on the connect-time stop
+                // and on no other. Branching on the `Option` let this disagree with the
+                // decoder, which reads those fields back for `BW_STOP_CONNECT_TIME` and
+                // never for the UDP variants.
+                if *request_type == BW_STOP_CONNECT_TIME {
+                    let data = payload.as_deref().unwrap_or(&[]);
                     dst.write_u8(0x08); // headerLength (with payload)
                     dst.write_u8(TYPE_ID_AUTODETECT_REQUEST);
                     dst.write_u16(*sequence_number);
@@ -361,10 +368,16 @@ impl Encode for AutoDetectRequest {
                 HEADER_MIN_SIZE + 2 /* payloadLength */ + payload.len()
             }
 
-            Self::BandwidthMeasureStop { payload, .. } => match payload {
-                Some(data) => HEADER_MIN_SIZE + 2 /* payloadLength */ + data.len(),
-                None => HEADER_MIN_SIZE,
-            },
+            // Mirrors `encode`: the payload fields are keyed off `requestType`.
+            Self::BandwidthMeasureStop {
+                request_type, payload, ..
+            } => {
+                if *request_type == BW_STOP_CONNECT_TIME {
+                    HEADER_MIN_SIZE + 2 /* payloadLength */ + payload.as_ref().map_or(0, Vec::len)
+                } else {
+                    HEADER_MIN_SIZE
+                }
+            }
 
             Self::NetworkCharacteristicsResult {
                 base_rtt_ms,
--- a/crates/ironrdp-pdu/src/rdp/autodetect.rs
+++ b/crates/ironrdp-pdu/src/rdp/autodetect.rs
@@ -84,6 +84,20 @@ pub const BW_STOP_RELIABLE_UDP: u16 = 0x0429;
 /// Bandwidth Measure Stop for continuous detection over lossy UDP.
 pub const BW_STOP_LOSSY_UDP: u16 = 0x0629;
 
+/// Which optional fields a Network Characteristics Result carries, as
+/// `(baseRTT, bandwidth)`.
+///
+/// [MS-RDPBCGR] 2.2.14.1.5 assigns these by `requestType` alone, so encode,
+/// `size` and decode all consult this rather than inspecting the values.
+fn netchar_fields(request_type: u16) -> (bool, bool) {
+    match request_type {
+        NETCHAR_RESULT_RTT => (true, false),
+        NETCHAR_RESULT_BW_RTT => (false, true),
+        NETCHAR_RESULT_ALL => (true, true),
+        _ => (false, false),
+    }
+}
+
 /// Network Characteristics Result: baseRTT + averageRTT (no bandwidth).
 ///
 /// [\[MS-RDPBCGR\] 2.2.14.1.5]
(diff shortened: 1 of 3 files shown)


crates/ironrdp-pdu/src/rdp/autodetect.rs
```diff
@@ -305,7 +305,14 @@ impl Encode for AutoDetectRequest {
...
             } => {
-                if let Some(data) = payload {
+                // Whether the payload fields appear on the wire is decided by
+                // `requestType`, not by whether a payload was supplied: MS-RDPBCGR
+                // 2.2.14.1.4 puts `payloadLength` and `payload` on the connect-time stop
+                // and on no other. Branching on the `Option` let this disagree with the
+                // decoder, which reads those fields back for `BW_STOP_CONNECT_TIME` and
+                // never for the UDP variants.
+                if *request_type == BW_STOP_CONNECT_TIME {
+                    let data = payload.as_deref().unwrap_or(&[]);
```
mamoreau-devolutions, line 338: MS-RDPBCGR §2.2.14.1.4 requires `payloadLength` to be present **and greater than zero** for `BW_STOP_CONNECT_TIME` (`0x002B`). This fallback emits zero when `payload` is `None`, producing a non-compliant PDU. Please reject `None`/an empty payload instead, and update the new test to assert that rejection.

mamoreau-devolutions (changes requested):
The request-type-based wire layout is correct, but the connect-time zero-payload case violates the mandatory non-zero `payloadLength` requirement in MS-RDPBCGR §2.2.14.1.4.

crates/ironrdp-pdu/src/rdp/autodetect.rs
glamberson: Correct, and the PR body advertised the defect as a feature: it said "an absent payload on a connect-time stop encodes as a zero length and reads back as an empty one", which describes a PDU 2.2.14.1.4 forbids. Keying the encoder off `requestType` rather than the `Option` was the right half of the fix; choosing `unwrap_or(&[])` as the fallback was not.

Quoting the sentence that settles it, since it is stronger than "present": "The payloadLength field MUST NOT be present if the value of the requestType field is not set to 0x002B. It MUST be present (and have a value greater than zero) if the value of the requestType field is set to 0x002B."

Encoding now refuses an absent or empty payload on a connect-time stop rather than emitting zero.

I also applied the same rule to the decoder, which you did not ask for but which has the same defect in the other direction. It read `payloadLength` and accepted any value, so it would have taken a zero-length connect-time stop that the spec says cannot exist, leaving us accepting what we refuse to emit. That is the same encoder/decoder asymmetry this PR exists to fix, so leaving it would have been odd.

I checked whether anything in-tree produces such a PDU and would newly hit the error: nothing does. `bw_stop_connect_time` is the only constructor and has no callers outside tests, so this changes no working path. Its doc now records the invariant, since the signature takes a `Vec` and cannot express it.

Tests assert both rejections, and I verified them separately rather than together: removing the encoder guard fails only the encode test, removing the decoder guard fails only the decode test, so neither is carried by the other's fix. The byte-stability test previously used an absent connect-time payload as one of its cases; that state now has no bytes to be stable about, so it uses a real payload and its comment explains why only the UDP-with-payload case remains.

glamberson:
Scope note, since this grew after the thread above was resolved.

Your finding prompted me to check whether the same shape appeared elsewhere in the file, and it did. `NetworkCharacteristicsResult` had the identical defect: its decoder keys the optional fields off `requestType` per 2.2.14.1.5, while its encoder and `size()` branched on which `Option`s were set. The consequence there is worse than the stop's. A `0x0840` result carrying `bandwidth_kbps` instead of `base_rtt_ms` wrote the bandwidth into the slot the decoder reads as baseRTT, so the value came back silently corrupted rather than failing to decode, and `headerLength` was always derived from `requestType`, so it could contradict its own body.

Both are fixed here rather than one now and one later, so the PR is retitled `fix(pdu): key auto-detect optional fields off requestType, not the Option`. The body has a new section for it.

mamoreau-devolutions (approved):
The connect-time payload invariant is now enforced on both encode and decode, and the related Network Characteristics Result field-layout fix correctly follows MS-RDPBCGR §2.2.14.1.5.

PR #1498, opened 2026-07-31
fix(egfx): add spec-compliant Planar frame sender

mamoreau-devolutions:
## Summary

Supersedes #1454.

Adds `GraphicsPipelineServer::send_planar_frame` for pre-encoded RDP6 Planar payloads. It emits the required `StartFrame` / `WireToSurface1` / `EndFrame` sequence, uses `RDPGFX_CODECID_PLANAR` (`0x000A`), and preserves the target surface pixel format.

Unlike #1454, this implementation rejects oversized destination dimensions before starting the frame instead of clipping the wire rectangle. `destRect` defines the dimensions of the decoded bitmap data, so clipping it independently from the pre-encoded Planar payload would be invalid.

## Validation

- `cargo test -p ironrdp-testsuite-core --test integration_tests_core planar_frame`
- `cargo test -p ironrdp-testsuite-core --test integration_tests_core egfx::server`

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -1604,6 +1604,60 @@ impl GraphicsPipelineServer {
         Some(frame_id)
     }
 
+    /// Queue a pre-encoded RDP6 Planar bitmap stream for transmission via EGFX.
+    ///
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
LGTM!

PR #1502, opened 2026-07-31
feat(connector): Hyper-V vmconnect support

irvingoujAtDevolution:
Re-land of #1475 onto **master**. That PR was accidentally merged into `build/update-toolchain` and never reached master.

Lets you connect to a Hyper-V VM console. Same RDP steps as usual; the host wants the front in a different order: Preconnection Blob with the VM id → TLS → CredSSP → X.224 last. That's the [Direct Approach](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/902b090b-9cb3-4efc-92bf-ee13373371e3), and since NLA already ran by then the host must select exactly `PROTOCOL_HYBRID`.

`ClientConnector::new_vmconnect(..)` picks the ordering, `ConfigBuilder::with_vmconnect(vm_id)` on the client, `--vmconnect <VM_ID>` on the viewer.

The whole difference between the two orderings sits in one `Front` enum — where we start, what X.224 advertises, what selected protocol we accept, and where negotiation/NLA each hand over. Only one new state (`PreconnectionBlobSendRequest`); both orderings share the existing X.224 initiation states. Tail is untouched. CredSSP is unchanged.

RDCleanPath is rejected for now (proxy does X.224 before this ordering is ready).

### Testing

Real Hyper-V host (nested NESTED-TGT on IT-HELP-DC), Server 2022 guest, desktop over the console on Direct. Unit tests cover both front orderings.

### DIFF
--- a/crates/ironrdp-client/src/config.rs
+++ b/crates/ironrdp-client/src/config.rs
@@ -58,6 +58,12 @@ pub struct Config {
     pub(crate) connector: ironrdp_connector::Config,
     pub(crate) destination: Destination,
     pub(crate) transport: Transport,
+
+    /// Hyper-V VM ID to connect to via vmconnect, if this is a vmconnect session.
+    ///
+    /// When set, the client sends a Preconnection Blob carrying this id and performs the Hyper-V
+    /// connection ordering (PCB → TLS → CredSSP → X.224 negotiation) instead of the standard one.
+    pub(crate) vm_id: Option<String>,
     pub(crate) kerberos_config: Option<ironrdp_connector::credssp::KerberosConfig>,
     pub(crate) fake_events_interval: Option<Duration>,
     pub(crate) channels: ChannelConfig,
@@ -100,6 +106,11 @@ impl Config {
         &self.transport
     }
 
+    /// Hyper-V VM ID for a vmconnect session, if any.
+    pub fn vm_id(&self) -> Option<&str> {
+        self.vm_id.as_deref()
+    }
+
     /// Optional Kerberos/KDC proxy configuration.
     pub fn kerberos_config(&self) -> Option<&ironrdp_connector::credssp::KerberosConfig> {
         self.kerberos_config.as_ref()
@@ -139,6 +150,7 @@ impl fmt::Debug for Config {
         s.field("connector", &self.connector);
         s.field("destination", &self.destination);
         s.field("transport", &self.transport);
+        s.field("vm_id", &self.vm_id);
         s.field("kerberos_config", &self.kerberos_config);
         s.field("fake_events_interval", &self.fake_events_interval);
         s.field("channels", &self.channels);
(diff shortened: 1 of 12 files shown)


crates/ironrdp-client/src/rdp.rs
```diff
@@ -473,8 +476,11 @@ async fn connect_gateway(
     let framed = ironrdp_tokio::TokioFramed::new(gw_stream);
 
     let connector = build_connector(config, client_addr, input_sender, cliprdr_factory);
-
-    tls_handshake_and_finalize(framed, connector, config).await
+    if config.vm_id().is_some() {
```
CBenoit, line 479: Enabling vmconnect on the gateway path exposes a destination-port loss: `GwConnectTarget` receives only `config.destination.name()`, while the gateway channel request fixes the target port at 3389. A vmconnect destination on 2179 is therefore routed to normal RDP. Do not support this path until the target port is forwarded, and add a 2179 gateway test.

crates/ironrdp-vmconnect/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        other => {
+            return Err(reason_err!(
+                "Initiation",
+                "expected EnhancedSecurityUpgrade after Hyper-V X.224 initiation, got {}",
+                other.name()
+            ));
+        }
+    };
+
+    if selected == REQUIRED_PROTOCOL {
```
CBenoit, line 121: The shared connector advertises both `HYBRID` and `HYBRID_EX`, but this exact-equality check rejects a valid `HYBRID_EX` selection after CredSSP. Either accept compatible CredSSP selections here or stop advertising `HYBRID_EX`; cover HYBRID, HYBRID_EX, TLS-only, and negotiation failure.

crates/ironrdp-client/src/config.rs
```diff
@@ -1119,6 +1149,12 @@ impl ConfigBuilder {
             }),
         };
 
+        // An RDCleanPath proxy negotiates X.224 for us before the Hyper-V ordering gets there. A
+        // gateway is only a tunnel, so the connector still drives the whole handshake itself.
+        if self.vm_id.is_some() && matches!(transport, Transport::RDCleanPath(_)) {
```
CBenoit, line 1154: This validates only RDCleanPath, so `with_vmconnect(...).with_credssp(false)` remains constructible even though vmconnect unconditionally runs CredSSP and later rejects the TLS-only negotiation. Reject CredSSP-disabled vmconnect configurations during build and cover configuration/CLI overrides.

CBenoit (commented):
This adds the required Direct Approach sequencing, but gateway routing, advertised protocol handling, and CredSSP-disabled configuration leave supported vmconnect paths unable to complete; the inline findings identify the required corrections and coverage.

CBenoit:
Human here. Hi @irvingoujAtDevolution, is it a duplicate with #1503? Please close it if it’s not relevant anymore. Thanks!

PR #1503, opened 2026-07-31
feat(connector): Hyper-V vmconnect support

irvingoujAtDevolution:
Adds Hyper-V VMConnect's direct ordering: PCB → TLS → CredSSP → X.224.

Enhanced Session is the default (`GUID;EnhancedMode=1`), with `--vmconnect-basic` for the synthetic console. Kept this separate in `ironrdp-vmconnect`; no SPN changes.

Tested against the nested Hyper-V lab:
- Enhanced: `HYBRID_EX`, rendered 1280×720
- Basic: `HYBRID`, rendered 1280×720
- `cargo xtask check fmt/lints/tests -v`


CBenoit (changes requested):
This PR requires special scrutiny as-is; back to it on Monday

mamoreau-devolutions (changes requested):
Requests changes: the Hyper-V preconnection blob must request enhanced mode (GUID;EnhancedMode=1) for the normal RDP stack to work.

mamoreau-devolutions (changes requested):
Requests changes: VMConnect still uses port 3389 through the gateway, and its mandatory TLS/CredSSP requirements are not enforced by configuration.

crates/ironrdp-vmconnect/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+        other => {
+            return Err(reason_err!(
+                "Initiation",
+                "expected EnhancedSecurityUpgrade after Hyper-V X.224 initiation, got {}",
+                other.name()
+            ));
+        }
+    };
+
+    if selected == SecurityProtocol::HYBRID || selected == SecurityProtocol::HYBRID_EX {
```
CBenoit: `HYBRID_EX` is not valid for this direct sequence without handling the Early User Authorization Result PDU that follows CredSSP. The normal CredSSP path has explicit handling for that exchange, but this path marks CredSSP complete immediately after X.224. Please reject `HYBRID_EX` here (or implement that PDU exchange) so the next read cannot be desynchronized.

crates/ironrdp-vmconnect/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+#[instrument(skip_all, fields(%vm_id, ?mode))]
+pub async fn send_preconnection_blob<S>(framed: &mut Framed<S>, vm_id: &str, mode: Mode) -> ConnectorResult<PcbSent>
+where
+    S: FramedWrite,
+{
+    let bytes = encode_preconnection_blob(vm_id, mode)?;
+
+    debug!(length = bytes.len(), "Send Preconnection Blob");
+    framed
+        .write_all(&bytes)
```
CBenoit, line 88: MS-RDPEPS requires the complete preconnection PDU within ten seconds of TCP connection creation. This awaited write is unbounded, so a stalled peer can violate the new connection sequence. Please apply a bounded deadline that covers this mandatory PCB transmission.

crates/ironrdp-client/src/config.rs
```diff
@@ -866,6 +886,22 @@ impl ConfigBuilder {
         self
     }
 
+    /// Connect to a Hyper-V VM console by VM GUID. Destination must use port
+    /// [`ironrdp_vmconnect::PORT`] (2179). Not supported over RDCleanPath.
+    #[must_use]
+    pub fn with_vmconnect(mut self, vm_id: impl Into<String>) -> Self {
+        self.vm_id = Some(vm_id.into());
```
CBenoit, line 1016: VMConnect always performs CredSSP before X.224, but this builder leaves a caller's `with_credssp(false)` setting intact. The later X.224 request can therefore advertise TLS while `connect_front` has already performed CredSSP and requires HYBRID. Please reject or normalize this incompatible configuration when enabling VMConnect.

crates/ironrdp-client/src/rdp.rs
```diff
@@ -473,8 +476,11 @@ async fn connect_gateway(
     let framed = ironrdp_tokio::TokioFramed::new(gw_stream);
 
     let connector = build_connector(config, client_addr, input_sender, cliprdr_factory);
-
-    tls_handshake_and_finalize(framed, connector, config).await
+    if config.vm_id().is_some() {
```
CBenoit: This routes VMConnect through the gateway even though `GwConnectTarget` does not carry the destination port and the adjacent TODO states the gateway may hard-code 3389. VMConnect requires port 2179, so this can open the wrong service. Please reject the gateway and VMConnect combination until the target port is propagated.

crates/ironrdp-client/src/rdp.rs
```diff
@@ -573,6 +579,65 @@ where
...
+    let mode = config
+        .vmconnect_mode()
+        .ok_or_else(|| ironrdp_connector::general_err!("vmconnect path requires a console mode"))?;
+
+    let pcb_sent = ironrdp_vmconnect::send_preconnection_blob(&mut framed, vm_id, mode).await?;
+
+    debug!("TLS upgrade");
+
+    let (initial_stream, leftover_bytes) = framed.into_inner();
+    let (tls_stream, tls_cert) = ironrdp_tls::upgrade(initial_stream, config.destination.name())
```
CBenoit, line 941: The VMConnect path unconditionally upgrades to TLS, but the public builder and viewer still allow TLS to be disabled. That silently ignores the resolved configuration. Please reject VMConnect with TLS disabled, or force the configuration value consistently when VMConnect is selected.

CBenoit (commented):
Thanks for the VMConnect support. I found several blocking correctness issues in the new direct connection path: it accepts HYBRID_EX without handling the required Early User Authorization Result, permits conflicting CredSSP and TLS configuration, does not bound the pre-TLS PCB write, and allows gateway routing although port 2179 is not carried. The inline comments identify the required corrections before this can be safely used.

crates/ironrdp-vmconnect/src/lib.rs
irvingoujAtDevolution: updated, ironrdp-vmconnect/src/lib.rs 40–41, 147, 188–194

crates/ironrdp-vmconnect/src/lib.rs
irvingoujAtDevolution: updated

crates/ironrdp-client/src/config.rs
irvingoujAtDevolution: updated

crates/ironrdp-client/src/rdp.rs
irvingoujAtDevolution: updated

crates/ironrdp-client/src/rdp.rs
irvingoujAtDevolution: updated

CBenoit (dismissed):
## Required changes

### 1. Make VMConnect an actual opt-in `ironrdp-client` capability

`ironrdp-vmconnect` is currently an unconditional dependency, unlike the other optional client extensions. Please add a `vmconnect = ["dep:ironrdp-vmconnect"]` feature, mark the dependency `optional = true`, and include it in `all` if that feature continues to mean all optional client capabilities.

Gate the VMConnect-specific config fields/accessors/build validation and handshake path behind that feature. The viewer needs to opt in explicitly; because it consumes the `ironrdp` facade, forward a `client-vmconnect` feature there and enable it from `ironrdp-viewer`.

### 2. Finish defining the new public boundary

`ironrdp-vmconnect` is a new public crate but is absent from `ARCHITECTURE.md` and not exposed by the `ironrdp` meta crate. Please document it as an extra-tier API boundary and either add an optional facade feature/re-export for direct supported use, or clearly constrain the supported entry point to `ironrdp-client`.

Also mark `Mode` as `#[non_exhaustive]`: its current exhaustive public enum makes a future console mode a downstream source-breaking addition.

### 3. Move VMConnect coverage into the integration suite

Add `test = false` next to `doctest = false` in `ironrdp-vmconnect/Cargo.toml`, remove the inline test module, and make `ironrdp-vmconnect` a direct dev-dependency of `ironrdp-testsuite-extra`.

The migrated tests should exercise the public contract: PCB V2 serialization for both modes, `send_preconnection_blob` output, the advertised deadline, and configuration/CLI validation. The existing extra-suite configuration tests already cover the latter category, but the new crate's public PCB/write API needs direct integration coverage. Keep the connector-internal state assertions private and test observable behavior instead.

### Compatibility note

No existing public function or type was removed, but `Credentials::username()` now suppresses an empty X.224 `mstshash` cookie. Please call out that wire-level behavioral change in release notes or API documentation if it is intentional.

irvingoujAtDevolution:
@CBenoit updated

crates/ironrdp-client/src/config.rs
```diff
@@ -1264,6 +1318,23 @@ impl ConfigBuilder {
             }),
         };
 
+        #[cfg(feature = "vmconnect")]
+        if self.vm_id.is_some() {
```
mamoreau-devolutions, line 1332: This validates TLS, CredSSP, and unsupported transports, but never validates the destination port. `Destination::new` defaults a bare host to 3389, so `--vmconnect <GUID> hyperv-host` proceeds against the ordinary RDP listener rather than the Hyper-V console. Please require `destination.port() == ironrdp_vmconnect::PORT` (or set 2179 automatically when enabling VMConnect) and add coverage for a bare/3389 destination. `MS-RDPEPS` §3.1.3 requires connecting to the IP address and port on which the server is listening; Hyper-V VMConnect listens on 2179.

mamoreau-devolutions (changes requested):
## Required change`n`nVMConnect must reject a destination other than TCP 2179, or default a VMConnect destination to 2179. This is the remaining correctness blocker.

crates/ironrdp-client/src/config.rs
irvingoujAtDevolution: Fixed in ae1b472a. A bare VMConnect destination now defaults to 2179. Explicit ports are kept as-is, so forwarded/custom endpoints still work. Added coverage for bare, explicit 3389, and custom ports.

mamoreau-devolutions (approved):
The VMConnect port handling now defaults an unqualified destination to TCP 2179 while preserving explicitly selected ports for port-forwarded or custom endpoints. This addresses the requested change.

mamoreau-devolutions (approved):

Issue #1464, opened 2026-07-21
[tracking] Client-side Graphics Pipeline (EGFX): wire the existing decode core into the connector and client stack


glamberson:
### Summary

IronRDP has a working client-side EGFX (MS-RDPEGFX Graphics Pipeline) decode core in `ironrdp-egfx`, but nothing above it was connected: the connector never advertised graphics-pipeline support, and neither `ironrdp-client` nor `ironrdp-web` opened the graphics channel or rendered what came back, so IronRDP clients couldn't use the Graphics Pipeline at all. I'm opening this to track wiring it end to end. The three foundational pieces are already filed as PRs (see the pipeline table): the client-side surface compositor (#1460), the session-side render that drains it (#1461), and native-client registration of the graphics channel (#1462). This issue tracks those, the remaining codec-coverage and negotiation-correctness work, and the coordination with the DVC and client refactors in flight. The sequencing below is a suggested cut, not a fixed plan.

### Why this is now high priority

Recent GNOME Remote Desktop (reproduced on 49 and reported on 50) removed its legacy RDP graphics paths and now **requires** the Graphics Pipeline. A client that doesn't advertise EGFX is rejected during capability exchange:

```
[RDP] Client did not advertise support for the Graphics Pipeline, closing connection
[rdp_peer_handle_state_demand_active]: [CONNECTION_STATE_CAPABILITIES_EXCHANGE_DEMAND_ACTIVE] Capabilities() callback failed
```

So this isn't an optional codec feature anymore; it's the requirement for connecting to a large and growing class of servers (every modern GNOME desktop, and the direction of travel is EGFX-mandatory more broadly). Remmina hit the identical wall (Remmina#3302), and #1446 is the concrete report. Both `ironrdp-web` and the native `ironrdp-client` are affected.

### A personal note

Since my products are server-focused, I haven't worried too much about the client side, but the need's becoming acute.

### What already exists (the hard part)

`ironrdp-egfx/src/client.rs` (`GraphicsPipelineClient`) is a complete state machine: capability negotiation (parses V8 through V10.7), full surface lifecycle (Create/Delete/Map/Reset), the frame loop with `FrameAcknowledge`, and it processes every `GfxPdu`. It natively decodes **AVC420** and **Uncompressed**, and codec dispatch for **ClearCodec (#1175)** and **RemoteFX Progressive (#1443)** is in review. The decode engine turns over; it's just not connected to anything.

### What is missing

[...]

PR #1507, opened 2026-07-31
feat(egfx): decode Planar bitmaps in the client

glamberson:
## What

Wires `RDPGFX_CODECID_PLANAR` (0x000A) into the EGFX client's `WireToSurface1` dispatch, alongside the existing ClearCodec and Uncompressed paths.

## Why

Planar reached the dispatch and fell through to `on_unhandled_pdu`, so a server sending Planar frames got nothing drawn. The decoder already existed: [MS-RDPEGFX] 2.2.2.1 defines the payload for this codec ID as the Planar Codec from [MS-RDPEGDI] 2.2.2.5.1 and 3.1.9, which is the `RDP6_BITMAP_STREAM` that `ironrdp-graphics` decodes and that `ironrdp-session` already uses for the fast-path bitmap route.

This is the same decode-core-exists-but-is-not-wired situation ClearCodec was in before #1175. #1498 makes it reachable in-tree: `send_planar_frame` means an IronRDP server can now send frames an IronRDP client would discard.

`ironrdp-egfx` already depends on `ironrdp-graphics` and already pulls `ClearCodecDecoder` from it, so this reuses `rdp6::BitmapStreamDecoder` rather than adding a second implementation.

## Three spec details that shape the decode

**Dimensions come from `destRect`.** [MS-RDPEGFX] 2.2.2.1 says the field carries both the target point and "the dimensions (width and height) of the bitmap data encapsulated in the bitmapData field", and that it "specifies a bounding rectangle if the codecId field contains the RDPGFX_CODECID_AVC420 (0x000B), RDPGFX_CODECID_AVC444 (0x000E), or the RDPGFX_CODECID_AVC444V2 (0x000F) identifier". Planar is not in that list, so the rectangle is exact and its extent is what the decoder is given.

**Pixels are opaque, deliberately.** An `RDP6_BITMAP_STREAM` can carry an alpha plane: [MS-RDPEGDI] 2.2.2.5.1 defines the `NA` bit as "Indicates if an alpha plane is present. If NA is equal to 1, there is no alpha plane... If NA is equal to 0, the alpha plane is sent as the first color plane." EGFX does not use that channel for opacity. It carries per-pixel opacity in a separate `RDPGFX_CODECID_ALPHA` (0x000C) command holding an `ALPHACODEC_BITMAP_STREAM` ([MS-RDPEGFX] 2.2.4.3), whose stated purpose is to specify "the opacity of each pixel in the encoded bitmap". So a Planar command contributes color only, and the decoder's RGB24 output is widened with a synthesized opaque alpha. That reasoning is in a comment at the conversion site so it does not read like an oversight.

**No vertical flip.** The RDP6 decoder emits top-down row-major, which is how surfaces are stored, so the rows are used as produced.

I also checked whether Planar needs capability negotiation before a server may send it. It does not: the only mentions of the codec in [MS-RDPEGFX] are the codec ID definition itself, unlike the AVC codecs which are capability-gated.

## Tests

The round-trip test builds its payload with IronRDP's own RDP6 encoder and drives a real `WireToSurface1` through the client, so it exercises the actual stream framing rather than a hand-built approximation. It covers both the RLE and raw plane forms, uses distinct per-channel values so a channel swap or row reversal fails, and asserts the update is not routed to `on_unhandled_pdu`. It fails against the previous code.

There is also a unit test for the RGB24 to RGBA widening.

## Scope

No public API change; `cargo semver-checks` reports no version bump required. Related to #1464.

## Verification

`cargo xtask check fmt/lints/tests/typos/locks` all pass.

### DIFF
--- a/crates/ironrdp-egfx/src/client.rs
+++ b/crates/ironrdp-egfx/src/client.rs
@@ -58,6 +58,7 @@ use std::collections::BTreeMap;
 use ironrdp_core::{Decode as _, ReadCursor, impl_as_any};
 use ironrdp_dvc::{DvcClientProcessor, DvcMessage, DvcProcessor};
 use ironrdp_graphics::clearcodec::ClearCodecDecoder;
+use ironrdp_graphics::rdp6::BitmapStreamDecoder;
 use ironrdp_graphics::zgfx;
 use ironrdp_pdu::geometry::ExclusiveRectangle;
 use ironrdp_pdu::{PduResult, decode_cursor, decode_err, pdu_other_err};
@@ -386,6 +387,7 @@ pub struct GraphicsPipelineClient {
     handler: Box<dyn GraphicsPipelineHandler>,
     h264_decoder: Option<Box<dyn H264Decoder>>,
     clearcodec_decoder: ClearCodecDecoder,
+    planar_decoder: BitmapStreamDecoder,
 
     decompressor: zgfx::Decompressor,
     decompressed_buffer: Vec<u8>,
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM!

PR #1510, opened 2026-07-31
fix(egfx): bound the compositor's dirty-region metadata

glamberson:
## Summary

- The compositor charges materialized pixel buffers against `MAX_COMPOSITOR_BYTES` but not the `DirtyRegion` entries that produce them, so `frame` grows outside the budget.
- The repeat filter is O(1) and compares only against the previous entry, so it collapses a rectangle repeated 65,535 times but not two rectangles alternating. The frame stays open until the peer sends `EndFrame`, and the peer chooses when that happens.
- `record_dirty` now charges one entry before pushing, and `EndFrame` releases the whole set before materializing so the pixel copies can spend what the metadata was holding.

## Cost to the peer

`RDPGFX_POINT16` (2.2.1.1) is four bytes on the wire; `RDPGFX_RECT16` (2.2.1.2) is eight. A `DirtyRegion` is ten bytes resident. Three commands loop over these arrays and record one dirty region each:

| PDU | Array element | Wire | Resident | Ratio |
|---|---|---|---|---|
| `RDPGFX_SOLIDFILL_PDU` (2.2.2.4) | `fillRects`, RECT16 | 8 | 10 | 1.25x |
| `RDPGFX_SURFACE_TO_SURFACE_PDU` (2.2.2.5) | `destPts`, POINT16 | 4 | 10 | 2.5x |
| `RDPGFX_CACHE_TO_SURFACE_PDU` (2.2.2.7) | `destPts`, POINT16 | 4 | 10 | 2.5x |

The ratios are modest. The point is that there was no ceiling at all, so a sustained stream grows the queue until the client is out of memory regardless of ratio.

## Validation

- New test alternates two rectangles past the budget and asserts the frame stops growing. Verified against a reverted fix: without the charge it reaches 128 entries, with it 64.
- `cargo xtask check fmt/lints/tests/typos/locks` all pass.

## Notes

- The charge counts logical entries, not `Vec` capacity. Since `Vec` grows by doubling, resident bytes for `frame` can reach roughly twice the charged figure. This matches the existing accounting, which charges `data.len()` for pixel buffers rather than capacity; flagging it rather than diverging from the established model.
- Refusing the charge drops the dirty region, so a starved client can show stale pixels. That is the contract `materialize` already follows for refused allocations, and `charge` logs the allocated and budget figures.
- Reported by Copilot on #1462. Follows #1460, which introduced the budget and the deferred materialization.

### DIFF
--- a/crates/ironrdp-egfx/src/compositor.rs
+++ b/crates/ironrdp-egfx/src/compositor.rs
@@ -409,7 +409,11 @@ impl Compositor {
     /// deleted or remapped mid-frame contributes nothing: its output region is
     /// repainted by whatever maps there next.
     pub(crate) fn end_frame(&mut self) {
-        for dirty in core::mem::take(&mut self.frame) {
+        let frame = core::mem::take(&mut self.frame);
+        // Released before materializing so the pixel copies can spend the budget the
+        // metadata was holding.
+        self.release(frame.len() * size_of::<DirtyRegion>());
+        for dirty in frame {
             self.materialize(&dirty);
         }
     }
@@ -499,6 +503,16 @@ impl Compositor {
             if last.surface_id == surface_id && covers(&last.rect, &rect) {
                 return;
             }
+        }
+
+        // The filter above only collapses consecutive repeats, so alternating
+        // rectangles still land one entry each and the frame stays open until the
+        // peer sends `EndFrame`. Charging the entry puts that growth under the same
+        // ceiling as the pixel buffers: `RDPGFX_POINT16` is four bytes on the wire
+        // against ten resident here, so an uncharged queue would let a peer trade
+        // its own bandwidth for 2.5 times as much of the client's memory.
+        if !self.charge(size_of::<DirtyRegion>()) {
+            return;
         }
         self.frame.push(DirtyRegion { surface_id, rect });
     }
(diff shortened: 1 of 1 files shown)


crates/ironrdp-egfx/src/compositor.rs
```diff
@@ -409,7 +409,11 @@ impl Compositor {
     /// deleted or remapped mid-frame contributes nothing: its output region is
     /// repainted by whatever maps there next.
     pub(crate) fn end_frame(&mut self) {
-        for dirty in core::mem::take(&mut self.frame) {
+        let frame = core::mem::take(&mut self.frame);
+        // Released before materializing so the pixel copies can spend the budget the
+        // metadata was holding.
+        self.release(frame.len() * size_of::<DirtyRegion>());
```
mamoreau-devolutions, line 415: This new `size_of::<DirtyRegion>()` call does not compile because `size_of` is not imported in this module. Please add `mem::size_of` to the `std` imports (I applied that fix locally while reviewing).

glamberson: `size_of` has been in the prelude since Rust 1.80 and the workspace pins 1.94.1 in `rust-toolchain.toml`, so this compiles as written. CI on this head is green across linux, macOS and Windows.

There are ~145 other bare `size_of::<...>()` call sites in the workspace on the same basis, e.g. `ironrdp-rdpeusb/src/pdu/header.rs:70`.

Happy to add the explicit `core::mem::size_of` import anyway if you would rather see it spelled out. Just say and I will push it.

mamoreau-devolutions (approved):

PR #1513, opened 2026-08-02
fix(rdpeusb): accept UsbDevice == 0 in AddDevice

clintcan:
## Problem

A real Windows client (mstsc) assigns `UsbDevice == 0` to a redirected device. Since #1321, `AddDevice::decode` rejects the `0x0..=0x3` range as a *"conflict with default interfaces"*, so **every mstsc `ADD_DEVICE` fails to parse and no device is ever presented**. FreeRDP happens to use an id `>= 0x4`, which is why this went unnoticed.

## Fix

Drop the special-case reject and let `InterfaceId::try_from` validate the field. This is both a correctness fix and a simplification:

- `UsbDevice` is a plain interface ID **scoped to the device's own per-device DVC**, so id 0 is unambiguous there and does not clash with the control-channel default interface IDs (`CAPABILITIES` / `DEVICE_SINK` / …).
- `InterfaceId::try_from` **already enforces the valid range** (`value <= 0x3FFF_FFFF`, which includes 0) and rejects the high/mask range — so the `0x0..=0x3` arm was redundant on top of being wrong.

FreeRDP (`>= 0x4`) and the invalid mask range are unchanged.

## Verification

Verified downstream (macrdp) against **real mstsc**: a redirected camera / HID device enumerates only once `UsbDevice == 0` is accepted. Added a round-trip test (`add_device_accepts_usb_device_zero`) that encodes + decodes an `ADD_DEVICE` carrying `UsbDevice == 0`.

Not a breaking change — the decoder simply accepts a value it previously rejected.

### DIFF
--- a/crates/ironrdp-rdpeusb/src/pdu/sink.rs
+++ b/crates/ironrdp-rdpeusb/src/pdu/sink.rs
@@ -101,7 +101,14 @@ impl AddDevice {
 
         ensure_size!(in: src, size: InterfaceId::FIXED_PART_SIZE);
         let usb_device = match src.read_u32() {
-            0x0..=0x3 => {
+            // `1..=3` are the RDPEUSB default Device Sink / Channel Notification
+            // interface IDs; a device announced with one collides with this crate's
+            // fixed dispatch paths (later completion PDUs would decode as the wrong
+            // message type), so keep them rejected. `0` is the deliberate exception:
+            // a real Windows client (mstsc) assigns `UsbDevice == 0` to a redirected
+            // device, addressed by its own per-device DVC, and rejecting it made every
+            // mstsc ADD_DEVICE fail to parse.
+            0x1..=0x3 => {
                 return Err(invalid_field_err!("UsbDevice", "conflict with default interfaces"));
             }
             value => InterfaceId::try_from(value)?,
--- a/crates/ironrdp-testsuite-core/tests/rdpeusb/device.rs
+++ b/crates/ironrdp-testsuite-core/tests/rdpeusb/device.rs
@@ -1,8 +1,9 @@
-use ironrdp_core::encode_vec;
+use ironrdp_core::{DecodeResult, decode, encode_vec};
 use ironrdp_rdpeusb::io::device::{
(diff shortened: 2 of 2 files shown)


crates/ironrdp-rdpeusb/src/pdu/sink.rs
```diff
@@ -100,12 +100,12 @@ impl AddDevice {
...
-                return Err(invalid_field_err!("UsbDevice", "conflict with default interfaces"));
-            }
-            value => InterfaceId::try_from(value)?,
-        };
+        // `UsbDevice` is a plain interface ID scoped to this device's own channel;
+        // `InterfaceId::try_from` already enforces the valid range (<= 0x3FFF_FFFF).
+        // Do not narrow it further here: a real Windows client (mstsc) assigns
+        // `UsbDevice == 0`, which is unambiguous in the device's per-device DVC and
+        // does not clash with the control-channel default interface IDs.
+        let usb_device = InterfaceId::try_from(src.read_u32())?;
```
mamoreau-devolutions: `InterfaceId::try_from` also admits `1..=3`, not just the observed zero value. Those are RDPEUSB default Device Sink/Channel Notification interface IDs; they collide with this crate's fixed dispatch paths and cause later callback/completion PDUs to be decoded as the wrong message type. Keep `1..=3` rejected and allow only `0` as the deliberate mstsc compatibility exception.

mamoreau-devolutions (changes requested):
Please narrow this compatibility relaxation to the observed `UsbDevice == 0` case.

clintcan:
Good catch — you're right, `1..=3` are the default Device Sink / Channel Notification IDs and a device announced with one would collide with the fixed dispatch, so my broad relaxation was wrong.

Narrowed in `3f4f10f`: the reject is now `0x1..=0x3`, and only `0` is allowed as the deliberate mstsc compatibility exception. Added a parametrized test (`add_device_rejects_reserved_default_interface`) asserting `1`/`2`/`3` still fail to decode, alongside the `UsbDevice == 0` round-trip. Thanks for the review.

mamoreau-devolutions (approved):

PR #1524, opened 2026-08-03
fix(session): preserve RDP6 bitmap orientation

mamoreau-devolutions:
## Summary
- restore bottom-up scanline composition for RDP6 bitmap updates
- add asymmetric raw/RLE bitmap-update regression coverage
- synchronize retained ActiveX DIB updates with GDI and surface copy failures

## Validation
- `cargo test -p ironrdp-session`
- `cargo test -p ironrdp-activex`
- `cargo build -p ironrdp-activex --release`
- verified the real `mstscex.exe -> mstsc.exe -> MsRdpEx.dll -> ironrdpax.dll` path renders the authorized test desktop upright without bands

### DIFF
--- a/crates/ironrdp-activex/src/control.rs
+++ b/crates/ironrdp-activex/src/control.rs
@@ -6020,7 +6020,7 @@ impl Control {
             // The GDI presenter has no hardware-cursor overlay, so cursor updates must be
             // composited into the decoded framebuffer before it receives image events.
             .with_pointer_software_rendering(true)
-            .with_legacy_rdp6_bitmap_order()
+            .with_bottom_up_rdp6_bitmap_order()
             .with_clipboard(if clipboard {
                 ClipboardType::Enable
             } else {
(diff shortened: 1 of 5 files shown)


crates/ironrdp-activex/src/control.rs
```diff
@@ -6020,6 +6020,7 @@ impl Control {
             // The GDI presenter has no hardware-cursor overlay, so cursor updates must be
             // composited into the decoded framebuffer before it receives image events.
             .with_pointer_software_rendering(true)
+            .with_bottom_up_rdp6_bitmap_order()
```
CBenoit: This enables the alternate row mapping for every ActiveX connection, without a negotiated capability, stream field, or peer-specific guard. The RDP6 stream structure does not carry a row-order selector, so a server that follows the existing decoder order will have every affected update vertically inverted. Please gate this compatibility path on demonstrated peer-specific behavior (with captured interoperability coverage), or preserve the existing mapping.

CBenoit (commented):
The review found one blocking interoperability concern: ActiveX now unconditionally reverses all compressed 32-bpp RDP6 bitmap rows, but the stream and negotiated capabilities do not select that interpretation, so connections to servers using the existing order can render affected rectangles upside down. The added unit coverage validates the local flip for raw and RLE streams but does not establish the required peer-specific behavior.

CBenoit (approved):
LGTM

PR #1527, opened 2026-08-03
refactor: move inline tests to file ends

CBenoit:
Inline test modules were interleaved with production code, which made the affected source files harder to follow.

Relocate the affected test modules to the ends of their files while preserving their contents and behavior. This groups both Fast-Path test modules together and applies the same layout to the PDU and EGFX modules.

Tests: `cargo test -p ironrdp-session -p ironrdp-pdu -p ironrdp-egfx`


CBenoit:
This is only moving code around; no behavioral change.

mamoreau-devolutions (approved):

PR #1529, opened 2026-08-03
fix(core): rename {Read,Write}Cursor::rewinded into rewound

RRRadicalEdward:


CBenoit (approved):
Thank you! Changed to `fix` since this is a change in the API, not just style. `fix!` would be the proper breaking change commit type, but it’s touching ironrdp-graphics as well, so let’s just leave that to cargo-semver-checks to figure out later

PR #1541, opened 2026-08-04
fix(pdu): don't reject a Share Data PDU whose totalLength is under-declared

GlassOnTin:
## The problem

VirtualBox's built-in RDP server (VRDP) cannot complete a connection with IronRDP. It fails on the last message of the handshake:

```
ironrdp_connector::connection_finalization: Server Control (Granted Control)
ironrdp_blocking::connector: Wait for PDU connector.state="ConnectionFinalization" hint=X224Hint
RDP protocol decode error: Error {
    context: "<ironrdp_pdu::rdp::headers::ShareControlHeader as ironrdp_core::decode::Decode<'_>>::decode",
    kind: NotEnoughBytes { received: 18, expected: 26 },
}
```

Nothing is missing. Those two numbers come from the `total_length` cross-check at the end of `ShareControlHeader::decode`, so `received` is the length the *server declared* and `expected` is the size the PDU *actually decoded to* — neither is a byte count off the wire:

```rust
if total_length < header_length {
    return Err(not_enough_bytes_err!(total_length, header_length));
}
```

For the Server Font Map PDU those sizes are 18 and 26: `SHARE_CONTROL_HEADER_SIZE` (10) + `ShareDataHeader::FIXED_PART_SIZE` (8) + `FontPdu` (8). VRDP declares `totalLength` as the two headers and never counts its own 8-byte body. All 26 bytes arrive; only the number is wrong.

The inner PDU has already decoded successfully out of those bytes by the time this check runs, which is what makes the rejection avoidable — the check is a consistency assertion, not a bounds check.

## The change

Only a server declaring **more** than was decoded leaves anything to do, and that case already has a handler — the trailing-padding path added for Windows. So the check collapses to:

```rust
if total_length > header_length {
    let padding = total_length - header_length;
    ensure_size!(in: src, size: padding);
    read_padding!(src, padding);
}
```

The `is_empty_output_pdu` special case goes with it: a `total_length` of 0 now falls through as the no-op it already was.

## Genuine truncation is unaffected

This is the part worth checking rather than taking on trust. A short buffer fails *earlier* — inside the inner PDU's own `Decode` impl, via `ensure_fixed_part_size!` — and never reaches this check. The existing test proves it and still passes:

```rust
fn from_header_only_buffer_rejects_rdp_pdu_client_font_list() {
    assert!(decode::<ShareControlHeader>(&CLIENT_FONT_LIST_BUFFER[..18]).is_err());
}
```

## Tests

Added one that fails without this change with the reported error verbatim:

```rust
fn from_buffer_with_under_declared_total_length_parses_rdp_pdu_server_font_map() {
    let mut buf = SERVER_FONT_MAP_BUFFER;
    buf[0] = 18;
    assert_eq!(SERVER_FONT_MAP.clone(), decode(buf.as_ref()).unwrap());
}
```

It sits next to `from_header_only_buffer_defaults_rdp_pdu_server_font_map`, which already tolerates an 18-byte *buffer* — this covers the neighbouring case of a 26-byte buffer with an 18-byte declaration.

`cargo test -p ironrdp-testsuite-core`: 1049 passed, 0 failed.

## Where this was found

Downstream in [Haven](https://github.com/GlassHaven/Haven), from a user who could not connect to VirtualBox at all ([GlassHaven/Haven#422](https://github.com/GlassHaven/Haven/issues/422)). The same fix also resolves an earlier report on that issue where VRDP under-declares a Pointer update mid-session as 24 bytes on an 8565-byte frame — same defect, different PDU, and one we had been working around client-side because it happened after connect. This one happens during finalization, where there is no session yet to work around it in.

I have not captured the VRDP bytes myself, so which PDU the server sends at that point rests on sequence position — it follows Granted Control, where the client waits for Font Map — rather than on a capture. The size arithmetic matches Font Map exactly, but a Control PDU is also 26 bytes, so I would not claim more than that. The fix does not depend on which it is.

### DIFF
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -1001,7 +1001,8 @@ mod tests {
             error.kind(),
             ironrdp_core::DecodeErrorKind::NotEnoughBytes {
                 received: 17,
-                expected: 18
+                expected: 18,
+                ..
             }
         ));
     }
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -1,7 +1,7 @@
 use bitflags::bitflags;
 use ironrdp_core::{
     Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteBuf, WriteCursor, cast_length, decode,
-    ensure_fixed_part_size, ensure_size, invalid_field_err, not_enough_bytes_err, other_err, read_padding,
+    ensure_fixed_part_size, ensure_size, invalid_field_err, other_err, read_padding,
     unsupported_value_err, write_padding,
 };
 use num_derive::FromPrimitive;
@@ -325,22 +325,15 @@ impl<'de> Decode<'de> for ShareControlHeader {
         };
 
         if pdu_type == ShareControlPduType::DataPdu {
+            // Servers get totalLength wrong in both directions. Some Windows versions append
+            // padding past the inner unit; VirtualBox's VRDP declares only the two headers of a
+            // Server Font Map PDU (18) and never counts its 8-byte body. An under-declared
+            // length is not a truncation — the inner PDU above has already decoded from bytes
+            // that were really present, and a genuinely short buffer fails there instead — so
+            // only a server claiming *more* than we consumed leaves anything to skip.
             let header_length = header.size();
 
-            let is_empty_output_pdu = matches!(
-                &header.share_control_pdu,
-                ShareControlPdu::Data(ShareDataHeader {
-                    share_data_pdu: ShareDataPdu::Update(data) | ShareDataPdu::Pointer(data),
-                    ..
-                }) if data.is_empty()
-            );
-
-            if header_length != total_length && !(total_length == 0 && is_empty_output_pdu) {
-                if total_length < header_length {
-                    return Err(not_enough_bytes_err!(total_length, header_length, in: src));
-                }
-
-                // Some Windows versions append padding that is not part of the inner unit.
+            if total_length > header_length {
                 let padding = total_length - header_length;
                 ensure_size!(in: src, size: padding);
                 read_padding!(src, padding);
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -1,8 +1,8 @@
 use bitflags::bitflags;
 use ironrdp_core::{
     Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteBuf, WriteCursor, cast_length, decode,
-    ensure_fixed_part_size, ensure_size, invalid_field_err, other_err, read_padding, unsupported_value_err,
-    write_padding,
+    ensure_fixed_part_size, ensure_size, invalid_field_err, not_enough_bytes_err, other_err, read_padding,
+    unsupported_value_err, write_padding,
 };
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive as _;
@@ -329,14 +329,28 @@ impl<'de> Decode<'de> for ShareControlHeader {
             // padding past the inner unit; VirtualBox's VRDP declares only the two headers of a
             // Server Font Map PDU (18) and never counts its 8-byte body. An under-declared
             // length is not a truncation — the inner PDU above has already decoded from bytes
-            // that were really present, and a genuinely short buffer fails there instead — so
-            // only a server claiming *more* than we consumed leaves anything to skip.
+            // that were really present, and a genuinely short buffer fails there instead — so a
+            // server claiming *more* than we consumed is the only case with anything left to skip.
+            //
+            // Zero stays rejected. That is not a server undercounting its own body, it is a
+            // length field never filled in; the one legitimate use is the no-op empty Update /
+            // Pointer PDU, so that case is carved out rather than the check dropped.
             let header_length = header.size();
+
+            let is_empty_output_pdu = matches!(
+                &header.share_control_pdu,
+                ShareControlPdu::Data(ShareDataHeader {
+                    share_data_pdu: ShareDataPdu::Update(data) | ShareDataPdu::Pointer(data),
+                    ..
+                }) if data.is_empty()
+            );
 
             if total_length > header_length {
                 let padding = total_length - header_length;
                 ensure_size!(in: src, size: padding);
                 read_padding!(src, padding);
+            } else if total_length == 0 && !is_empty_output_pdu {
+                return Err(not_enough_bytes_err!(total_length, header_length));
             }
         }
 
--- a/crates/ironrdp-pdu/src/rdp/headers.rs
+++ b/crates/ironrdp-pdu/src/rdp/headers.rs
@@ -325,18 +325,15 @@ impl<'de> Decode<'de> for ShareControlHeader {
         };
 
         if pdu_type == ShareControlPduType::DataPdu {
-            // Servers get totalLength wrong in both directions. Some Windows versions append
-            // padding past the inner unit; VirtualBox's VRDP declares only the two headers of a
-            // Server Font Map PDU (18) and never counts its 8-byte body. An under-declared
-            // length is not a truncation — the inner PDU above has already decoded from bytes
-            // that were really present, and a genuinely short buffer fails there instead — so a
-            // server claiming *more* than we consumed is the only case with anything left to skip.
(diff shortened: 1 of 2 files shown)


crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -321,22 +321,15 @@ impl<'de> Decode<'de> for ShareControlHeader {
...
-                }) if data.is_empty()
-            );
-
-            if header_length != total_length && !(total_length == 0 && is_empty_output_pdu) {
-                if total_length < header_length {
-                    return Err(not_enough_bytes_err!(total_length, header_length));
-                }
-
-                // Some Windows versions append padding that is not part of the inner unit.
+            if total_length > header_length {
```
mamoreau-devolutions, line 355: `MS-RDPBCGR` 2.2.8.1.1.1.1 defines `totalLength` as the packet length including the Share Control Header, and 3.2.5.2 requires checking it for consistency (a discrepancy should cause the connection to be dropped). This condition now accepts `totalLength < header_length` for every Data PDU, including `totalLength == 0`; the existing `rdp::headers::tests::reject_zero_length_non_output_data_pdu` fails because a `ShutdownDenied` PDU with a zero length is accepted. Please retain the prior exception only for empty Update/Pointer PDUs, or narrow the compatibility handling to a complete known VRDP Font Map body rather than accepting arbitrary under-declared lengths.

mamoreau-devolutions (changes requested):
Request changes: the compatibility relaxation must not accept arbitrary under-declared Data PDUs. The existing zero-length non-output regression test now fails, contrary to the Share Control length consistency requirement in MS-RDPBCGR.

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -331,15 +341,12 @@ impl<'de> Decode<'de> for ShareControlHeader {
                 }) if data.is_empty()
             );
 
-            if header_length != total_length && !(total_length == 0 && is_empty_output_pdu) {
-                if total_length < header_length {
-                    return Err(not_enough_bytes_err!(total_length, header_length));
-                }
-
-                // Some Windows versions append padding that is not part of the inner unit.
+            if total_length > header_length {
```
mamoreau-devolutions, line 355: Still request changes: this condition leaves every positive `totalLength < header_length` accepted. That includes malformed non-output Data PDUs (for example, a header-only PDU declared with 1–17 bytes), not just the complete VRDP Font Map compatibility case. `MS-RDPBCGR §3.2.5.2` requires the server to validate `totalLength` consistency; please narrow the under-declared exception to the validated Server Font Map case, or otherwise prove that the complete inner PDU is present before accepting it.

mamoreau-devolutions (changes requested):
Request changes: the new zero-length guard fixes the specific regression, but the generalized under-declared-length acceptance remains too broad for protocol correctness.

GlassOnTin:
Thanks — the second round was right, and I've taken the first of your two options: the exception is now keyed on the Server Font Map PDU type rather than on a length comparison. A header-only non-output PDU declaring 1–17 is rejected again, so relative to the branch you reviewed this is a net *tightening*, not a further relaxation.

There is one thing worth surfacing, because it is a defect in `master` independent of VRDP.

`header_length` is `header.size()` — the **re-encoded** size, not the bytes that were on the wire. For a Font Map the two differ, because `ShareDataPdu::from_type` substitutes `FontPdu::default()` when the cursor is empty:

```rust
ShareDataPduType::FontMap => {
    let font_pdu = if src.is_empty() { FontPdu::default() } else { FontPdu::decode(src)? };
```

So a header-only Server Font Map re-encodes to 26 while only 18 bytes ever existed. A server sending 18 bytes and honestly declaring `totalLength = 18` is measured against 26, and `master` refuses it. That is a spec-conformant encoding being rejected, and it is why I could not narrow this by comparing lengths — the quantity being compared against is not the one `totalLength` describes.

I did try the derived version you'd expect, accepting `totalLength` when it equals the bytes actually consumed. It is not in the diff, because it turned out to be dead code: the Font Map is the only PDU whose re-encoded size exceeds what it consumed, so the type check always reached the case first. I mutated the clause away and no test changed result, so I removed it rather than ship an unreachable branch with a confident comment on it.

The over-declared branch is untouched, `ensure_size!` included.

Two tests, and I checked both can actually fail rather than assuming:

- `from_header_only_buffer_with_matching_total_length_parses_rdp_pdu_server_font_map` — the 18-byte Font Map declaring 18, i.e. the conformant case `master` refuses. This is the test your automation suggested; it was a good catch.
- `reject_under_declared_non_output_data_pdu` — a `ShutdownDenied` declaring 17, the exact example from your review.

Removing the Font Map exception fails both Font Map tests; restoring the open-ended acceptance fails both rejection tests, including the existing `reject_zero_length_non_output_data_pdu` that my first revision broke. `cargo test --workspace` passes (17 suites) and clippy is clean.

If you would rather not carry a per-server exception at all, the alternative I can offer is to drop the VRDP tolerance and keep only the header-only fix — that alone makes `master` accept the conformant 18-byte Font Map, and it leaves VirtualBox broken. I have a preference but not a strong one; it is your call which side of that line the crate should sit on.

crates/ironrdp-pdu/src/rdp/headers.rs
mamoreau-devolutions: Addressed in `cca62c90`. The compatibility path now requires a Server Font Map with the exact 18-byte header-only declaration. Other under-declared lengths, including zero, are rejected.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/ironrdp-pdu/src/rdp/headers.rs
mamoreau-devolutions: Addressed in `cca62c90`. The exception is now limited to `pduType2 = Font Map` and `totalLength = 18`; tests cover rejection of every other under-declared Font Map length.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/ironrdp-pdu/src/rdp/headers.rs
```diff
@@ -321,6 +321,16 @@ impl<'de> Decode<'de> for ShareControlHeader {
...
+            // Servers get totalLength wrong in both directions. Some Windows versions append
+            // padding past the inner unit; VirtualBox's VRDP declares only the two headers of a
+            // Server Font Map PDU (18) and never counts its 8-byte body. An under-declared
+            // length is not a truncation — the inner PDU above has already decoded from bytes
+            // that were really present, and a genuinely short buffer fails there instead — so a
+            // server claiming *more* than we consumed is the only case with anything left to skip.
+            //
+            // Zero stays rejected. That is not a server undercounting its own body, it is a
+            // length field never filled in; the one legitimate use is the no-op empty Update /
+            // Pointer PDU, so that case is carved out rather than the check dropped.
```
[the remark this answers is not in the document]
mamoreau-devolutions: Addressed in `cca62c90`. The comment now notes that this decoder also handles untrusted client-to-server PDUs, and the compatibility rule requires the exact Font Map type and 18-byte declaration.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

crates/ironrdp-testsuite-core/tests/pdu/rdp.rs
```diff
@@ -78,6 +78,16 @@ fn from_header_only_buffer_defaults_rdp_pdu_server_font_map() {
...
 
+/// VirtualBox's VRDP declares `totalLength` as the size of the two headers (18) and does not
+/// count the 8-byte Font Map body that follows it, so the PDU is complete but under-declared.
+#[test]
+fn from_buffer_with_under_declared_total_length_parses_rdp_pdu_server_font_map() {
+    let mut buf = SERVER_FONT_MAP_BUFFER;
+    buf[0] = 18;
+
+    assert_eq!(SERVER_FONT_MAP.clone(), decode(buf.as_ref()).unwrap());
+}
```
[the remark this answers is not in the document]
mamoreau-devolutions: Addressed. The header-only Font Map with `totalLength = 18` remains covered, and `cca62c90` adds boundary coverage rejecting all other under-declared Font Map lengths.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

mamoreau-devolutions (approved):
The protocol-correctness concerns are addressed in cca62c90: the VRDP exception is limited to Font Map with totalLength 18, nonzero under-declared empty output PDUs are rejected again, and boundary regressions are covered.

PR #1543, opened 2026-08-04
Extract reusable RDP daemon support

mamoreau-devolutions:
## Summary
- extract the local RPC schema, codecs, framing, and transport into unpublished `ironrdp-rpc`
- extract persistent RDP daemon/session ownership into unpublished `ironrdp-daemon`
- retain `ironrdp-agent` as the CLI and preserve its existing daemon workflow

This intentionally excludes viewer and ActiveX RPC backends and launcher/CI work.

## Validation
- `cargo clippy --no-deps -p ironrdp-rpc -p ironrdp-daemon -p ironrdp-agent --all-targets -- -D warnings`
- `cargo test -p ironrdp-agent --test ipc`
- `cargo test -p ironrdp-testsuite-extra --test integration_tests_extra agent::`
- `cargo xtask check locks -v`


mamoreau-devolutions:
The only failing check is Check public API compatibility. Its log shows the runner cannot execute 
ustc after switching to 
obody (Permission denied) and then reports the checkout path is not a directory; the same failure reproduced after rerun before any project compilation. All build, test, formatting, lint, FFI, fuzzing, and localhost automation checks pass.

CBenoit (approved):
LGTM!

The warning comment is more of a nitpick

crates/ironrdp-rpc/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+//! Shared local RPC protocol and transport used by IronRDP local clients.
+
+pub mod ipc;
+pub mod transport;
+pub mod wire;
```
CBenoit, line 6: nitpick: Could we keep `wire` internal to the crate (or feature-gated for the shared test suite)? This makes all of the low-level codec helpers a permanent public API, whereas before extraction they were private unless the explicitly internal test feature was enabled. The tests are the only external consumer, so preserving that boundary would leave room to refactor the wire implementation without downstream compatibility constraints. Tests can be kept by keeping the test-dedicated feature flag.

PR #1545, opened 2026-08-04
feat(viewer): host agent RPC endpoint

mamoreau-devolutions:
## Summary
- host the existing `ironrdp-agent` RPC contract from `ironrdp-viewer --rpc`
- retain one shared daemon state for GUI and RPC input, frames, session lifecycle, screenshots, logs, and NOW operations
- use the agent's default endpoint so `ironrdp-agent` itself remains unchanged

## Usage
Start `ironrdp-viewer --rpc` before invoking `ironrdp-agent`; the viewer owns the usual agent endpoint until its window closes. Use matching explicit `--rpc-endpoint` and agent `--endpoint` values for a custom endpoint.

## Validation
- `cargo fmt --check`
- `cargo test -p ironrdp-viewer`
- `cargo test -p ironrdp-daemon -p ironrdp-agent --lib`
- `cargo check -p ironrdp-viewer -p ironrdp-daemon -p ironrdp-agent`


README.md
CBenoit, line 1: We are extending README.md a bit too eagerly; most of the details can be kept in the individual README.md so we don’t flood the front page.
Also less likely to generate conflicts.

mamoreau-devolutions: Trimmed the root README to the existing concise daemon overview; viewer/backend-specific behavior remains in the crate READMEs.

crates/ironrdp-daemon/src/daemon.rs
```diff
@@ -560,6 +635,12 @@ impl Daemon {
         Response::ok()
     }
 
+    /// Requests the RPC server stop after replying to the caller.
+    pub fn shutdown(&self) -> Response {
+        let _ = self.shutdown.send(());
```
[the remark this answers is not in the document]
mamoreau-devolutions: Shutdown is now represented separately until `handle_connection` writes the success response; only then does it publish the shutdown signal. Added a regression test for the deferred signal.

crates/ironrdp-viewer/src/main.rs
```diff
@@ -69,6 +78,75 @@ fn main() -> anyhow::Result<()> {
...
+                NonZeroU32::new(u32::from(frame.height)),
+            ) else {
+                continue;
+            };
+            if frame_proxy
+                .send_event(ViewerEvent::Frame {
+                    buffer: frame.pixels,
+                    width,
+                    height,
+                })
```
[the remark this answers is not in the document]
mamoreau-devolutions: The daemon retains only the latest frame, and the forwarding task now coalesces a lightweight `FrameAvailable` wakeup with an atomic pending flag. `RpcApp` acknowledges it before reading that retained frame.

crates/ironrdp-viewer/src/app.rs
```diff
@@ -33,9 +51,21 @@ pub struct App {
 
 impl App {
     pub fn new(
-        event_loop: &EventLoop<RdpOutputEvent>,
+        event_loop: &EventLoop<ViewerEvent>,
```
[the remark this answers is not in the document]
mamoreau-devolutions: Restored `App::new(&EventLoop<RdpOutputEvent>, ...)` as the direct-mode compatibility wrapper. RPC mode now uses the separate `RpcApp` type with `EventLoop<ViewerEvent>`.

crates/ironrdp-viewer/Cargo.toml
```diff
@@ -28,6 +28,8 @@ qoi = ["ironrdp/qoi"]
 qoiz = ["ironrdp/qoiz"]
 
 [dependencies]
+ironrdp-daemon = { path = "../ironrdp-daemon", version = "0.1" }
```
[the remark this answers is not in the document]
mamoreau-devolutions: Marked `ironrdp-daemon` as `# public` because `RpcApp::new` exposes `Arc<Daemon>` in its public signature.

crates/ironrdp-daemon/src/daemon.rs
```diff
@@ -512,16 +563,37 @@ impl Daemon {
...
-            );
+            ),
+            Err(ResizeError::NoSession | ResizeError::Closed) => Response::typed_error(
+                crate::ipc::AgentErrorCategory::Unavailable,
+                "session input channel is unavailable",
+            ),
+            Err(ResizeError::Full) => Response::typed_error(
+                crate::ipc::AgentErrorCategory::Unavailable,
+                "session input channel is full",
+            ),
```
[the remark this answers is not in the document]
mamoreau-devolutions: Split the cases: `NoSession` now reports `no active session`, while `Closed` continues to report an unavailable input channel. Added coverage for the no-session response.

crates/ironrdp-agent/README.md
```diff
@@ -2,13 +2,22 @@
...
+local RPC backends:
 
 - **Daemon** (`ironrdp-agent daemon-start`): a long-lived, foreground process implemented by
   [`ironrdp-daemon`] that owns the
   [`ironrdp-client`] engine and one RDP session. It stays alive across many CLI invocations and
   serves requests over a local IPC transport (a Unix domain socket on Unix, a named pipe on
   Windows).
+- **Viewer** (`ironrdp-viewer --rpc`): the same RPC contract hosted by the visible
+  [`ironrdp-viewer`]. The viewer window and CLI share one RDP session, framebuffer, and input path.
+
```
CBenoit: This is not an `ironrdp-agent` thing, it’s part of `ironrdp-viewer`, thus we don’t need to update ironrdp-agent’s README.md at all.
PR title and body must also be adjusted in consequence, and please follow the conventional commit spec (use other commits as reference)

crates/ironrdp-viewer/src/app.rs
```diff
@@ -19,8 +21,27 @@ use winit::window::{CursorIcon, CustomCursor, Window, WindowAttributes};
...
+pub enum ViewerEvent {
+    FrameAvailable,
+    Shutdown,
+}
+
+/// Where local window input is sent.
+enum InputTarget {
+    Direct(RdpInputSender),
+    Rpc(Arc<Daemon>),
+}
```
CBenoit, line 34: I think we already abstracted how inputs are sent to the RDP client (via tokio channels). My intuition is that all we need to support RPC calls is to clone the sender to the RPC server, and then we don’t need to modify anything on the frontend, and everything is mostly self-contained. Verify this intuition, and report back

CBenoit (commented):
The intuition I have that could make this much simpler:

It should be possible to modify ONLY ironrdp-viewer, and keep ironrdp-agent as-is. Simply expose an RPC endpoint (a named pipe I recall) from ironrdp-viewer that ironrdp-agent will use as if it was an ironrdp-agent server. Minimal changes, works the same (except ironrdp-viewer binary must be called instead).

crates/ironrdp-agent/README.md
mamoreau-devolutions: Reworked the PR so the agent is unchanged: its README, CLI, help, and transport changes were removed. Viewer-specific instructions now live only in the viewer README. The branch is a single conventional `feat(viewer): host agent RPC endpoint` commit, and the PR title/body now match that scope.

crates/ironrdp-viewer/src/app.rs
mamoreau-devolutions: Verified the intuition: cloning `RdpInputSender` alone cannot preserve the RPC contract. The sender is created only after `Connect`, while the daemon also owns configuration/session lifecycle, retained frames, screenshots, logs, NOW operations, and the shared `Database` that correctly maps GUI and RPC input operations to FastPath events. The simplification now leaves the agent unchanged: `ironrdp-viewer --rpc` manually hosts that existing `Daemon` contract at the agent default endpoint, and both GUI and RPC input use the same daemon state.

CBenoit (commented):
Another question I have: how necessary the new shutdown action is? Feel free to keep if it’s load bearing, just avoid extra complexity if not required

CBenoit (approved):
Approving for velocity, as this is looking mostly good to me

crates/ironrdp-viewer/src/app.rs
mamoreau-devolutions: Removed the wire-level `Request::Shutdown` action. It was only needed for the agent-side `stop` command, which is no longer part of this viewer-only design. Closing the viewer still directly signals its in-process daemon to stop, so the local lifecycle behavior remains without expanding the RPC schema.

CBenoit:
This finding is a bit concerning:

> RDP_PASSWORD is now shown with its current value in clap-generated help unless environment values are hidden. This can print the credential when a user runs ironrdp-viewer --help; match the agent CLI's secret handling (crates/ironrdp-agent/src/cli.rs:306).

The other suppressed comments are also interesting

mamoreau-devolutions:
Addressed the suppressed findings: `RDP_PASSWORD` is hidden in viewer help; frame wakeups no longer clone or inspect a framebuffer before coalescing; RPC mode hides Winit's local cursor; viewer RPC server failures are now returned from `run_rpc`; and grouped RPC input reserves once and emits one FastPath message. I also ran the exact stable formatter check that previously failed, plus the targeted viewer/daemon tests and agent/viewer/daemon compile check.

crates/ironrdp-viewer/src/cli.rs
```diff
@@ -504,3 +540,20 @@ fn resolve_clipboard_type(cli: ClipboardType, redirect_clipboard: bool) -> Resol
         ClipboardType::Stub => ResolvedClipboardType::Stub,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::ViewerConfig;
```
CBenoit: Inlined tests are against ARCHITECTURE.md. Either place in the testsuite, or remove. I think this test is low value and could be deleted.

mamoreau-devolutions: Removed the low-value inline CLI test. I also fixed the two Clippy violations that failed the Linux, Windows, and macOS checks (`core` atomic import and a single inherent `RpcApp` implementation block).

PR #1556, opened 2026-08-05
fix(async): reject a zero-length unmatched PDU in read_by_hint

clintcan:
## Problem

`Framed::read_by_hint` loops to skip PDUs that don't match the requested hint:

```rust
loop {
    match hint.find_size(self.peek()).map_err(io::Error::other)? {
        Some((matched, length)) => {
            let bytes = self.read_exact(length).await?.freeze();
            if matched { return Ok(bytes); } else { debug!("Received and lost an unexpected PDU"); }
        }
        None => { /* read more / handle EOF */ }
    }
}
```

If a `PduHint` reports `Some((false, 0))` — an unmatched, **zero-length** PDU — then `read_exact(0)` consumes nothing and performs no I/O, so the loop spins forever at 100% CPU: it never `.await`s a real read, so it never yields to the runtime and never observes EOF. Because it doesn't yield, it also starves whatever task drives it (e.g. an acceptor's connection loop), so a single malformed frame can take a whole server down.

I hit this in practice via the built-in X.224 hint: a 2-byte fast-path-shaped header (first byte with `& 0b11 == 0`, e.g. `04 00`) made `find_size` return a zero-length fast-path `PduInfo`, `read_by_hint` treated it as an unexpected PDU, and spun — unauthenticated, before TLS, from one connection.

#1515 hardened `find_size` so the built-in X.224 / fast-path hint no longer returns a zero-length PDU, which resolves that concrete trigger. This PR is defense-in-depth for the primitive itself: `read_by_hint` accepts arbitrary `dyn PduHint` implementations, so it shouldn't rely on every hint avoiding the degenerate case to avoid an infinite loop.

## Fix

Guard the single non-progress case — an unmatched zero-length PDU can't be skipped, so fail instead of looping:

```rust
if length == 0 && !matched {
    return Err(io::Error::new(
        io::ErrorKind::InvalidData,
        "PduHint reported a zero-length unmatched PDU; cannot make progress",
    ));
}
```

The invariant becomes: `read_by_hint` always makes forward progress or returns an error.

## Testing

`ironrdp-async` and `ironrdp-tokio` are both `test = false`, and there are currently no async-framing unit tests in the repo (`find_size`, where the parsing lives, has the regression coverage added in #1515). Verified manually that `04 00` / `00 00` no longer spin. Happy to add a `read_by_hint` regression test in `ironrdp-testsuite-core` (it would need `ironrdp-async` / `ironrdp-tokio` dev-deps + a small mock `PduHint`) if you'd prefer one included.

### DIFF
--- a/crates/ironrdp-async/src/framed.rs
+++ b/crates/ironrdp-async/src/framed.rs
@@ -170,6 +170,18 @@ where
         loop {
             match hint.find_size(self.peek()).map_err(io::Error::other)? {
                 Some((matched, length)) => {
+                    // Guard against a zero-length unmatched PDU. Skipping it (the `else`
+                    // branch below) consumes nothing via `read_exact(0)` and performs no
+                    // I/O, so the loop would spin forever at 100% CPU — never yielding to
+                    // the runtime and never observing EOF. This is reachable from a single
+                    // malformed frame if any `PduHint` reports `Some((false, 0))`, so fail
+                    // instead of hanging rather than relying on every hint to avoid it.
+                    if length == 0 && !matched {
+                        return Err(io::Error::new(
+                            io::ErrorKind::InvalidData,
+                            "PduHint reported a zero-length unmatched PDU; cannot make progress",
+                        ));
+                    }
                     let bytes = self.read_exact(length).await?.freeze();
                     if matched {
                         return Ok(bytes);


clintcan:
**Real-world impact + reproduction (from a downstream server)**

Adding context that this is a live, exploitable issue rather than just a defensive nicety — I hit it in [macrdp](https://github.com/clintcan/macrdp), a downstream IronRDP-based RDP server.

**Impact:** unauthenticated, pre-TLS, remote. A single **2-byte** fast-path frame (`04 00` / `00 00` — first byte `& 0b11 == 0`, length `0`) makes `Framed::read_by_hint` spin at 100% CPU: `find_size` returns `Some((matched: false, length: 0))`, the "skip the unmatched PDU" branch does `read_exact(0)` (no I/O, no forward progress, no `.await` yield), and the loop never terminates. Because it never yields, it also **starves the executor driving it** — for a server that's the acceptor's connection loop, so the whole process stops accepting new connections = full outage. It's reachable before TLS/CredSSP, so connection-level auth / rate-limiting never sees it, and a liveness watchdog that probes the runtime can miss it (the spin pegs one worker while the probe still runs on the others).

**Verified** on unfixed vs fixed builds: firing that exact 2-byte trigger at the unpatched server pegs a worker at ~100% CPU and wedges the accept loop (needs a restart to recover); with this guard, CPU stays at 0%, the accept loop stays live, and each malformed connection is cleanly rejected.

**Re #1515:** that hardened `find_size` so the built-in X.224 / fast-path hint no longer returns a zero-length PDU — which covers the *built-in* path. This guards the framing primitive itself, so it also covers arbitrary `dyn PduHint` impls (which downstreams provide) rather than relying on every hint to avoid the degenerate case. Either fix alone breaks the loop; together they're belt-and-suspenders.

macrdp is shipping this as a vendored patch in the meantime. Happy to add a unit test here — a stub `PduHint` returning `(false, 0)`, asserting `read_by_hint` errors instead of hanging — or adjust anything if that helps it land. Thanks!

PR #1564, opened 2026-08-07
fix(egfx): advertise only capability sets the client can decode

### LINKED ISSUE #1563: egfx client advertises AVC444 it cannot decode, so frames are dropped
### Summary

`GraphicsPipelineHandler::capabilities()` defaults to advertising `CapabilitySet::V10_7`, which tells the server AVC444 is available. `GraphicsPipelineClient::handle_wire_to_surface1` has no AVC444 decoder — it matches `Codec1Type::Avc444 |Codec1Type::Avc444v2`, logs "AVC444 codec not yet implemented" and forwards the PDU to `on_unhandled_pdu`. Per MS-RDPEGFX 3.2.5.19 the server picks the highest set the client advertised, so on any host that prefers AVC444 the client asks for a codec it then discards. Every frame of the desktop is lost and the session looks frozen.

This is item 5 of the tracking issue #1464 ("advertise only what we can decode", which already names AVC444 as the example) filed as a standalone bug with a reproduction and server-side measurements, since that item is currently a design note scheduled as step 4 of the suggested order. It is **not** #1455: that one preserves `AVC_DISABLED` through negotiation in `GraphicsPipelineServer`, on the other side of the wire.

Verified present on `main` today, not only in the published 0.3.0.

### Where

`crates/ironrdp-egfx/src/client.rs`:

- default `capabilities()` returns `V10_7 { SMALL_CACHE }` first. `AVC_DISABLED` is not set, so `CodecCapabilities::from_capability_set` reports `avc444: true` for it.
- `handle_wire_to_surface1` has no AVC444 arm; both variants go to `on_unhandled_pdu`.
- `DvcProcessor::start` does filter AVC-implying sets, but only in the `else` branch taken when  `h264_decoder.is_none()`. With a decoder present — the configuration where AVC is wanted — the  `V10_7` offer goes out unfiltered.

### Reproduction

Client: `GraphicsPipelineClient::new(handler, Some(openh264_decoder))` with the default `capabilities()`, registered as a DVC via `DrdynvcClient::with_dynamic_channel`.
Server: Windows 11 Pro 25H2 (build 26200.8875), no GPU (WARP software rendering), same LAN.

The channel negotiates, `on_capabilities_confirmed` fires, surfaces are created and mapped, and `StartFrame`/`EndFrame` arrive with `FrameAcknowledge` going back — but no pixels are ever painted, because every `WireToSurface1` carries AVC444. At `debug` level the log fills with "AVC444 codec not yet implemented".

Downstream report with the full measurements: totoshko88/RustConn#262.

### Server-side evidence

Sampled on the server with `Get-Counter "\RemoteFX Graphics(*)\*"` over 24 s while interacting:

| Client | session res | output fps avg/max | avg encoding time |
|---|---|---|---|
| this client, default caps (AVC444 selected) | 3360x1930 | **0.00 / 0.00** | 4.50 ms |
| FreeRDP `sdl-freerdp`, same host, minutes apart | 3360x1930 | 9.31 / 24.00 | 5.25 ms |

`frames skipped/second - insufficient server resources` and `- insufficient client resources` were both 0.00, and guest CPU stayed at 1-3%: the server is encoding fine and is not throttling.
FreeRDP's near-identical encoding time on the same pipeline confirms the server side is healthy.

### Why it is hard to diagnose from the client

Nothing reaches `on_bitmap_updated`, so a consumer watching for empty `BitmapUpdate`s to detect a failing decoder sees no signal at all — the counter simply stays at zero while the screen is blank. The only trace is a `debug!` line. Two suggestions, independent of the fix:

- log the first occurrence at `warn!` with the `codec_id`, rather than `debug!` per frame
- consider a typed callback for "content arrived in a codec this client cannot decode", so  consumers can fall back deliberately instead of inferring it from an absence

### Minimal fix

Drop `V10_7` from the default until an AVC444 decoder exists:

```rust
fn capabilities(&self) -> Vec<CapabilitySet> {
    vec![
        CapabilitySet::V8_1 {
            flags: CapabilitiesV81Flags::AVC420_ENABLED | CapabilitiesV81Flags::SMALL_CACHE,
        },
        CapabilitySet::V8 {
            flags: CapabilitiesV8Flags::SMALL_CACHE,
        },
    ]
}
```

Windows then confirms V8.1 and sends AVC420, which decode_avc420 handles. An alternative that keeps V10.x is to advertise it with AVC_DISABLED set, but that also gives up AVC420, so V8.1 is the better default while AVC444 is missing.

The proper fix is AVC444 itself: RFX_AVC444_BITMAP_STREAM is a 4-byte header carrying which subframes are present plus the size of the first YUV420p stream, followed by one or two AVC420 subframes (luma, then the chroma auxiliary view) encoded with the same profile. Both go through the existing AVC420 path; what is new is the split and the 4:4:4 chroma reconstruction. Happy to test a branch against the 25H2 host above — the counter harness is scripted and repeatable.

### A related gap in the no-decoder path
When h264_decoder is None the filter leaves V8 only, and Windows then answers with RFX Progressive. WireToSurface2 currently goes to on_wire_to_surface2, whose default is a no-op, so that configuration produces the same blank screen by a different route until #1443 lands. Worth noting on #1464 item 4: without a progressive decoder, opening the graphics channel at all is worse for a consumer than staying on the RemoteFX path.

totoshko88:
## Summary

`GraphicsPipelineHandler::capabilities()` defaulted to advertising `CapabilitySet::V10_7`, which
tells the server AVC444 is available unless `AVC_DISABLED` is set. `GraphicsPipelineClient` has no
AVC444 decoder — `handle_wire_to_surface1` routes `Codec1Type::Avc444` and `Avc444v2` to
`on_unhandled_pdu`. The server picks one of the advertised sets and prefers the most capable, so on
a host that supports AVC444 the client asked for a codec it then discarded: every frame of the
desktop was lost, the session looked frozen, and nothing reached `on_bitmap_updated` at all.

This drops V10.7 from the default. V8.1 keeps `AVC420_ENABLED`, which *does* decode, so no H.264
capability is lost, and V8 remains the no-AVC fallback. V10.7 can come back the moment AVC444
decodes — the doc comment and the new test both say so.

Closes #1563. This is item 5 of the #1464 tracking list ("advertise only what we can decode", which
already names AVC444 as the example), addressed for the AVC444 case only.

## Why the smaller set is not a downgrade

| set | AVC420 | AVC444 | decodable today |
|---|---|---|---|
| V10.7 (removed) | yes | yes | **no** — AVC444 has no decoder |
| V8.1 `AVC420_ENABLED` | yes | no | yes, via `decode_avc420` |
| V8 | no | no | yes (Uncompressed, ClearCodec, Planar) |

An alternative that keeps V10.x is to advertise it with `AVC_DISABLED` set, but per
`CodecCapabilities::from_capability_set` that flag clears `avc420` as well, so it gives up H.264
entirely. V8.1 is the better default while AVC444 is missing.

## Changes

- `crates/ironrdp-egfx/src/client.rs` — remove `V10_7` from the default `capabilities()`; document
  why V10.x is absent and what has to happen before it returns.
- Two unit tests: the advertised sets imply no AVC444, and AVC420 stays advertised. They assert
  against `CodecCapabilities::from_capability_set` rather than naming versions, so they keep holding
  if the set changes shape.
- `crates/ironrdp-testsuite-core/tests/egfx/client.rs` — `client_keeps_avc_caps_with_decoder` now
  expects two sets instead of three.

## Validation

Reproduced and measured against Windows 11 Pro 25H2 (build 26200.8875), no GPU (WARP software
rendering), same LAN, `base tcp rtt` 5 ms. Server-side counters sampled with
`Get-Counter "\RemoteFX Graphics(*)\*" -SampleInterval 2 -MaxSamples 12` while interacting:

| client | session res | output fps avg/max | avg encoding time |
|---|---|---|---|
| default caps, AVC444 selected | 3360x1930 | **0.00 / 0.00** | 4.50 ms |
| FreeRDP `sdl-freerdp`, same host, minutes apart | 3360x1930 | 9.31 / 24.00 | 5.25 ms |

`frames skipped/second — insufficient server resources` and `— insufficient client resources` were
both 0.00 and guest CPU stayed at 1–3%: the server was encoding and not throttling. FreeRDP's
near-identical encoding time on the same pipeline confirms the server side was healthy, which is
what places the fault in the client's advertisement.

Local checks, all clean:

- `cargo xtask check fmt -v`
- `cargo xtask check tests -v` — 22 suites, 0 failed
- `cargo clippy -p ironrdp-egfx --all-targets --all-features` — 0 warnings
- `cargo doc -p ironrdp-egfx --no-deps` — the two remaining warnings (`THIRD_PARTY_NOTICES`,
  `compositor` → private `Compositor`) predate this branch

`cargo xtask check typos -v` was skipped: `typos-cli` is not installed in this environment.

## Deliberately not in this PR

- **AVC444 itself.** It needs YUV plane access that `H264Decoder`/`DecodedFrame` do not expose —
  the auxiliary stream is a packed chroma carrier, not an image, so it cannot be reconstructed from
  two RGBA buffers — plus a second decoder context, since each subframe is its own H.264 sequence.
  That is a public-API decision for the maintainers; I sketched a non-breaking shape
  (`fn decode_avc444(..) -> DecoderResult<DecodedFrame>` defaulting to an error, implemented for
  `OpenH264Decoder`) in #1563 and would rather agree on it before writing the reconstruction.
  I can validate a branch against the 25H2 host above; the counter harness is scripted.
- **Diagnosability.** #1563 also suggests raising the "AVC444 codec not yet implemented" `debug!` to
  a first-occurrence `warn!` and adding a typed callback for undecodable content. Both are useful
  and both are separate from this fix, so I left them out rather than mix concerns.
- **The no-decoder path.** With `h264_decoder: None` the filter leaves V8 only and a Windows host
  answers with RFX Progressive, which `on_wire_to_surface2` drops by default — the same blank screen
  by another route until #1443 lands. Out of scope here, noted for #1464 item 4.

### DIFF
--- a/crates/ironrdp-egfx/src/client.rs
+++ b/crates/ironrdp-egfx/src/client.rs
@@ -220,18 +220,25 @@ pub struct BitmapUpdate {
 pub trait GraphicsPipelineHandler: Send {
     /// Returns the capability sets to advertise to the server
     ///
-    /// The default advertises V10.7 (AVC420+AVC444), V8.1 (AVC420 only),
-    /// and V8 (no AVC) as fallback.
+    /// The default advertises V8.1 (AVC420) and V8 (no AVC) as fallback.
+    ///
+    /// V10.x is deliberately absent. Those versions signal AVC444 support
+    /// unless `AVC_DISABLED` is set, and [`GraphicsPipelineClient`] has no
+    /// AVC444 decoder: [`Codec1Type::Avc444`] and [`Codec1Type::Avc444v2`]
+    /// reach [`GraphicsPipelineHandler::on_unhandled_pdu`] instead of being
+    /// decoded. The server picks one of the advertised sets ([capability
+    /// negotiation]) and prefers the most capable, so offering AVC444 makes a
+    /// host that supports it send nothing the client can paint. Add V10.7 back
+    /// once AVC444 decodes.
     ///
     /// Note: AVC-capable versions are automatically filtered out at
     /// advertisement time if no H.264 decoder is configured on the
     /// [`GraphicsPipelineClient`]. If all returned sets require AVC
     /// and no decoder is available, a V8-only fallback is used.
+    ///
+    /// [capability negotiation]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/31c6e2b1-335b-4a75-9454-bb2309958c21
     fn capabilities(&self) -> Vec<CapabilitySet> {
         vec![
-            CapabilitySet::V10_7 {
-                flags: CapabilitiesV107Flags::SMALL_CACHE,
-            },
             CapabilitySet::V8_1 {
                 flags: CapabilitiesV81Flags::AVC420_ENABLED | CapabilitiesV81Flags::SMALL_CACHE,
             },
@@ -1371,5 +1378,34 @@ mod tests {
         let rgba = convert_uncompressed_to_rgba(&wire_pixels);
         // Expected: [R, G, B, 0xFF] per pixel (alpha forced to opaque)
         assert_eq!(rgba, vec![0xFF, 0x80, 0x00, 0xFF, 0x30, 0x20, 0x10, 0xFF]);
+    }
+
+    /// Every capability set the default advertises must correspond to a codec
+    /// `handle_wire_to_surface1` can actually decode.
+    ///
+    /// The server picks one of the advertised sets and prefers the most capable,
+    /// so a set that implies AVC444 makes the server send `Avc444`/`Avc444v2`,
+    /// which fall through to `on_unhandled_pdu` and leave the screen blank.
+    #[test]
+    fn default_capabilities_do_not_advertise_avc444() {
+        for cap in TestHandler.capabilities() {
+            assert!(
+                !CodecCapabilities::from_capability_set(&cap).avc444,
+                "advertised set implies an AVC444 decoder that does not exist: {cap:?}"
+            );
+        }
+    }
+
+    /// AVC420 must stay advertised: it is the H.264 flavour that does decode, and
+    /// dropping V10.x must not cost it.
+    #[test]
+    fn default_capabilities_still_advertise_avc420() {
+        assert!(
+            TestHandler
+                .capabilities()
+                .iter()
+                .any(|cap| CodecCapabilities::from_capability_set(cap).avc420),
+            "no advertised set enables AVC420"
+        );
     }
 }
--- a/crates/ironrdp-testsuite-core/tests/egfx/client.rs
+++ b/crates/ironrdp-testsuite-core/tests/egfx/client.rs
@@ -178,12 +178,13 @@ fn client_keeps_avc_caps_with_decoder() {
     let caps_pdu = decode_caps_from_message(&messages[0]);
     assert_eq!(
         caps_pdu.0.len(),
-        3,
-        "expected all three capability sets with decoder present"
+        2,
+        "expected both capability sets with decoder present"
     );
-    assert_eq!(caps_pdu.0[0].version, CapabilityVersion::V10_7);
-    assert_eq!(caps_pdu.0[1].version, CapabilityVersion::V8_1);
-    assert_eq!(caps_pdu.0[2].version, CapabilityVersion::V8);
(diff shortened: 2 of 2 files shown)


CBenoit (approved):
LGTM!

PR #1569, opened 2026-08-07
fix(graphics): don't emit a value after a run that ends the input

glamberson:
RL mode codes the following value's magnitude minus one, so it cannot express
zero. #1179 began coding a zero there when the run consumed the rest of the
input, and the decoder reconstructs magnitude as the coded value plus one, so
every input ending in a zero run gained a trailing 1.

That is what fails `progressive_fractional_base_quantization_reconstructs_rgb`
from #1499 on master: quantized coefficients end in a zero run, so the phantom
value lands in HH1 of each component.

The trailing zeros are the run and nothing follows them. #1179's other two
fixes are untouched.

2000 randomized round trips per entropy mode, 0 failures. Workspace suite green.


CBenoit (approved):
LGTM

PR #1586, opened 2026-08-08
fix(rdpeusb): enforce terminal state after retraction

uchouT:
## Summary

Make device retraction a terminal state for `UrbdrcDeviceServer` until the containing dynamic virtual channel is closed.

- Track the device lifecycle as awaiting device, ready, or retracted.
- Reject new I/O and cancellation requests after `RETRACT_DEVICE`.
- Ignore client device messages received after retraction.
- Preserve pending request metadata until channel teardown.
- Notify the device backend when the DVC closes.
- Add regression coverage for the retract and close lifecycle.

## Rationale

Previously, `retract_device` cleared the pending request state but left the device processor usable. New requests could still be constructed, while late client completions could be treated as mismatched requests.

According to [MS-RDPEUSB section 3.3.5.3.8](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/8a86b746-b361-4e4a-8e47-53c90089893d), the client should terminate the dynamic virtual channel after receiving `RETRACT_DEVICE`. Messages received while the channel is closing are therefore ignored as out-of-sequence, following [section 3.1.5](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/f31cc9ef-a8c3-4a4d-b64d-f027ed0752b0).

Closing the containing DVC remains the responsibility of the server or application layer.

### DIFF
--- a/crates/ironrdp-rdpeusb/src/server.rs
+++ b/crates/ironrdp-rdpeusb/src/server.rs
@@ -220,16 +220,27 @@ pub trait UrbdrcDeviceServerBackend: Send {
         request_id: RequestId,
         completion: TransferOutCompletionResult,
     ) -> PduResult<()>;
+
+    /// Notifies the backend that the per-device dynamic channel has closed.
+    fn close(&mut self, _channel_id: u32) {}
 }
 
 pub struct UrbdrcDeviceServer {
     msg_alloc: IdAllocator,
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
LGTM! Thanks

PR #1589, opened 2026-08-08
ci: remove transient package manifest

mamoreau-devolutions:
Remove an untracked root package.json after each Claude invocation before github-script loads trusted validators.

The Claude action can leave an invalid manifest that Node rejects while resolving validator modules. Tracked manifests remain untouched.


CBenoit:
Superseded by #1591

PR #1626, opened 2026-08-11
feat(rdpemt): add the RDPEMT multitransport tunnel crate

glamberson:
# feat(rdpemt): add the RDPEMT multitransport tunnel crate

First of six. Adds `ironrdp-rdpemt`, the tunnel described by [MS-RDPEMT] that
carries dynamic virtual channel traffic over a sideband connection.

This is the smallest and most self-contained piece of the RDP-UDP work
@mamoreau-devolutions asked for on #140, and it stands alone: nothing here
depends on the UDP transport, and the tunnel is driven by decoded PDUs rather
than by a socket.

## What is here

**PDUs.** The tunnel create request and response, the data PDU with its
subheader chain, and the Initiate Multitransport request and response from
[MS-RDPBCGR] 2.2.15 that begin the exchange.

**Tunnel state machine.** Both sides of the handshake: a client that sends the
create request and waits for a response, a server that checks the security
cookie it receives against the one it issued, and the established state where
both sides simply carry data. Sans-I/O, so it takes PDUs in and produces events
out; `no_std` with an `alloc` dependency.

## Notes for review

`TunnelData::encode` returns an error rather than truncating when the header or
payload length will not fit its field. The lengths are `u8` and `u16` on the
wire and the payload comes from a higher layer, so a cast would have silently
produced a PDU claiming a size it did not have.

The security cookie comparison on the server side is what [MS-RDPEMT] 3.2.5.1
asks for, matching the cookie the client presents against the one carried in
the Initiate Multitransport Request.

## What comes next

The remaining five PRs add the RDP-UDP transport this tunnel runs over: the
MS-RDPEUDP v1 handshake PDUs, the MS-RDPEUDP2 v2 data-transfer PDUs, the
connection state machine, a tokio driver, and fuzz targets. Each is stacked on
the previous one and will be linked here as it is filed.

## Test plan

`cargo xtask check fmt/lints/tests/typos/locks`

54 tests in `ironrdp-testsuite-core`, covering every PDU round trip plus the
tunnel state machine on both sides including the cookie-mismatch path.


mamoreau-devolutions (commented):
## Protocol review (MS-RDPEMT / MS-RDPBCGR §2.2.15)

**Baseline:** Core wire layouts for CreateRequest / CreateResponse / Data / header packing match MS-RDPEMT §2.2 and the §4 examples. Action values `0x0` / `0x1` / `0x2` (not the §3.1.5.4 `0x3` typo), LE fields, 16-byte cookie, and BCGR multitransport payload constants look correct. No blocking endianness or fixed-layout defects found.

**Please address before treating this as integration-ready:**

1. Server may enter `Established` and allow Data before CreateResponse is actually sent (MS-RDPEMT §3.1.5.5).
2. Create PDU decode should validate `PayloadLength` (24 / 4).
3. `SubHeaderLength` encode can silently truncate.
4. Tunnel SM drops SubHeaders on RX and cannot send them (auto-detect over sideband).
5. Prefer reusing `ironrdp-pdu` multitransport types (or make payload-only unmistakable) to avoid missing `SEC_TRANSPORT_*` headers.

Inline notes cover the above plus smaller HRESULT / cookie-compare hygiene items.

> [!NOTE]

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+                    });
+                }
+            }
+
+            // Server in Created receives CreateRequest
+            (TunnelState::Created, Side::Server, TunnelPdu::CreateRequest(req)) => {
+                if req.request_id == self.config.request_id && req.security_cookie == self.config.security_cookie {
+                    self.state = TunnelState::Established;
+                    self.enqueue_create_response(TunnelCreateResponse::S_OK);
+                    self.events.push_back(TunnelEvent::Established);
```
mamoreau-devolutions, line 220: **Protocol (MS-RDPEMT §3.1.5.5):** On a successful cookie match the server immediately enters `Established` and enqueues CreateResponse. `send_data` only checks `Established`, so a driver that reacts to `TunnelEvent::Established` before draining `poll_pdu()` can put a Data PDU on the wire **before** CreateResponse.

The server MUST NOT send data until it has received CreateRequest, processed it, **and sent** CreateResponse with a successful HRESULT.

Suggested fix: hold a post-accept state until CreateResponse is dequeued (or refuse `send_data` while CreateResponse is still at the front of `outgoing`).

> [!NOTE]

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+                    // E_ACCESSDENIED
+                    let hr = 0x80070005u32;
+                    self.enqueue_create_response(hr);
+                    self.events.push_back(TunnelEvent::Failed { hr_response: hr });
+                }
+            }
+
+            // Either side in Established receives Data
+            (TunnelState::Established, _, TunnelPdu::Data(data_pdu)) => {
+                self.events.push_back(TunnelEvent::Data(data_pdu.higher_layer_data));
```
mamoreau-devolutions: **Protocol (MS-RDPEMT §2.2.1.1.1 / §3.1.5.3):** Decode preserves SubHeaders on `TunnelData`, but the tunnel only emits `higher_layer_data`. Combined with `send_data` always using empty `sub_headers`, sideband auto-detect (BW measure start/stop/results via RDP_TUNNEL_SUBHEADER) is neither visible nor sendable through `RdpemtTunnel`.

Surface subheaders on receive (dedicated event or richer `Data`) and allow attaching them on send.

> [!NOTE]

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+                    self.state = TunnelState::Failed;
+                    self.events.push_back(TunnelEvent::Failed {
+                        hr_response: resp.hr_response,
+                    });
+                }
+            }
+
+            // Server in Created receives CreateRequest
+            (TunnelState::Created, Side::Server, TunnelPdu::CreateRequest(req)) => {
+                if req.request_id == self.config.request_id && req.security_cookie == self.config.security_cookie {
```
mamoreau-devolutions: **Security hygiene (MS-RDPEMT §3.2.5.1 / §5.1):** Cookie match is the binding check between the main session and this UDP tunnel. Prefer a constant-time compare for the 16-byte cookie (and ideally the request-id + cookie decision) rather than `==`.

Also note: product behavior often drops the connection on mismatch instead of sending a failure CreateResponse; `E_ACCESSDENIED` below is still protocol-legal.

> [!NOTE]

crates/ironrdp-rdpemt/src/pdu/create_request.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        if header.header_length != 4 {
+            return Err(ironrdp_core::DecodeError::invalid_field(
+                Self::NAME,
+                "HeaderLength",
+                "must be 4 for CreateRequest",
+            ));
+        }
+
+        ironrdp_core::ensure_size!(in: src, size: PAYLOAD_SIZE);
```
mamoreau-devolutions, line 114: **Protocol (MS-RDPEMT §2.2.1.1, §3.1.5.2, §4.1):** CreateRequest payload is fixed at 24 bytes (`PayloadLength = 0x0018`). Decode checks `HeaderLength == 4` but never validates `header.payload_length`.

Please reject when `payload_length != 24` so the length field cannot disagree with the body (message-mode framing depends on it). Same check for CreateResponse with `payload_length == 4`.

> [!NOTE]

crates/ironrdp-rdpemt/src/pdu/create_response.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// S_OK: tunnel created successfully.
+    pub const S_OK: u32 = 0x0000_0000;
+
+    const NAME: &'static str = "RDP_TUNNEL_CREATERESPONSE";
+
+    /// Total wire size: 4 (header) + 4 (payload) = 8 bytes.
+    const WIRE_SIZE: usize = TunnelHeader::MIN_SIZE + PAYLOAD_SIZE;
+
+    /// Whether this response indicates successful tunnel creation.
+    pub fn is_success(&self) -> bool {
```
mamoreau-devolutions, line 54: **Interop edge (MS-RDPEMT §3.1.5.5 / §3.3.5.1):** Spec text requires a *successful* HRESULT (severity bit clear), not specifically `S_OK`. Windows always sends `S_OK` on accept, so this is fine for product interop, but `SUCCEEDED(hr)` / `(hr as i32) >= 0` would match the normative wording if non-Windows peers are in scope.

> [!NOTE]

crates/ironrdp-rdpemt/src/pdu/subheader.rs
```diff
@@ -0,0 +1,129 @@
...
+        let total = self.wire_size();
+        ironrdp_core::ensure_size!(in: dst, size: total);
+
+        // SubHeaderLength includes itself and SubHeaderType
+        #[expect(
+            clippy::as_conversions,
+            clippy::cast_possible_truncation,
+            reason = "sub-headers are small; max 257 bytes"
+        )]
+        let sub_header_length: u8 = total as u8;
```
mamoreau-devolutions: **Protocol (MS-RDPEMT §2.2.1.1.1):** `SubHeaderLength` is one byte and includes length + type + data. `total as u8` silently wraps when `2 + data.len() > 255`, producing a header that disagrees with the following bytes.

`TunnelData` already refuses oversized `HeaderLength` / `PayloadLength` — please use the same `u8::try_from(total)` pattern here and return an encode error on overflow.

> [!NOTE]

crates/ironrdp-rdpemt/src/pdu/multitransport_request.rs
```diff
@@ -0,0 +1,237 @@
+//! Server Initiate Multitransport Request PDU payload.
+//!
+//! MS-RDPBCGR Section 2.2.15.1.
+//!
+//! This is the **inner payload** after the TPKT + X224 + MCS + security
```
mamoreau-devolutions: **API / wire footgun (MS-RDPBCGR §2.2.15.1):** These types are the *inner payload only*. Full message-channel PDUs already exist in `ironrdp-pdu::rdp::multitransport` with the required `BasicSecurityHeader` (`SEC_TRANSPORT_REQ` / `SEC_TRANSPORT_RSP`).

Payload layout looks correct, but encoding this type alone onto the MCS message channel would omit the security header. Prefer reusing / wrapping the `ironrdp-pdu` types, or make the payload-only nature unmistakable in the type name and call sites.

> [!NOTE]

glamberson: Deleted multitransport_request.rs and multitransport_response.rs entirely, including the re-exports and the testsuite-core test for the response type. Confirmed ironrdp-pdu::rdp::multitransport::MultitransportRequestPdu/MultitransportResponsePdu already cover this, already validate BasicSecurityHeader's SEC_TRANSPORT_REQ/SEC_TRANSPORT_RSP flags, and are already wired into the connector's MultitransportBootstrapping state with a fuzz oracle. Nothing in this crate consumed the duplicates: TunnelConfig is a plain two-field struct any caller can build directly from the real ironrdp-pdu type's fields, so to_tunnel_config wasn't load-bearing either. This also resolves the test-location inconsistency flagged separately, since the whole module is gone.

crates/ironrdp-rdpemt/src/pdu/multitransport_request.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn to_tunnel_config(&self) -> TunnelConfig {
+        TunnelConfig {
+            request_id: self.request_id,
+            security_cookie: self.security_cookie,
+        }
+    }
+}
+
+impl Encode for MultitransportRequest {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
```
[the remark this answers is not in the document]
glamberson: Deleted multitransport_request.rs and multitransport_response.rs entirely, including the re-exports and the testsuite-core test for the response type. Confirmed ironrdp-pdu::rdp::multitransport::MultitransportRequestPdu/MultitransportResponsePdu already cover this, already validate BasicSecurityHeader's SEC_TRANSPORT_REQ/SEC_TRANSPORT_RSP flags, and are already wired into the connector's MultitransportBootstrapping state with a fuzz oracle. Nothing in this crate consumed the duplicates: TunnelConfig is a plain two-field struct any caller can build directly from the real ironrdp-pdu type's fields, so to_tunnel_config wasn't load-bearing either. This also resolves the test-location inconsistency flagged separately, since the whole module is gone.

crates/ironrdp-rdpemt/src/pdu/multitransport_request.rs
```diff
@@ -0,0 +1,237 @@
...
+            requested_protocol: RequestedProtocol::LossyUdp,
+            security_cookie: [0xFF; 16],
+        };
+
+        let encoded = ironrdp_core::encode_vec(&original).expect("encode");
+        let decoded: MultitransportRequest = ironrdp_core::decode(&encoded).expect("decode");
+        assert_eq!(decoded, original);
+        assert_eq!(decoded.requested_protocol, RequestedProtocol::LossyUdp);
+    }
+}
```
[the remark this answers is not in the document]
glamberson: Deleted multitransport_request.rs and multitransport_response.rs entirely, including the re-exports and the testsuite-core test for the response type. Confirmed ironrdp-pdu::rdp::multitransport::MultitransportRequestPdu/MultitransportResponsePdu already cover this, already validate BasicSecurityHeader's SEC_TRANSPORT_REQ/SEC_TRANSPORT_RSP flags, and are already wired into the connector's MultitransportBootstrapping state with a fuzz oracle. Nothing in this crate consumed the duplicates: TunnelConfig is a plain two-field struct any caller can build directly from the real ironrdp-pdu type's fields, so to_tunnel_config wasn't load-bearing either. This also resolves the test-location inconsistency flagged separately, since the whole module is gone.

crates/ironrdp-rdpemt/src/tunnel.rs
glamberson: I checked this empirically rather than by inspection alone: send_data can't be called until handle_pdu returns, and by then CreateResponse is already enqueued at the front of the same outgoing FIFO send_data appends to. I reproduced the exact scenario (react to Established, call send_data before draining poll_pdu) in a test, and CreateResponse still goes out first. I confirmed the test is meaningful by mutating send_data's push_back to push_front, which does break the order and fails it.

I didn't change the state-machine logic since nothing's broken, but I added the scenario as a permanent regression test and a doc note on outgoing explaining why the ordering holds, so a future refactor that split the queue wouldn't break this silently.

crates/ironrdp-rdpemt/src/tunnel.rs
glamberson: I switched the cookie comparison to subtle::ConstantTimeEq, which was already resolvable in the lock file transitively, so this adds no new crate to the dependency graph. I left request_id as an ordinary comparison since it isn't a secret. I left the failure response as E_ACCESSDENIED rather than dropping the connection, since that's still protocol-legal and changing it would be a behavior change beyond this finding's scope. I verified this clean under both default and no_std feature sets, and the existing mismatch test still passes.

crates/ironrdp-rdpemt/src/pdu/header.rs
```diff
@@ -0,0 +1,158 @@
...
+        Ok(())
+    }
+
+    fn name(&self) -> &'static str {
+        Self::NAME
+    }
+
+    fn size(&self) -> usize {
+        Self::MIN_SIZE + self.sub_headers.iter().map(TunnelSubHeader::size).sum::<usize>()
+    }
```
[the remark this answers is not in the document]
glamberson: I made encode derive header_length fresh from sub_headers, matching what TunnelData::encode already did correctly for its own header construction, instead of trusting the stored field. Same error path as TunnelData uses: it refuses rather than truncates when the sub-headers exceed one byte of length. I added your exact scenario as a regression test (header_length: 4 with a non-empty sub-header) and confirmed it fails without the fix by reverting encode to trust the stored field directly.

crates/ironrdp-rdpemt/src/error.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn decode(error: DecodeError) -> Self;
+    fn encode(error: EncodeError) -> Self;
+    fn invalid_state(context: &'static str) -> Self;
+    fn unknown_action(context: &'static str, action: u8) -> Self;
+    fn non_zero_flags(context: &'static str, flags: u8) -> Self;
+    fn header_too_small(context: &'static str, length: u8) -> Self;
+    fn invalid_header_length(context: &'static str, action: TunnelAction, length: u8) -> Self;
+    fn tunnel_rejected(context: &'static str, hr_response: u32) -> Self;
+    fn cookie_mismatch(context: &'static str) -> Self;
+}
```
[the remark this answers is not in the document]
glamberson: I confirmed all six are genuinely dead: nothing constructs them anywhere in the crate. The actual failure paths report differently than these variants assumed: cookie/request-id mismatch goes through TunnelEvent::Failed, not a returned Err, and header/action decode failures raise plain ironrdp_core::DecodeError directly rather than these dedicated kinds. I removed all six variants, their trait methods, and the now-unused TunnelAction import, leaving Decode/Encode/InvalidState, which are the three actually reachable.

crates/ironrdp-rdpemt/src/pdu/subheader.rs
glamberson: I switched to u8::try_from(total), erroring rather than truncating, matching the pattern TunnelData already uses. The removed #[expect] comment's own reasoning was wrong: it said "max 257 bytes" and 257 doesn't fit in a u8 to begin with. I added a regression test at the actual overflow boundary (254 bytes of data, 2+254=256) and one confirming 253 bytes (the 255 boundary) still encodes, and confirmed the overflow test fails against the reverted truncating cast.

crates/ironrdp-rdpemt/src/tunnel.rs
glamberson: I checked whether this would collide with the not-yet-filed driver PR: its write path bypasses send_data entirely by constructing TunnelData directly, so it's unaffected either way. TunnelEvent::Data now carries sub_headers alongside data, and I added send_data_with_sub_headers beside the existing send_data, which now delegates to it with an empty vec, so none of the existing call sites needed to change. I marked TunnelEvent #[non_exhaustive] while touching it, since that's exactly the case the second concern raised. I added two regression tests: one confirming a received sub-header now surfaces through the event, one confirming a sent sub-header round-trips through the peer's handle_pdu.

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+                    let hr = 0x80070005u32;
+                    self.enqueue_create_response(hr);
+                    self.events.push_back(TunnelEvent::Failed { hr_response: hr });
+                }
+            }
+
+            // Either side in Established receives Data
+            (TunnelState::Established, _, TunnelPdu::Data(data_pdu)) => {
+                self.events.push_back(TunnelEvent::Data(data_pdu.higher_layer_data));
+            }
```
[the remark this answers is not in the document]
glamberson: I checked whether this would collide with the not-yet-filed driver PR: its write path bypasses send_data entirely by constructing TunnelData directly, so it's unaffected either way. TunnelEvent::Data now carries sub_headers alongside data, and I added send_data_with_sub_headers beside the existing send_data, which now delegates to it with an empty vec, so none of the existing call sites needed to change. I marked TunnelEvent #[non_exhaustive] while touching it, since that's exactly the case the second concern raised. I added two regression tests: one confirming a received sub-header now surfaces through the event, one confirming a sent sub-header round-trips through the peer's handle_pdu.

crates/ironrdp-rdpemt/src/pdu/create_request.rs
glamberson: Added the check to both decoders: CreateRequest against the fixed 24 bytes, CreateResponse against the fixed 4. I added a regression test per PDU that flips PayloadLength's low byte in the spec example wire capture while leaving the real body untouched, and confirmed both fail against the code with the check removed.

crates/ironrdp-rdpemt/src/pdu/create_response.rs
glamberson: Added the check to both decoders: CreateRequest against the fixed 24 bytes, CreateResponse against the fixed 4. I added a regression test per PDU that flips PayloadLength's low byte in the spec example wire capture while leaving the real body untouched, and confirmed both fail against the code with the check removed.

crates/ironrdp-rdpemt/src/pdu/create_request.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        let request_id = src.read_u32();
+        let _reserved = src.read_u32();
+        let security_cookie: [u8; SECURITY_COOKIE_LEN] = src.read_array();
+
```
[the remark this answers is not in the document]
glamberson: Added the check to both decoders: CreateRequest against the fixed 24 bytes, CreateResponse against the fixed 4. I added a regression test per PDU that flips PayloadLength's low byte in the spec example wire capture while leaving the real body untouched, and confirmed both fail against the code with the check removed.

crates/ironrdp-rdpemt/src/pdu/create_response.rs
glamberson: Apologies, my previous reply on this thread was misfiled and answered a different finding (payload_length validation). Reopening it now to actually address this one.

I changed is_success to check the severity bit (hr_response & 0x80000000 == 0) instead of exact equality to S_OK, matching the Win32 SUCCEEDED(hr) macro and the spec's normative wording. I added a regression test confirming S_FALSE (0x00000001, severity bit clear) now reads as success, and confirmed it fails against the exact-equality check.

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+            events: VecDeque::new(),
+        }
+    }
+
+    /// Process a received (already TLS-decrypted) PDU.
+    ///
+    /// After calling this, use `poll_event()` to retrieve any events
+    /// and `poll_pdu()` to retrieve any response PDUs.
+    pub fn handle_pdu(&mut self, data: &[u8]) -> Result<(), RdpemtError> {
+        let pdu: TunnelPdu = ironrdp_core::decode(data).map_err(RdpemtError::decode)?;
```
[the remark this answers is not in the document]
glamberson: Checked the driver PR (not yet filed): its read_tunnel_pdu already reads the fixed 4-byte header first, then reads exactly HeaderLength + PayloadLength more bytes off the stream before ever calling handle_pdu, so it already frames one PDU per call rather than handing over a buffer that could span two. Documented that contract explicitly on handle_pdu instead of reworking the signature, since the actual consumer already honors it.

crates/ironrdp-rdpemt/src/pdu/subheader.rs
```diff
@@ -0,0 +1,129 @@
...
+            return Err(ironrdp_core::DecodeError::invalid_field(
+                Self::NAME,
+                "SubHeaderLength",
+                "sub-header length must be at least 2",
+            ));
+        }
+
+        let type_raw = src.read_u8();
+        let sub_header_type = SubHeaderType::from_u8(type_raw)
+            .ok_or_else(|| ironrdp_core::DecodeError::unexpected_message_type(Self::NAME, type_raw))?;
```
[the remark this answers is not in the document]
glamberson: Agreed with the security-first reading, and the pinning test you asked for already exists: decode_rejects_unknown_type in the testsuite covers exactly this case. No code change needed.

crates/ironrdp-rdpemt/src/tunnel.rs
```diff
@@ -0,0 +1,290 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TunnelEvent {
+    /// The tunnel has been established (CreateResponse with S_OK).
+    Established,
+    /// Higher-layer data received through the tunnel.
+    Data(Vec<u8>),
+    /// The tunnel creation failed (non-recoverable).
+    Failed {
+        /// The HRESULT from the CreateResponse, or 0xFFFFFFFF for server-side
+        /// cookie mismatch.
```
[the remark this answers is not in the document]
glamberson: Fixed both. TunnelEvent::Failed now documents the actual 0x80070005 (E_ACCESSDENIED) sentinel instead of the never-occurring 0xFFFFFFFF. The lifecycle diagram now states that client() constructs directly into AwaitingResponse with CreateRequest already queued, rather than implying a Created-to-AwaitingResponse transition on poll_pdu that a client never actually goes through.

glamberson:
I addressed all sixteen review threads from mamoreau and the automated review, thirteen items total.

The one that matters most: this crate defined its own MultitransportRequest/MultitransportResponse types duplicating what ironrdp-pdu::rdp::multitransport already provides, minus the mandatory security header, so encoding one alone onto the wire would have produced an invalid PDU. Removed both, along with their tests. Also fixed a real encoding bug where TunnelHeader's length field could disagree with its actual sub-headers, another where SubHeaderLength silently wrapped past 255, switched the cookie comparison to constant-time, widened the create-response success check to match the spec's actual wording instead of an exact S_OK match, and added the missing PayloadLength validation on both create PDUs.

Sub-headers now surface on receive and can be attached on send, closing the gap that made sideband bandwidth-measurement traffic (MS-RDPBCGR 2.2.14) unreachable through this API. Removed six unreachable error-kind variants and their constructors.

Three items needed no code change: I reproduced the theoretical Established-before-CreateResponse race in a test and confirmed the shared FIFO queue already prevents it, locked that in as a regression test; confirmed the existing strict-rejection test already pins the unknown-SubHeaderType choice the automated review asked to have pinned; and documented rather than reworked the one-PDU-per-call contract on handle_pdu, since the driver code that will consume this already frames PDUs correctly before calling it.

#1627 merged today. This one and the rest of the stack follow the same pattern.

glamberson:
Noted on the size threshold. This one is almost entirely net-new code for a new crate (2171+/0-, 22 files), not a sprawling change across existing systems, so I don't think splitting it serves review quality here the way it would for a refactor. mamoreau's review above already covered the substantive ground the automated pass would otherwise add. Happy to reconsider if a maintainer would rather see it split regardless.

mamoreau-devolutions:
@glamberson yeah agreed on the size limitation, we'll need to find a way to bypass it when it doesn't make sense

PR #1627, opened 2026-08-11
feat(rdpeudp): add the RDP-UDP crate and the v1 handshake PDUs

glamberson:
# feat(rdpeudp): add the RDP-UDP crate and the v1 handshake PDUs

Second of six, and the first of the RDP-UDP transport @mamoreau-devolutions
asked for on #140. Independent of the RDPEMT PR; the two can land in either
order.

Adds `ironrdp-rdpeudp` with the structures [MS-RDPEUDP] section 2.2 defines for
the three-way handshake: the FEC header and flags, the SYN and extended SYN
payloads that negotiate initial sequence numbers, MTU and protocol version, the
ACK vector with its run-length encoding, the AckOfAcks header, the correlation
ID payload, and the composite datagram that assembles them in the order 2.2.2
requires.

## The convention that matters when reading this

Everything here is **big-endian**, and the diagrams number bits **most
significant first**. Section 2.2: "all of the messages written to the network or
read from the network MUST be in network byte order."

MS-RDPEUDP2, which the data transfer switches to, is little-endian and numbers
diagram bits least significant first. The two documents are opposite on both
counts. That is why the two wire formats are split across two PRs rather than
reviewed together.

## Two readings I would particularly like checked

**The ACK flag does not always announce an ACK vector.** 2.2.2.1 defines it that
way, but 3.1.5.1.3 builds the SYN+ACK as a plain SYN with the flag set and
`snSourceAck` filled in, and the capture in 4.1.2 confirms it: `uFlags` is
`0x0005` and the SYNDATA payload follows the 8-byte header directly, with no
vector between them. So on a SYN the flag says only that `snSourceAck` is
meaningful. `encode` refuses a SYN carrying a vector rather than writing bytes a
peer would read as the start of SYNDATA.

**The ACK vector's element layout and padding** were settled against the ACK
packet capture in 4.2.3 (`00 01 04 00`): a big-endian `uAckVectorSize`, the
state in the top two bits of each element, and padding out to a DWORD boundary.

Both captures are decoded byte by byte in `pdu_v1_datagram.rs` rather than only
round tripped against our own encoder. A mistake made symmetrically in encode
and decode is invisible to a round trip, so the captures are the only thing that
can catch it.

## What is not here

No connection state machine, no v2 data transfer format. Those are the next two
PRs in the stack.

## Test plan

`cargo xtask check fmt/lints/tests/typos/locks`

43 tests in `ironrdp-testsuite-core`, including the section 4.1.1, 4.1.2 and
4.2.3 captures decoded whole.


crates/ironrdp-rdpeudp/src/pdu/v1_ack.rs
```diff
@@ -0,0 +1,497 @@
...
+
+        // Two size bytes are not on a DWORD boundary, so two pad bytes follow.
+        assert_eq!(encoded.as_slice(), &[0x00, 0x00, 0x00, 0x00]);
+        assert_eq!(header.size(), 4);
+    }
+
+    #[test]
+    fn encode_ack_vector_with_elements() {
+        let header = V1AckVectorHeader {
+            elements: vec![
```
mamoreau-devolutions, line 369: `cargo test -p ironrdp-rdpeudp` cannot compile this test module because `vec!` is not in scope in this `no_std` crate. Import `alloc::vec` alongside `Vec` so the crate's unit tests can run.
> [!NOTE]

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
```diff
@@ -0,0 +1,299 @@
...
+        let udp_ver_raw = src.read_u16_be();
+        let udp_ver = UdpVersion::from_u16(udp_ver_raw).ok_or_else(|| {
+            ironrdp_core::invalid_field_err!("RDPUDP_SYNDATAEX_PAYLOAD", "uUdpVer", "unknown protocol version")
+        })?;
+
+        // cookieHash is present only for v3 and when there are enough remaining bytes.
+        // Per spec: MUST be present in client→server SYN with v3, MUST NOT be present otherwise.
+        // We detect its presence by checking for remaining bytes because the caller may not
+        // know the role (client vs server) at the PDU layer.
+        let cookie_hash = if udp_ver == UdpVersion::V3 && src.len() >= Self::COOKIE_HASH_SIZE {
```
mamoreau-devolutions, line 309: This cannot use remaining length to decide whether `cookieHash` is present. SYN and SYN+ACK datagrams are padded to the negotiated MTU, so a v3 server SYN+ACK will decode 32 padding bytes as a hash even though MS-RDPEUDP forbids `cookieHash` in that direction. Thread endpoint direction into decoding, or parse the hash only in a caller that knows it is handling a client SYN.
> [!NOTE]

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
```diff
@@ -0,0 +1,299 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl Encode for SynDataPayload {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
+        ironrdp_core::ensure_fixed_part_size!(in: dst);
+
+        dst.write_u32_be(self.initial_sequence_number);
```
mamoreau-devolutions, line 78: Encoding writes arbitrary public MTU values, while decoding rejects values outside `1132..=1232`. Validate both MTUs before writing so callers cannot emit a non-conformant SYN/SYN+ACK that this crate itself would reject on decode.
> [!NOTE]

crates/ironrdp-testsuite-core/tests/rdpeudp/pdu_v1_datagram.rs
```diff
@@ -0,0 +1,444 @@
...
+        ack_of_acks: None,
+        syn_data: None,
+        correlation_id: None,
+        syn_data_ex: None,
+    };
+    let mut encoded = encode_vec(&datagram).expect("encode");
+
+    // Manually set DATA flag in the wire bytes
+    // Flags are at offset 6..8 in FecHeader (after snSourceAck(4) + windowSize(2))
+    let flags_raw = u16::from_le_bytes([encoded[6], encoded[7]]);
```
mamoreau-devolutions: `FecHeader` flags are big-endian. This little-endian conversion writes wire bytes for `CORRELATION_ID` (`0x0800`), not `DATA` (`0x0008`), so the test passes on an unrelated truncated-correlation error rather than the DATA rejection. Use big-endian conversion and assert the expected error path.
> [!NOTE]

crates/ironrdp-rdpeudp/src/pdu/v1_header.rs
```diff
@@ -0,0 +1,69 @@
+//! RDPUDP_FEC_HEADER: the mandatory header for every v1 datagram.
+//!
+//! MS-RDPEUDP Section 2.2.2.1.
+//! Wire layout: `snSourceAck(4)` + `uReceiveWindowSize(2)` + `uFlags(2)` = 8 bytes, little-endian.
```
mamoreau-devolutions: The header codec uses `read_u*_be`/`write_u*_be`, and MS-RDPEUDP requires network byte order. Change this to big-endian (or network byte order) to avoid contradicting the implementation and protocol.
> [!NOTE]

crates/ironrdp-rdpeudp/src/error.rs
```diff
@@ -0,0 +1,110 @@
+//! Error types for the RDP-UDP connection state machine.
+//!
+//! These surface at the public API boundary of [`RdpeudpConnection`], and carry
+//! both wire-level decode and encode failures from `ironrdp-core` and
+//! protocol-level failures from the state machine itself.
+//!
+//! [`RdpeudpConnection`]: crate::RdpeudpConnection
```
mamoreau-devolutions: `RdpeudpConnection` is not defined or re-exported by this crate, so this link and the `poll_transmit` link below render as unresolved rustdoc warnings. Remove the links or defer this connection-state-machine API documentation until that API exists.
> [!NOTE]

crates/ironrdp-rdpeudp/README.md
```diff
@@ -0,0 +1,20 @@
...
+Reliable UDP transport implemented as described in [MS-RDPEUDP] and
+[MS-RDPEUDP2].
+
+The two documents divide the work. [MS-RDPEUDP] defines the handshake that
+opens a connection and negotiates a protocol version; from version 3 onward
+that handshake leads into the data transfer defined by [MS-RDPEUDP2], which is
+the one this crate implements. Note that the documents take opposite byte
+orders, and number the bits in their diagrams in opposite directions.
+
+Sans-I/O: the state machine is driven by datagrams and by a caller-supplied
```
mamoreau-devolutions: This crate currently exposes only PDU codecs; it has no state machine, caller-supplied instant, or transmit API. Reword this as planned design or scope the README to the implemented handshake-PDU layer so users are not promised unavailable functionality.
> [!NOTE]

mamoreau-devolutions (changes requested):
Please address the inline protocol, build, test, and documentation defects before merging.
> [!NOTE]

crates/ironrdp-rdpeudp/README.md
CBenoit, line 1: Please, can you update this file using the following skills (in the .agents folder):
- docs-compressor
- docs-writer
- markdown-writer
- prose-verifier
- prose-writer

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
glamberson: Fixed. validate_mtu is now generic over the error type, and encode calls it for both fields before writing, so this crate's own encoder can't produce a SYN or SYN+ACK its own decoder would reject. Added two encode-side tests mirroring the existing decode-side ones, and confirmed both fail against the prior code.

crates/ironrdp-rdpeudp/src/error.rs
glamberson: Removed error.rs and its lib.rs re-export entirely. RdpeudpError/RdpeudpErrorKind/RdpeudpErrorExt are used nowhere outside that file, so this PR has no consumer for them. It belongs in the state machine PR that will actually construct and return it. Also dropped the now-unused ironrdp-error dependency from Cargo.toml.

crates/ironrdp-rdpeudp/src/error.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[track_caller]
+    fn connection_closed(context: &'static str) -> Self {
+        Self::new(context, RdpeudpErrorKind::ConnectionClosed)
+    }
+
+    #[track_caller]
+    fn invalid_packet(context: &'static str, reason: &'static str) -> Self {
+        Self::new(context, RdpeudpErrorKind::InvalidPacket { reason })
+    }
+}
```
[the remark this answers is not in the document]
glamberson: Removed error.rs and its lib.rs re-export entirely. RdpeudpError/RdpeudpErrorKind/RdpeudpErrorExt are used nowhere outside that file, so this PR has no consumer for them. It belongs in the state machine PR that will actually construct and return it. Also dropped the now-unused ironrdp-error dependency from Cargo.toml.

crates/ironrdp-rdpeudp/src/pdu/v1_ack.rs
glamberson: Added use alloc::vec; gated the same way as the existing Vec import (#[cfg(not(feature = "std"))]), so it's in scope under the default no_std build but doesn't produce a redundant-import warning when std is on. cargo test -p ironrdp-rdpeudp now builds and its 20 tests pass; confirmed the prior code failed with "cannot find macro vec in this scope".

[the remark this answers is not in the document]
glamberson: Added use alloc::vec; gated the same way as the existing Vec import (#[cfg(not(feature = "std"))]), so it's in scope under the default no_std build but doesn't produce a redundant-import warning when std is on. cargo test -p ironrdp-rdpeudp now builds and its 20 tests pass; confirmed the prior code failed with "cannot find macro vec in this scope".

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
glamberson: Fixed by threading direction into decoding, as suggested. V1Datagram::decode already distinguishes a client SYN from a server SYN+ACK via header.flags (SYN without ACK vs SYN+ACK, the same test the ACK-vector gate above it uses), so it now calls a new SynDataExPayload::decode_directional(src, is_client_syn) instead of the generic Decode impl. cookieHash is read only when is_client_syn && udp_ver == V3, exact bytes, nothing inferred from what remains. The standalone Decode impl is unchanged (still correct for exact-length buffers, keeps its existing tests), now documented as not for use on bytes carved out of a padded datagram.

Added a regression test constructing a v3 SYN+ACK, encoding it, padding it to the MTU, and decoding: confirmed it fabricates a 32-byte all-zero hash on the prior code and correctly reads None now.

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
```diff
@@ -0,0 +1,299 @@
...
+        let udp_ver = UdpVersion::from_u16(udp_ver_raw).ok_or_else(|| {
+            ironrdp_core::invalid_field_err!("RDPUDP_SYNDATAEX_PAYLOAD", "uUdpVer", "unknown protocol version")
+        })?;
+
+        // cookieHash is present only for v3 and when there are enough remaining bytes.
+        // Per spec: MUST be present in client→server SYN with v3, MUST NOT be present otherwise.
+        // We detect its presence by checking for remaining bytes because the caller may not
+        // know the role (client vs server) at the PDU layer.
+        let cookie_hash = if udp_ver == UdpVersion::V3 && src.len() >= Self::COOKIE_HASH_SIZE {
+            Some(src.read_array::<32>())
```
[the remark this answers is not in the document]
glamberson: Fixed by threading direction into decoding, as suggested. V1Datagram::decode already distinguishes a client SYN from a server SYN+ACK via header.flags (SYN without ACK vs SYN+ACK, the same test the ACK-vector gate above it uses), so it now calls a new SynDataExPayload::decode_directional(src, is_client_syn) instead of the generic Decode impl. cookieHash is read only when is_client_syn && udp_ver == V3, exact bytes, nothing inferred from what remains. The standalone Decode impl is unchanged (still correct for exact-length buffers, keeps its existing tests), now documented as not for use on bytes carved out of a padded datagram.

Added a regression test constructing a v3 SYN+ACK, encoding it, padding it to the MTU, and decoding: confirmed it fabricates a 32-byte all-zero hash on the prior code and correctly reads None now.

crates/ironrdp-rdpeudp/src/pdu/v1_ack.rs
```diff
@@ -0,0 +1,497 @@
...
+
+    /// Bytes of padding needed so the structure ends on a DWORD boundary.
+    fn padding_len(element_count: usize) -> usize {
+        (Self::FIXED_PART_SIZE + element_count).next_multiple_of(4) - Self::FIXED_PART_SIZE - element_count
+    }
+}
+
+impl Encode for V1AckVectorHeader {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
+        let count: u16 = ironrdp_core::cast_length!("RDPUDP_ACK_VECTOR_HEADER", "uAckVectorSize", self.elements.len())?;
```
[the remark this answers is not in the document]
glamberson: Enforced on both sides. Encode now rejects more than 2048 elements before writing; decode rejects a claimed count above 2048 right after reading it, before allocating. Added tests for both directions, plus one confirming exactly 2048 still encodes (the spec ceiling, not an overflow). The decode test carries enough element and padding bytes to isolate the new check from the pre-existing insufficient-bytes rejection; confirmed both new tests fail without the fix.

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// [MS-RDPEUDP2]. 1.3.2.2 states it from the other direction, that the
+    /// MS-RDPEUDP data transfer messages "MUST be used only when the version
+    /// negotiated in the UDP connection initialization phase is version 1 or
+    /// version 2".
+    pub fn uses_v2_wire_format(self) -> bool {
+        matches!(self, Self::V3)
+    }
+
+    /// Minimum retransmission timeout for this version.
+    pub fn min_retransmit_ms(self) -> u32 {
```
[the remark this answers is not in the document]
glamberson: Removed both accessors and the test that checked them. Neither has a consumer in this PR, and checking the MS-RDPEUDP2 corpus, the V3 figures were more than uncertain: 3.1.1.2.3 defaults the ack-delay to half the round trip time, not a fixed constant, so a hardcoded value would have been wrong regardless of which number I picked. Logged the same problem against the not-yet-filed connection PR, which already hardcodes a 50ms ack-delay default citing MS-RDPEUDP2 incorrectly.

crates/ironrdp-rdpeudp/src/pdu/v1_syn.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn decode(src: &mut ReadCursor<'_>) -> DecodeResult<Self> {
+        ironrdp_core::ensure_fixed_part_size!(in: src);
+
+        let syn_ex_flags_raw = src.read_u16_be();
+        let syn_ex_flags = SynExFlags::from_bits_truncate(syn_ex_flags_raw);
+
+        let udp_ver_raw = src.read_u16_be();
+        let udp_ver = UdpVersion::from_u16(udp_ver_raw).ok_or_else(|| {
+            ironrdp_core::invalid_field_err!("RDPUDP_SYNDATAEX_PAYLOAD", "uUdpVer", "unknown protocol version")
+        })?;
```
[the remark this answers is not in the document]
glamberson: Deliberate, for now. Checked MS-RDPEUDP 1.7: the negotiation MUST clause only requires supporting versions at or below what a peer advertises, not understanding anything above it, so failing to interpret an unrecognized future value isn't itself non-conformant. What is a real gap is that this crate can't currently participate in that at all, since the unrecognized value never reaches a caller: UdpVersion is closed and decode hard-fails. Fixing that means widening the enum, which is a public API decision I don't want to make inside this PR. Logged it as a requirement for whichever PR first needs interop with an unrecognized version.

crates/ironrdp-testsuite-core/tests/rdpeudp/pdu_v1_datagram.rs
glamberson: Fixed the byte order: the test patched the flags little-endian, but the wire format is big-endian, so setting DATA (0x0008) landed on the wire as 0x0800 (CORRELATION_ID) instead. Confirmed by disabling the actual DATA/FEC rejection checks in V1Datagram::decode: the old test still passed, because it was really exercising CORRELATION_ID-without-payload truncation, not the rejection it named. Fixed to big-endian, and added the missing FEC counterpart. Both new/fixed tests now fail when the rejection checks are disabled, confirming they test what they claim to.

crates/ironrdp-testsuite-core/tests/rdpeudp/pdu_v1_datagram.rs
```diff
@@ -0,0 +1,444 @@
...
+    let modified = flags_raw | V1Flags::DATA.bits();
+    let [low, high] = modified.to_le_bytes();
+    encoded[6] = low;
+    encoded[7] = high;
```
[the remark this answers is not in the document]
glamberson: Fixed the byte order: the test patched the flags little-endian, but the wire format is big-endian, so setting DATA (0x0008) landed on the wire as 0x0800 (CORRELATION_ID) instead. Confirmed by disabling the actual DATA/FEC rejection checks in V1Datagram::decode: the old test still passed, because it was really exercising CORRELATION_ID-without-payload truncation, not the rejection it named. Fixed to big-endian, and added the missing FEC counterpart. Both new/fixed tests now fail when the rejection checks are disabled, confirming they test what they claim to.

crates/ironrdp-rdpeudp/README.md
glamberson: Rewritten to describe what's here today: v1 handshake PDU types only, no state machine, no MS-RDPEUDP2 data transfer. Kept the byte-order/bit-numbering warning since that's a real, still-relevant gotcha for anyone touching both spec documents. Word count dropped from about 114 to about 84.

crates/ironrdp-rdpeudp/README.md
```diff
@@ -0,0 +1,20 @@
...
+instant, and returns the datagrams it wants sent. It performs no I/O and reads
+no clock, so it can be driven by any runtime.
```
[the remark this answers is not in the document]
glamberson: Rewritten to describe what's here today: v1 handshake PDU types only, no state machine, no MS-RDPEUDP2 data transfer. Kept the byte-order/bit-numbering warning since that's a real, still-relevant gotcha for anyone touching both spec documents. Word count dropped from about 114 to about 84.

crates/ironrdp-rdpeudp/README.md
glamberson: Rewrote it through the chain: markdown-writer and prose-writer for the draft, docs-compressor and prose-verifier as review criteria. One sentence per source line, no mid-sentence hard-wraps, shallowest heading level, and the content itself now matches what the crate actually contains rather than the planned end state.

crates/ironrdp-rdpeudp/src/pdu/v1_header.rs
glamberson: Fixed. The code was already correct (write_u32_be/write_u16_be, network byte order); the module doc just said the opposite. Swept the rest of the crate for the same leftover and found no other instances.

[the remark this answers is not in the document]
glamberson: Fixed. The code was already correct (write_u32_be/write_u16_be, network byte order); the module doc just said the opposite. Swept the rest of the crate for the same leftover and found no other instances.

crates/ironrdp-rdpeudp/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+[features]
+# Matches `ironrdp-pdu` and `ironrdp-rdpeusb`: `no_std` unless a consumer asks
+# for `std`. The `alloc` crate is not optional here, because the variable-length
+# payloads this protocol carries are heap-allocated, so there is no useful build
+# without it and no `alloc` feature is offered.
+default = []
+std = ["ironrdp-core/std", "ironrdp-error/std"]
+# Structure-aware fuzzing support, mirroring `ironrdp-pdu`.
+# `arbitrary` needs the standard library, so structure-aware fuzzing implies `std`.
+arbitrary = ["dep:arbitrary", "std", "bitflags/arbitrary"]
```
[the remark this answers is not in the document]
glamberson: Kept arbitrary and the derives: checked the fuzz PR that follows this one, and it's a real, already-written consumer, not abandoned surface. Fixed the dangling part instead. The test doc comment crediting rdpeudp_pdu_round_trip now says that oracle is filed separately rather than implying it's present in this tree.

crates/ironrdp-rdpeudp/src/pdu/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+        if self.syn_data_ex.is_some() {
+            flags |= V1Flags::SYNEX;
+        }
+
+        flags
+    }
+}
+
+impl Encode for V1Datagram {
+    fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
```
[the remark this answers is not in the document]
glamberson: Documented on the type: encode doesn't pad, and the caller sending these bytes on the wire is responsible for padding to min(upstream_mtu, downstream_mtu) before transmitting. Checked both downstream PRs that could plausibly own this instead: neither the connection state machine nor the async driver currently pads either. Logged that as a real, verified gap rather than leaving it implicit.

glamberson:
I was away from IronRDP most of this week and am now working through the backlog, starting with this PR.

Addressed all review threads from mamoreau, CBenoit, and the automated review, twelve items total. The one that matters most: cookieHash presence in the SYN extension was inferred from remaining buffer length, which fabricated a 32 byte hash out of MTU padding on any v3 SYN+ACK. Fixed by threading the datagram's own direction (already available via the header flags) into decoding instead of guessing from what bytes happen to remain. Also fixed a no_std build break, a test that patched flags little-endian and silently exercised the wrong rejection path, added encode-side validation matching what decode already required for MTU and ACK vector size, and removed a public error API this PR never uses.

The README is rewritten to describe what this crate actually contains today rather than the planned end state, run through the repo's own docs-writer and docs-compressor skill chain per CBenoit's request.

Two gaps surfaced that belong to PRs further down the stack rather than here: the connection state machine's ack-delay default doesn't match what MS-RDPEUDP2 actually specifies, and nothing in the local stack currently pads a SYN or SYN+ACK to the negotiated MTU before it would go on the wire. Both are tracked for those PRs.

mamoreau-devolutions (approved):

mamoreau-devolutions:
Thanks @glamberson I've just merged this PR

PR #1689, opened 2026-08-17
fix(pdu)!: accept the full documented range of Client Core Data keyboardType values

glamberson:
## Summary

KeyboardType (TS_UD_CS_CORE's keyboardType field, MS-RDPBCGR 2.2.1.3.2) was a closed enum covering discriminants 1 through 7, transcribed from the field's other documentation in 2.2.7.1.6 (TS_INPUT_CAPABILITYSET), which omits value 8 (Korean keyboard). The table this type actually decodes against, 2.2.1.3.2, documents 1 through 8. Any client outside the narrower range, including every genuine Korean keyboard, had its whole Client Core Data parse hard rejected before the connection started.

Converted KeyboardType to a repr(transparent) struct KeyboardType(pub u32) with named constants for the eight documented values, mirroring RdpVersion one field above it in the same struct, and made decode infallible at all three sites that previously disagreed with each other: gcc::core_data::client hard rejected, rdp::capability_sets::input silently dropped to None, and ironrdp-activex's COM setter returned E_INVALIDARG.

FreeRDP hit the identical gap in its own documentation-only enum until a two-line fix (FreeRDP#11035). Neither FreeRDP nor xrdp gate connection admission on this field at all; both decode it as a raw integer. Windows' own GetKeyboardType additionally documents 0x51 for generic HID keyboards, which a narrower fix adding only Korean would not survive.

Marked as breaking since KeyboardType's public shape changes for any downstream consumer of ironrdp-pdu outside this workspace. Note that cargo-semver-checks in this repo's PR automation only runs against the facade ironrdp crate, which will not see a break confined to ironrdp-pdu's own API, so the automated breaking-change label may not apply here even though this is a real one.

## Validation

cargo xtask check fmt/lints/tests/typos/locks all pass. Extended ironrdp-activex's existing inline COM boundary test to cover value 8 round tripping and a negative value still being rejected.


crates/ironrdp-pdu/src/gcc/core_data/client.rs
```diff
@@ -156,10 +156,7 @@ impl<'de> Decode<'de> for ClientCoreData {
             .trim_end_matches('\u{0}')
             .into();
 
-        let keyboard_type = src
-            .read_u32()
-            .pipe(KeyboardType::from_u32)
-            .ok_or_else(|| invalid_field_err!("keyboardType", "invalid keyboard type"))?;
+        let keyboard_type = src.read_u32().pipe(KeyboardType);
```
mamoreau-devolutions, line 159: Please add a deterministic Core Data decode/re-encode regression test for `keyboardType = 8` and `0x51` here. The touched PDU fixtures still exercise only `4`, while the ActiveX test covers only the COM getter/setter, so neither would catch the former closed-enum validation being reintroduced at this wire boundary. MS-RDPBCGR section 2.2.1.3.2 documents both values for Client Core Data.

> [!NOTE]

glamberson: Added a decode/re-encode round-trip test for KOREAN (8) and 0x51 in `ironrdp-testsuite-core/tests/pdu/gcc.rs`, next to the existing Core Data fixtures. Verified it catches the regression: mutated the decode site back to reject keyboardType 8, confirmed the test fails, reverted.

mamoreau-devolutions (approved):

PR #1690, opened 2026-08-17
test(core): add a rdpeusb (URBDRC) decode fuzz target

clintcan:
## Summary

`ironrdp-rdpeusb` has no fuzz coverage, even though it decodes
attacker-controlled **client→server** PDUs off two DVC channels. The device
channel in particular carries URB completions, IO-control completions and
interface-info results whose **length-prefixed payloads must be bounds-checked
before they're read** — historically a good place for unchecked-read panics.
(A downstream server built on `ironrdp-server`, [macrdp](https://github.com/clintcan/macrdp),
hit exactly that class of panic in these decoders in the past, which is what
motivated adding coverage here.)

This adds a fuzz target for the two top-level client PDU families a server
decodes off the wire.

## What's added

- **Oracle** `oracles::rdpeusb_decode` — decodes the two client→server entry
  points via `ironrdp_core::decode`:
  - `UrbdrcClientControlPdu` (main-channel family)
  - `UrbdrcClientDevicePdu<Raw>` (per-device family — recursively exercises the
    URB / IO-control / interface-info result payloads)
- **Fuzz target** `fuzz/fuzz_targets/rdpeusb_decode.rs` (auto-discovered by
  `cargo xtask fuzz list`, so it joins the CI matrix with no further wiring).
- **Regression check** `check_rdpeusb_decode` in `fuzz_regression.rs`, plus the
  `rdpeusb_decode/` corpus folder, so any crash found becomes a permanent
  replay test.

## Verification

- `cargo build -p ironrdp-fuzzing` — the oracle compiles against the real API.
- `cargo test -p ironrdp-testsuite-core check_rdpeusb_decode` — passes.
- `cargo fmt --all -- --check` (stable + nightly) clean; `cargo clippy -p ironrdp-fuzzing` clean.
- `cargo xtask fuzz list` lists `rdpeusb_decode`.

Purely additive (a new oracle, target, regression check + the `ironrdp-rdpeusb`
dev-dependency of `ironrdp-fuzzing`); no existing code changes. Complements the
in-flight rdpeusb server work by giving the crate's decode surface ongoing
fuzzing.

### DIFF
--- a/crates/ironrdp-fuzzing/src/oracles/mod.rs
+++ b/crates/ironrdp-fuzzing/src/oracles/mod.rs
@@ -577,6 +577,40 @@ pub fn cliprdr_channel_process(input: &[u8]) {
     let _ = cliprdr.process(input);
 }
 
+/// The URBDRC (MS-RDPEUSB / USB redirection) client→server PDUs a server
+/// decodes off the wire. `UrbdrcClientControlPdu` is the main-channel family;
+/// `UrbdrcClientDevicePdu<Raw>` is the per-device family, which carries the URB
+/// completions, IO-control completions and interface-info results.
+///
+/// Decoding `UrbdrcClientDevicePdu<Raw>` only slurps each URB *result* body into
+/// `Raw` — the operation-specific reinterpretation of those length-prefixed
+/// payloads (`SELECT_CONFIGURATION` / `SELECT_INTERFACE` / interface-info / isoch
+/// results, the historical home of unchecked-read panics) happens later in the
+/// stateful server handlers via `into_expected`, so `<Raw>` alone never reaches
+/// them. Those result decoders are all public, so we also drive them directly on
+/// the raw input to keep the whole bounds-checked decode surface fuzzed.
+pub fn rdpeusb_decode(data: &[u8]) {
+    use ironrdp_core::{ReadCursor, decode};
+    use ironrdp_rdpeusb::pdu::completion::ts_urb_result::{
+        Raw, TsUrbGetCurrFrameNumResult, TsUrbIsochTransferResult, TsUrbSelectConfigResult, TsUrbSelectInterfaceResult,
+        TsUsbdInterfaceInfoResult, TsUsbdPipeInfoResult,
+    };
+    use ironrdp_rdpeusb::pdu::{UrbdrcClientControlPdu, UrbdrcClientDevicePdu};
+
+    // Top-level client→server PDU families, off the two DVC channels.
+    let _ = decode::<UrbdrcClientControlPdu>(data);
+    let _ = decode::<UrbdrcClientDevicePdu<Raw>>(data);
+
+    // The length-prefixed URB / interface-info *result* payload decoders the
+    // per-device family reinterprets `Raw` into once correlated with a request.
+    let _ = TsUrbSelectConfigResult::decode(&mut ReadCursor::new(data));
+    let _ = TsUrbSelectInterfaceResult::decode(&mut ReadCursor::new(data));
+    let _ = TsUrbGetCurrFrameNumResult::decode(&mut ReadCursor::new(data));
+    let _ = TsUrbIsochTransferResult::decode(&mut ReadCursor::new(data));
+    let _ = decode::<TsUsbdInterfaceInfoResult>(data);
+    let _ = decode::<TsUsbdPipeInfoResult>(data);
+}
+
 /// Minimal backend for fuzzing that enables file transfer capabilities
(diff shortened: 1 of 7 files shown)


crates/ironrdp-testsuite-core/test_data/fuzz_regression/rdpeusb_decode/seed-empty.bin
CBenoit, line 1: Do we really need an empty seed here?

CBenoit (approved):
LGTM! Thank you for this. I left a question-comment, and Copilot also left an interesting comment. Could you consider it before we merge?

crates/ironrdp-fuzzing/src/oracles/mod.rs
```diff
@@ -577,6 +577,20 @@ pub fn cliprdr_channel_process(input: &[u8]) {
...
+/// `UrbdrcClientDevicePdu<Raw>` is the per-device family, which carries the URB
+/// completions, IO-control completions and interface-info results whose
+/// length-prefixed payloads must be bounds-checked before they are read.
+pub fn rdpeusb_decode(data: &[u8]) {
+    use ironrdp_core::decode;
+    use ironrdp_rdpeusb::pdu::completion::ts_urb_result::Raw;
+    use ironrdp_rdpeusb::pdu::{UrbdrcClientControlPdu, UrbdrcClientDevicePdu};
+
+    let _ = decode::<UrbdrcClientControlPdu>(data);
+    let _ = decode::<UrbdrcClientDevicePdu<Raw>>(data);
```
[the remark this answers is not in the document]
clintcan: Good catch — you're right, and this is exactly the class macrdp hit historically. `UrbdrcClientDevicePdu<Raw>` only slurps each URB result body into `Raw`; the length-prefixed reinterpretation happens later in the stateful handlers via the `pub(crate) into_expected`, so `<Raw>` alone never reaches those decoders.

Fixed in the latest push: the oracle now also drives the public result decoders directly on the raw input — `TsUrbSelectConfigResult`, `TsUrbSelectInterfaceResult`, `TsUrbGetCurrFrameNumResult`, `TsUrbIsochTransferResult` (inherent `decode`), plus `TsUsbdInterfaceInfoResult` / `TsUsbdPipeInfoResult` (via `Decode`). Those are the length-prefixed / count-prefixed reads (`TsUsbdInterfaceInfoResult`'s `Length`-slice, `TsUrbSelectConfigResult`'s `NumInterfaces` loop, `TsUrbIsochTransferResult`'s `NumberOfPackets` loop) that are the motivating attack surface. Driving the stateful completion path would also work but needs the request-ID correlation state, which doesn't fit a stateless byte-slice target — so I decode the payload types directly instead.

crates/ironrdp-testsuite-core/test_data/fuzz_regression/rdpeusb_decode/seed-empty.bin
clintcan: It's needed for a mechanical reason: the `check!` macro does `std::fs::read_dir(folder).unwrap()`, so the corpus folder has to exist in-tree — and git won't track an empty directory, so it needs at least one file. `seed-empty.bin` is the existing convention for a target with no crash reproducer yet: `egfx_round_trip`, `egfx_multi_frame`, `egfx_avc420_decode`, `bulk_decompress_ncrush`, `bulk_decompress_xcrush`, `bulk_round_trip`, and `pdu_round_trip` all carry the same placeholder. Empty input is also a legitimate trivial case (decode-of-nothing must not panic). Happy to drop it and `.gitkeep` instead, or leave the whole regression check out until there's a real crash to replay — whichever you prefer.

CBenoit: I would lean towards .gitkeep, thank you!

clintcan: Done — swapped `seed-empty.bin` → `.gitkeep` (the `check!` macro just needs the corpus dir to exist in-tree; `.gitkeep` reads as the empty-input case exactly the same, and `check_rdpeusb_decode` still passes). Also rebased onto master to clear the conflict with #1707 (its rdpeudp/rdpemt fuzz targets touched the same `fuzz/Cargo.toml` / `ironrdp-fuzzing` dep list — additive, kept both). PR is green + mergeable again; oracle still compiles against the rebased tree (incl. #1683's usb module).

PR #1708, opened 2026-08-19
fix(mstsgu): report EOF and stop polling a completed work task

mamoreau-devolutions:
GwClient polled its background work JoinHandle on every read and write,
which tokio panics on once the task completes, and it never reported
end-of-stream when the gateway stream closed. A caller that holds a read
across the connection lifetime (for example a bidirectional relay) hit
`JoinHandle polled after completion` or hung forever instead of seeing
EOF.

Track work completion with a flag so the handle is polled at most once,
and surface UnexpectedEof once the task has ended and no data remains.


mamoreau-devolutions:
Opened against the wrong head branch (the full stacked branch instead of the focused fix branch). Re-opening from the correct branch.

PR #1731, opened 2026-08-20
fix(egfx): honor the server's AVC_DISABLED during negotiation

hojjatabdollahi:
`AVC_DISABLED` is an inverted-sense flag: a peer sets it to say it cannot handle AVC/H.264. #1490 stopped AND-ing away the *client's* copy, but the server's copy is still dropped whenever the client leaves it clear.

That misreports servers which cannot produce AVC at all, such as one streaming only RemoteFX Progressive: it sets AVC_DISABLED, negotiation clears it, and the capabilities confirmation promises the client a codec that will never arrive.

OR the flag from both sides instead, so it survives when either peer disables it, and cover the server case in the negotiation tests.

For positive flags they are still AND-ed but for AVC_DISABLED, which is negative it is OR-ed so if either side sets it, it's enough.

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -713,7 +713,13 @@ fn negotiate_capabilities(client_caps: &[CapabilitySet], server_caps: &[Capabili
     None
 }
 
-/// Intersect positive flags while preserving the client AVC decoder constraint.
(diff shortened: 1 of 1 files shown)


hojjatabdollahi:
Draft, until I make sure my assumption was correct.

mamoreau-devolutions:
> Draft, until I make sure my assumption was correct.

was your assumption correct in the end?

PR #1742, opened 2026-08-21
feat(cliprdr): add a chunked file-contents fetch primitive

glamberson:
## Summary

- MS-RDPECLIP's file-contents protocol is receiver-driven: fetching a
  file means sending a sequence of byte-range FileContentsRequests and
  reassembling the FileContentsResponses. Nothing in this crate helps
  with that sequencing, so every consumer that wants a whole file ends
  up writing its own request/response loop by hand.
- Adds ChunkedFetch: a small state machine that produces the next
  request to send and consumes each response, tracking offset and
  buffered bytes so the caller doesn't have to. The size phase is
  optional: construct with a known size (the common case, since it's
  usually already available from an earlier FileGroupDescriptorW
  exchange) to skip straight to range requests, or without one to
  query it first.
- Every request carries an explicit clip_data_id, the id
  CliprdrBackend::on_remote_file_list already hands the caller for the
  file's locked snapshot. request_file_contents can auto-fill this from
  current_lock_id when left unset, but a fetch spans multiple requests
  and a FormatList between them can expire the old lock and install a
  new one under the same field, so auto-filling would let a fetch
  already in flight silently retarget to whichever lock happens to be
  current at send time.
- A caller-provided max_total_size bounds how large a file this type
  will try to buffer in memory, checked as soon as the size is known
  (either the caller-supplied total_size, or the peer's own SIZE
  response), independent of what that size claims to be.
- Two defensive cases beyond the happy path: an empty range response
  before the file is complete fails the fetch rather than
  re-requesting the same range forever, since the protocol has no
  "not ready yet" signal that would legitimately produce one, and a
  response longer than what's left is clamped rather than trusted,
  since accepting it as-is would grow the buffer past the file's
  total size.

## Validation

`cargo xtask check fmt/lints/tests/typos/locks` all pass. Tests live
in ironrdp-testsuite-core since this crate sets [lib] test = false.

## Notes

No public API break: this is a new module with entirely new public
types, no existing signature changes.

### DIFF
--- a/crates/ironrdp-cliprdr/src/chunked_fetch.rs
+++ b/crates/ironrdp-cliprdr/src/chunked_fetch.rs
@@ -70,6 +70,11 @@ pub struct ChunkedFetch {
     next_offset: u64,
     state: State,
     awaiting_response: bool,
+    /// `requested_size` of the outstanding `RANGE` request, if any. Set by
+    /// [`Self::next_request`], checked by [`Self::on_response`]: MS-RDPECLIP 2.2.5.3 defines
+    /// `cbRequested` as the maximum a `RANGE` responder may return, so a response longer than
+    /// this is a protocol violation, not extra data to accept.
+    last_requested_size: Option<u32>,
 }
 
 impl ChunkedFetch {
@@ -92,6 +97,7 @@ impl ChunkedFetch {
                 State::Fetching
             },
             awaiting_response: false,
+            last_requested_size: None,
         }
     }
 
@@ -108,6 +114,7 @@ impl ChunkedFetch {
             next_offset: 0,
             state: State::AwaitingSize,
             awaiting_response: false,
+            last_requested_size: None,
         }
     }
 
@@ -149,6 +156,7 @@ impl ChunkedFetch {
                 let requested_size = remaining.min(u64::from(self.chunk_size));
                 // Infallible: bounded above by self.chunk_size (a u32) via the min() just above.
                 let requested_size = u32::try_from(requested_size).unwrap_or(self.chunk_size);
+                self.last_requested_size = Some(requested_size);
 
                 FileContentsRequest {
                     stream_id: self.stream_id,
@@ -189,17 +197,25 @@ impl ChunkedFetch {
             },
             State::Fetching => {
                 let data = response.data();
+                let exceeds_requested = self
+                    .last_requested_size
+                    .is_some_and(|requested| u64::from(requested) < u64::try_from(data.len()).unwrap_or(u64::MAX));
 
                 // A compliant peer never sends an empty range response before the file is
                 // complete (MS-RDPECLIP defines no "not ready yet" signal here). Treating
                 // one as failure rather than InProgress avoids looping forever re-requesting
                 // the same range against a peer that keeps answering with nothing.
-                if data.is_empty() {
+                //
+                // A response longer than cbRequested is a protocol violation, not extra data
+                // to accept: [MS-RDPECLIP] 2.2.5.3 defines cbRequested as the maximum a RANGE
+                // responder may return, and bytes beyond it were never validated as the range
+                // this fetch actually asked for.
+                if data.is_empty() || exceeds_requested {
                     self.state = State::Failed;
                 } else {
-                    // Clamp rather than trust the peer's byte count: a response longer than
-                    // what's left would otherwise grow the buffer past total_size and desync
-                    // next_offset from what was actually requested.
+                    // Still clamp against total_size defensively: a response within
+                    // cbRequested that nonetheless overruns a (possibly attacker-influenced)
+                    // total_size must not grow the buffer past it or desync next_offset.
                     let data_len = u64::try_from(data.len()).unwrap_or(u64::MAX);
                     let remaining = self
                         .total_size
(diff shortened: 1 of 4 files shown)


crates/ironrdp-cliprdr/src/chunked_fetch.rs
```diff
@@ -0,0 +1,237 @@
...
+                        .min(data_len);
+                    // Infallible: just clamped to `data.len()` via `data_len`/`min()` above.
+                    let take = usize::try_from(remaining).unwrap_or(data.len());
+
+                    self.received.extend_from_slice(&data[..take]);
+                    self.next_offset = self.next_offset.saturating_add(remaining);
+
+                    if self.next_offset >= self.total_size.unwrap_or(0) {
+                        self.state = State::Complete;
+                    }
```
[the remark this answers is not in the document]
glamberson: Agreed, and I went further than a doc note. I traced `Cliprdr::process_svc_message`'s response dispatch directly: It does have an exact-length check for SIZE responses, but for RANGE responses it just forwards whatever came back with no length validation, and `FileTransferState` doesn't even store `requested_size`, so there was nothing to check against even if it wanted to. This confirms the gap is pre-existing, not something my diff introduced.

I tightened `ChunkedFetch` itself rather than just updating the doc comment. It now tracks `last_requested_size`, set when `next_request()` builds a RANGE request, and treats a response exceeding it as a protocol violation: It fails the fetch rather than accepting the extra bytes. The concern is not just the byte count. Accepting unrequested bytes means folding data into the buffer that was never validated against the specific range this fetch actually asked for, and this crate has to treat wire input as untrusted. The existing total_size-based clamp stays as a defensive backstop for the case where `on_response` gets called without a prior `next_request()`, a caller-side misuse rather than the normal path.

I updated two tests. The old "oversized response is clamped" test now correctly expects `Failed`, since that response exceeds cbRequested. A new test isolates the backstop clamp on its own by exercising the no-prior-request case directly.

crates/ironrdp-cliprdr/src/chunked_fetch.rs
```diff
@@ -0,0 +1,253 @@
...
+                let requested_size = u32::try_from(requested_size).unwrap_or(self.chunk_size);
+                self.last_requested_size = Some(requested_size);
+
+                FileContentsRequest {
+                    stream_id: self.stream_id,
+                    index: self.file_index,
+                    flags: FileContentsFlags::RANGE,
+                    position: self.next_offset,
+                    requested_size,
+                    data_id: None,
```
mamoreau-devolutions: Please preserve this fetch's `clipDataId` instead of leaving every request unset. `request_file_contents` fills `None` from `current_lock_id`, but a later `FormatList` expires the old lock and clears or replaces `current_lock_id` while deliberately keeping the old lock alive for in-flight downloads. If the clipboard changes between chunks, the next request can target a different snapshot and silently assemble bytes from two files. `ChunkedFetch` should capture the `clip_data_id` passed to `on_remote_file_list` and use it for both SIZE and RANGE requests.

> [!NOTE]

crates/ironrdp-cliprdr/src/chunked_fetch.rs
```diff
@@ -0,0 +1,253 @@
...
+                    let data_len = u64::try_from(data.len()).unwrap_or(u64::MAX);
+                    let remaining = self
+                        .total_size
+                        .unwrap_or(0)
+                        .saturating_sub(self.next_offset)
+                        .min(data_len);
+                    // Infallible: just clamped to `data.len()` via `data_len`/`min()` above.
+                    let take = usize::try_from(remaining).unwrap_or(data.len());
+
+                    self.received.extend_from_slice(&data[..take]);
```
mamoreau-devolutions, line 271: Please add an aggregate size limit or a streaming boundary before appending here. `new_with_size_query` accepts any remote-supplied `u64`, and every response is retained in this `Vec`; a peer can report `u64::MAX` and keep satisfying chunk requests until allocation failure terminates the process. Chunking bounds each response, not total memory. Require a caller-provided maximum and fail before RANGE fetching when it is exceeded, or yield chunks to the caller instead of retaining the whole file.

> [!NOTE]

mamoreau-devolutions (changes requested):
Requesting changes for two correctness issues: each fetch must remain bound to its original clipboard-lock snapshot across `FormatList` changes, and remote-controlled file sizes must not drive unbounded aggregate memory growth. Details are inline.

> [!NOTE]

crates/ironrdp-cliprdr/src/chunked_fetch.rs
glamberson: Added an explicit clip_data_id parameter to both constructors, set on every request instead of leaving data_id unset for request_file_contents to auto-fill. The id comes from CliprdrBackend::on_remote_file_list, so a fetch stays bound to the same snapshot for its whole sequence regardless of what current_lock_id points at by the time each request goes out.

crates/ironrdp-cliprdr/src/chunked_fetch.rs
glamberson: Added a caller-provided max_total_size to both constructors. It's checked as soon as total_size is known, either at construction for the known-size path, or when the SIZE response arrives for the query path, and fails the fetch before any RANGE request is ever issued if exceeded. Clamping against total_size already bounded overrun past a claimed size, but total_size itself is remote-controlled either way (a FileGroupDescriptorW entry, or the peer's own SIZE answer), so that clamp alone never capped how large a claimed size this type would try to satisfy.

mamoreau-devolutions (approved):

PR #1754, opened 2026-08-21
test(core): relocate autodetect's inline tests into testsuite-core

glamberson:
ironrdp-server sets `[lib] test = false`, so `autodetect.rs`'s inline `#[cfg(test)]` module never runs under `cargo test --workspace`. Six of its ten tests duplicate coverage already present in `ironrdp-testsuite-core/tests/server/autodetect.rs`, and a seventh (`sequence_number_wraps`) checks the same wrap-around behavior the existing `sequence_number_wraps_at_u16_max` already covers externally, just via a direct write to the private `next_sequence` field instead of the public API.

This moves the three tests with no external equivalent (`backwards_clock_yields_a_zero_sample`, `backwards_clock_keeps_the_probe_pending`, `an_enormous_gap_clamps_to_u32_max`) into testsuite-core and deletes the now-fully-redundant inline module. All ten tests' behavior is preserved: three run under CI for the first time, seven were already covered by the external suite.


CBenoit (approved):
Thank you, LGTM!

PR #1755, opened 2026-08-21
test(extra): relocate CLI argument-parsing tests into testsuite-extra

glamberson:
ironrdp-agent sets `[lib] test = false`, so `cli.rs`'s inline `#[cfg(test)]` tests never run under `cargo test --workspace`. This moves the five tests that exercise only the public `Cli` surface (clap's `Parser`/`CommandFactory` traits, no private fields or enum variants) into `ironrdp-testsuite-extra/tests/agent.rs`, alongside the existing `remote_exit_status` coverage for the same module.

`sandbox.rs` and `sandbox_grpc.rs` stay inline: both are declared `pub(crate)` in `lib.rs`, so every item inside them is unreachable from an external crate regardless of its own visibility. The remaining `cli.rs` tests stay inline for the same reason, touching private `Command`/`SandboxCommand` enum variants or private helper functions.

### DIFF
--- a/crates/ironrdp-testsuite-extra/Cargo.toml
+++ b/crates/ironrdp-testsuite-extra/Cargo.toml
@@ -24,6 +24,7 @@ harness = true
 [dev-dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+clap = { version = "4.6", features = ["derive", "cargo", "env"] }
 ironrdp = { path = "../ironrdp", features = ["server", "pdu", "connector", "session", "svc", "dvc", "echo"] }
 ironrdp-async.path = "../ironrdp-async"
 ironrdp-agent.path = "../ironrdp-agent"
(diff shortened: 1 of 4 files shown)


crates/ironrdp-testsuite-extra/Cargo.toml
```diff
@@ -24,6 +24,7 @@ harness = true
 [dev-dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+clap = { version = "4.6", features = ["derive", "cargo", "env"] }
```
CBenoit, line 27: I don’t really want to have clap in dependencies here. I’m not very satisfied about the tests in ironrdp-agent, I would prefer we don’t try to relocate these for now.

glamberson:
Closing per your feedback above. Makes sense, especially on not wanting clap as a testsuite-extra dependency. Thanks for the quick look.

PR #1756, opened 2026-08-21
test(extra): relocate config, rail, and input tests into testsuite-extra

glamberson:
ironrdp-client sets `[lib] test = false`, so `config.rs`, `rail.rs`, and `rdp.rs`'s inline `#[cfg(test)]` tests never run under `cargo test --workspace`. This moves the nine tests that exercise only public API into `ironrdp-testsuite-extra`:

- Eight `ConfigBuilder` tests (RemoteApp mode, monitor layout, sound and audio-capture modes) into the existing `client_config.rs`, alongside a new `complete_builder()` helper matching the original's.
- One `RailClient` input-sender test into a new `client_input.rs`.

A new `client_rail.rs` adds six more `RailClient` tests (handshake, queued-execute release, desktop synchronization, client status flags) that were also inline-only and untested under CI.

Enabling `ironrdp-client`'s `sound` feature on the `testsuite-extra` dev-dependency lets the relocated audio-mode tests build without the crate-level `#[cfg(feature = "sound")]` gates, which only made sense inside `ironrdp-client` itself; the audio dependencies (`cpal`, `alsa`) were already pulled in by `testsuite-extra`'s existing `ironrdp-rdpsnd-native` dev-dependency, so this adds no new weight.

The remaining inline tests in `config.rs` (webauthn, the RemoteApp Execute-queuing test touching the private `rail_initial_execute` field), `rail.rs` (`queued_executes_do_not_exceed_pending_limit`, which reads the private `MAX_PENDING_EXECUTES` const), and all of `rdp.rs` except the one relocated test stay inline: each touches a private field, private helper function, or private type with no public equivalent.


crates/ironrdp-testsuite-extra/tests/main.rs
```diff
@@ -5,6 +5,8 @@ mod agent;
 mod async_framed;
 mod capture_helpers;
 mod client_config;
+mod client_input;
+mod client_rail;
```
CBenoit: suggestion: Group into a `client` module with `config`, `input` and `rail` submodules.

CBenoit:
LGTM, but address first my comment here: https://github.com/Devolutions/IronRDP/pull/1756#discussion_r3831690963

crates/ironrdp-testsuite-extra/tests/main.rs
glamberson: Done, in 25cb4aa5. Grouped into a client module with config/input/rail submodules, matching the server module's existing shape.

mamoreau-devolutions (approved):

PR #1764, opened 2026-08-22
Enhance RDP features and expose additional WASM options

tonguyenducmanh:


crates/iron-remote-desktop/src/lib.rs
```diff
@@ -243,6 +243,38 @@ macro_rules! make_bridge {
                 Self($crate::SessionBuilder::render_canvas(&self.0, canvas))
             }
 
+            // tdmanh1 28/03/2026 expose thêm 1 số hàm để set các thông tin từ web
```
mamoreau-devolutions: Comments in English please

tonguyenducmanh: Thanks for the feedback! I've updated the code according to your review. Please take another look when you have time.

CBenoit (changes requested):
Thank you for the PR.

As per the design principle [documented here](https://github.com/tonguyenducmanh/IronRDP/tree/08e7d427fa8ff7a5ce64aadbce9ba4edeb79e53d/crates/iron-remote-desktop#design-philosophy), I think these should be provided as extensions instead of be part of the common abstraction.

PR #1769, opened 2026-08-22
feat(server)!: give mouse button events a position, fix X1/X2 and middle-button/hwheel handling

### LINKED ISSUE #1466: ironrdp-server: button mouse PDUs drop their coordinates — clicks land at the stale cursor position for clients that send a button with no preceding MOVE
## Summary

In `ironrdp-server`, a mouse **button** event delivered to `RdpServerInputHandler::mouse` carries **no position**, even though the wire PDU always does. As a result, a click lands wherever the server's cursor happened to be — not where the client actually clicked — for any client that sends a button PDU **without** a preceding move PDU.

Most clients (mstsc, Microsoft Remote Desktop / the desktop Windows App) send a separate `MOVE` PDU immediately before the button, so their clicks already land correctly and the bug is masked. It bites clients that send a tap as a **single button PDU**. A concrete repro is below.

## Root cause

The `MouseEvent` variants for buttons are position-less unit variants — only `Move`/`Scroll`/`RelMove` carry coordinates:

`crates/ironrdp-server/src/handler.rs:27`
```rust
pub enum MouseEvent {
    Move { x: u16, y: u16 },
    RightPressed, RightReleased,
    LeftPressed,  LeftReleased,
    MiddlePressed, MiddleReleased,
    Button4Pressed, Button4Released,
    Button5Pressed, Button5Released,
    VerticalScroll { value: i16 },
    Scroll { x: i32, y: i32 },
    RelMove { x: i32, y: i32 },
}
```

The `From<MousePdu>` / `From<MouseXPdu>` conversions therefore **discard** the PDU's `x_position`/`y_position` whenever a button flag is set — the coordinates survive only when the PDU degenerates to a `Move`:

`crates/ironrdp-server/src/handler.rs:149` (`From<MousePdu>`) and `:176` (`From<MouseXPdu>`) both do
```rust
if flags.contains(LEFT_BUTTON) { /* -> LeftPressed / LeftReleased, x/y dropped */ }
else if flags.contains(RIGHT_BUTTON) { /* -> Right..., x/y dropped */ }
...
else { MouseEvent::Move { x, y } }   // position only survives here
```

The dispatch then forwards that position-less event straight through — every mouse arm calls `handler.mouse(mouse.into())`:

`crates/ironrdp-server/src/server.rs` (fast-path `:1548`/`:1552`/`:1556`, slow-path `:1708`/`:1712`).

So the position a button PDU carried is lost before the handler ever sees it, on both the regular (`MousePdu`) and extended (`MouseXPdu`) absolute-position paths.

## Reproduction

- Client: **Windows App on iOS/iPadOS, touch mode.**
- A tap is sent as a **single button PDU** (button DOWN then UP at the tap coordinates) with **no preceding `MOVE` PDU**.
- Result: the click registers at the *previous* cursor position, not where the user tapped — taps feel "off" / "not working."
- A mouse-driven client (mstsc, desktop Windows App) is unaffected because it emits `MOVE` → button, so the position is already current when the button arrives.

## Impact

Any `ironrdp-server` consumer serving a client that sends button-without-preceding-move gets mis-placed clicks. It's silent (no error) and easy to misattribute to coordinate scaling.

## Suggested fixes

Two shapes, and I think the choice is a maintainer call because one touches the public API:

1. **Minimal / non-breaking — synthesize a `Move` before the button** at the point of conversion/dispatch (apply the PDU's `x_position`/`y_position` as a `Move`, then the button). No public-API change. Downside: it has to be repeated at each dispatch arm (or each `From`), it emits a redundant `Move` for the common move-then-click client, and the "which flags are position-less" knowledge ends up duplicated alongside the `From` impls.

2. **Root-cause / breaking — give the button events a position.** e.g. a single `MouseEvent::Button { x, y, button, pressed }`, or coordinate fields on the existing pressed/released variants. This fixes `MousePdu`, `MouseXPdu` (and structurally the `MouseRelPdu` button branches) **uniformly in one place**, with no synthetic event and no duplicated flag knowledge. The cost is a breaking change to the public `MouseEvent` enum / `RdpServerInputHandler` contract, which affects downstream server implementations — hence flagging it rather than sending a patch.

## Downstream reference

Observed and worked around in a downstream `ironrdp-server` consumer (macrdp) via option (1) — a small helper that prepends the `Move` on the regular `MousePdu` path — which confirms the fix but also surfaced why (2) is cleaner (the same gap is reachable through `MousePdu`, `MouseXPdu`, and the `MouseRel` button branches, so (1) is a per-arm band-aid). Happy to open a PR in whichever shape you prefer.

## Environment

- Upstream `master` @ `079b4842` (verified the code above against it).
- Not related to #507, which is the native **client** showing a wrong position *to* a Windows host (`A-native-client`); this is the **server** dropping an incoming button PDU's position.

glamberson:
## Summary

- Fixes #1466: `MouseEvent`'s button variants were position-less, so `From<MousePdu>` and `From<MouseXPdu>` silently discarded the wire PDU's x/y whenever a button flag was set (position only survived for a pure Move). A client that sends a tap as a single button PDU with no preceding Move (confirmed: Windows App on iOS/iPadOS, touch mode) clicks at the stale cursor position rather than where it tapped.
- While fixing that, found two more bugs in the same conversion layer, both confirmed against MS-RDPBCGR: `From<MousePdu>` never checked `MIDDLE_BUTTON_OR_WHEEL` or `HORIZONTAL_WHEEL`, so middle-click and horizontal wheel silently fell through to the Move fallback. `From<MouseXPdu>` mapped `PointerXFlags::BUTTON1`/`BUTTON2` to Left/Right, but per 2.2.8.1.2.2.4 these are "Extended mouse button 1 (also referred to as button 4)" and "...2 (...button 5)", the X1/X2 side buttons. `MouseRelPdu`'s own sibling `From` impl already distinguishes these correctly via `XBUTTON1`/`XBUTTON2`, so this was an inconsistency, not a deliberate choice.
- Checked FreeRDP, xrdp, KDE krdp, and GNOME Remote Desktop before settling on a shape. All four check explicit MOVE flags rather than falling through on else, all four handle middle button and both wheel flags, none conflate X1/X2 with primary buttons. xrdp's own source has a maintainer comment confirming real clients exercise the HWHEEL gap: "As mstsc does MOUSE not MOUSEX for horizontal scrolling, PTRFLAGS_HWHEEL must be handled here."
- `MouseEvent::Button { x, y, button, pressed }` replaces the ten position-less `LeftPressed`/`LeftReleased`/.../`Button5Released` variants, scoped to the PDU types that actually carry absolute position (`MousePdu`, `MouseXPdu`, `ainput::MousePdu`, which already carried x/y on every event but never applied it to buttons either). `MouseRelPdu` gets `MouseEvent::ButtonRel { button, pressed }` instead, since it only has relative deltas. Both share a new `MouseButton` enum (`Left`/`Right`/`Middle`/`X1`/`X2`). Added `MouseEvent::HorizontalScroll` to pair with the existing `VerticalScroll`. Breaking anyway, so `MouseEvent` and `MouseButton` are both `#[non_exhaustive]`.
- There's independent, unmerged prior art for the middle-button half of this on branch `probakowski/xrdp` (commit `e924272`, ad hoc xrdp-interop debugging, never opened as a PR) that arrived at the same `MIDDLE_BUTTON_OR_WHEEL` fix. Crediting it here since I found it while researching this.

## Validation

`cargo xtask check fmt/lints/tests/typos/locks` all pass.

## Notes

Single file, `crates/ironrdp-server/src/handler.rs`. Grepped the whole workspace for `MouseEvent::` usage; this file is the only one that constructs or matches it, no other call sites need updating for this PR. Downstream, I'm updating my own `lamco-rdp-input`/`lamco-rdp-server` consumers to the new shape as a follow-up.

### DIFF
--- a/crates/ironrdp-server/src/handler.rs
+++ b/crates/ironrdp-server/src/handler.rs
@@ -24,21 +24,55 @@ pub enum KeyboardEvent {
 /// Describes a mouse event received from the client
 ///
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum MouseEvent {
-    Move { x: u16, y: u16 },
-    RightPressed,
-    RightReleased,
-    LeftPressed,
-    LeftReleased,
-    MiddlePressed,
-    MiddleReleased,
-    Button4Pressed,
-    Button4Released,
-    Button5Pressed,
-    Button5Released,
-    VerticalScroll { value: i16 },
-    Scroll { x: i32, y: i32 },
-    RelMove { x: i32, y: i32 },
+    Move {
+        x: u16,
+        y: u16,
+    },
+    /// A button press or release at the absolute position the event source
+    /// carries (`MousePdu`, `MouseXPdu`, `ainput::MousePdu`).
+    Button {
+        x: u16,
+        y: u16,
+        button: MouseButton,
+        pressed: bool,
+    },
+    /// A button press or release with no absolute position. `MouseRelPdu`
+    /// only carries relative motion deltas, so its buttons have nothing to
+    /// report a position with; position accumulates from `RelMove`.
+    ButtonRel {
+        button: MouseButton,
+        pressed: bool,
+    },
+    VerticalScroll {
+        value: i16,
+    },
+    HorizontalScroll {
+        value: i16,
+    },
+    Scroll {
+        x: i32,
+        y: i32,
+    },
+    RelMove {
+        x: i32,
+        y: i32,
+    },
+}
+
+/// Mouse button identity, shared by [`MouseEvent::Button`] and
+/// [`MouseEvent::ButtonRel`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum MouseButton {
+    Left,
+    Right,
+    Middle,
+    /// Extended button 1, also known as button 4 or the "back" button.
+    X1,
+    /// Extended button 2, also known as button 5 or the "forward" button.
+    X2,
 }
 
 /// Input Event Handler for an RDP server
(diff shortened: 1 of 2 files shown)


clintcan:
Thanks for running with it, @glamberson — no problem that you took it; the broader scope you found is exactly why it's better in your hands than a narrow patch from me.

Corroborating from the downstream side (macrdp is where #1466 came from): our vendored fix deliberately scoped to LEFT/RIGHT only (the reported iOS touch-tap, a left-click), but our divergence note explicitly flagged that **middle-click and the MouseX (X1/X2) buttons have the same latent coord-drop, left unaddressed** — so fixing all the button paths uniformly matches what we'd independently seen, and it's the right call over a per-button patch.

The two extra bugs you folded in (`From<MousePdu>` skipping `MIDDLE_BUTTON_OR_WHEEL`/`HORIZONTAL_WHEEL`, `From<MouseXPdu>` mapping X1/X2 → Left/Right) are consistent with how brittle that layer is — good catches. Option 2 (position on the button variants, `#[non_exhaustive]` on `MouseEvent`) is fine by us as a downstream `MouseEvent` consumer — we laid out both shapes on #1466 to leave that call to you — and we'll adapt at the pin bump.

Happy to validate the button behavior end-to-end on macrdp (mstsc + Windows App, incl. middle / X-buttons) once the shape settles, if a real-client check helps.

mamoreau-devolutions (approved):

crates/ironrdp-server/src/handler.rs
```diff
@@ -148,87 +182,120 @@ impl From<SyncToggleFlags> for KeyboardEvent {
...
 impl From<MouseRelPdu> for MouseEvent {
     fn from(value: MouseRelPdu) -> Self {
+        let pressed = value.flags.contains(PointerRelFlags::DOWN);
+
         if value.flags.contains(PointerRelFlags::BUTTON1) {
-            if value.flags.contains(PointerRelFlags::DOWN) {
-                MouseEvent::LeftPressed
-            } else {
-                MouseEvent::LeftReleased
+            MouseEvent::ButtonRel {
```
mamoreau-devolutions, line 280: Could `ButtonRel` retain the PDU's `x_delta` and `y_delta`? MS-RDPBCGR section 2.2.8.1.1.3.1.1.7 says a relative button press occurs at the position produced by applying those deltas to the previous position. Returning only the button state here discards movement from a combined relative button event, so the handler can click at a stale position. Including the deltas in `ButtonRel` would preserve the wire semantics.

> [!NOTE]

glamberson: You're right, and I checked MS-RDPBCGR 2.2.8.1.1.3.1.1.7 directly to confirm: xDelta/yDelta are present on every relative event, and PTRFLAGS_DOWN's own definition ties the press to the position produced by applying that same delta, not a separate move. My original design comment ("position accumulates from RelMove") was wrong for the combined case: no RelMove is ever emitted for a button-flagged PDU, so that motion was being dropped outright.

Added x/y to ButtonRel (matching RelMove's naming) and populated them in all five button branches, not just the fallback.

PR #1779, opened 2026-08-23
feat(rdpdr): make the PDU layer bidirectional for a future server

glamberson:
## Summary

- ironrdp-rdpdr only decodes what a client receives from a server (6 PacketIds) and only encodes what a client sends to a server. A server needs the opposite. This makes the PDU layer bidirectional: decode() added to every response type the client currently only encodes, and encode() added to every request type the client currently only decodes.
- CoreCapability::decode already branched on packet_id for both directions and just needed the client-capability arm wired into RdpdrPdu::decode_body, alongside three new decodes (ClientNameRequest, ClientDeviceListAnnounce, ClientDeviceListRemove).
- CoreDeviceIoCompletion is deliberately not wired into the automatic RdpdrPdu::decode dispatch. Fourteen different response variants share that one PacketId (see the header() method's match), and nothing on the wire disambiguates which one a given completion answers. That requires knowing the MajorFunction (and, for DirectoryControl, MinorFunction) of the original request, which only a future IRP-tracking layer will have. Added RdpdrPdu::decode_io_completion as an explicit entry point for that layer to call once it exists, taking the major/minor function and, for the three completions whose body layout depends on it, the FileInformationClass the original request asked for.
- DeviceControlResponse's output_buffer is a Box<dyn rpce::Encode> on the encode side, since it carries IOCTL-specific NDR content only the issuer of the original request can interpret. That can't be decoded back into a concrete type. Decode reads it as opaque bytes through a private RawOutputBuffer implementing the same trait, so the decoded value has the same shape without pretending to know a structure it can't actually know.
- Filled two response-decode gaps that were there before this PR touched the file, found while checking the new decode paths against MS-RDPEFS directly: FileStandardInformation and FileAttributeTagInformation are two of the exact three classes 2.2.3.3.8 allows for a Query Information response (the third, Basic, already round-tripped) and had encode with no decode. FileDirectoryInformation, FileFullDirectoryInformation, FileBothDirectoryInformation, and FileNamesInformation are the complete set 2.2.3.3.10 allows for a Query Directory response and were entirely unwired in FileInformationClass::decode, so directory listing had no working decode path at all.
- VersionAndIdPdu::decode always tagged PacketId::CoreClientidConfirm as ServerClientIdConfirm. Per MS-RDPEFS 2.2.1.1, that PacketId is shared between Server Client ID Confirm (2.2.2.6) and Client Announce Reply (2.2.2.3): correct for a client, which never decodes its own reply, but wrong for a server decoding the client's reply. Added decode_client_announce_reply as an explicit entry point for that case, mirroring decode_io_completion's precedent for the same class of PacketId ambiguity, and left decode() itself untouched.
- DeviceControlRequest<T> had decode and decode_with_input_buffer but no encode, the one server-sent I/O request missing it. Adding encode required tightening IoCtlCode to also require Into<u32>, which surfaced that ScardIoCtlCode (the trait's only other implementor) was missing that conversion too.

## Validation

`cargo xtask check fmt/lints/tests/typos/locks` all pass. Round-trip tests added for every new decode()/encode() pair in ironrdp-testsuite-core/tests/rdpdr/mod.rs, including one per Query Information and Query Directory class, plus the two additions above.

## Notes

PDU-codec layer only. The RdpdrServer state machine, IRP tracking, and ironrdp-server integration (a factory trait plus attach_channels wiring, the same shape as the recently landed RdpeiServerFactory) are a follow-up PR once this lands, so the server-relevant surface can be reviewed in a size a reviewer can actually hold in their head at once.

### DIFF
--- a/crates/ironrdp-rdpdr/src/pdu/efs.rs
+++ b/crates/ironrdp-rdpdr/src/pdu/efs.rs
@@ -112,6 +112,27 @@ impl VersionAndIdPdu {
             }
         };
 
+        Self::decode_with_kind(kind, src)
+    }
+
+    /// Decodes a `PacketId::CoreClientidConfirm` body received by a server, where it carries
+    /// a Client Announce Reply rather than the Server Client ID Confirm [`decode`] assumes.
+    ///
+    /// Per MS-RDPEFS 2.2.1.1, PAKID_CORE_CLIENTID_CONFIRM (0x4343) is the shared PacketId for
+    /// both the Client Announce Reply (2.2.2.3, client-to-server) and the Server Client ID
+    /// Confirm (2.2.2.6, server-to-client): the wire alone cannot disambiguate which one a
+    /// given caller is decoding, only the caller knows which side of the connection it is on.
+    /// [`decode`] serves the client, which never decodes its own reply, so it always tags this
+    /// PacketId [`VersionAndIdPduKind::ServerClientIdConfirm`]. A server decoding the client's
+    /// reply must call this instead to get the correct [`VersionAndIdPduKind::ClientAnnounceReply`]
+    /// tag; the fixed-part layout is identical either way.
+    ///
+    /// [`decode`]: Self::decode
+    pub fn decode_client_announce_reply(src: &mut ReadCursor<'_>) -> DecodeResult<Self> {
+        Self::decode_with_kind(VersionAndIdPduKind::ClientAnnounceReply, src)
+    }
+
+    fn decode_with_kind(kind: VersionAndIdPduKind, src: &mut ReadCursor<'_>) -> DecodeResult<Self> {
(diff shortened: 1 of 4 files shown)


glamberson:
This PR is self-contained and appropriately scoped as a single PR.

Every decode/encode pair here serves one goal: PDU-codec bidirectionality. Splitting decode from encode per type would fragment review without reducing what a reviewer has to hold in mind. The two pre-existing decode gaps I fixed (FileInformationClass::decode missing Query Information and Query Directory classes) are corrections to the same type family this PR already touches, found while adding the new decode paths.

Stacked pull requests are not available to me here since my branches live on a fork. I split what naturally splits: The RdpdrServer orchestration built on this PDU-codec layer is a separate PR, #1783, stacked on this one.

mamoreau-devolutions (approved):

Issue #195, opened 2023-09-20
Check whether RDPDR initialization sequence occurs twice


ibeckermayer:
> > Unfortunately these structs rely on taking ownership of the underlying
> > data when it's used. This would work very nicely/efficiently, except
> > under normal circumstances ironrdp-client connects and then reconnects
> > during a session, which means that in the second run (the one that the user
> > actually interacts with) the data is gone, and empty lists are sent for
> > both of these values.
> 
> This is very surprising by the way. Currently ironrdp-client is reconnecting from scratch and the Rdpdr channel is thus constructed from scratch as well, meaning `with_smartcard` is called again. I would expect the data to be populated again. Why are we actually ending up with empty lists?

@CBenoit you're correct, I was misinterpreting the logs. What I'm seeing on that commit is that we're going through the RDPDR initialization sequence twice in a single connection:

```
2023-09-20T21:49:23.076710Z TRACE ironrdp_rdpdr: received RdpdrPdu(VersionAndIdPdu { version_major: 1, version_minor: 13, client_id: 3, kind: ServerAnnounceRequest })
2023-09-20T21:49:23.076720Z TRACE ironrdp_rdpdr: sending RdpdrPdu(VersionAndIdPdu { version_major: 1, version_minor: 12, client_id: 3, kind: ClientAnnounceReply })
2023-09-20T21:49:23.076727Z TRACE ironrdp_rdpdr: sending RdpdrPdu(Unicode("IronRDP"))
[...]

PR #1783, opened 2026-08-23
feat(rdpdr): add RdpdrServer orchestration on top of the bidirectional PDU layer

glamberson:
## Depends on #1779

This branch is stacked on `feat/rdpdr-server-pdu-decode` (#1779) and only makes sense once that lands: `RdpdrServer` is built entirely on the decode/encode surface #1779 adds. #1779's own Notes section already flags this as the intended follow-up.

## Summary

- `RdpdrServer` is a state machine (`Initializing`, `AwaitingAnnounce`, `CapabilityExchange`, `Active`) driven by `SvcProcessor::process`, following the handshake sequence MS-RDPEFS 3.1.5/3.2.5 describe. `CapabilityExchange` and `Active` both accept `CoreDevicelistAnnounce` since real clients don't always send it strictly during capability exchange. A client re-sending its Announce Reply while already `Active` re-initializes (issue #195's duplicate-init case) rather than erroring: devices and in-flight IRPs clear, the backend sees a fresh `on_device_announce` batch.
- Outstanding drive I/O is tracked by `CompletionId` in an `IrpTracker`, mirroring `ironrdp-rdpeusb`'s `RequestIdAllocator`/`pending_io` precedent. Each `PendingIrp` variant carries exactly the context `RdpdrPdu::decode_io_completion` needs (`MajorFunction`, `MinorFunction`, and the file/filesystem information class where the response layout depends on it). An unrecognized `CompletionId` is logged and ignored, not treated as an error.
- `RdpdrServerBackend` has one completion callback per `PendingIrp` variant (fourteen, matching `MajorFunction` exactly except the deliberately unsupported `SetVolumeInformation`) plus `on_device_announce`, `on_device_remove`, `on_client_name`. `NoopRdpdrServerBackend` accepts every device and no-ops the rest.
- Fourteen `drive_*` methods build and send each request type. Thirteen share a `DriveRequestBody` trait and a `send_device_io_request` helper; `device_control` needs its own wrapper since `DeviceControlRequest`'s encode doesn't include the input buffer that follows it on the wire, matching its decode-side counterpart.

## Validation

`cargo xtask check fmt/lints/tests/typos/locks` all pass. Full handshake, per-drive-method completion round trips (all fourteen), device accept/reject, duplicate-init, and orphaned-completion-id tests added in `ironrdp-testsuite-core/tests/rdpdr/mod.rs`.

## Notes

`ironrdp-server` integration (a factory trait plus `attach_channels` wiring, matching `RdpeiServerFactory`'s recently landed shape) is a further follow-up once this lands.

### DIFF
--- a/crates/ironrdp-rdpdr/src/lib.rs
+++ b/crates/ironrdp-rdpdr/src/lib.rs
@@ -19,6 +19,7 @@ use tracing::{debug, trace, warn};
 
 pub mod backend;
 pub mod pdu;
+pub mod server;
 
 pub use self::backend::noop::NoopRdpdrBackend;
 pub use self::backend::{
--- a/crates/ironrdp-rdpdr/src/server.rs
+++ b/crates/ironrdp-rdpdr/src/server.rs
@@ -1,0 +1,1021 @@
+//! Server-side [\[MS-RDPEFS\]] orchestration: the half of RDPDR that initiates the
+//! handshake, tracks outstanding I/O requests by `CompletionId`, and routes client
+//! completions back to an application-supplied [`RdpdrServerBackend`].
+//!
+//! This module only builds and interprets PDUs; it knows nothing about where a
+//! `drive_read` completion's bytes end up or which local file a `drive_create`
+//! corresponds to. That is the backend's job, matching the layering already used
+//! by `crate::backend` on the client side and by `ironrdp-rdpeusb`'s
+//! `UrbdrcDeviceServerBackend` on the server side of a different channel.
+//!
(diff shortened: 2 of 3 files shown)


glamberson:
This PR is self-contained and appropriately scoped as a single PR.

Every decode/encode pair here serves one goal: PDU-codec bidirectionality. Splitting decode from encode per type would fragment review without reducing what a reviewer has to hold in mind. The two pre-existing decode gaps I fixed (FileInformationClass::decode missing Query Information and Query Directory classes) are corrections to the same type family this PR already touches, found while adding the new decode paths.

Stacked pull requests are not available to me here since my branches live on a fork. I split what naturally splits: The RdpdrServer orchestration built on this PDU-codec layer is a separate PR, #1783, stacked on this one.

mamoreau-devolutions (approved):

Issue #1712, opened 2026-08-20
Windows App (macOS) drops AVC sessions on master: EGFX region-bounds fixes exist only in a fork


maryny4:
> **Corrected below:** this is not a regression between two upstream commits — master is missing three EGFX commits that only exist in a fork, and #1530 is not involved. The original report is kept as written; see the first comment for the corrected diagnosis and evidence.

Downstream server report (hypr-rdp, an ironrdp-server embedder; Hybrid security, EGFX AVC420, no multitransport). At revision 0568dfb0 sessions from Windows App on macOS (the current Microsoft client) work. Built against current master (04320a0c) with the same embedder code, every Windows App connection dies right after connection setup, while sdl-freerdp 3.30 sessions against the same build work fine.

Observed sequence with `ironrdp_*=debug` (full log available): CredSSP/NTLM completes, licensing StatusValidClient, static channels start, DVC caps V3 negotiated, four DVC create responses arrive — one channel refused with CreationStatus(0xC0000001) (possibly benign, we have no pre-regression debug log to compare), the others created. ~600ms later the client sends a DVC Close for the graphics channel ("EGFX channel closed"), then drops TCP without close_notify; the client UI shows "Your session ended because of a protocol error".

The window is 10 commits (0568dfb0..04320a0c). The prime suspect from the diff is #1530 (`feat(connector)!: pass frame arrival time into Sequence::step`), which also reworks `ironrdp-egfx/src/server.rs` (−149/+39) — the same channel the client closes. Nothing else in the window touches the graphics wire path; the server-side DVC registration is unchanged.

Disclosure: our master build carries two API-compat shims (fork-only `client_keyboard_layout` input override removed; the AVC444 send path stubbed — AVC420 was negotiated, so neither is on the failing wire path). Happy to bisect the window and test patches — the failing client setup is at hand.

maryny4:
Correction: this is not a regression between two upstream commits, and the "after DVC create" framing in the report was a red herring. Bisecting from the other side showed the opposite — the fork this server pins is *ahead* of master by three EGFX commits that upstream does not have: `feat(egfx): add explicit AVC444v2 frame sender`, `fix(egfx): use exclusive AVC region bounds` and `fix(egfx): bound both AVC444 streams` (MuNeNICK/IronRDP). Master carries only five commits the fork lacks, none of them touching graphics.

Evidence, same server code and the same client each time:

[...]

PR #1788, opened 2026-08-23
fix(egfx): use exclusive AVC region bounds

maryny4:
MS-RDPEGFX 2.2.1.2 defines `RDPGFX_RECT16` with exclusive `right` and `bottom`, and 2.2.4.4.1 gives `regionRects` in `RFX_AVC420_METABLOCK` that same type. `Avc420Region` documented its edges as inclusive, `full_frame()` built `width - 1` / `height - 1`, and `to_rectangle()` handed those to an `InclusiveRectangle` that encoded them unchanged, so every `regionRects` entry went out a pixel short on the right and bottom edge. `GraphicsPipelineServer::compute_dest_rect()`, used for the `WireToSurface1` destination, does add that pixel back: the two call sites disagreed about whether the conversion had already happened.

`Avc420Region` is exclusive throughout now and `to_rectangle` produces an `ExclusiveRectangle`, so the asymmetric adjustment disappears. `fix(egfx): bound both AVC444 streams` then computes the destination rectangle from both streams rather than from the luma one, so a chroma region larger than luma is no longer cut short. `feat(egfx): add explicit AVC444v2 frame sender` separates the AVC444v2 path from AVC444; both fixes build on it.

What the defect costs: with `ironrdp-server` on master, Windows App on macOS drops the connection right after the encoder starts, the server reporting `peer closed connection without sending TLS close_notify`. Pinning the session resolution, so that no deactivation-reactivation happens at all, changes nothing. With these three commits the same server build and the same client run normally. FreeRDP-based clients accept the frames either way, which is why this shows only against the Microsoft client. That is not proof this field is the trigger - 2.2.4.4.1 also calls the metablock informational - only that the field is wrong on the wire and that these commits make that client work.

`avc.rs` lost its inline test module to #1736 while these sat in a fork, so the `full_frame` expectation is updated in `testsuite-core/tests/egfx/avc.rs` instead.

Please rebase-merge rather than squash. The three commits are @MuNeNiCK's with his authorship intact, and the breaking change is confined to the middle one; squashing would collapse both.

BREAKING CHANGE: `Avc420BitmapStream::rectangles` is now `Vec<ExclusiveRectangle>`, and `Avc420Region::to_rectangle` returns `ExclusiveRectangle`. Callers that build or match `InclusiveRectangle` there have to switch, and a region built with `Avc420Region::full_frame` no longer subtracts one from the width and height.

Issue: #1712

### DIFF
--- a/crates/ironrdp-egfx/src/pdu/avc.rs
+++ b/crates/ironrdp-egfx/src/pdu/avc.rs
@@ -2,7 +2,7 @@ use core::fmt;
 
 use bit_field::BitField as _;
 use bitflags::bitflags;
-use ironrdp_pdu::geometry::InclusiveRectangle;
+use ironrdp_pdu::geometry::ExclusiveRectangle;
 use ironrdp_pdu::{
     Decode, DecodeResult, Encode, EncodeResult, ReadCursor, WriteCursor, cast_length, ensure_fixed_part_size,
     ensure_size, invalid_field_err,
@@ -77,7 +77,8 @@ impl<'de> Decode<'de> for QuantQuality {
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[derive(Clone, PartialEq, Eq)]
 pub struct Avc420BitmapStream<'a> {
-    pub rectangles: Vec<InclusiveRectangle>,
+    /// Region masks encoded as `RDPGFX_RECT16` with exclusive right/bottom bounds.
+    pub rectangles: Vec<ExclusiveRectangle>,
     pub quant_qual_vals: Vec<QuantQuality>,
     pub data: &'a [u8],
 }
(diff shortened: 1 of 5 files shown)


MuNeNiCK:
Thank you for taking the time to validate these changes and submit them upstream. Sorry I missed your earlier messages. I’m happy for these changes to be upstreamed through this PR.

glamberson:
Thanks for filing this, and glad to see MuNeNiCK confirm he's happy for it to go through this PR. Good fix.

maryny4:
Context that does not belong in the commit message.

**Authorship.** These three commits are @MuNeNiCK's, cherry-picked onto master with his authorship intact; I am the committer only. I asked him twice whether he would rather send them himself before opening this, and he has since confirmed here that he is happy for them to go upstream through this PR. That is the reason for the rebase-merge request in the body: a squash would put my name on his work.

**Provenance of the analysis.** The exclusive/inclusive mismatch was diagnosed by @glamberson in #1712. I checked each step of it against master before sending this rather than taking it on trust.

**What I ran locally.** `cargo xtask check fmt` is clean. `ironrdp-egfx` is 46 passed / 0 failed and the egfx part of the testsuite is 52 passed / 0 failed. I could not run `cargo xtask ci` in full on this machine: the repo pins toolchain 1.94.1 and the installed one is 1.98.0, under which `cargo clippy --workspace -- -D warnings` fails in `ironrdp-bulk` and `ironrdp-str` - crates this PR does not touch, and it fails there on unmodified master too - and `typos-cli` is not installed here. The CI on this head is green: 35 passed, 8 skipped, none failed.

mamoreau-devolutions (approved):

PR #1790, opened 2026-08-24
feat(viewer): add /microphone command-line switch

manio:
Expose a `--microphone` CLI flag in ironrdp-viewer to enable the AUDIO_INPUT (MS-RDPEAI) dynamic virtual channel, redirecting the local microphone into the RDP session. It wires into the existing `ConfigBuilder::with_audio_capture(true)` and is off by default, matching `ChannelConfig::default().audio_capture == false`.

The switch name is in pair with FreeRDP's own `/microphone` option, for consistency.

### DIFF
--- a/crates/ironrdp-viewer/src/cli.rs
+++ b/crates/ironrdp-viewer/src/cli.rs
@@ -195,6 +195,13 @@ struct Args {
     #[clap(long = "no-webauthn", action = clap::ArgAction::SetTrue, overrides_with = "webauthn")]
     no_webauthn: bool,
 
+    /// Audio input (microphone)
+    ///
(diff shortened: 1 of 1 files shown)


manio:
@mamoreau-devolutions

mamoreau-devolutions:
What are the current options we have for audio output? We don't have to match FreeRDP exactly. I wonder if we shouldn't call it "audio input" and "audio output" instead

manio:
> What are the current options we have for audio output? We don't have to match FreeRDP exactly. I wonder if we shouldn't call it "audio input" and "audio output" instead

No idea to be honest. I tested this only on my microphone...
In regard to naming convention I am following @CBenoit [comment](https://github.com/Devolutions/IronRDP/pull/1200#issuecomment-4177832571) to keep it in sync where possible.

mamoreau-devolutions (approved):

PR #1793, opened 2026-08-24
perf(graphics): optimize RemoteFX decoding

zmb3:
Adds a benchmark suite for the various stages of the RemoteFX decode pipeline, then implements a couple changes that result in noticeable improvements to the metrics.

The optimized bit reader can likely be reused in a few places in the EGFX decode pipeline, but I'll save that for later to keep the change focused on RemoteFX.

This is my first foray into Rust benchmarking, and I also wanted a way to run the benchmarks in WASM since the IronRDP client I work with most runs in the browser. Let me know if I've done something non-standard or if you have any suggestions for improvements.

### DIFF
--- a/crates/ironrdp-graphics/Cargo.toml
+++ b/crates/ironrdp-graphics/Cargo.toml
@@ -25,7 +25,7 @@ ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.9", features = ["std"] } #
 byteorder = "1.5" # TODO: remove
 num-derive.workspace = true # TODO: remove
 num-traits.workspace = true # TODO: remove
-yuv = { version = "0.8", features = ["rdp"] }
+yuv = { version = "0.8.17", features = ["rdp"] }
 wide = "0.7" # portable SIMD for the inverse DWT (wasm/x86/ARM); `std::simd` is still nightly-only
 
 [dev-dependencies]
--- a/crates/ironrdp-graphics/src/rlgr.rs
+++ b/crates/ironrdp-graphics/src/rlgr.rs
@@ -6,8 +6,6 @@ use bitvec::prelude::*;
 use ironrdp_pdu::codecs::rfx::EntropyAlgorithm;
 use yuv::YuvError;
 
-use crate::utils::Bits;
-
 const KP_MAX: u32 = 80;
 const LS_GR: u32 = 3;
 const UP_GR: u32 = 4;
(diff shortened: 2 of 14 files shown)


crates/ironrdp-graphics/src/rlgr.rs
```diff
@@ -311,8 +412,12 @@ pub fn decode(mode: EntropyAlgorithm, tile: &[u8], mut output: &mut [i16]) -> Re
                     EntropyAlgorithm::Rlgr3 => {
                         let n_index = compute_n_index(code_remainder);
 
-                        let val1 = load_be_u32(try_split_bits!(bits, n_index));
-                        let val2 = code_remainder - val1;
```
zmb3, line 315: This looked like a possible underflow.

crates/ironrdp-graphics/Cargo.toml
```diff
@@ -25,7 +25,7 @@ ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.9", features = ["std"] } #
 byteorder = "1.5" # TODO: remove
 num-derive.workspace = true # TODO: remove
 num-traits.workspace = true # TODO: remove
-yuv = { version = "0.8", features = ["rdp"] }
+yuv = { version = "0.8.17", features = ["rdp"] }
```
CBenoit, line 28: issue: Avoid specifying minor versions.

suggestion: Bump in Cargo.lock only instead.

rationale: I assume the performance is impacted by this specific patch, but I don’t think it’s critical enough for us to lock in a specific version in the Cargo.toml manifest.

CBenoit (commented):
The way testing is done in this PR is going against the general practices of IronRDP, as it’s typical with LLM-generated code in general, even though I’ve created skills and it should be mentioned in the AGENTS.md file, but I will say: it’s okay for now. I see a lot of similar code was merged recently, so I would rather do an intentional clean up later. I enabled the automatic reviewing for oversized PRs, that being said I think this can be split into smaller PRs: the benchmarks in one, and the actual optimizations in another. Ideally, I would like to keep some form of documentation of this effort, maybe a .md file in the `docs/` folder? Open to ideas here. Also, I think we already have a benchmarking crate, so I wonder if it’s really necessary to create a new one again. I assume it’s because you need a WASM harness? We may need to think about how we organize benchmarks across the workspace in general.

CBenoit (commented):
[above comment was duplicated]

PR #1797, opened 2026-08-25
fix(client): re-register AUDIO_INPUT listener so microphone can reopen

manio:
The AUDIO_INPUT (MS-RDPEAI) dynamic virtual channel was registered via `with_dynamic_channel`, which wraps the processor in an `OnceListener`. `OnceListener::create` hands out its `RdpeaiClient` exactly once: after the first DVC Close, it keeps returning `None`, so any later Create Request for AUDIO_INPUT is answered with `NO_LISTENER` and the microphone silently stops working for the rest of the session.

Windows opens and closes AUDIO_INPUT repeatedly as apps grab/release the mic, so a single-shot registration is not sufficient.

Switch to `with_listener` with a small `RdpeaiListener` factory that builds a fresh `RdpeaiClient` (and a fresh `RdpeaiCaptureBackend`/cpal stream) on every Create Request, matching the pattern already used for the RDPEWA listener. Gate the new type behind `feature = "sound"` and widen the existing `DvcChannelListener` import's cfg accordingly, since it is now also needed outside the `windows` + `dvc-com-plugin` path.

@mamoreau-devolutions

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -11,8 +11,8 @@ use ironrdp_connector::{ConnectionResult, ConnectorResult};
 use ironrdp_core::WriteBuf;
 use ironrdp_displaycontrol::client::DisplayControlClient;
 use ironrdp_displaycontrol::pdu::MonitorLayoutEntry;
-#[cfg(all(windows, feature = "dvc-com-plugin"))]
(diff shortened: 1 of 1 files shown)


mamoreau-devolutions (approved):

manio:
Is this OK for merging now?

PR #1799, opened 2026-08-25
fix(server): disable Nagle on accepted connections

maryny4:
RDP output is a stream of small writes the peer is waiting on: a frame, a pointer update, a channel PDU. Nagle holds the trailing partial segment of each until the previous one is acknowledged, so against a peer using delayed acknowledgements every such write waits on a timer that has nothing to do with the encoder or the link. `run` sets `SO_REUSEADDR` on the listening socket and nothing on the accepted ones.

A failure is logged rather than propagated: the session works either way, only less promptly, and refusing a connection over a socket option would cost more than the latency it avoids.

`run_connection` and `run_connection_with` take a stream from the caller and cannot set this themselves, so their documentation now says whose job it is. For reference, `ironrdp-client` sets it on its RDCleanPath transport and not on a direct connection.

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -1194,6 +1194,11 @@ impl RdpServer {
     /// Run a single RDP connection over `stream`, performing the
     /// IronRDP-managed TLS handshake on `ShouldUpgrade` (standard TCP+TLS).
     ///
+    /// Socket options on `stream` are the caller's to set. In particular RDP
+    /// is a stream of small, latency-sensitive writes, so a TCP stream should
+    /// have `TCP_NODELAY` set; [`RdpServer::run`] does that for the
+    /// connections it accepts itself.
+    ///
     /// Equivalent to [`run_connection_with`](Self::run_connection_with) with
     /// [`TransportTls::Managed`].
     pub async fn run_connection<S>(&mut self, stream: S) -> ServerResult<()>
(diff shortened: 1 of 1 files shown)


glamberson:
The write path (`ironrdp-tokio/src/lib.rs:87-96`) already calls `flush()` after every `write_all()`, with no `BufWriter` and no coalescing anywhere in `ironrdp-server/src/`. Someone already decided this codebase should never sit on output. It just doesn't work in practice. Calling `flush()` only pushes bytes to the kernel; it never touches Nagle, which lives one layer down. TCP_NODELAY is what actually enforces the intent the code already committed to.

Because writes are already this unbatched (a busy frame fragments into 600+ separate `write_all`s at `fast_path.rs:7`'s 16,374-byte boundary), disabling Nagle changes nothing about packet count. The kernel was always going to segment those the same way; it only removes the stall on the trailing sub-MSS remainder waiting on a delayed ACK. So the "cost more than the latency it avoids" line in the PR body is not just reasonable, it is effectively free: Same packets, no delay.

Philosophically: Nagle and delayed ACK are throughput-era defaults (1984 and 1989, scarce links, bulk transfer). RDP optimizes for perceived immediacy, not bytes per packet, and every interactive protocol re-litigates this fight for the same reason. IronRDP already made this call once, at the UDP layer: RDPEUDP2 cut its own delayed-ACK timer from 300ms to 50ms for exactly this. This PR is the TCP-side sibling of a decision the protocol's own designers already made.

Scope is honest: Server's outgoing direction only, correctly leaves caller-owned streams alone. No downside found. Approve, no reservations.

mamoreau-devolutions (approved):

PR #1803, opened 2026-08-26
fix(rdpsnd): stop blocking the real-time playback callback

manio:
`RxBuffer::fill()` ran on the cpal real-time audio thread and called `Receiver::recv_timeout(Duration::from_millis(4000))` whenever it ran out of buffered PCM. Any pause in incoming Wave PDUs longer than one callback period — completely normal whenever nothing is currently playing remotely — blocked the audio device's callback thread for up
to 4 seconds, starving its ring buffer and triggering a cpal "Buffer underrun/overrun occurred" error plus a "timed out waiting on channel" warning, repeating every ~4s for as long as playback stayed idle.

Replace the `mpsc::Receiver<Vec<u8>>` between the producer (network/decode) and the consumer (cpal callback) with a lock-free SPSC byte ring buffer (`ringbuf`). `RxBuffer::fill()` now calls `pop_slice`, which is non-blocking and returns immediately with whatever is available; any shortfall is filled with correct silence (0x80 for unsigned 8-bit PCM, 0x00 otherwise) instead of blocking or leaving stale bytes. Volume is still applied at pop time, right before playback, preserving the existing responsiveness of VolumePdu changes.

This also drops the old `last`/`idx` partial-block bookkeeping in `RxBuffer` (the ring buffer handles partial reads natively), and moves Opus decoding off the `stream_handle` thread into its own thread spawned directly from `wave()`, pushing decoded PCM into the same ring buffer non-blockingly.

Fixes for this kind of errors:
```
2026-08-26T04:19:43.004416Z ERROR ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:43.055013Z ERROR ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:43.104671Z ERROR ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:47.109771Z  WARN ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:540: error=timed out waiting on channel
2026-08-26T04:19:47.109844Z ERROR ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:48.950520Z  WARN ThreadId(54) crates/ironrdp-rdpsnd-native/src/cpal.rs:540: error=channel is empty and sending half is closed
2026-08-26T04:19:49.375421Z ERROR ThreadId(56) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:49.489871Z ERROR ThreadId(56) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:49.593537Z ERROR ThreadId(56) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
2026-08-26T04:19:49.679003Z ERROR ThreadId(56) crates/ironrdp-rdpsnd-native/src/cpal.rs:473: error=Buffer underrun/overrun occurred.
```

### DIFF
--- a/crates/ironrdp-rdpsnd-native/Cargo.toml
+++ b/crates/ironrdp-rdpsnd-native/Cargo.toml
@@ -28,7 +28,7 @@ ironrdp-error = { path = "../ironrdp-error", version = "0.2", features = ["std"]
 ironrdp-rdpeai = { path = "../ironrdp-rdpeai", version = "0.1", optional = true } # public
 ironrdp-rdpsnd = { path = "../ironrdp-rdpsnd", version = "0.9" } # public
 opus2 = { version = "0.4", optional = true, features = ["bundled"] }
-ringbuf = "0.5"
+ringbuf = "0.5" # public
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
--- a/crates/ironrdp-rdpsnd-native/Cargo.toml
+++ b/crates/ironrdp-rdpsnd-native/Cargo.toml
@@ -28,6 +28,7 @@ ironrdp-error = { path = "../ironrdp-error", version = "0.2", features = ["std"]
 ironrdp-rdpeai = { path = "../ironrdp-rdpeai", version = "0.1", optional = true } # public
 ironrdp-rdpsnd = { path = "../ironrdp-rdpsnd", version = "0.9" } # public
 opus2 = { version = "0.4", optional = true, features = ["bundled"] }
+ringbuf = "0.5"
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
--- a/crates/ironrdp-rdpsnd-native/Cargo.toml
+++ b/crates/ironrdp-rdpsnd-native/Cargo.toml
@@ -28,7 +28,7 @@ ironrdp-error = { path = "../ironrdp-error", version = "0.2", features = ["std"]
 ironrdp-rdpeai = { path = "../ironrdp-rdpeai", version = "0.1", optional = true } # public
 ironrdp-rdpsnd = { path = "../ironrdp-rdpsnd", version = "0.9" } # public
 opus2 = { version = "0.4", optional = true, features = ["bundled"] }
-ringbuf = "0.5" # public
+ringbuf = "0.5"
 tracing = { version = "0.1", features = ["log"] }
 
 [dev-dependencies]
--- a/crates/ironrdp-rdpsnd-native/src/cpal.rs
+++ b/crates/ironrdp-rdpsnd-native/src/cpal.rs
@@ -1,8 +1,7 @@
 use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
-use core::time::Duration;
 use std::borrow::Cow;
 use std::sync::Arc;
-use std::sync::mpsc::{self, Receiver, Sender};
+use std::sync::mpsc::{self, Sender};
 use std::thread::{self, JoinHandle};
 
 use cpal::traits::{DeviceTrait as _, HostTrait as _, StreamTrait as _};
@@ -10,6 +9,8 @@ use cpal::{SampleFormat, Stream, StreamConfig, SupportedStreamConfigRange};
 use ironrdp_error::bail;
 use ironrdp_rdpsnd::client::RdpsndClientHandler;
 use ironrdp_rdpsnd::pdu::{AudioFormat, AudioFormatFlags, PitchPdu, VolumePdu, WaveFormat};
+use ringbuf::traits::{Consumer as _, Producer as _, Split as _};
+use ringbuf::{HeapCons, HeapProd, HeapRb};
 use tracing::{debug, error, trace, warn};
 
 use crate::error::{RdpsndNativeError, RdpsndNativeErrorKind, RdpsndNativeResult};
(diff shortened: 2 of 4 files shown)


manio:
CC: @mamoreau-devolutions

crates/ironrdp-rdpsnd-native/src/cpal.rs
```diff
@@ -363,73 +509,11 @@ impl RdpsndClientHandler for RdpsndBackend {
...
-                rx = dec_rx;
-            }
-            WaveFormat::PCM => {}
-            _ => bail!(
-                "matching server-requested wave format",
-                RdpsndNativeErrorKind::UnsupportedFormat,
-            ),
-        }
-
+    pub fn new(rx_format: &AudioFormat, consumer: HeapCons<u8>, volume: Arc<AtomicU32>) -> RdpsndNativeResult<Self> {
```
mamoreau-devolutions: This public signature now requires `HeapCons<u8>`, but `examples/cpal.rs` still passes an `mpsc::Receiver`, so `cargo test -p ironrdp-rdpsnd-native --locked --no-run` fails with E0308. Please update the example to construct and split a `HeapRb`, or keep the ring-buffer type behind a crate-private API.

> [!NOTE]

crates/ironrdp-rdpsnd-native/src/cpal.rs
```diff
@@ -330,11 +455,30 @@ impl RdpsndClientHandler for RdpsndBackend {
...
+        match self.ingest.as_mut() {
+            Some(Ingest::Pcm(producer)) => {
+                let written = producer.push_slice(&data);
+                let dropped = data.len() - written;
+                if dropped > 0 {
+                    // A handful of dropped bytes every so often is normal clock-drift
+                    // absorption between the server's audio timeline and the local
+                    // device (no resampling here); only worth a WARN once it's large
+                    // enough to suggest an actual scheduling stall on the audio thread.
+                    if dropped > format.n_avg_bytes_per_sec as usize / 100 {
```
mamoreau-devolutions: This cast triggers the workspace `clippy::as_conversions` lint, which becomes an error under `-D warnings`. Please use a checked conversion, matching `ring_buffer_capacity`, such as `usize::try_from(format.n_avg_bytes_per_sec).unwrap_or(usize::MAX)`.

> [!NOTE]

crates/ironrdp-rdpsnd-native/src/cpal.rs
```diff
@@ -1,15 +1,16 @@
 use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
-use core::time::Duration;
 use std::borrow::Cow;
 use std::sync::Arc;
-use std::sync::mpsc::{self, Receiver, Sender};
+use std::sync::mpsc::{self, Sender};
```
mamoreau-devolutions, line 5: These imports are now used only by `#[cfg(feature = "opus")]` code. Without default features they are unused and fail Clippy with `-D warnings`; please gate this import with the same feature.

> [!NOTE]

crates/ironrdp-rdpsnd-native/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ironrdp-error = { path = "../ironrdp-error", version = "0.2", features = ["std"]
 ironrdp-rdpeai = { path = "../ironrdp-rdpeai", version = "0.1", optional = true } # public
 ironrdp-rdpsnd = { path = "../ironrdp-rdpsnd", version = "0.9" } # public
 opus2 = { version = "0.4", optional = true, features = ["bundled"] }
+ringbuf = "0.5"
```
mamoreau-devolutions: `ringbuf::HeapCons` is exposed by the public `DecodeStream::new` signature, so repository dependency policy requires this to be marked `# public`. If that exposure is unintended, making the constructor crate-private would avoid coupling the public API to ringbuf instead.

> [!NOTE]

mamoreau-devolutions (changes requested):
Requesting changes for the example and workspace-lint build failures, plus the feature-gating and public-dependency issues noted inline.

> [!NOTE]

manio:
Addressed, thanks!

mamoreau-devolutions (approved):

mamoreau-devolutions:
@manio can you fix the failing CI checks

crates/ironrdp-rdpsnd-native/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ironrdp-error = { path = "../ironrdp-error", version = "0.2", features = ["std"]
 ironrdp-rdpeai = { path = "../ironrdp-rdpeai", version = "0.1", optional = true } # public
 ironrdp-rdpsnd = { path = "../ironrdp-rdpsnd", version = "0.9" } # public
 opus2 = { version = "0.4", optional = true, features = ["bundled"] }
+ringbuf = "0.5" # public
```
CBenoit: ringbuf itself looks sufficiently lightweight and is a good fit for the real-time SPSC path. However, can we keep it implementation-private? Otherwise, a breaking change for them, is a breaking change for us as well.

CBenoit (commented):
Thank you for the fix!

~~In addition to fixing the CI, before we merge, can you make sure you follow the repository conventions about PR/commits?~~

~~We are providing skills to help with that:~~
- ~~https://github.com/Devolutions/IronRDP/blob/master/.agents/skills/commit-and-pr-conventions/SKILL.md~~
- ~~https://github.com/Devolutions/IronRDP/blob/master/.agents/skills/commit-scope/SKILL.md~~

~~I think Claude is supposed to read them when necessary, I’m not sure why it did not happen here. If you see how we could improve our setup, let us know.~~

EDIT: Looks like you updated accordingly while I was reviewing, thank you! 🙂

crates/ironrdp-rdpsnd-native/Cargo.toml
manio: It was, but I did this after @mamoreau-devolutions review

manio: > ringbuf itself looks sufficiently lightweight and is a good fit for the real-time SPSC path. However, can we keep it implementation-private? Otherwise, a breaking change for them, is a breaking change for us as well.

Changed, please see if it is now OK

PR #1804, opened 2026-08-26
fix(client): send shutdown before resize-fallback reconnect

manio:
Hi
This is my series of commits that fixes the connection issues to the server. Unfortunately, one of my target hosts is not exactly a speed demon, and to make matters worse, my corporate setup adds full antivirus protection and other stuff on top of it. As a result, only maybe 2–3 out of 10 connection attempts would succeed.

You had to hit the timing just right because IronRDP’s initial sequence was quite aggressive (interestingly, FreeRDP didn’t have this issue).

These two commits improve the situation dramatically - now virtually all the connections I've tried work reliably.

### DIFF
--- a/crates/ironrdp-client/src/rdp.rs
+++ b/crates/ironrdp-client/src/rdp.rs
@@ -2185,6 +2185,27 @@ fn poll_deferred_rdpdr_output(active_stage: &mut ActiveStage) -> SessionResult<O
     }
 }
 
+/// Best-effort notice to the server that this session is being abandoned before the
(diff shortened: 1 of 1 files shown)


manio:
I also ran some additional tests in a completely different environment, and the results were actually as expected (although somewhat counterintuitive, which is a good thing). Increasing the timeout to 8 seconds makes the session appear around 5-6 seconds faster.
I did around 30 test connections using both the original 3-second value and the new 8-second value, and the results were consistent. So, from my perspective, the change looks absolutely valid.

p.s. Is this PR good to merge?

PR #1805, opened 2026-08-26
fix(egfx): do not track frames a suspended client cannot acknowledge

maryny4:
A client that sets queueDepth to SUSPEND_FRAME_ACKNOWLEDGEMENT stops
sending Frame Acknowledge until it opts back in ([MS-RDPEGFX] 2.2.2.13),
so nothing is coming to take a frame sent during that pause back out of
`unacknowledged`. `begin_frame` inserted every one of them regardless.

On resume the map therefore stood above `max_in_flight` with no way
down: `should_backpressure` holds every frame back, and with no frame
going out there is no End Frame left for the client to acknowledge. The
resuming PDU removes exactly one entry, which is not enough to break
out, and 3.3.5.13 has the client report only its most recently processed
frame -- there is no backlog of acknowledgements to catch up on. The
session stays frozen until it is torn down.

Backpressure is switched off while the client is suspended, so the entry
has no work to do there in the first place -- the fix is to stop making
it.

Frames already in flight when the suspension arrives are dropped because
3.2.5.13 requires it: "the server MUST clear the Unacknowledged Frames
ADM element and MUST NOT expect any further RDPGFX_FRAME_ACKNOWLEDGE_PDU
messages from the client. In this mode, the server MUST NOT wait or
block on unacknowledged frames". That happens after the acknowledgement
this PDU carries has been taken out, not before -- a suspending Frame
Acknowledge is still an acknowledgement, and the frame it names carries
the round-trip sample and byte count the QoE report is built from.

What the accounting does lose while suspended: `in_flight` reads zero,
and frames sent during the pause never reach `total_acked`,
`total_bytes_sent` or the round-trip series, since their
acknowledgements never arrive. Any consumer of those counters sees the
divergence, and it does not close on resume.

Three tests: frames keep flowing after a resume, the whole window is
available again afterwards, and a suspending acknowledgement still
counts. The suspend path had no coverage at all before.

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -525,15 +525,20 @@ impl FrameTracker {
         let frame_id = self.next_frame_id;
         self.next_frame_id = self.next_frame_id.wrapping_add(1);
 
-        self.unacknowledged.insert(
-            frame_id,
-            FrameInfo {
+        // A suspended client sends no acknowledgement until it opts back in
+        // ([MS-RDPEGFX] 2.2.2.13), so nothing would take this frame back out.
+        // Backpressure is off while suspended, so the entry has no work to do.
+        if !self.ack_suspended {
+            self.unacknowledged.insert(
                 frame_id,
-                timestamp,
-                sent_at: Instant::now(),
-                size_bytes: 0,
-            },
-        );
+                FrameInfo {
+                    frame_id,
+                    timestamp,
+                    sent_at: Instant::now(),
+                    size_bytes: 0,
+                },
+            );
+        }
 
         self.total_sent += 1;
         frame_id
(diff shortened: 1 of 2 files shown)


maryny4:
Context that does not belong in the commit message.

Found while chasing a freeze in a downstream server that embeds this crate. I have no capture of the deadlock itself — the trace level that would have shown `ack_suspended` was off at the time — so the accounting above is read out of the code and the spec, not out of a reproduction. What is reproduced is the tests: they fail against `master`.

That downstream server has the same defect in its own counter and I am fixing that separately, but its send path gates on `FrameTracker::should_backpressure` as well, so the downstream fix alone cannot help.

Neither client I can test against ever suspends: 8986 frames of instrumented tracing, `client_queue_depth` 0 and `ack_suspended` false throughout, no `0xFFFFFFFF` anywhere. mstsc answers a minimised window with a Suppress Output PDU ([MS-RDPBCGR] 2.2.11.3) instead. So this is a real defect with no trigger I can reach, which is why it comes with tests rather than a repro.

Issue #1330, opened 2026-05-27
client: drop-policy classification for RdpOutputEvent + drop-oldest/coalesce channel wrapper


glamberson:
@CBenoit: picking up the invitation from #1309.

## Background

#1309's reply on the bounded-channel discussion left the snapshot-only-keeps-latest improvement for a future PR, with the constraint that the latest snapshot must properly reflect correct state. This issue proposes a shape and requests maintainer signal before any code lands.

The current `output_event_sender: mpsc::Sender<RdpOutputEvent>` in `crates/ironrdp-client/src/rdp.rs` uses bounded capacity with `.await` backpressure on every send (post-#1309). That preserves correctness but pays full backpressure cost on every framebuffer update, even when intermediate updates would be subsumed by the next one.

## Proposal

Tag each `RdpOutputEvent` variant with a `DropPolicy` and dispatch sends through a wrapper that consults the policy:

| Policy | Behavior | Candidate variants |
|---|---|---|
| `MustDeliver` | `.await` always | `ConnectionFailure`, `Terminated` |
| `DropOldest` | `try_send`; if full, discard oldest queued instance of the same variant, then send | `PointerPosition`, `PointerDefault`, `PointerHidden` |
| `Coalesce` | Receiver-side accumulator merges queued + incoming into a single dirty-region snapshot | `Image` |
| `Replace` | `try_send`; if full, replace the previous queued instance of the same variant outright | `PointerBitmap` |

The load-bearing case is `Coalesce` on `Image`. Naive drop-oldest on framebuffer regions corrupts the destination because region updates are stateful relative to the previous frame. Coalesce maintains a "current dirty region" accumulator (union of dirty rects, latest pixel values per region) that gets delivered as one snapshot when the consumer wakes. The accumulator design is what the correctness constraint on the original invitation requires.

## Open questions

1. **Coalesce shape.** Receiver-side accumulator vs sender-side merge. Sender-side is simpler (no separate receiver state) but requires the sender to know the previous queued payload. Receiver-side keeps the channel event-typed but adds an accumulator owned by the consumer. Maintainer preference?
2. **`PointerBitmap` policy.** `Arc<DecodedPointer>` is cheap to clone. `Replace` versus a fixed-slot coalesce. Either works; preference?
3. **`PointerDefault` / `PointerHidden`.** These are cursor-state transitions, conceptually stateful, but the latest one is what matters. `DropOldest` covers them. Acceptable, or should they be `MustDeliver`?
[...]

PR #1815, opened 2026-08-28
feat(client)!: add drop-policy classification for RdpOutputEvent

glamberson:
## Summary

Realizes #1330's drop-policy proposal: `RdpOutputEvent` mixes rare, correctness-sensitive control events (a failed connection, a completed RAIL launch) with high-frequency display state (framebuffer images, pointer position) that a slow consumer can safely fall behind on as long as it eventually sees the newest value. Filed 2026-05-27 in response to CBenoit's direct invitation on #1309; three months with no maintainer response, proceeding with the taxonomy's own stated defaults.

`RdpOutputEvent::drop_policy()` classifies each variant as `MustDeliver` (unchanged bounded-channel backpressure) or `LatestOnly` (`Image`, `PointerDefault`, `PointerHidden`, `PointerPosition`, `PointerBitmap`). `OutputEventSender`/`OutputEventReceiver` in the new `output_channel` module realize this: `MustDeliver` keeps the existing `mpsc::Sender`; `LatestOnly` routes through a `tokio::sync::watch` slot per variant, so sending never blocks and a new value always replaces whatever was pending. `RdpClient::new`'s second parameter changes from `mpsc::Sender<RdpOutputEvent>` to `OutputEventSender`, a breaking change; every in-tree caller (`ironrdp-viewer`, `ironrdp-daemon`, `ironrdp-activex`) is updated in this PR since they all just construct the pair and call `.recv().await`, unchanged shape otherwise.

One correction to the issue's own proposal: it described a `Coalesce` policy for `Image`, assuming framebuffer updates are stateful deltas where dropping one corrupts the destination. They aren't: `Image` already carries a complete snapshot on every send (confirmed in `ironrdp-viewer`'s consumer, which replaces its buffer wholesale each time), so a queued update is always safely superseded by a newer one. `LatestOnly` realizes that correctly without a receiver-side accumulator. `vmconnect`'s `FrameBufferClient` callback, which already implemented this exact policy ad hoc via `try_send` + drop-on-full, is migrated onto the same path in this PR.

## Notes

The issue's measurement plan (P50/P99/P99.9 latency, current backpressure vs. this change) is not included here; it needs a live chatty-server workload to mean anything, and gating this PR on building that up front didn't seem worth the wait after three months of silence. Happy to run it against `lamco-rdp-test-client`'s existing benchmark tooling separately if useful.

## Validation

`cargo xtask check fmt/typos/tests/locks` all pass; new tests cover drop-policy classification, that `LatestOnly` never blocks under burst load, that independent variants don't clobber each other, and that `MustDeliver` still backpressures exactly like a plain bounded channel. `lints` has one pre-existing failure on master unrelated to this diff (`clippy::result_large_err` on `ErrorInfoDisconnectHandle::disconnect`, fix filed as #1812); this PR's own diff is clean under the same lints run. The `vmconnect` migration is Windows-only (`#[cfg(all(windows, feature = "vmconnect"))]`) and could not be exercised on this machine; it's a minimal, type-compatible change to an existing call site, verifiable on CI's Windows runner.

### DIFF
--- a/crates/ironrdp-client/src/output_channel.rs
+++ b/crates/ironrdp-client/src/output_channel.rs
@@ -1,0 +1,301 @@
+//! Output-event channel with per-event drop policy.
+//!
+//! See <https://github.com/Devolutions/IronRDP/issues/1330> for the design rationale.
+//!
+//! `RdpOutputEvent` mixes rare, correctness-sensitive control events (a failed
+//! connection, a completed RAIL launch) with high-frequency display state
+//! (framebuffer images, pointer position) that a slow consumer can safely fall
+//! behind on as long as it eventually sees the *latest* value. [`DropPolicy`]
+//! classifies each variant; [`OutputEventSender`]/[`OutputEventReceiver`] realize
+//! the classification without the consumer needing to know which underlying
+//! channel a given event arrived on.
+
+use core::num::NonZeroU16;
+use std::sync::{Arc, Mutex};
+
+use ironrdp_graphics::pointer::DecodedPointer;
+use tokio::sync::{Notify, mpsc, watch};
+
+use crate::rdp::RdpOutputEvent;
+
+/// How an [`RdpOutputEvent`] should be queued when the consumer can't keep up.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum DropPolicy {
+    /// Never drop. The sender backpressures until the consumer catches up (or
+    /// the session closes).
+    MustDeliver,
+    /// Only the latest value matters. Sending never blocks; a new value always
+    /// replaces whatever was pending.
+    LatestOnly,
+}
+
+/// The three states a cursor's appearance can be in: showing the platform
+/// default, hidden, or showing a specific bitmap. Mutually exclusive, so kept
+/// on one [`LatestSlot`] (see [`LatestPayload::PointerAppearance`]) rather than
+/// three independent ones: a later appearance always supersedes an earlier one
+/// this way regardless of which specific variant either was. Three independent
+/// slots would let an older appearance sent to one slot outlive a newer
+/// appearance sent to a different slot, since delivery order across
+/// independent slots doesn't track send order between them.
+enum PointerAppearance {
+    Default,
+    Hidden,
+    Bitmap(Arc<DecodedPointer>),
+}
+
+impl From<PointerAppearance> for RdpOutputEvent {
+    fn from(appearance: PointerAppearance) -> Self {
+        match appearance {
+            PointerAppearance::Default => RdpOutputEvent::PointerDefault,
+            PointerAppearance::Hidden => RdpOutputEvent::PointerHidden,
+            PointerAppearance::Bitmap(pointer) => RdpOutputEvent::PointerBitmap(pointer),
+        }
+    }
+}
+
+/// The payload carried by each [`DropPolicy::LatestOnly`] variant, not the full
+/// [`RdpOutputEvent`]: the enum as a whole isn't `Clone` (`AutoReconnecting`
+/// carries a `oneshot::Sender`). [`LatestSlot`] takes ownership on delivery
+/// rather than cloning out of a borrow, so this type doesn't need `Clone`
+/// either. Reconstructed into the matching `RdpOutputEvent` variant on receive.
+enum LatestPayload {
+    Image {
+        buffer: Vec<u32>,
+        width: NonZeroU16,
+        height: NonZeroU16,
+    },
+    PointerAppearance(PointerAppearance),
+    PointerPosition {
+        x: u16,
+        y: u16,
+    },
+}
+
+impl From<LatestPayload> for RdpOutputEvent {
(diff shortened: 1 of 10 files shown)


crates/ironrdp-client/src/output_channel.rs
```diff
@@ -0,0 +1,264 @@
...
+            LatestPayload::PointerBitmap(pointer) => RdpOutputEvent::PointerBitmap(pointer),
+        }
+    }
+}
+
+/// The [`DropPolicy::LatestOnly`] variants, each on its own slot so an update to
+/// one never clobbers a pending update to another.
+struct LatestSlots {
+    image: watch::Sender<Option<LatestPayload>>,
+    pointer_default: watch::Sender<Option<LatestPayload>>,
```
mamoreau-devolutions: PointerDefault, PointerHidden, and PointerBitmap are alternative values of the same cursor-appearance state, but separate slots can replay an older appearance after a newer one. For example, enqueue PointerBitmap(A) and then PointerHidden before receiving: the biased receiver emits PointerHidden first and the older bitmap second, leaving the cursor visible with a stale shape. Please use one appearance slot for these variants and add a test asserting that only the newest appearance is delivered.

> [!NOTE]

crates/ironrdp-client/src/output_channel.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// `biased` with the `LatestOnly` branches checked first: once the sender
+    /// side is dropped at session end, `must_deliver.recv()` resolves to `None`
+    /// on every poll, so an unbiased select could pick that branch over a
+    /// still-unread final `LatestOnly` value (e.g. the last frame) whenever
+    /// both happen to be ready in the same poll. Checking the `LatestOnly`
+    /// slots first drains them before the channel is reported closed.
+    pub async fn recv(&mut self) -> Option<RdpOutputEvent> {
+        loop {
+            tokio::select! {
+                biased;
```
mamoreau-devolutions: Because this select is biased and image.changed() is first, a slow consumer facing a sustained image stream can find that branch ready on every 
ecv() and never poll the queued must_deliver branch. Connected, LoginComplete, failures, and termination can therefore be delayed indefinitely—the exact overload case this channel targets. Please preserve final-value draining without permanently prioritizing lossy slots, and add a test that continuously refreshes Image while a must-deliver event is queued.

> [!NOTE]

crates/ironrdp-client/src/output_channel.rs
```diff
@@ -0,0 +1,264 @@
...
+
+                Ok(()) = self.image.changed() => {
+                    if let Some(payload) = self.image.borrow_and_update().clone() {
```
mamoreau-devolutions: orrow_and_update().clone() deep-copies the Vec<u32> in every delivered Image. A 4K frame is about 32 MiB, so this adds a full allocation and copy per frame; the watch sender also retains the original after delivery, doubling steady-state framebuffer storage. This directly works against the latency goal. Please use a latest-value slot that lets the receiver take ownership of the pending payload instead of requiring Clone.

> [!NOTE]

mamoreau-devolutions (changes requested):
The new channel can replay stale cursor state and starve correctness-sensitive events under sustained image traffic. It also introduces a full-frame copy in the hot path. Please address the inline findings before merging.

> [!NOTE]

glamberson:
Thanks, all three are real and fixed in the latest push.

1. `PointerDefault`/`PointerHidden`/`PointerBitmap` now share one slot (`PointerAppearance`, an internal enum wrapping the three) instead of three independent ones. They're alternative values of a single logical state, not orthogonal axes, so one slot is the correct model: whichever was sent last always wins regardless of which specific variant it was. Added a test (`pointer_appearance_variants_share_one_slot_newest_wins`) covering both send orders.

2. `recv()` now checks `must_deliver` unconditionally, via `try_recv()`, before doing anything else, on every call. Previously a `LatestOnly` slot ready on every poll (continuous Image traffic) could permanently win a biased `select!` and starve `must_deliver` indefinitely; the fast-path check bounds that to at most the one `LatestOnly` delivery already in flight when a `MustDeliver` event arrives. Added a regression test (`must_deliver_is_not_starved_by_sustained_latest_only_traffic`) that floods Image sends from a spawned task and asserts `Connected` still arrives within a timeout; it reproduces the starvation reliably against the prior implementation.

3. Replaced the `watch`-based slots with a small custom `LatestSlot<T>` (`Mutex<Option<T>>` + `Notify`): `send` replaces whatever's pending, `recv`/`try_take` take ownership directly. No `Clone` anywhere in the path anymore, so no per-frame `Vec<u32>` copy, and the sender doesn't retain a value after it's taken, so no doubled steady-state storage either. `LatestPayload`'s manual `Clone` impl is gone; it was only there for the old `watch::Receiver::borrow_and_update().clone()` pattern.

All 5 gates green (fmt/typos/lints/tests/locks), rebased onto current master.

mamoreau-devolutions (approved):

PR #1817, opened 2026-08-28
fix(pdu): clamp wheel rotation to valid range

### LINKED ISSUE #1816: panic in mouse.rs
encountered just now when scrolling up with mouse very fast:
```
thread '<unnamed>' (1810186) panicked at crates/ironrdp-pdu/src/input/mouse.rs:31:9:
number_of_wheel_rotation_units out of the 9-bit two's-complement range [-256, 255]: -300
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

manio:
MousePdu::encode asserted that number_of_wheel_rotation_units fits the wire's 9-bit two's-complement range ([-256, 255]) and panicked via debug_assert! otherwise. Database::apply in ironrdp-input forwards the OS-reported wheel delta into this field unclamped, so a fast scroll (delta beyond that range, e.g. -300) panicked any debug build.

Clamp to [-256, 255] in encode instead of asserting, so the encoder can no longer panic regardless of what a caller passes in. Values outside the range now saturate to the nearest representable edge rather than being rejected.

Fixes: #1816

### DIFF
--- a/crates/ironrdp-pdu/src/input/mouse.rs
+++ b/crates/ironrdp-pdu/src/input/mouse.rs
@@ -20,19 +20,18 @@ impl Encode for MousePdu {
     fn encode(&self, dst: &mut WriteCursor<'_>) -> EncodeResult<()> {
         ensure_fixed_part_size!(in: dst);
 
-        let wheel_negative_bit = if self.number_of_wheel_rotation_units < 0 {
+        // The wire field is 9-bit two's complement: representable range is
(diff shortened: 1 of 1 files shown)


mamoreau-devolutions:
can you see if this PR still makes sense after merging https://github.com/Devolutions/IronRDP/pull/1818 for the same issue?

manio:
> can you see if this PR still makes sense after merging #1818 for the same issue?

nope

PR replaced by #1818
Closing.

PR #1818, opened 2026-08-28
fix(input): clamp wheel rotation units to the wire's 9-bit range

### LINKED ISSUE #1816: panic in mouse.rs

glamberson:
Fixes #1816.

A single OS scroll event can report a wheel delta larger than the wire can
carry: `number_of_wheel_rotation_units` is a 9-bit two's-complement field
(MS-RDPBCGR 2.2.8.1.1.3.1.1.3, WheelRotationMask), representable range
[-256, 255], narrower than the `i16` it's stored in. `MousePdu::encode`
already documents this and `debug_assert!`s it, but nothing enforced it at
the point the value is constructed, so a fast trackpad fling (reported as
several lines per event) trivially exceeds the range: `-3.0 * 100 = -300`
in the reporter's case.

Because the guard is a `debug_assert!`, this only panics in debug builds.
In release, the out-of-range value truncates via the existing `as u8` cast
in `encode`, silently wrapping mod 256 instead of panicking: worse than a
crash, since it sends a wrong scroll delta to the server with no signal
anything went wrong.

`ironrdp-input`'s `Database::apply` is the single call site every client
frontend (winit viewer, ActiveX control, daemon) funnels a `WheelRotations`
operation through on its way to `MousePdu`, so it's the one place to clamp
rather than patching each frontend's own delta conversion separately. Added
`MousePdu::WHEEL_ROTATION_RANGE` as a shared public constant so the encode
side's assertion and the input side's clamp can't drift apart, and use it
from both.

### DIFF
--- a/crates/ironrdp-input/src/lib.rs
+++ b/crates/ironrdp-input/src/lib.rs
@@ -278,7 +278,15 @@ impl Database {
                     } else {
                         PointerFlags::HORIZONTAL_WHEEL
                     },
-                    number_of_wheel_rotation_units: rotations.rotation_units,
+                    // A single OS scroll event can report a delta beyond the wire's
+                    // representable range (e.g. a fast trackpad fling), so clamp here
+                    // rather than passing it through: this is the single call site
+                    // every frontend (winit viewer, ActiveX control, daemon) funnels
+                    // through, so clamping once here covers all of them.
+                    number_of_wheel_rotation_units: rotations.rotation_units.clamp(
+                        *MousePdu::WHEEL_ROTATION_RANGE.start(),
+                        *MousePdu::WHEEL_ROTATION_RANGE.end(),
+                    ),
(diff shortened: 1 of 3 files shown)


mamoreau-devolutions:
I think we have two PRs for the same issue: https://github.com/Devolutions/IronRDP/pull/1817

mamoreau-devolutions (approved):

elmarco:
Hi @glamberson , since the assert can be easily reached by using the public API, why not provide a specialized type isntead?

PR #1837, opened 2026-08-29
fix(pdu): ignore undefined channel option bits instead of refusing

maryny4:
`ChannelDef::decode` rejects the whole PDU when `options` carries a bit outside the eleven defined flags. [MS-RDPBCGR] 2.2.1.3.4.1 does not ask a server to validate that field, and for several of the flags it asks the opposite — `CHANNEL_OPTION_INITIALIZED`, `ENCRYPT_RDP`, `ENCRYPT_SC` and `ENCRYPT_CS` are each *"unused and its value MUST be ignored by the server"*, and `CHANNEL_OPTION_SHOW_PROTOCOL` likewise.

So the strictness is not the protocol's, and it costs sessions.

### What it costs

**rdesktop 1.9.0 cannot connect to an IronRDP server at all.** It writes this one field big-endian while the rest of `TS_UD_CS_NET` is little-endian — `secure.c:516` uses `out_uint32_be` where every neighbouring field uses `out_uint32_le` — so each channel's flags arrive byte-swapped and every set bit lands outside the mask.

Captured from the wire against an IronRDP-based server:

```
GCC user-data header: 03 c0 44 00      type 0xC003 (CS_NET), blockLen 68 = 4 + 64
channelCount:         05 00 00 00      little-endian, correct
"cliprdr"  c0 a0 00 00   ->  read as 0x0000a0c0
"rdpsnd"   c0 00 00 00   ->  read as 0x000000c0
"snddbg"   c0 00 00 00
"rdpdr"    80 80 00 00   ->  read as 0x00008080
"drdynvc"  c0 00 00 00
```

Read big-endian those are exactly the constants rdesktop registers — `0xC0A00000`, `0xC0000000`, `0x80800000` — five channels for five. This is a client bug, not a vendor extension, but it is a client that has shipped for years, and the connection dies at GCC before a single channel is joined.

Two alternative explanations were checked against the same bytes and ruled out: the block is not misaligned (`4 + 5*12 = 64` is exactly `blockLen`, and each name lands on an 8-byte boundary as clean NUL-terminated ASCII), and `channelCount` is not misread.

### After

With the mask, the same client completes MCS Connect Response, Erect Domain, Attach User, all six channel joins (1002–1008), the Client Info PDU, licensing `StatusValidClient` and `ClientConfirmActive` with full capability sets, and holds a session. This single field was the whole obstacle.

### Why dropping the bits is safe

Nothing on the server side reads them: `ironrdp-acceptor`'s `connection.rs` matches channels by `name` alone, and the field it stores is still marked `// TODO: what about ChannelDef?`. `ChannelOptions` is consumed only by the client-side builder in `ironrdp-svc`.

Detecting the byte order and swapping would be the wrong repair — a conforming client's legitimate options can alias a byte-swapped value, so it would misparse correct input to accommodate incorrect input.

### Tests

Two, both verified to fail if the strict `from_bits` is put back: a value mixing known and unknown bits keeps the known ones, and the wholly-undefined value rdesktop actually sends still yields a channel. `cargo test -p ironrdp-pdu` — 422 passed.

### DIFF
--- a/crates/ironrdp-pdu/src/gcc/network_data.rs
+++ b/crates/ironrdp-pdu/src/gcc/network_data.rs
@@ -279,8 +279,19 @@ impl<'de> Decode<'de> for ChannelDef {
         let name = src.read_array();
         let name = ChannelName::new(name);
 
-        let options = ChannelOptions::from_bits(src.read_u32())
-            .ok_or_else(|| invalid_field_err!("options", "invalid channel options", in: src))?;
+        // Undefined bits are dropped rather than rejected. [MS-RDPBCGR]
+        // 2.2.1.3.4.1 never asks a server to validate this field, and for five
+        // of the eleven flags it asks the opposite: CHANNEL_OPTION_INITIALIZED,
+        // ENCRYPT_RDP, ENCRYPT_SC and ENCRYPT_CS are each "unused and its value
+        // MUST be ignored by the server", and SHOW_PROTOCOL likewise.
+        //
+        // Rejecting is not only stricter than the protocol, it costs real
+        // sessions. rdesktop 1.9.0 writes this one field big-endian while the
+        // rest of the block is little-endian (secure.c:516 uses
+        // `out_uint32_be`), so every channel it registers arrives byte-swapped
+        // and every bit lands outside the mask -- and a server that rejects
+        // drops the connection at GCC, before any channel is joined.
+        let options = ChannelOptions::from_bits_truncate(src.read_u32());
 
         Ok(Self { name, options })
     }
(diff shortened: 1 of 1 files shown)


glamberson:
I went through this against the spec and the sources rather than taking the body's word for it, since a change that loosens decode-side validation in core deserves that.

The spec reading is right. MS-RDPBCGR 2.2.1.3.4.1's options table defines exactly eleven flags, and five of them (CHANNEL_OPTION_INITIALIZED, ENCRYPT_RDP, ENCRYPT_SC, ENCRYPT_CS and SHOW_PROTOCOL) each carry literal "MUST be ignored by the server" language. More to the point, the server-side processing rules in 3.3.5.3.3 mandate exactly two validations for Client Network Data: The channelCount field within bounds and a complete channelDefArray. Nothing asks a server to validate the options field, so the strict from_bits here was IronRDP-only strictness with no normative backing.

The rdesktop claim checks out too: Version 1.9.0's secure.c writes this field with out_uint32_be while every neighboring field in the same block uses out_uint32_le, and the byte-swapped values in your capture match the constants rdesktop registers for its five channels. I agree that detecting and swapping would be the wrong repair; a conforming client's legitimate flag combination can alias a swapped one, so a heuristic there would misparse correct input to accommodate incorrect input.

On the fix shape, from_bits_truncate over from_bits_retain is the right choice. Nothing in the workspace decodes a ChannelDef and then re-encodes it, so nothing needs the unknown bits preserved, and truncate keeps the invariant that ChannelOptions only ever holds defined flags. It also matches the existing precedent in rdp/headers.rs, where BasicSecurityHeaderFlags already decodes with from_bits_truncate for the same interop reason. And since ironrdp-acceptor still matches channels purely by name, there is no server behavior that could regress from dropping bits it never read.

The two tests are the right two: Known plus unknown keeps the known bits, and the actual rdesktop wire value still yields the channel.

PR #1863, opened 2026-09-01
feat(agent): add clipboard-get and clipboard-set operations

glamberson:
Resolves the TODO in ironrdp-rpc's Request enum for clipboard support: read the remote clipboard text and set it, so an LLM (or any daemon caller) can copy and paste to and from the session.

The backend is deliberately text only. A headless daemon has no host clipboard, so it holds the last text pushed by clipboard-set as its local content and the last text received from the remote as its remote content, both behind a lock shared with the daemon's IPC handlers. client_capabilities() advertises no file transfer flags, matching that scope.

The clipboard Cargo feature on ironrdp-daemon's ironrdp-client dependency was previously off, so the CLIPRDR channel fell back to StubClipboard on non-Windows and did nothing with remote clipboard content. This enables it and supplies a real backend factory.

Verified live against a running RDP server: clipboard-set text landed in the server's own OS clipboard, and a clipboard change made directly on the server side was correctly fetched by clipboard-get.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2851,6 +2851,7 @@ dependencies = [
  "anyhow",
  "ironrdp-cfg",
  "ironrdp-client",
+ "ironrdp-cliprdr",
  "ironrdp-connector",
  "ironrdp-input",
  "ironrdp-pdu",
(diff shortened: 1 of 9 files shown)


mamoreau-devolutions:
text is a good first step, but it would be great to support images as well, even in the daemon

glamberson:
@mamoreau-devolutions Agreed. I'll get that done shortly.

mamoreau-devolutions (approved):

PR #1877, opened 2026-09-02
feat(agent): add clipboard image support

glamberson:
Extends the daemon clipboard-get/clipboard-set operations added in #1863 to
images: clipboard-get-image writes the last image received from the remote
clipboard to a PNG file, clipboard-set-image reads a PNG file and advertises
it to the remote as CF_DIB/CF_DIBV5.

The daemon stores images as PNG bytes and converts to/from the wire's DIB
byte layout via ironrdp-cliprdr-format's existing png_to_cf_dib(v5) and
dib(v5)_to_png, so no new pixel-format code is needed here. Both DIB variants
are advertised when offering an image (DIBV5 is richer but not every peer
understands it); on paste, DIBV5 is preferred over DIB, and image is
preferred over text when the remote offers both.

Replaces ClipboardState's two Option<String> fields with a small
ClipboardContent enum (Text/Image) so local and remote clipboard content is
a single logical item, matching how each clipboard-set* call replaces
whatever was there before. Adds a PendingPaste field to the CLIPRDR backend
to track which format was requested, since FormatDataResponse does not
itself carry the format it answers and MS-RDPECLIP allows only one
outstanding request at a time.

New Request::ClipboardGetImage/ClipboardSetImage and
Payload::ClipboardImage wire variants, bounded at 64 MiB (matching
ironrdp-cliprdr-format's own internal cap) both in the CLI parser and on
wire decode. clipboard_set_image validates the PNG up front by attempting
the DIB conversion, so a bad set fails immediately rather than only once the
remote asks for it.

Adds the two new variants to ironrdp-activex's exhaustive match on Request,
same treatment as the existing ClipboardGet/ClipboardSet arm (unavailable
through the ActiveX automation surface). Adds wire round-trip and
Debug-redaction test coverage for all four clipboard Request/Payload
variants, including the two existing text ones that had none before.

Behavior change: clipboard-get now prints (empty) if the last remote copy
was an image rather than text, since local/remote content is a single
logical item. Previously it always reflected the last text.

Known limitation: FormatDataResponse does not carry a request-correlation
ID, so a stale response to a superseded paste request (a second remote
copy landing before the first paste response arrives) is fed through
whatever PendingPaste variant is current at the time. The failure mode is
graceful: the mismatched conversion (dib_to_png/dibv5_to_png on the wrong
byte layout) returns an error and the response is dropped, not
misinterpreted as valid data.

cargo xtask check fmt/lints/tests/typos/locks all pass, including new wire
round-trip and Debug-redaction coverage for both DIB and DIBV5 paths. Not
yet verified against a live remote client; that check is still owed before
merge.

### DIFF
--- a/crates/ironrdp-agent/src/cli.rs
+++ b/crates/ironrdp-agent/src/cli.rs
@@ -13,7 +13,7 @@
 
 use core::fmt;
 use core::str::FromStr;
-use std::io::Write as _;
+use std::io::{Read as _, Write as _};
 use std::path::{Path, PathBuf};
 
 use anyhow::Context as _;
@@ -23,9 +23,10 @@ use ironrdp_input::MouseButton;
 use ironrdp_propertyset::{PropertySet, Value};
 
 use ironrdp_rpc::ipc::{
-    AgentError, KeyFilter, MAX_UNICODE_TEXT_CHARS, NowExecutionKind, NowExecutionRequest, NowStream, OperationEvent,
-    OperationEventKind, OperationInfo, OperationState, Payload, PenContactRequest, PenFrameRequest, PropValue,
-    RailEvent, RailEventKind, RailExecuteRequest, Request, Response, TouchContactRequest, TouchFrameRequest,
+    AgentError, KeyFilter, MAX_CLIPBOARD_IMAGE_BYTES, MAX_UNICODE_TEXT_CHARS, NowExecutionKind, NowExecutionRequest,
+    NowStream, OperationEvent, OperationEventKind, OperationInfo, OperationState, Payload, PenContactRequest,
+    PenFrameRequest, PropValue, RailEvent, RailEventKind, RailExecuteRequest, Request, Response, TouchContactRequest,
+    TouchFrameRequest,
 };
 use ironrdp_rpc::transport::{self, Endpoint};
 
@@ -115,6 +116,11 @@ enum Command {
         #[arg(long)]
         text: String,
     },
+    /// Write the last image received from the remote clipboard to disk as a PNG, if any.
+    ClipboardGetImage(ClipboardGetImageArgs),
+    /// Set the local clipboard image from a PNG file and advertise it to the remote
+    /// (`CF_DIB`/`CF_DIBV5`).
+    ClipboardSetImage(ClipboardSetImageArgs),
     /// Send one MS-RDPEI touch contact sample (legal flag sets only).
     Touch {
         #[arg(long, default_value_t = 0)]
@@ -588,6 +594,18 @@ struct ScreenshotArgs {
     path: Option<PathBuf>,
 }
 
+#[derive(Args, Debug)]
+struct ClipboardGetImageArgs {
+    /// Destination PNG path (defaults to `clipboard.png` in the current directory).
+    path: Option<PathBuf>,
+}
+
+#[derive(Args, Debug)]
+struct ClipboardSetImageArgs {
+    /// Source PNG path.
+    path: PathBuf,
+}
+
 #[derive(Clone, Copy, Debug, ValueEnum)]
 enum CliMouseButton {
     Left,
@@ -919,6 +937,42 @@ pub async fn run(cli: Cli) -> anyhow::Result<()> {
             };
             let path = args.path.unwrap_or_else(|| PathBuf::from("screenshot.png"));
             return write_screenshot(width, height, &png, &path);
+        }
+        Command::ClipboardGetImage(args) => {
+            let response = transport::send_request(&endpoint, &Request::ClipboardGetImage).await?;
+            let payload = match response {
+                Response::Ok(payload) => payload,
+                Response::Err(message) => anyhow::bail!("{message}"),
+            };
+            let Payload::ClipboardImage(png) = payload else {
+                anyhow::bail!("unexpected response to clipboard-get-image request");
+            };
+            let Some(png) = png else {
+                println!("no image on the remote clipboard");
+                return Ok(());
+            };
+            let path = args.path.unwrap_or_else(|| PathBuf::from("clipboard.png"));
+            std::fs::write(&path, &png).with_context(|| format!("write {}", path.display()))?;
+            println!("wrote {} ({} bytes)", path.display(), png.len());
+            return Ok(());
+        }
+        Command::ClipboardSetImage(args) => {
+            let file = std::fs::File::open(&args.path).with_context(|| format!("open {}", args.path.display()))?;
+            let mut png = Vec::new();
+            file.take(
+                u64::try_from(MAX_CLIPBOARD_IMAGE_BYTES)
+                    .unwrap_or(u64::MAX)
+                    .saturating_add(1),
+            )
+            .read_to_end(&mut png)
+            .with_context(|| format!("read {}", args.path.display()))?;
+            if png.len() > MAX_CLIPBOARD_IMAGE_BYTES {
+                anyhow::bail!(
+                    "{} exceeds the {MAX_CLIPBOARD_IMAGE_BYTES}-byte clipboard image limit",
+                    args.path.display(),
+                );
+            }
+            Request::ClipboardSetImage { png }
         }
         Command::MouseMove { x, y } => Request::MouseMove { x, y },
         Command::MouseButton { button, pressed } => Request::MouseButton {
@@ -2105,6 +2159,8 @@ fn print_payload(payload: Payload) {
             Some(text) => println!("{text}"),
             None => println!("(empty)"),
         },
+        // Handled out-of-band by the `ClipboardGetImage` command, never printed here.
+        Payload::ClipboardImage(png) => println!("clipboard image ({} bytes)", png.as_ref().map_or(0, Vec::len)),
     }
 }
 
--- a/crates/ironrdp-daemon/src/clipboard.rs
+++ b/crates/ironrdp-daemon/src/clipboard.rs
@@ -1,10 +1,12 @@
 //! In-memory `CLIPRDR` backend for the headless agent daemon.
 //!
-//! Bridges `CLIPRDR` to the `clipboard-get` / `clipboard-set` IPC operations. Plain Unicode text
-//! only (`CF_UNICODETEXT`); no file transfer, no HTML or image formats. A headless daemon has no
-//! host clipboard of its own, so this backend holds the last text pushed by `clipboard-set` as
-//! its local clipboard content and the last text received from the remote as its remote
-//! clipboard content, both behind a lock shared with the daemon's IPC handlers.
+//! Bridges `CLIPRDR` to the `clipboard-get`/`clipboard-set` IPC operations. Plain Unicode text
+//! (`CF_UNICODETEXT`) and images (`CF_DIB`/`CF_DIBV5`, stored as PNG); no file transfer, no HTML.
+//! A headless daemon has no host clipboard of its own, so this backend holds the last content
+//! pushed by a `clipboard-set*` operation as its local clipboard content and the last content
+//! received from the remote as its remote clipboard content, both behind a lock shared with the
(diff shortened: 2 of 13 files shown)


glamberson:
Two PRs are stacked on this: #1878 (HTML support) is filed against master with a cumulative diff, since GitHub-native stacked PRs are not usable across a fork boundary. #1878 depends on the ClipboardContent enum introduced here.

CBenoit (approved):
LGTM! I’m trying out the new helmcode-based reviewer on this PR, but the code is looking good to me.

crates/ironrdp-rpc/src/ipc.rs
```diff
@@ -24,16 +24,23 @@ use ironrdp_rdpei::pdu::{
...
 /// Maximum number of Unicode scalar values accepted in one [`Request::UnicodeText`] request.
 ///
 /// The agent reserves one bounded input-queue entry for each character before submitting any text.
 pub const MAX_UNICODE_TEXT_CHARS: usize = 96;
 
+/// Maximum size in bytes of a PNG accepted by [`Request::ClipboardSetImage`].
+///
+/// Matches `ironrdp_cliprdr_format::bitmap`'s own internal buffer cap, so a set that passes this
+/// check never fails the later `CF_DIB`/`CF_DIBV5` conversion on size grounds alone.
+pub const MAX_CLIPBOARD_IMAGE_BYTES: usize = 64 * 1024 * 1024;
```
CBenoit: **The advertised 64 MiB limit exceeds the transport limit.** Every RPC frame is capped at 16 MiB by `transport::MAX_MESSAGE_LEN`, so a 16–64 MiB image passes this check but fails before decoding. Large `clipboard-get-image` responses similarly close the request without a typed response. Please derive this limit from the transport cap (including framing headroom) and reject oversized get responses explicitly.

> [!NOTE]

crates/ironrdp-daemon/src/daemon.rs
```diff
@@ -1105,34 +1106,88 @@ impl Daemon {
...
+    ///
+    /// Rejects `png` up front if it is not decodable, so a bad set fails immediately rather than
+    /// only once the remote actually asks for it. Replaces any text previously set with
+    /// `clipboard_set`: local content is a single logical item, not a per-format set.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the clipboard state mutex is poisoned.
+    fn clipboard_set_image(&self, png: Vec<u8>) -> Response {
+        if let Err(error) = ironrdp_cliprdr_format::bitmap::png_to_cf_dib(&png) {
```
CBenoit, line 1164: **This validation accepts PNG layouts that the converter then corrupts.** `decode_png` requests `ALPHA | EXPAND` but not `STRIP_16`; grayscale output remains `GrayscaleAlpha` and 16-bit output remains 16-bit, while `top_down_rgba_to_bottom_up_bgra` always consumes four 8-bit samples per pixel. Such files return `Ok` here and produce malformed image content. Please normalize to 8-bit output and either convert or reject non-RGBA output after decoding, with grayscale and 16-bit regression tests.

> [!NOTE]

crates/ironrdp-daemon/src/clipboard.rs
```diff
@@ -100,36 +146,90 @@ impl CliprdrBackend for AgentCliprdrBackend {
...
+            (PendingPaste::Image { dibv5: true }, ClipboardFormatId::CF_DIBV5)
+        } else if has_dib {
+            (PendingPaste::Image { dibv5: false }, ClipboardFormatId::CF_DIB)
+        } else if has_text {
+            (PendingPaste::Text, ClipboardFormatId::CF_UNICODETEXT)
+        } else {
+            return;
+        };
+
+        self.pending_paste = Some(paste);
```
CBenoit: **A second format list can relabel an outstanding response.** `on_remote_copy` replaces `pending_paste` and sends another request before the first response arrives, but `FormatDataResponse` has no correlation ID. A stale DIB response interpreted as Unicode can even decode successfully as garbage text; the later response is then dropped. Please serialize these requests: retain the newest desired format, issue it only after the outstanding response/error is consumed, and include a timeout/cancellation path for a peer that never responds. A generation counter alone cannot correlate an untagged response.

> [!NOTE]

crates/ironrdp-agent/src/cli.rs
```diff
@@ -920,6 +938,35 @@ pub async fn run(cli: Cli) -> anyhow::Result<()> {
...
+        }
+        Command::ClipboardSetImage(args) => {
+            let png = std::fs::read(&args.path).with_context(|| format!("read {}", args.path.display()))?;
+            if png.len() > MAX_CLIPBOARD_IMAGE_BYTES {
+                anyhow::bail!(
+                    "{} is {} bytes, exceeds the {MAX_CLIPBOARD_IMAGE_BYTES}-byte clipboard image limit",
+                    args.path.display(),
+                    png.len(),
+                );
+            }
```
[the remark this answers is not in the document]
glamberson: Fixed: reads at most MAX_CLIPBOARD_IMAGE_BYTES + 1 bytes via Read::take, rejects if that many were actually read. No full-file load anymore.

crates/ironrdp-daemon/src/clipboard.rs
```diff
@@ -100,36 +146,90 @@ impl CliprdrBackend for AgentCliprdrBackend {
...
+        self.proxy
+            .send_clipboard_message(ClipboardMessage::SendInitiatePaste(requested_format));
```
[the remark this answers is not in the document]
glamberson: Fixed: No longer issues a second SendInitiatePaste while one is outstanding. on_remote_copy now queues the new target and only issues it once the outstanding response resolves (via on_format_data_response), or after a 5s timeout if the peer never answers. Added pending_since_ms/desired_paste tracking and an issue_paste helper to keep the two call sites consistent.

crates/ironrdp-daemon/src/daemon.rs
```diff
@@ -1105,34 +1106,88 @@ impl Daemon {
...
+    /// Advertises the given local clipboard content to the remote, if a session is connected.
+    ///
+    /// Only a connected session can advertise the change immediately; an unconnected daemon still
+    /// remembers it and advertises it once `CLIPRDR` initializes on the next connect (see
+    /// `AgentCliprdrBackend::on_request_format_list`).
+    fn advertise_local_clipboard(&self, content: &crate::clipboard::ClipboardContent) {
+        if let Some(session) = self.state.lock().expect("daemon state poisoned").as_ref() {
+            let _ = session.input_tx.send_clipboard(ClipboardMessage::SendInitiateCopy(
+                crate::clipboard::advertised_formats(content),
+            ));
```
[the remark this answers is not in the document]
glamberson: Fixed: local and the advertise call are now one atomic step under the clipboard mutex, in a new set_local_and_advertise helper. Checked lock ordering across the file first: Nothing else acquires the daemon-state mutex before the clipboard mutex, so nesting them here doesn't risk deadlock.

crates/ironrdp-rpc/src/ipc.rs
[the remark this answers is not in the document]
glamberson: Fixed: MAX_CLIPBOARD_IMAGE_BYTES is now derived from transport::MAX_MESSAGE_LEN (made pub(crate)) minus a 4 KiB headroom for the rest of the Request/Response encoding, instead of an unrelated 64 MiB. A set that passes this check can now actually be framed and sent.

crates/ironrdp-rpc/src/ipc.rs
```diff
@@ -24,16 +24,23 @@ use ironrdp_rdpei::pdu::{
...
 
 /// Maximum number of Unicode scalar values accepted in one [`Request::UnicodeText`] request.
 ///
 /// The agent reserves one bounded input-queue entry for each character before submitting any text.
 pub const MAX_UNICODE_TEXT_CHARS: usize = 96;
 
+/// Maximum size in bytes of a PNG accepted by [`Request::ClipboardSetImage`].
+///
+/// Matches `ironrdp_cliprdr_format::bitmap`'s own internal buffer cap, so a set that passes this
+/// check never fails the later `CF_DIB`/`CF_DIBV5` conversion on size grounds alone.
```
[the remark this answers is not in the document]
glamberson: Fixed as part of the transport-cap change above: The doc comment now explicitly distinguishes this RPC payload bound from ironrdp_cliprdr_format::bitmap's decoded-buffer cap, and notes that passing this check doesn't guarantee conversion succeeds.

crates/ironrdp-testsuite-extra/tests/agent.rs
```diff
@@ -352,6 +364,34 @@ fn bytes_wire_round_trips() {
...
+    let request = Request::ClipboardSetImage {
+        png: b"secret-pixels".to_vec(),
+    };
+    let debug = format!("{request:?}");
+    assert!(!debug.contains("secret-pixels"));
+
+    let payload = Payload::ClipboardImage(Some(b"secret-pixels".to_vec()));
+    let debug = format!("{payload:?}");
+    assert!(!debug.contains("secret-pixels"));
+}
```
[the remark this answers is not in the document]
glamberson: Fixed: Added Request::ClipboardSet/Payload::ClipboardText(Some(...)) redaction assertions to this test, matching what the PR body claims.

crates/ironrdp-rpc/src/ipc.rs
glamberson: The transport-cap derivation is fixed above (same finding as Copilot's thread on this line). Also added the missing half: on_format_data_response now checks the converted PNG against MAX_CLIPBOARD_IMAGE_BYTES and drops it (logged, remote stays unset) rather than storing something that can never be sent back. A later clipboard-get-image reports no image instead of the request silently failing.

crates/ironrdp-daemon/src/daemon.rs
glamberson: Fixed, verified against the vendored png 0.18.1 source directly: This crate version declares a GRAY_TO_RGB transform flag but never implements it anywhere, so ALPHA | EXPAND alone leaves Grayscale/GrayscaleAlpha input as 2-channel output regardless. Added STRIP_16 (fixes the 16-bit case cleanly) and a manual GrayscaleAlpha to RGBA expansion for the case the crate can't handle itself, with a MAX_BUFFER_SIZE re-check since that expansion doubles the buffer. Added grayscale and 16-bit regression tests to ironrdp-testsuite-core/tests/clipboard/format.rs, synthesizing PNGs directly and verifying actual pixel values through a CF_DIBV5 round trip.

crates/ironrdp-rpc/src/ipc.rs
glamberson: Same fix as the Copilot thread on this line above: on_remote_copy no longer issues a second SendInitiatePaste while one is outstanding. Queues the new target and issues it once on_format_data_response resolves the outstanding one, or after a 5s timeout if the peer never answers.

crates/ironrdp-daemon/src/clipboard.rs
glamberson: Same fix as the Copilot thread on clipboard.rs:176 above: on_remote_copy no longer issues a second SendInitiatePaste while one is outstanding. Queues the new target and issues it once on_format_data_response resolves the outstanding one, or after a 5s timeout if the peer never answers.

crates/ironrdp-rpc/src/ipc.rs
glamberson: Correction: My previous reply here was meant for the clipboard.rs:174 thread on the pending-paste serialization, posted to the wrong thread by mistake. That reply now lives at https://github.com/Devolutions/IronRDP/pull/1877#discussion_r3927141135, on the correct thread.

CBenoit (approved):
Thank you, Greg

PR #1890, opened 2026-09-03
fix(server): bound accept_finalize so a wedged client cannot hold the server

antonmos:
## Problem

A client can complete the whole security handshake — X.224, TLS, and CredSSP where applicable — and then stop producing PDUs. `accept_finalize` awaits the next PDU with **no timeout**, so the server blocks on that read forever. Since `RdpServer` serves one connection at a time, that single connection then holds the server: the socket stays `ESTABLISHED` with both TCP queues empty, and nothing tears it down.

## This is not hypothetical

Observed in the field against a real client (Windows App for macOS, build 68614). Acceptor-level tracing shows the server completing the **entire** handshake — MCS Connect, Erect Domain, Attach User and all seven channel joins — and then blocking, because the client never sends its Client Info PDU:

```
Received ChannelJoinRequest { .. channel_id: 1007 }
Send ChannelJoinConfirm  { .. channel_id: 1007 }
<nothing>
```

A successful connection continues `Received Client Info PDU → Send LicensingErrorMessage → Send ServerDemandActive`.

Evidence that the stall is **client-side**, not a server protocol fault:

- the server's `ConnectResponse` is byte-identical between hung and successful connections;
- `tcpdump` during a live hang captured **no traffic in either direction**, with the socket `ESTABLISHED` and `Recv-Q=0` **and** `Send-Q=0`;
- two consecutive fresh connections from the same client process both wedged, and only restarting the client cleared it.

But the server had no way out of it. One such connection sat there for **45 minutes**.

## Fix

Bound the call with `FINALIZE_TIMEOUT` (30 s).

**Placement is load-bearing.** The bound goes on the inner `ironrdp_acceptor::accept_finalize` call, *not* on `RdpServer::accept_finalize` or its callers. That method is a loop whose body also runs `client_accepted`, which drives the entire live session — a timeout hoisted any higher would cap session length. Applying it per pass also gives a deactivation-reactivation its own budget instead of sharing one with the initial handshake, and bounds a reactivation that wedges the same way.

**Why 30 s.** Measured across 197 successful connections in production logs: median finalize **0.07 s**, all but 13 samples under **1 s**, slowest ever **4.97 s** (a mobile link with heavy retransmits). So this is ~6× the observed worst case. It is deliberately generous because the same budget covers reactivations, whose tail is harder to measure. A false timeout is cheap — the connection drops and the client reconnects, immediately if it holds an auto-reconnect cookie (#1405).

The error surfaces as `ServerErrorKind::Io` with `ErrorKind::TimedOut`, plus a `warn!` naming the timeout.

## Relationship to #1476

Independent and complementary. #1476 adds `CANDIDATE_NEGOTIATION_TIMEOUT`, which bounds a *candidate* being negotiated alongside a live session on the preemption path. This bounds the connection actually being **served**, which is unbounded on `master` today regardless of whether #1476 lands. The two don't conflict; if #1476 merges first this applies unchanged (it would then also cover its `serve_negotiated` call site, since both route through the same inner call).

## Testing

`cargo check`/`clippy -p ironrdp-server --all-features` clean (the one remaining clippy warning, in `ironrdp-egfx`, is pre-existing on `master`), `cargo fmt` applied.

**No test is included here, deliberately** — `ironrdp-server` has no test harness for driving a client through a real handshake, and its dev-dependencies are currently just `tokio/sync`; adding one would mean pulling in `ironrdp-connector`, `ironrdp-tokio`, `tokio-rustls` and a cert generator as dev-deps, which felt like a bigger call than this fix should make unilaterally.

The fix does carry a regression test downstream, which I'm happy to port if you'd like the dev-deps: a client completes X.224 + TLS and then deliberately never sends its Connect Initial, and the server must drop it rather than park on the read. Verified to fail without the fix (the test hangs) and pass with it in 0.04 s, using `#[tokio::test(start_paused = true)]` so the 30 s bound costs no wall-clock time — with the caveat that paused time proves *a bound exists*, not the specific duration.

### DIFF
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -2940,7 +2940,7 @@ impl RdpServer {
                 Err(_) => {
                     warn!(
                         timeout = ?FINALIZE_TIMEOUT,
-                        "Client stopped responding during the finalize handshake — dropping the connection"
+                        "Client stopped responding during the finalize handshake, dropping the connection"
                     );
                     return Err(ServerError::io(
                         "timed out waiting for the client during finalize",
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -2940,7 +2940,7 @@ impl RdpServer {
                 Err(_) => {
                     warn!(
                         timeout = ?FINALIZE_TIMEOUT,
-                        "client stopped responding during the finalize handshake — dropping the connection"
+                        "Client stopped responding during the finalize handshake — dropping the connection"
                     );
                     return Err(ServerError::io(
                         "timed out waiting for the client during finalize",
--- a/crates/ironrdp-server/src/server.rs
+++ b/crates/ironrdp-server/src/server.rs
@@ -72,6 +72,24 @@ use ironrdp_rdpeusb::{InterfaceAlloc, server::UrbdrcControlServer, server::Urbdr
 const LISTENER_BACKLOG: u32 = 1024;
 const AUTO_RECONNECT_COOKIE_UPDATE_INTERVAL: Duration = Duration::from_secs(60 * 60);
 
+/// How long a single [`ironrdp_acceptor::accept_finalize`] pass may take before
+/// the connection is dropped.
+///
+/// A client can complete the whole security handshake — X.224, TLS, and CredSSP
+/// where applicable — and then stop producing PDUs, leaving the server blocked
+/// on a socket read with no timeout. Because `RdpServer` serves one connection
+/// at a time, that connection then holds the server indefinitely: the socket
+/// stays ESTABLISHED with both TCP queues empty, and nothing tears it down.
+/// This has been observed in the field against a real client, which completed
+/// MCS Connect, Erect Domain, Attach User and every channel join and then never
+/// sent its Client Info PDU; the connection sat there for 45 minutes.
+///
+/// Generous on purpose: a healthy finalize is sub-second, and even over a slow
+/// mobile link with heavy retransmits the whole pass stays within a few
+/// seconds. A false timeout is cheap — the connection is dropped and the client
+/// reconnects (immediately, if it holds an auto-reconnect cookie).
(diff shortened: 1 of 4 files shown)


antonmos:
Both addressed in f1dc8a2.

**Regression coverage** — added `crates/ironrdp-testsuite-core/tests/server/finalize_timeout.rs`. Thanks for the pointer to `ironrdp-testsuite-core`; it does already depend on `ironrdp-server`, so this needed only tokio's `test-util` feature in its dev-dependencies.

I took a slightly simpler route than suggested: with `with_no_security()` the negotiation resolves to plain RDP and `accept_begin` returns `BeginResult::Continue`, which reaches `accept_finalize` directly — so no TLS and no certificate generator, over a `tokio::io::duplex` pair. (`TransportTls::AlreadyDone` is still passed, but the `Continue` branch makes it moot.)

It covers exactly the hazard you named — hoisting the bound to the outer session loop would still pass CI — since that loop also runs `client_accepted`, so a timeout there would cap session length rather than the handshake. The test is written so that case fails loudly rather than hanging: it uses an outer deadline longer than `FINALIZE_TIMEOUT`, so under paused time the runtime advances to *whichever* deadline exists. Verified both ways:

- with the bound: passes in 0.00 s
- reverting the bound: fails in 0.00 s with `server never gave up on a client that stalled during finalize` — no hang

One honest caveat noted in the test: paused time proves *a bound exists*, not the specific duration.

Implementation note — `RdpServer`'s future is `!Send`, so it needs `LocalSet`/`spawn_local` rather than `tokio::spawn`.

**Logging style** — fixed; the message now starts with a capital letter per `STYLE.md`. I confirmed the rule and that the surrounding `server.rs` logs all follow it.

Full `ironrdp-testsuite-core` suite green (1627 passed), `clippy --all-features --all-targets` clean for both crates, `cargo fmt` applied.

CBenoit:
Hi! Thank you, looks like the CI is red though, can you fix that before I review it? Thanks!

CBenoit (approved):
LGTM, thank you!

PR #1910, opened 2026-09-05
fix(web): restore default remote cursor rendering

### LINKED ISSUE #1900: ironrdp-web: remote cursor shapes never applied — the enable_server_pointer rename inverted the flag but not the literal
## Summary

In the wasm web client, **no remote cursor shape is ever applied**. Every pointer — resize double-arrows, hand, I-beam — renders as the browser's plain arrow. `canvas.style.cursor` is never written even once, not even the initial `default` that a normal RDP session produces within the first seconds.

This looks like an unintended regression from the `no_server_pointer` → `enable_server_pointer` rename, not a deliberate default.

## Regression: the rename inverted the flag but not the literal

**2023-08-04 — [`1f401a13`](https://github.com/Devolutions/IronRDP/commit/1f401a13) (#168, "pointer processing logic")** added pointer support, and the web client was configured with:

```rust
no_server_pointer: false,      // "no server pointer" = false  →  cursors ON
```

**2025-08-04 — [`218fed03`](https://github.com/Devolutions/IronRDP/commit/218fed03c7993af0f958453e3944c58bcf9f43cb) ("rename option no_server_pointer into enable_server_pointer")** flipped the polarity of the flag. In that same commit:

```diff
# ironrdp-client — negation added, correct
- no_server_pointer: args.no_server_pointer,
+ enable_server_pointer: !args.no_server_pointer,

# ironrdp-web — literal kept as-is, polarity silently flipped
- no_server_pointer: false,
+ enable_server_pointer: false,
```

The meaning went from "don't disable" to "don't enable", so the web client went from **cursors ON** to **cursors OFF**. The CLI call site in the very same commit got the `!`, which suggests the inversion was understood — the web one was likely missed because it is a bare literal rather than an expression.

`enable_server_pointer: false` is still what `crates/ironrdp-web/src/session.rs::build_config()` has on `master` today.

## Why it is invisible

```
build_config: enable_server_pointer = false
  → ActiveStage receives false
  → ironrdp-session/src/fast_path.rs::process_pointer_update()
      returns on its first line: `if !self.enable_server_pointer { return ... }`
  → every pointer update is dropped silently — no error, no log at any level
  → set_cursor_style() is never called
  → canvas.style.cursor is never written
```

It is not a failure path, so nothing is logged. We only found it by reading the source after ruling everything else out.

The feature itself is fully implemented and works — the JS callback plumbing, the PNG encoding of pointer bitmaps into `data:` URLs, and the down-scaling of sprites larger than 32×32 are all present and correct. Only the switch that feeds them is off.

## Verification

We rebuilt `@devolutions/iron-remote-desktop-rdp` from tag `npm-iron-remote-desktop-rdp-v0.7.0` with **exactly one line changed** (`enable_server_pointer: false` → `true`), using the upstream recipe: `wasm-pack build --target web` with the xtask RUSTFLAGS (`-Ctarget-feature=+simd128,+bulk-memory --cfg getrandom_backend="wasm_js"`), then the package's own vite build. The result is interface-identical to the published release (`index.d.ts` 740 lines, same top-level exports, same version).

**With that one line changed, all remote cursor shapes work correctly** — verified against a real Windows Server host. `canvas.style.cursor` receives `url(data:image/png;base64,…)` values as expected.

## Suggested fix

Restore the pre-rename behaviour by setting `enable_server_pointer: true` in `ironrdp-web::build_config`.

Optionally also expose it on `SessionBuilder`, so embedders can turn it off deliberately. Today there is no way to reach it from JavaScript — the published typings expose only `setCursorStyleCallback`, `setCursorStyleCallbackContext` (on `SessionBuilder`) and `setCursorStyleOverride` (on `UserInteraction`).

## What we ruled out before reading the source

| Hypothesis | Result |
|---|---|
| No cursor support in the wasm build | Ruled out — the inlined wasm contains `FastPathUpdate::decode_cursor_with_code`, the PNG encoding path, the `default`/`hidden`/`url` kind literals, and a `set cursor style callback failed:` error path |
| The web component does not register the callback | Ruled out — `connect()` calls `setCursorStyleCallbackContext(this)` and `setCursorStyleCallback(this.setCursorStyleCallback)` |
| The embedder overrides the cursor | Ruled out — `setCursorStyleOverride` is never called in our codebase |
| A newer release fixes it | Ruled out — 0.11.0 / 0.7.0 were the latest published versions |
| Mouse position never reaches the server | Ruled out — the component's `mousemove` handler calls `updateMousePosition` with `backing/CSS` scaling, and our canvas is sized so that ratio equals `devicePixelRatio` exactly |
| Windows experience/performance flags suppress it | Ruled out — the same host with performance mode enabled shows correct cursors under the native `mstscax` control. Experience flags govern wallpaper / full-window drag / theming / animations / font smoothing; pointer shape travels in its own Pointer Update PDU |
| The callback is invoked but throws | Ruled out — we forward the wasm's `console.error`/`console.warn` into our application log; zero wasm errors across many sessions |
| The server does not send pointer updates | Ruled out — the native `mstscax` control on the same host shows every cursor shape |

Instrumentation used: `init('info')`, console forwarding of `error`/`warn` in full plus `log`/`info`/`debug` filtered on `/cursor|pointer/i`, and sampling `canvas.style.cursor` every 150 ms for the whole session. **No wasm output related to cursors or pointers at any level**, and the style stayed `""`.

## Environment

- `@devolutions/iron-remote-desktop@0.11.0`, `@devolutions/iron-remote-desktop-rdp@0.7.0`
- Extensions in use: `displayControl(true)`, `enableCredssp(true)`. DisplayControl works — dynamic resize and `DesktopScaleFactor` are both confirmed working end to end.
- Clipboard and clipboard file transfer both work.
- Host: Windows Server, RDP on a non-standard port, reached through an RDCleanPath proxy.
- Client: WebView2 (Tauri 2) on Windows 11.

glamberson:
`build_config()` has hardcoded `enable_server_pointer: false` since the
`no_server_pointer` to `enable_server_pointer` rename (218fed03) inverted
the flag's polarity without updating this literal. The CLI call site in
the same commit correctly negated the old value (`!args.no_server_pointer`);
this bare literal in `ironrdp-web` was missed.

The native client already defaults to `true`
(`ClientBuilder::enable_server_pointer.unwrap_or(true)`), so this restores
parity. Today, remote cursor shapes silently never render in the web
client: `ironrdp-session::fast_path::process_pointer_update` returns on
its first line when the flag is `false`, with no error at any log level,
so the regression is invisible until someone reads the source.

Fixes #1900, which independently rebuilt the wasm package with this exact
one-line change and verified all remote cursor shapes work correctly
against a real Windows Server host.

Not bundled here: the issue also suggests exposing this setting on
`SessionBuilder` so embedders can turn it off deliberately, since nothing
in the current public API reaches it. Leaving that as a separate,
smaller follow-up rather than widening this fix's scope.

`cargo xtask check fmt/lints/tests/typos/locks` and `cargo xtask wasm check` all pass.

### DIFF
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -1503,7 +1503,7 @@ fn build_config(
         client_dir: "C:\\Windows\\System32\\mstscax.dll".to_owned(),
         platform: ironrdp::pdu::rdp::capability_sets::MajorPlatformType::UNSPECIFIED,
         compression_type: None,
-        enable_server_pointer: false,
+        enable_server_pointer: true,
         autologon: false,
         enable_audio_playback: false,
         enable_audio_capture: false,


CBenoit (approved):
LGTM

PR #1912, opened 2026-09-05
fix(egfx): do not track frames a suspended client cannot acknowledge

maryny4:
When a client sends a Frame Acknowledge with queue depth `SUSPEND_FRAME_ACKNOWLEDGEMENT` ([MS-RDPEGFX] 2.2.2.13) it stops acknowledging frames until it opts back in. `FrameTracker::begin_frame` keeps inserting every frame it sends into `unacknowledged` throughout the pause, and nothing takes those entries out again: the client will never acknowledge them, and per [MS-RDPEGFX] 3.2.5.13 the server *"MUST clear the Unacknowledged Frames ADM element"* on suspension anyway.

`should_backpressure` already gates on `ack_suspended`, so the encoder is not throttled *during* the pause. The cost lands on **resume**: the frames that piled up while suspended are still in the map, so `in_flight` is back over `max_in_flight` the moment the client resumes, and backpressure latches on for the rest of the connection — with no frame going out, there is no End Frame left for the client to acknowledge, so it never clears. A short pause becomes a permanently frozen surface.

### The fix

Two changes to `FrameTracker`:

- `begin_frame` does not record a frame while acknowledgement is suspended.
- A suspending `acknowledge` clears the map after taking out the frame it carries (that one still feeds the QoE report).

### Tests

Two: frames sent across a suspend/resume cycle must not hold backpressure on afterwards, and a suspension must not shrink the in-flight window for the rest of the connection. Both fail on current master.

*(Reopened after my fork was briefly deleted, which auto-closed the original #1805. Rebased onto current master; since the original PR, master added the `should_backpressure` gate that removes the steady-state throttle, so this now targets the post-resume wedge and the spec's MUST-clear specifically.)*

### DIFF
--- a/crates/ironrdp-egfx/src/server.rs
+++ b/crates/ironrdp-egfx/src/server.rs
@@ -591,15 +591,20 @@ impl FrameTracker {
         let frame_id = self.next_frame_id;
         self.next_frame_id = self.next_frame_id.wrapping_add(1);
 
-        self.unacknowledged.insert(
-            frame_id,
-            FrameInfo {
+        // A suspended client sends no acknowledgement until it opts back in
+        // ([MS-RDPEGFX] 2.2.2.13), so nothing would take this frame back out.
+        // Backpressure is off while suspended, so the entry has no work to do.
+        if !self.ack_suspended {
+            self.unacknowledged.insert(
                 frame_id,
-                timestamp,
-                sent_at: Instant::now(),
-                size_bytes: 0,
-            },
-        );
+                FrameInfo {
+                    frame_id,
+                    timestamp,
+                    sent_at: Instant::now(),
+                    size_bytes: 0,
+                },
+            );
+        }
 
         self.total_sent += 1;
         // Edge-trigger after insert so a frame that pushes us to max_in_flight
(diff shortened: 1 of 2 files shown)


glamberson:
I checked this against MS-RDPEGFX 3.2.5.13 directly rather than taking the citation on faith. The text is exact: On a queue depth of SUSPEND_FRAME_ACKNOWLEDGEMENT, the server MUST clear the Unacknowledged Frames ADM element and MUST NOT expect any further Frame Acknowledge messages, and separately MUST NOT wait or block on unacknowledged frames during the pause. Nothing in that requirement is approximate, and the fix matches it precisely.

The FrameTracker changes are correct on both sides. Skipping the insert in `begin_frame` while `ack_suspended` is set stops the leak at its source, and clearing `unacknowledged` in `acknowledge` only on the suspending call, after removing the one frame that call actually carries, keeps the round-trip sample for that frame in the QoE report instead of losing it to the clear.

The three tests isolate the failure modes cleanly: Backpressure surviving past a suspend and resume, the in-flight window shrinking permanently instead of restoring in full, and the suspending acknowledgment itself being dropped from the QoE data. Each one matches a specific consequence of the bug rather than a generic regression check. This is a sound fix.

CBenoit (approved):
Thank you!

Issue #327, opened 2023-12-07
Expose an API from the WASM module to configure how the security protocol is negotiated


CBenoit:
We need to expose options similar to [the ones in RDM](https://twitter.com/awakecoding/status/1651609994394062851).

Currently we only support enabling (or disabling) these two:

- "Winlogon" (PROTOCOL_SSL)
- "CredSSP" (PROTOCOL_HYBRID + PROTOCOL_HYBRID_EX)

For some background: https://github.com/Devolutions/IronRDP/issues/314#issuecomment-1843925374

PR #1914, opened 2026-09-06
feat(web): expose enable_server_pointer from the WASM module

Ryzero:
Follow-up to #1900 / #1910: make `enable_server_pointer` reachable from JS instead of being a hardcoded constant, and default it to `true`.

### Background

`build_config` hardcodes `enable_server_pointer: false`. That value reaches `ActiveStage`, and `process_pointer_update()` in `ironrdp-session` returns on its first line, so every Pointer Update PDU is dropped **silently** — no error, no log. The visible result is that remote cursor shapes never reach the JS `set_cursor_style` callback: resize arrows, hand, I-beam all render as the plain arrow. Everything else on that path is already implemented (PNG encoding of the pointer bitmap, >32×32 scaling, the JS callbacks). Full write-up in #1900.

@glamberson filed #1910 with the one-line default flip and suggested exposing the flag on `SessionBuilder` as a follow-up. This is that follow-up.

### What this does

Follows the existing `enable_credssp` / `legacy_graphics` pattern:

- `SessionBuilderInner.enable_server_pointer`, defaulting to `true`
- an `extension_match!` arm for it
- applied to `connector::Config` in `connect()`, right next to `enable_credssp`
- `enableServerPointer(enable)` exported from `iron-remote-desktop-rdp`'s `main.ts`

One item of the `TODO(#327): expose these options from the WASM module`.

### Relationship to #1910

This deliberately **does not touch** the `enable_server_pointer: false` literal in `build_config`. `connect()` overwrites it unconditionally, so the effective default is already `true`, and leaving that line alone means **zero textual conflict with #1910**. Either can merge first.

If you would rather have the literal cleaned up here too, or would prefer to take #1910 alone and have this reduced, happy to adjust.

### Verification

- `cargo check -p ironrdp-web --target wasm32-unknown-unknown` passes.
- The extension identifier matches on both sides: `extension_match!` keys on `stringify!` of the closure parameter, so `|enable_server_pointer: bool|` pairs with `new Extension('enable_server_pointer', …)`. Checked mechanically rather than by eye — a mismatch there would fall through to the `Unknown extension` warning and the new API would be silently dead.
- The runtime effect of the flag itself was verified on real hardware in #1900, by rebuilding the wasm package from tag `npm-iron-remote-desktop-rdp-v0.7.0` with only that constant flipped: cursor shapes became correct immediately. This PR does not change that behaviour, it only makes the flag settable.

One caveat on my side: I compile-checked with Rust 1.98 rather than the pinned 1.94.1 (a local toolchain install problem, not related to this change). The diff adds no new syntax or imports, and CI will cover the pinned toolchain.

### DIFF
--- a/crates/ironrdp-web/src/session.rs
+++ b/crates/ironrdp-web/src/session.rs
@@ -91,6 +91,7 @@ struct SessionBuilderInner {
 
     use_display_control: bool,
     enable_credssp: bool,
+    enable_server_pointer: bool,
     legacy_graphics: bool,
     outbound_message_size_limit: Option<usize>,
 }
@@ -134,6 +135,7 @@ impl Default for SessionBuilderInner {
 
             use_display_control: false,
             enable_credssp: true,
+            enable_server_pointer: true,
             legacy_graphics: false,
             outbound_message_size_limit: None,
         }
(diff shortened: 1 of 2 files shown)


glamberson:
Thanks for picking this up, and for keeping it clean against #1910. Routing the default through `connect()` instead of touching the `build_config` literal is the right way to avoid a textual conflict while #1910 is still unreviewed, and I do not have a preference on which merges first. Either order gets the same end state.

The extension wiring looks right to me: Matching `enable_server_pointer` through `extension_match!`'s `stringify!` on the closure parameter is exactly the existing `enable_credssp` pattern. Checking that mechanically rather than by eye is the correct instinct too, since a silent mismatch there would only fall through to the unknown-extension warning rather than an error.

CBenoit (approved):

CBenoit:
LGTM, thank you!

PR #1919, opened 2026-09-08
feat(rdpeudp): MS-RDPEUDP version 1/2 reliable data transfer

AKolenda:
Every Windows host tested on a LAN answers the RDPUDP SYN with uUdpVer 0x0002: they speak MS-RDPEUDP (version 1/2), not MS-RDPEUDP2 (version 3). A client that only implements the version 3 framing therefore always times out on the UDP handshake and falls back to TCP, so the multitransport path never carries data against Windows.

This adds the version 1/2 reliable data path to the sans-I/O connection state machine, negotiated from the SYN+ACK and reusing the existing send/receive windows, RTT estimator, loss detection and timers:

- pdu: RDPUDP_SOURCE_PAYLOAD_HEADER (2.2.2.4) and the data payload on a version 1/2 Source Packet. DATA datagrams decode; FEC (lossy) datagrams are rejected.
- connection: NegotiatedParams records the wire format. Version 1/2 uses a single Source sequence space from ISN+1; snCoded numbers each transmission, snSourceAck drives an ACK vector walked down from the highest received sequence (3.1.1.4), ACK-of-ACKs resets the receiver's vector base (2.2.2.6), a Congestion Notify is answered with CWR at most once per RTT (3.1.1.8), and ACKDELAYED marks timer-driven ACKs (3.1.6.3). Timers take the version's floors.
- The 6-bit ACK-vector run length is encoded as count minus one on the wire, which MS-RDPEUDP 2.2.2.7.1 leaves ambiguous but Windows requires: element 0x00 acknowledges one Source Packet and 0x02 acknowledges three. Reading it literally leaves the first Source Packet unacknowledged forever, collapsing the congestion window.
- seq: 32-bit wrap-aware sequence reconstruction and narrowing.
- ConnectionConfig::offer_version lets a client offer version 2 or 1 outright (no cookieHash), which is what the measured hosts answer immediately.
- RdpeudpConnection::v1_stats exposes retransmit, ACK and datagram counters for observability.

Verified against the spec's 4.2.1 example capture and new state-machine tests; the existing tests are unchanged.

**Note on MS-RDPEUDP 4.2.3.** The spec's worked example annotates ACK vector element 0x04 as "4 datagrams received," a literal reading of the six-bit run length. Captures against Windows Server show the field is the run length minus one: after one Source Packet the server answers `snSourceAck = 1` with element 0x00, after four it answers `snSourceAck = 4` with 0x03. Read literally, the first is an empty run and the second never acknowledges seq 1, which stalls the sender. This implementation follows the wire, not the example. Evidence: https://gist.github.com/AKolenda/51931b094dd454ef0eb1430503629c8a

### DIFF
--- a/crates/ironrdp-rdpeudp/src/connection.rs
+++ b/crates/ironrdp-rdpeudp/src/connection.rs
@@ -1229,7 +1229,9 @@ impl RdpeudpConnection {
             ack_vector: Some(V1AckVectorHeader {
                 elements: vec![V1AckVectorElement {
                     state: VectorElementState::DatagramReceived,
-                    length: 1,
+                    // One datagram (the SYN+ACK): wire runs are count - 1, see
+                    // process_v1_acknowledgement.
+                    length: 0,
                 }],
             }),
             ack_of_acks: None,
@@ -2723,6 +2725,85 @@ mod v1_tests {
             .and_then(|t| decode::<V1Datagram>(&t.contents).ok())
             .is_some_and(|d| d.data.is_some());
         assert!(!is_data, "acknowledged packet must not be retransmitted");
+    }
+
+    /// The client's final handshake ACK acknowledges exactly the SYN+ACK, so
+    /// its single run must be encoded as 0x00, the same way build_v1_ack_parts
+    /// encodes a run of one.
+    #[test]
+    fn the_final_handshake_ack_encodes_one_datagram_as_zero() {
+        let mut conn = RdpeudpConnection::connect(config(), at(0)).unwrap();
+        let _ = conn.poll_transmit(at(0)).expect("client SYN");
+        let mut wire = syn_ack(2);
+        conn.handle_datagram(&mut wire, at(20)).unwrap();
+
+        let out = conn.poll_transmit(at(20)).expect("final ACK");
+        let datagram: V1Datagram = decode(&out.contents).unwrap();
+        assert_eq!(datagram.header.sn_source_ack, REMOTE_ISN);
+        assert_eq!(
+            datagram.ack_vector.unwrap().elements,
+            vec![V1AckVectorElement {
+                state: VectorElementState::DatagramReceived,
+                length: 0,
+            }]
+        );
+    }
+
+    /// ACKs captured from Windows Server (10.0.0.12) over an MS-RDPEUDP
+    /// version 2 tunnel, verbatim apart from `snSourceAck`, which is rebased
+    /// onto this test's ISN:
+    ///
+    /// * after our first Source Packet (the TLS ClientHello, seq 1), the
+    ///   ServerHello came back as `00 00 00 01 00 c8 00 0c 00 01 00 b5 ...`:
+    ///   `snSourceAck` 1 and one element 0x00;
+    /// * after four Source Packets (seq 1..=4, all of which the server acted
+    ///   on: the TLS handshake finished), `00 00 00 04 00 c6 00 04 00 01 03 b5`:
+    ///   `snSourceAck` 4 and one element 0x03.
+    ///
+    /// Read literally 0x00 would be an empty run and 0x03 would leave seq 1
+    /// unacknowledged forever; Windows means "one" and "four", so the six-bit
+    /// run length is count - 1.
+    #[test]
+    fn windows_ack_vector_runs_are_count_minus_one() {
+        fn windows_ack(sn_source_ack: u32, element: u8, window: u16, flags: u16) -> Vec<u8> {
+            let mut bytes = sn_source_ack.to_be_bytes().to_vec();
+            bytes.extend_from_slice(&window.to_be_bytes());
+            bytes.extend_from_slice(&flags.to_be_bytes());
+            bytes.extend_from_slice(&[0x00, 0x01, element, 0xb5]);
+            bytes
+        }
+        fn pending(conn: &RdpeudpConnection) -> usize {
+            conn.send_window.as_ref().unwrap().pending_entries().count()
+        }
+
+        // One Source Packet, acknowledged by element 0x00.
+        let mut conn = established(2);
+        conn.send(b"ClientHello".to_vec()).unwrap();
+        let _ = conn.poll_transmit(at(30)).unwrap();
+        assert_eq!(pending(&conn), 1);
+        let mut ack = windows_ack(LOCAL_ISN + 1, 0x00, 0x00c8, 0x000c);
+        // The captured datagram also carried data; a bare ACK is enough here.
+        ack[7] = 0x04;
+        conn.handle_datagram(&mut ack, at(50)).unwrap();
+        assert_eq!(pending(&conn), 0, "element 0x00 acknowledges exactly one datagram");
+
+        // Four Source Packets, acknowledged by element 0x03.
+        let mut conn = established(2);
+        for payload in [&b"1"[..], b"2", b"3", b"4"] {
+            conn.send(payload.to_vec()).unwrap();
+            let _ = conn.poll_transmit(at(30)).unwrap();
+        }
+        assert_eq!(pending(&conn), 4);
+        let mut ack = windows_ack(LOCAL_ISN + 4, 0x03, 0x00c6, 0x0004);
+        conn.handle_datagram(&mut ack, at(50)).unwrap();
+        assert_eq!(pending(&conn), 0, "element 0x03 acknowledges seq 4 down to seq 1");
+
+        conn.handle_timeout(at(5_000));
+        let retransmit = conn
+            .poll_transmit(at(5_000))
+            .and_then(|t| decode::<V1Datagram>(&t.contents).ok())
+            .is_some_and(|d| d.data.is_some());
+        assert!(!retransmit, "nothing left to retransmit");
     }
 
     #[test]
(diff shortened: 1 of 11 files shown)


crates/ironrdp-rdpeudp/src/connection.rs
```diff
@@ -1893,14 +2073,301 @@ impl RdpeudpConnection {
...
+        let reference = send_window.next_channel_seq().saturating_sub(1);
+        let highest_seen = seq::reconstruct_seq32(header.sn_source_ack, reference);
+
+        let mut acked: Vec<(u64, bool, MonotonicInstant)> = Vec::new();
+        if let Some(vector) = ack_vector {
+            let mut current = Some(highest_seen);
+            for element in &vector.elements {
+                // Windows encodes a run of n datagrams as n - 1 (an element of 0x00
+                // acknowledges exactly one), which 2.2.2.7.1's prose leaves open.
+                for _ in 0..=element.length {
```
glamberson, line 2159: This counts one more datagram than the peer confirmed, for the reasons in the review body above.

crates/ironrdp-rdpeudp/src/connection.rs
```diff
@@ -1893,14 +2073,301 @@ impl RdpeudpConnection {
...
+                let chunk = u8::try_from(remaining.min(u64::from(V1AckVectorElement::MAX_LENGTH) + 1))
+                    .expect("clamped to 6-bit range plus one");
+                elements.push(V1AckVectorElement {
+                    state: if received {
+                        VectorElementState::DatagramReceived
+                    } else {
+                        VectorElementState::DatagramNotYetReceived
+                    },
+                    // Wire length is count - 1; see process_v1_acknowledgement.
+                    length: chunk - 1,
```
glamberson, line 2287: This is the encode side of the same issue.

crates/ironrdp-rdpeudp/src/connection.rs
```diff
@@ -2106,3 +2573,227 @@ mod tests {
...
+            ack_vector: Some(V1AckVectorHeader {
+                elements: received
+                    .iter()
+                    .map(|(r, n)| V1AckVectorElement {
+                        state: if *r {
+                            VectorElementState::DatagramReceived
+                        } else {
+                            VectorElementState::DatagramNotYetReceived
+                        },
+                        length: n - 1, // wire runs are count - 1
```
glamberson, line 2673: Test fixtures bake in the same convention, which is why the suite doesn't catch it against the spec's own example.

glamberson (commented):
I think there's an off-by-one in the ACK vector run-length encoding here. connection.rs:2131-2132 reads:

> Windows encodes a run of n datagrams as n - 1 (an element of 0x00 acknowledges exactly one), which 2.2.2.7.1's prose leaves open.

MS-RDPEUDP 2.2.2.7.1 says the six least significant bits "specify the length of a continuous sequence of datagrams," and all three of the spec's own worked examples (4.2.1, 4.2.2, 4.2.3) gloss the same byte the same way: raw value 4 is "length of the vector, 4 datagrams received," a direct literal reading with no minus-one. FreeRDP's Wireshark dissector (tools/wireshark/rdp-udp.lua) reads it the same way: local rleLen = bit32.band(b, 0x3f), no adjustment.

This crate's own pre-existing enqueue_final_ack (unchanged by this PR) sends V1AckVectorElement { length: 1, .. } to acknowledge exactly one datagram, a literal encoding. That's the opposite of what this PR's new code assumes, so after this merges the same field means two different things depending on which code path built it.

The 4.2.1-capture test added here (SourcePayloadHeader round-trip) doesn't actually exercise the ACK vector byte from that same example, only the header bytes next to it. Round-tripping 0x04 from that capture through build_v1_ack_parts/process_v1_acknowledgement would show 4 datagrams going in and 5 coming out.

If I'm wrong about this and there's a real packet capture showing Windows requiring the minus-one reading, I'd like to see it, since it contradicts the spec text, all three of its own examples, and FreeRDP's independent implementation. As written, process_v1_acknowledgement's 0..=element.length will count one more datagram as acknowledged than the peer actually confirmed, which on a lossy link means retiring a send-window entry the peer never received.

AKolenda:
Thanks for the close read. The minus-one encoding is measured Windows behaviour, not a guess. Capture and byte-level decode here: https://gist.github.com/AKolenda/51931b094dd454ef0eb1430503629c8a

Short version, from a session against Windows Server (MS-RDPEUDP protocol version 2, which uses the same FEC-header framing and 2.2.2.7.1 ACK vector). After our first Source Packet (seq 1, the TLS ClientHello) the ServerHello came back with `snSourceAck = 1` and ACK vector `00 01 00 b5`, one element 0x00. After four Source Packets (seq 1..4, TLS handshake complete on both sides) the ACK was `00 00 00 04 00 c6 00 04 00 01 03 b5`: `snSourceAck = 4`, one element 0x03. Later `snSourceAck = 5` with element 0x04. No ACK-of-ACKs flag in any of them. Read literally, 0x00 is an empty run and 0x03 leaves seq 1 unacknowledged forever, which is exactly the stall we hit before this change: seq 1 retransmitted with a doubling RTO until the window collapsed. Read as count-1, every ACK is consistent and the session runs.

On the sources: the 2.2.2.7.1 prose never gives the numeric mapping, and 4.2.1/4.2.2/4.2.3 show the 0x04 byte without the peer's sequence numbers, so they cannot distinguish 4 from 5. The `rdp-udp.lua` `rleLen = bit32.band(b, 0x3f)` code and the `packet-rdpudp.c` `base_seq + rle_len` code are both the RDPEUDP2 ACK vector (`dissectV2` / `RDPUDP2_ACKVEC`), a different format. The Wireshark v1 path only renders the 0xc0 and 0x3f bit fields and does no sequence arithmetic, so it takes no position.

You are right that `enqueue_final_ack` was left on the old convention, sending a run of 1 where the new builder sends 0. Fixed in c5d3a71 with a test on the wire byte, plus a regression test that feeds the captured Windows ACKs through `process_v1_acknowledgement` and checks the send window drains. `V1AckVectorElement::length` is now documented as the raw wire value.

glamberson:
Thanks for the detailed reply and the capture. That's genuinely convincing evidence.

I want to flag something for the record, though: MS-RDPEUDP's own worked example in section 4.2.3 explicitly reads a single ACK vector element with value 0x04 as "4 datagrams received," literally, with no minus-one adjustment. That's Microsoft's own annotation on their own example, and it directly conflicts with the count-minus-one convention your capture demonstrates and the fix now encodes.

I traced the actual walk in `process_v1_acknowledgement` against your two examples (snSourceAck=4/element=0x03, and the two-element gap test) and it's internally self-consistent under the count-minus-one model: It doesn't double-count across multiple elements, since `current` only ever anchors to `snSourceAck` once, at the start of the walk. So the implementation checks out given the premise.

But the premise is where the two sources disagree, and that conflict shouldn't pass silently. A live capture against a real Windows Server is stronger evidence than a documentation example (Microsoft's open specs have known errata, and the literal reading reproduces the exact stall this PR fixes), so I'd trust your fix over the worked example here. Might be worth a line in the PR body noting the discrepancy, for whoever reads section 4.2.3 next and wonders why the code doesn't match it.

AKolenda:
Agreed on all counts, and thanks for tracing the walk. I've added a note to the PR body: 2.2.2.7.1's prose gives no numeric mapping, 4.2.3's annotation reads 0x04 as four datagrams literally, and Windows Server on the wire encodes a run of n as n-1 (0x00 acknowledges one). The code follows the wire, with the capture linked as the evidence, so readers of 4.2.3 know why the two differ.

AKolenda:
I believe on all counts based on the back and forth that we have, I think we're ready to merge. Maintainers agree?

CBenoit (approved):
Agreed; I started the CI workflow and Copilot review. I’ll likely merge regardless, but I would appreciate follow up PRs if relevant. Thank you!

CBenoit:
Looks like this is breaking the testsuite crates compilation.

AKolenda:
Sorry about the testsuite break; 098d801 fixes it. `ironrdp-testsuite-core` builds `V1Datagram` and `ConnectionConfig` by hand and had not picked up the new `data` / `offer_version` fields. Locally `cargo xtask check tests --no-run`, `cargo xtask check tests` (rdpeudp, rdpeudp-tokio, testsuite-core: 195 + 53 + 1649 pass), `check lints` and `check fmt` are all green on 1.94.1. One pre-existing test, `a_client_rejects_a_syn_ack_that_settles_below_version_3`, asserted the behaviour this PR replaces; it now checks that the client follows the server down onto version 2 (MS-RDPEUDP 1.7) and that a version above the SYN's offer is refused.

Copilot review, same commit:

* `connection.rs:1160` (version selection): agreed. `handle_syn_ack` now rejects a version other than 1/2/3 and any selection above `ConnectionConfig::offer_version`, with tests for both.
* `connection.rs:2142` (per-sequence window scan): agreed. The pending entries are indexed once and the walk stops at the lowest pending sequence, so a 2048-element vector costs one pass over the window plus a lookup per named sequence. Test added with a 2048 x 64 vector.
* `connection.rs:2202` (ACK-of-ACKs reset ahead of `highest_seq`): agreed, a reset above what we have received is dropped, with a test that a later Source Packet is still delivered.
* `pdu/mod.rs:131`, `pdu/mod.rs:272`, `v1_data.rs:22`, `seq.rs:306`: all applied (docs, lowercase error message, named addends, functions moved above the test module and the seq.rs module doc corrected).

Deferred to follow-up PRs, as suggested above, because each changes on-the-wire behaviour I want to validate against Windows before landing:

* `connection.rs:2128` gating new Source Packets on the peer's advertised `receive_window_size`;
* `connection.rs:2243` / `2303` how ACK-vector state is retired: the Windows captures never carried an ACK-of-ACKs, so holding reception state until one arrives interacts with the 32-element cap and needs measuring rather than reasoning about;
* `connection.rs:2270` / `2183` receiver-side CN generation and sequence-tracked CWR suppression.

@CBenoit the workflow will need re-approving for the new commit.

PR #1923, opened 2026-09-08
fix(egfx): decode AVC420 with full-range BT.709

### LINKED ISSUE #1924: fix(egfx): encode AVC420 with full-range BT.709
## Problem

The OpenH264 AVC420 **encoder** path (`OpenH264Encoder::encode` in `crates/ironrdp-egfx/src/encode.rs`) converts RGBA input to YUV420 via `openh264::formats::YUVBuffer::from_rgb_source`, which applies a **limited-range (studio swing) BT.601** matrix internally (openH264 `rgb2yuv`: luma offset +16, coefficients 66/129/25 >> 8).

MS-RDPEGFX §3.3.8.3.1 normatively mandates **full-range BT.709** for `RFX_AVC420_BITMAP_STREAM` color conversion (forward matrix: `Y = (54·R + 183·G + 18·B) >> 8`, no 16–235 swing). A stream produced by the current encoder therefore violates the spec: luma occupies only [16, 235] instead of [0, 255].

## Inconsistency after PR #1923

The client-side decoder was fixed to be spec-conformant in PR #1923 (`fix(egfx): decode AVC420 with full-range BT.709`): it now converts YUV→RGBA with full-range BT.709, matching the spec's normative inverse matrix and FreeRDP's `prim_YUV.c` reference.

This makes an IronRDP-server → IronRDP-client round trip **color-distorted by construction**: the server encodes luma into [16, 235], the client correctly assumes [0, 255], and the resulting frames appear washed out / crushed at the extremes. Before the decoder fix the two errors partially cancelled; now the mismatch is fully visible.

## Expected fix

Convert RGBA→YUV420 with the yuv crate's forward full-range BT.709 420 variant (e.g. `yuv::rgba_to_yuv420(..., YuvRange::Full, YuvStandardMatrix::Bt709)`), the mirror of the decoder-side change, instead of `YUVBuffer::from_rgb_source`. Notes:

- `yuv` is already an optional dependency of `ironrdp-egfx` wired to the `openh264` feature, so no new dependency is needed.
- Chroma siting for the 4:2:0 downsample should be checked for consistency with the yuv crate's convention.
- Add a chromatic round-trip test (a saturated red/green/blue tile encoded then decoded should retain its channel extremes); the existing round-trip test only asserts dimensions, and neutral luma cannot discriminate the matrix.

## References

- MS-RDPEGFX §3.3.8.3.1 Color Conversion (forward ARGB→AYUV: "based on full-range BT.709", "MUST be clamped to the range 0...255 inclusive")
- FreeRDP `libfreerdp/primitives/prim_YUV.c` `general_YUV420ToRGB_8u_P3AC4R` (spec matrix verbatim)
- Decoder-side fix: #1923

> [!NOTE]
> > Human-tuned, LLM-assisted content.

ciron-nine:
Replaces openh264's built-in write_rgba8 in the AVC420 decoder path with an explicit full-range BT.709 conversion from the yuv crate.

MS-RDPEGFX 3.3.8.3.1 mandates full-range BT.709 for RFX_AVC420_BITMAP_STREAM color conversion, but openh264 hardcodes a limited-range (studio swing) BT.601 matrix: near-white luma is over-expanded past 255 and clamps, near-black luma is crushed below 0, and saturated colors shift hue (BT.601 inverse renders pure red at R~232 instead of 255). Upgrading openh264 does not help; the range bias is hardcoded in every released version.

The conversion now uses yuv420_to_rgba with YuvRange::Full and YuvStandardMatrix::Bt709, matching the spec's normative inverse matrix and FreeRDP's prim_YUV.c reference. Real per-plane strides from openh264 replace the previous packed assumption, so macroblock-aligned plane padding no longer produces diagonal streaks. Conversion failures surface as DecoderError instead of a panic from openh264's internal asserts.

The optional yuv dependency is implementation-only (no public API exposure) and is wired to the openh264 feature. Cargo.lock is unchanged since yuv already exists in the dependency tree through ironrdp-graphics.

A round-trip regression test encodes near-white (Y=245) and near-black (Y=10) frames and asserts neither clamps to 255 or 0; pure black and white pass under both conversions, so the discriminating values sit just inside the extremes.

The openh264 encoder path still emits limited-range BT.601 via YUVBuffer::from_rgb_source and is now inconsistent with this conforming decoder; that fix will follow separately.

Encoder-side fix: #1925 (fixes #1924)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2947,6 +2947,7 @@ dependencies = [
  "ironrdp-pdu",
  "openh264",
  "tracing",
+ "yuv",
 ]
 
 [[package]]
--- a/crates/ironrdp-egfx/Cargo.toml
+++ b/crates/ironrdp-egfx/Cargo.toml
@@ -25,10 +25,11 @@ ironrdp-graphics = { path = "../ironrdp-graphics", version = "0.9" } # public
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.9" } # public
 openh264 = { version = "0.9", optional = true, default-features = false }
 tracing = { version = "0.1", features = ["log"] }
+yuv = { version = "0.8", default-features = false, optional = true }
 
 [features]
 arbitrary = ["dep:arbitrary", "bitflags/arbitrary", "ironrdp-pdu/arbitrary"]
-openh264 = ["dep:openh264"]
+openh264 = ["dep:openh264", "dep:yuv"]
 openh264-bundled = ["openh264", "openh264/source"]
 openh264-libloading = ["openh264", "openh264/libloading"]
 
--- a/crates/ironrdp-egfx/src/decode.rs
+++ b/crates/ironrdp-egfx/src/decode.rs
@@ -183,6 +183,7 @@ pub trait H264Decoder: Send {
 
 #[cfg(feature = "openh264")]
 mod openh264_impl {
+    use openh264::formats::YUVSource;
     use tracing::warn;
 
     use super::{DecodedFrame, DecoderError, DecoderResult, H264Decoder};
@@ -253,7 +254,8 @@ mod openh264_impl {
                 .map_err(|e| DecoderError::new("OpenH264 decode failed", e))?
                 .ok_or_else(|| DecoderError::msg("OpenH264 returned no picture"))?;
 
-            let (width, height) = openh264::formats::YUVSource::dimensions(&yuv);
+            let (width, height) = YUVSource::dimensions(&yuv);
+            let (y_stride, u_stride, v_stride) = YUVSource::strides(&yuv);
 
             #[expect(
                 clippy::as_conversions,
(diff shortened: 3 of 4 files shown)


glamberson (approved):
This is a really nice catch and a clean fix. Thanks for tracking it down and for grounding it in the spec section instead of just patching the symptom.

A few things I particularly like: Catching the packed-buffer assumption too (the real per-plane strides fix is its own bug, separate from the color-range one, and it's easy to miss), converting the panic path to a proper DecoderError, and having the regression test round-trip through a live openh264 encoder/decoder pair instead of asserting against hand-built bytes.

The body already flags the encoder-side half (#1924) as a known follow-up, which is the right way to scope this. I'm just flagging it again here in case it helps: it might be worth keeping that note prominent in the body as this PR moves toward merge, since landing the decoder half alone temporarily makes any full AVC420 round trip (any IronRDP server talking to this now-fixed decoder) look worse instead of better, until the encoder side lands too. This isn't a blocker on this PR. It's just context for anyone reviewing.

ciron-nine:
The encoder-side half is ready: #1924 now has a fix stacked on this PR (branch `fix/egfx-avc420-encoder-bt709`, PR to follow or already open).

   It mirrors this change: the encoder converts RGBA→YUV420 with `yuv::rgba_to_yuv420(YuvRange::Full, YuvStandardMatrix::Bt709)` instead of openh264's limited-range internal conversion, hands the planes to openh264 as pixel data, and adds round-trip tests through the live encoder/decoder pair — including a cold-channel leakage bound that discriminates a full-range BT.601 matrix regression, which hot-channel assertions alone cannot detect.

   Until it lands, full AVC420 round trips between IronRDP servers and this fixed decoder render worse than before, since the encoder's limited-range output no longer cancels against the old decoder. Reviewing/merging the two together is worth considering.

   > [!NOTE]

glamberson:
Following up now that the encoder side (#1925) has landed: It uses the same yuv-crate full-range BT.709 approach as this PR, so both halves of the round trip are consistent with each other and with the spec. Investigating this also surfaced the same defect in our own downstream encoder; we're fixing it now using the same pattern. Thanks again for catching this.

CBenoit (commented):
I think you need to update the Cargo.lock file to make CI green, and address the Copilot findings if relevant. Otherwise, this is looking good to me, thank you!

ciron-nine:
Context on how I ran into this: I was using IronRDP as the RDP client library against a Windows server. After connecting, I right-clicked the desktop to open the context menu, and hovering over menu items produced no highlight — the items simply stayed the same color as the background.

   Digging in, the symptom pointed at the color pipeline rather than the renderer: the menu highlight color is a very light gray (luma close to white), and under the incorrect conversion it got expanded past 255 and clamped into the same value as the pure-white background, so the highlight became indistinguishable. That is the classic signature of running a full-range stream through a limited-range conversion, and following that thread led to the conflict between MS-RDPEGFX 3.3.8.3.1 and openh264's built-in matrix.

   I'm adding this here mostly so it's searchable: if anyone else hits "hover highlights disappear / gradients band / whites look blown out" against a Windows server, this is very likely the same root cause.

ciron-nine:
Side-by-side of the actual symptom on a Windows server: with the fix, hovering menu items shows the light-gray highlight (left); before the fix, the highlight color was clamped into the white background and the hover state became invisible (right).
 
rdp-hover-comparison

CBenoit (approved):
Good catch! Thank you for the write-up, and side-by-side comparison; it’s really helpful.

PR #1925, opened 2026-09-08
fix(egfx): encode AVC420 with full-range BT.709

### LINKED ISSUE #1924: fix(egfx): encode AVC420 with full-range BT.709

ciron-nine:
Replaces openh264's built-in RGBA-to-YUV420 conversion in the AVC420 encoder path with an explicit full-range BT.709 conversion from the yuv crate, completing the decoder-side fix in #1923.

   MS-RDPEGFX 3.3.8.3.1 mandates full-range BT.709 for RFX_AVC420_BITMAP_STREAM color conversion, but openh264's YUVBuffer::from_rgb_source hardcodes a limited-range (studio swing) BT.601 matrix internally (luma offset +16, coefficients 66/129/25 >> 8), placing luma in [16, 235] instead of [0, 255]. The decoder fix makes the mismatch visible: an IronRDP server talking to the now-conformant decoder produces washed-out, crushed output where the two errors previously partially cancelled.

   The conversion now runs rgba_to_yuv420 with YuvRange::Full and YuvStandardMatrix::Bt709 into planes allocated by YuvPlanarImageMut::alloc, and the converted planes are handed to openh264 as pixel data via YUVSlices, so openh264 no longer touches color. Conversion failures surface as EncoderError instead of an internal panic.

   Two round-trip regression tests encode solid colors through the live encoder/decoder pair: hot-channel bounds fail under the old encoder (white decodes to 235), and a cold-channel leakage bound discriminates a full-range BT.601 matrix
 regression, which the hot-channel test alone cannot detect.

   Stacked on #1923 (decoder side) and should land shortly after it: until the encoder half lands, full AVC420 round trips between IronRDP servers and the fixed decoder look worse than before, since the errors no longer cancel.

   Fixes: #1924

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2947,6 +2947,7 @@ dependencies = [
  "ironrdp-pdu",
  "openh264",
  "tracing",
+ "yuv",
 ]
 
 [[package]]
--- a/crates/ironrdp-egfx/Cargo.toml
+++ b/crates/ironrdp-egfx/Cargo.toml
@@ -25,10 +25,11 @@ ironrdp-graphics = { path = "../ironrdp-graphics", version = "0.9" } # public
 ironrdp-pdu = { path = "../ironrdp-pdu", version = "0.9" } # public
 openh264 = { version = "0.9", optional = true, default-features = false }
 tracing = { version = "0.1", features = ["log"] }
+yuv = { version = "0.8", default-features = false, optional = true }
 
 [features]
 arbitrary = ["dep:arbitrary", "bitflags/arbitrary", "ironrdp-pdu/arbitrary"]
-openh264 = ["dep:openh264"]
+openh264 = ["dep:openh264", "dep:yuv"]
 openh264-bundled = ["openh264", "openh264/source"]
 openh264-libloading = ["openh264", "openh264/libloading"]
 
--- a/crates/ironrdp-egfx/src/decode.rs
+++ b/crates/ironrdp-egfx/src/decode.rs
@@ -183,6 +183,7 @@ pub trait H264Decoder: Send {
 
 #[cfg(feature = "openh264")]
 mod openh264_impl {
+    use openh264::formats::YUVSource;
     use tracing::warn;
 
     use super::{DecodedFrame, DecoderError, DecoderResult, H264Decoder};
@@ -253,7 +254,8 @@ mod openh264_impl {
                 .map_err(|e| DecoderError::new("OpenH264 decode failed", e))?
                 .ok_or_else(|| DecoderError::msg("OpenH264 returned no picture"))?;
 
-            let (width, height) = openh264::formats::YUVSource::dimensions(&yuv);
+            let (width, height) = YUVSource::dimensions(&yuv);
+            let (y_stride, u_stride, v_stride) = YUVSource::strides(&yuv);
 
             #[expect(
                 clippy::as_conversions,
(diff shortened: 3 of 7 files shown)


glamberson (commented):
Really solid follow-through, thank you for closing the loop on this so fast. Verified the forward-transform claim directly against MS-RDPEGFX 3.3.8.3.1 (full-range BT.709, clamped 0..255) and it matches. The `yuv::rgba_to_yuv420` call with `YuvRange::Full, YuvStandardMatrix::Bt709` is the correct mirror of #1923's decoder-side fix, and handing openh264 the converted planes directly via `YUVSlices` means it never touches color at all now, which is the right way to close this off for good.

The two new tests are the right shape too: The hot-channel bounds catch the old limited-range behavior, and the cold-channel leakage bound specifically catches a residual BT.601-matrix regression that the hot-channel test alone couldn't see.

This also caught something on our end: Our own downstream encoder had the identical defect (same root cause, openh264's internal conversion), and we're fixing it now with the same approach. Appreciate the clean, well-tested reference to work from.

ciron-nine:
@CBenoit
This is not a duplicate of #1923: #1923 covers YUV to RGBA in `decode.rs`, whereas this implements RGBA to YUV in `encode.rs`. The conversion directions are reversed and they reside in separate files.
Please re-review this PR and merge it into the main branch.

PR #1927, opened 2026-09-08
fix(rdpeusb): stop releasing the device sink interface

uchouT:
mstsc closes the per-device channel right after ADD_DEVICE when the server releases it, aborting every redirected device. A Windows server never sends this release.

So we align with windows server.


CBenoit (approved):
LGTM! Feel free to open a follow up PR if necessary for the AI finding 🙂

PR #1930, opened 2026-09-09
fix(web): report invalid image extraction regions

CBenoit:
- Return a contextual error when the web renderer receives a malformed or out-of-image extraction region, rather than panicking while slicing the framebuffer.
- Propagate extraction failures before drawing while preserving `WriteBuf` reuse, cursor commit behavior, valid output rectangles, and the small-rectangle versus whole-row heuristic.
- Run the regression coverage through the centralized extra testsuite without changing Cargo test flags.

Current graphics producers are expected to supply valid regions, so this is defensive error handling for an invariant violation, not a demonstrated reachable protocol crash or a general panic-freedom guarantee.

This keeps the useful error-propagation direction from #993 and the [review decision](https://github.com/Devolutions/IronRDP/pull/993#issuecomment-5592549297), with thanks to @AlexYusiuk for the original contribution. It deliberately excludes the global `clippy::indexing_slicing` rollout and changes to already-safe canvas pixel indexing or internally-sized destination slices.

### DIFF
--- a/crates/ironrdp-testsuite-extra/Cargo.toml
+++ b/crates/ironrdp-testsuite-extra/Cargo.toml
@@ -24,7 +24,7 @@ harness = true
 [dev-dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
-ironrdp = { path = "../ironrdp", features = ["server", "pdu", "connector", "session", "svc", "dvc", "echo"] }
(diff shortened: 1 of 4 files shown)


CBenoit:
Closing this focused follow-up after reassessing the current producer chain. Bitmap, pointer, and EGFX paths already validate or clip regions before emitting `GraphicsUpdate` values. The change would add renderer-side errors and test wiring only to re-check an internal invariant, without a demonstrated reachable failure. That defensive benefit does not justify the added noise here.

> [!NOTE]

PR #1956, opened 2026-09-11
ci(pr-automation): clarify review repair failures

CBenoit:
General-review failures combine vague validation errors with repair instructions, making it difficult to diagnose rejected output or correct it within the repair budget.

Derive factual diagnostics in the authoritative final-review validator and report disposition-map problems together, including missing, unknown, or duplicate candidates and unusable rationales. Identify candidates by trusted aggregate positions without echoing model-authored text, and keep reasons within 230 bytes so runtime telemetry and terminal errors preserve them intact.

Clarify which specialist findings require dispositions and cover the repair path through the real runtime with a mock provider. Acceptance rules, normalization, reviewer budgets, and workflow gates remain unchanged.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.github/PR_AUTOMATION.md
+++ b/.github/PR_AUTOMATION.md
@@ -46,6 +46,7 @@ Protocol candidates also carry structured protocol references.
 One specialist never receives another specialist's output.
 
 The general reviewer independently inspects the pull request, attempts to falsify every candidate, and records exactly one `accepted`, `refined`, or `rejected` disposition per candidate.
+A candidate is one entry in the findings of a reviewer the aggregate reports as valid, so a reviewer that failed or reported nothing contributes none.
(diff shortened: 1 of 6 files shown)


CBenoit:
> Duplicate detection only consults `byCandidate`, which is populated after the rationale passes `normalizeText`.

Fixed in e4c17e8b: known references are tracked independently, so duplicate entries and invalid rationales are reported together regardless of entry order.
The diagnostics now distinguish forbidden controls from accepted whitespace controls and name the title and rationale byte limits explicitly.
Added regression coverage for both entry orders, two invalid rationales, accepted whitespace normalization, and correction through the real runtime with a mock provider.

> [!NOTE]

PR #1966, opened 2026-09-14
fix(pdu): add missing error info codes

zmb3:
We were missing enum variants for the shutdown and reboot error codes: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/a21a1bd9-2303-49c1-90ec-3932435c248c


CBenoit (approved):
Thank you!

PR #1970, opened 2026-09-16
feat(web): support WebSocket subprotocols

MuNeNiCK:
## Motivation

WebSocket subprotocol negotiation is a standard part of the opening handshake. Endpoints commonly use it for application protocol selection, version negotiation, and handshake authentication, especially in browser clients that cannot attach arbitrary request headers.

`ironrdp-web` currently creates its WebSocket internally with only the proxy URL, so consumers cannot connect to endpoints that require one or more subprotocols.

## Changes

- expose a `webSocketProtocols()` configuration extension from the RDP web package
- accept and validate a JavaScript string array in `ironrdp-web`
- open the connection with `WebSocket::open_with_protocols` when protocols are configured
- preserve the existing `WebSocket::open` path when the option is omitted or empty
- document the new configuration helper

## Validation

- `cargo xtask wasm check -v`
- `cargo xtask web check -v`
- `cargo xtask check fmt -v`
- `cargo xtask check locks -v`


crates/ironrdp-web/Cargo.toml
```diff
@@ -61,7 +61,7 @@ web-sys = { version = "0.3", features = [
     "Window",
 ] }
 js-sys = "0.3"
-gloo-net = { version = "0.7", default-features = false, features = ["websocket", "http", "io-util"] }
+gloo-net = { version = "0.7", default-features = false, features = ["websocket", "http", "io-util", "json"] }
```
CBenoit: question: Why json feature is necessary now?

CBenoit (commented):
Hi! Thank you for the PR.

I think this is something I would like to have in the main "iron-remote-desktop" API, not as an extension. Could you change that before we merge?

Thank you!

MuNeNiCK:
Thank you! I moved the option into the main `iron-remote-desktop` API. Consumers now configure it with `ConfigBuilder.withWebSocketProtocols()`, and the value flows through the common `SessionBuilder.webSocketProtocols()` API. The RDP-specific extension was removed, while an empty list still preserves the existing connection path.

crates/ironrdp-web/Cargo.toml
MuNeNiCK: The `json` feature was required only because `gloo-net::WebSocket::open_with_protocols` serializes the protocol slice through Serde. I changed the implementation to construct the browser WebSocket with `web_sys::WebSocket::new_with_str_sequence` and then wrap it with the existing gloo WebSocket `TryFrom` implementation. The `json`, `serde`, and `serde_json` dependencies are no longer added.

web-client/iron-remote-desktop-rdp/README.md
```diff
@@ -22,6 +22,17 @@ Otherwise, you can run `npm install` targeting the `dist/` folder directly.
 
 Import the `iron-remote-desktop-rdp.umd.cjs` from `node_modules/` folder.
 
+## WebSocket subprotocols
+
+WebSocket endpoints use subprotocol negotiation for application protocol
+selection, versioning, and handshake authentication. Pass
+`webSocketProtocols()` to the configuration builder when the endpoint requires
+one or more subprotocols during the opening handshake:
```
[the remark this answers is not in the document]
MuNeNiCK: Fixed. The documentation is now in the main `iron-remote-desktop` README, and each sentence is kept on one source line.

PR #1977, opened 2026-09-18
fix(egfx): stop the desktop from tearing when the server resizes graphics output

asjdf:
Fixes screen tearing / ghosting after the server changes the graphics output size.

Rebased onto current `master`, which already resizes the session framebuffer on `ResetGraphics` via `take_output_reset` and [`reset_preserving_pointer`](https://github.com/Devolutions/IronRDP/blob/b149f50/crates/ironrdp-session/src/image.rs#L204) before compositor deltas are applied. **This PR does not change `active_stage.rs`.** It fixes the EGFX/graphics decode path and `ironrdp-web` so the pixels written into that buffer are correct.

## The bug

When the server resizes the graphics output (EGFX `ResetGraphics`, or a Deactivation-Reactivation Sequence), stale pixels survive the transition. What you see is part of the old desktop still on screen, or blocks that never repaint until something happens to overdraw them.

Several independent faults contribute; any one of them can produce the symptom alone:

1. **`ResetGraphics` dropped the wrong caches.** The bitmap cache was cleared, but MS-RDPEGFX 3.3.5.14 only redefines the output buffer — cache slots are connection-scoped. Windows restores the desktop after a resolution change almost entirely from slots filled *before* the reset, so dropping them makes hundreds of `CacheToSurface` blits silent no-ops.
2. **Progressive tile references outlived the surfaces they belonged to.** A later surface reusing the same id would difference against the previous desktop.
3. **SRL streams were decoded incorrectly.** Zero-padding was not honoured and a zero run spanning a single event was over-consumed, so a run could terminate early and leave the tail of a tile unwritten — or the session aborted with `Srl(MissingTerminator)` on real Windows hosts.
4. **Session framebuffer sizing (already on upstream `master`).** Deltas for the new geometry must land in a buffer already sized for the new output. Upstream now does this with `take_output_reset` + `reset_preserving_pointer` before `composite_graphics_updates`. This PR assumes that base; it does not reimplement it.

On top of that, the web client never opted into MS-RDPEGFX, and its canvas never followed a size the server picked on its own — so even with the protocol side correct, `ironrdp-web` still tore.

## Where to look first

Most of the diff is not the fix. `examples/rdp_stress.rs` is 1299 lines of offline grading harness with no product code in it, `ironrdp-testsuite-core` adds integration tests, and roughly a third of the remaining product diff is comment recording spec reasoning. The behavioural changes in **this** PR are small; the table is everything that still differs from `master`.

| # | Where | The change | Why that fixes the tear |
|---|---|---|---|
| 1 | [`compositor.rs::reset`](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-egfx/src/compositor.rs#L182) | stop clearing the bitmap cache on reset; charge `allocated_bytes` to surviving slots | Windows repaints a resized desktop almost entirely with `CacheToSurface` from slots filled *before* the reset. Clearing them turns hundreds of those blits into silent no-ops, so the old pixels stay where they were. |
| 2 | [`client.rs::handle_reset_graphics`](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-egfx/src/client.rs#L700) and [`progressive.rs::clear_tile_references`](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-graphics/src/progressive.rs#L1572) | clear progressive tile references on reset; keep CONTEXT and the ClearCodec glyph cache | A reset implicitly destroys every surface. Stale coefficient buffers make a later surface reusing the same id difference against the previous desktop. CONTEXT stays because 3.3.5.14 redefines only the output buffer and Windows does not re-send SYNC. |
| 3 | [`srl.rs`](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-graphics/src/srl.rs#L59) | honour zero-padding; consume zero runs one event at a time | Without this, many real servers never get past the first EGFX frames (`Srl(MissingTerminator)`), and static tiles never refresh when streams end early. |
| 4 | [`ironrdp-web` EGFX enable](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-web/src/session.rs#L1608) | `support_dyn_vc_gfx_protocol: true` and register the EGFX DVC | The browser client must enter the same pipeline these fixes live in. |
| 5 | [`sync_canvas_to_image`](https://github.com/juanjiTech/IronRDP/blob/51e956207acd616f2d92b40080fd9df6b194f26f/crates/ironrdp-web/src/session.rs#L1507) | match the HTML canvas to `DecodedImage` before drawing; full repaint when the backing store size changes | `ActiveStage::process` can resize the image in the same frame; assigning canvas `width`/`height` clears it. Dirty-rectangle painting alone would leave the rest blank. |

Rows 1–3 are protocol/decode faults; rows 4–5 are web embedding. Each of 1–3 can produce visible tearing or a dead session on its own.

## What changed

**Protocol (`ironrdp-egfx`, `ironrdp-graphics`) — this diff**
- `ResetGraphics` retains the bitmap cache, keeps the progressive CONTEXT and the ClearCodec glyph cache, and clears progressive tile references.
- Deleting a surface drops that surface's progressive tile references, so a difference tile before a new base tile fails cleanly with `MissingTileReference`.
- SRL decoding honours zero-padding and consumes a zero run one event at a time; erroneous terminator and eager cap logic is gone.
- `SessionErrorKind::Pdu` carries the inner error; `ProgressiveDecodeError` implements `core::error::Error`.

**Session (`ironrdp-session`) — upstream base, not modified here**
- Framebuffer resize on reset: `take_output_reset` → `reset_preserving_pointer` → `composite_graphics_updates`, plus output-dimension limits via `Compositor::materializable_output_size`.

**Web (`ironrdp-web`) — this diff**
- Enable the graphics pipeline and register the EGFX DVC.
- Sync the canvas backing store to `DecodedImage` before the frame's `GraphicsUpdate` is drawn.
- Repaint the full image when the canvas backing store actually changed; sync after reactivation when the image is replaced.
- `Canvas::resize` reports whether the size changed; `Session::desktop_size()` tracks the size in effect.

Resize *requests* deliberately do not touch the canvas: it follows the size the server actually applies. Notably `SuppressOutput`/`RefreshRect` are not sent on reset — with RDPGFX those PDUs do not invalidate the surface cache ([FreeRDP#12723](https://github.com/FreeRDP/FreeRDP/issues/12723)). `examples/rdp_stress.rs --redraw` exists to re-check that against a real server.

## Testing

| Area | Test |
|---|---|
| bitmap cache survives reset | `compositor::tests::reset_keeps_the_bitmap_cache` |
| stale progressive tile references | `client::tests::reset_graphics_drops_tile_references_of_implicitly_destroyed_surfaces` |
| reset + compositing integration | `session::active_stage::egfx_reset_resizes_the_image_before_compositing_the_same_payload` (full `ActiveStage` + EGFX; fill outside the pre-reset image, inside the new one) |
| framebuffer + pointer on reset | `session::active_stage::reset_graphics_preserves_software_pointer_state` (upstream; unchanged by this PR) |
| SRL decode | `progressive::tests::upgrade_pass_completes_when_the_srl_stream_ends_early`, `tile_upgrade_keeps_all_components_on_srl_error` |

Integration tests live in `ironrdp-testsuite-core` because `ironrdp-session` sets `[lib] test = false`.

### Grading a real session

`examples/rdp_stress.rs` drives resolution changes against a live server and grades the decoded framebuffer (black / stale / seam metrics). Black *and* stale count toward failure so a fully painted but ghosted resize does not pass.

```shell
cargo run --example=rdp_stress --features "session,connector,graphics,dvc,displaycontrol" -- \
    --host <host> -u <user> --rounds 10 --out-dir /tmp/rdp-stress
```

On a Windows 11 host with RDPGFX, cycling `1300x820` ↔ `1828x1004`, 5 rounds:

| Build | Result |
|---|---|
| this branch (on current `master`) | 9/9 real resizes at `stale_tiles=0.00%`; frame size matches the requested dimensions |
| `master` without this PR | opening EGFX frames abort with `Srl(MissingTerminator)` (fault 3), so resize grading never runs |

The harness connects at the first `--sizes` entry, so round 1's opening step can show high stale % on an idle desktop (no server repaint) — a metric artefact, not leftover tearing.


uchouT:
Sorry for off-topic here. It's surprise to see my college alumni here :)

PR #1979, opened 2026-09-18
fix(server): drive clipboard channel timeouts

pierophp:
## Problem

`Cliprdr::drive_timeouts` is documented as something the embedder must call from a periodic timer:

> Callers must drive this method from a periodic timer in their event loop (e.g., every 5 seconds). The returned PDUs must be sent on the CLIPRDR channel.

`ironrdp-client` (`crates/ironrdp-client/src/rdp.rs`) and `ironrdp-web` (`crates/ironrdp-web/src/session.rs`) both do. **`ironrdp-server` never calls it from anywhere.**

On the server role that means the three sweeps inside `drive_timeouts` never run:

1. **Outgoing locks are never released.** `send_lock` creates a lock when a file `FormatList` arrives (2.2.4.1). A later `FormatList` moves it to `Expired` — deliberately without sending `Unlock`, since downloads from the previous clipboard may still be in flight. The `Unlock` is left to the sweep. With no sweep, it never happens: `outgoing_locks` grows monotonically until `MAX_OUTGOING_LOCKS` (100), after which `send_lock` logs `"Too many outgoing locks, skipping automatic lock"` and silently stops locking for the rest of the session.
2. **Abandoned file contents requests are never failed.** Backends keep waiting on a response that cannot arrive.
3. **Locked file list snapshots are never dropped.**

## Change

Adds a `drive_cliprdr_timeouts` arm to `client_loop`, mirroring the existing `send_heartbeats` and `refresh_auto_reconnect_cookie` timers:

- `CLIPRDR_DRIVE_TIMEOUTS_INTERVAL = 5s`, matching the interval the docs suggest.
- `MissedTickBehavior::Skip`, since a stalled write should not produce a burst of back-to-back sweeps for work that is idempotent.
- A sweep failure is logged and swallowed. Best-effort maintenance must not disconnect an otherwise healthy session.

**What a backend will start seeing.** Driving the sweep on the server also activates the two other sweeps inside `drive_timeouts`, which until now only ran for the client and web roles. Both are governed by the same `transfer_timeout` (60s by default):

- a `FileContentsRequest` we sent that stays pending for the timeout is dropped, and the backend gets a synthetic error through `on_file_contents_response(FileContentsResponse::new_error(stream_id))`;
- a locked file list snapshot with no `FileContentsRequest` activity for the timeout is dropped, and the backend gets `on_unlock(LockDataId(..))`.

Neither is new code — both are existing `drive_timeouts` sweeps that the server simply never reached. But a server `CliprdrBackend` that treated `on_unlock` as "the peer sent `UnlockData`" will now also see it fire on a timeout. That is the intended behaviour, and the point of it (releasing file handles held for a transfer the peer abandoned), but it is worth calling out for backend implementors.

## Tests

Two tests in `crates/ironrdp-server/src/server.rs`:

- `an_expired_lock_is_released_when_the_sweep_is_driven` — negotiates `CAN_LOCK_CLIPDATA`, feeds a file `FormatList` (asserting the `LockData` PDU goes out), expires the lock with a text `FormatList`, then drives the sweep across a controllable `CliprdrBackend::now_ms` clock: nothing before the inactivity timeout, an `Unlock` PDU after it.
- `a_session_without_a_clipboard_channel_sweeps_quietly` — the timer fires for every connection, so a session with no clipboard channel must not be an error path.

**Scope note:** the tests cover `drive_cliprdr_timeouts` itself, not the `tokio::select!` arm that calls it — driving that would need a harness that reaches the active stage, which this crate does not currently have (`finalize_timeout.rs` stops at the finalize handshake). The arm is six lines mirroring two existing timers.


glamberson:
Thanks for this, it's a good catch and a clean fix. I checked the premise against master and it holds. ironrdp-client (rdp.rs) and ironrdp-web (session.rs) both call drive_timeouts on a timer and ironrdp-server never does, so on the server role locks are never released and the 100-lock cap does silently end automatic locking. This affects me directly, since lamco-rdp-server is a long-lived server with file transfer in both directions, so I'm glad to see it fixed.

I also merged the branch with current master and ran the two new tests and the fmt and lints gates, and everything passes. I liked that the expired-lock test drives a controllable clock instead of sleeping, and it's a real test too. When I turn drive_cliprdr_timeouts into a no-op, it fails.

One suggestion for the description. Driving the sweep on the server also turns on two behaviors that already exist for the client and web roles. A file contents request pending for 60 seconds now gets a synthetic error response, and a locked file list snapshot with no requests for 60 seconds is dropped with an on_unlock call to the backend. I think both are right for the server too, but a backend will start seeing them, so a line about it would help reviewers.

A heads-up that it conflicts textually with my #1954 in client_loop, since each adds a writer clone and a select! arm. Keeping both resolves it, so whichever lands second needs a small rebase.

pierophp:
Thanks for taking the time to verify this independently — especially the no-op mutation. That is the part I could not easily demonstrate myself, and it is the right check for a test whose subject is "something now happens on a timer".

Good call on the two extra behaviours; I have added a paragraph to the description. For the record, both are governed by the same `transfer_timeout` (60s by default): a `FileContentsRequest` pending past it is dropped with a synthetic `FileContentsResponse::new_error` delivered through `on_file_contents_response`, and a locked file list snapshot with no request activity past it is dropped with `on_unlock`. Neither is new code, but as you say, a server backend that read `on_unlock` as "the peer sent `UnlockData`" will now also see it fire on a timeout, so it belongs in the description rather than only in the cliprdr docs. I also put the same note on the `CLIPRDR_DRIVE_TIMEOUTS_INTERVAL` doc comment (9eca237), since that is where someone writing a server backend is most likely to look.

On #1954: agreed, it is the writer clone plus the `select!` arm, and keeping both is the whole resolution. Since #1954 is stacked on #1953 and #1951, this one will probably land first; if it lands the other way around I am happy to rebase.

PR #1981, opened 2026-09-18
fix(cliprdr): validate locked file requests against the list their lock covers

pierophp:
> **Stacked on #1980.** GitHub cannot base a cross-fork PR on another fork branch, so the diff here also carries #1980's commit. The change under review is the second commit, `fix(cliprdr): validate locked file requests against the list their lock covers`. Rebasing onto `master` once #1980 lands makes it standalone.

## Problem

[MS-RDPECLIP] 3.1.5.4.5:

> If the clipDataId field is present, then the locked File Stream data associated with the ID MUST be used to service the request.

**The sender side already honours this.** A Lock PDU received from the remote snapshots `local_file_list` into `locked_file_lists`, and an incoming `FileContentsRequest` carrying a `clipDataId` is served from that snapshot — with the spec text quoted inline at the use site.

**The receiver side has no equivalent.** A Lock PDU *we* send asks the remote to keep its File Stream data alive across a clipboard change. `request_file_contents` even documents the contract:

> Caller can override by setting `request.data_id` explicitly before calling this method. This allows the receiver to correlate the request with locked File Stream data **even if the clipboard has changed since the lock was sent.**

…and twelve lines later validates the index against `remote_file_list`, which `handle_format_list` clears on every clipboard change and the next `FormatDataResponse` overwrites:

```rust
if let Some(ref file_list) = self.remote_file_list {
    if file_list.files.len() <= validated_file_index {
        return Err(/* "file index out of bounds for remote file list" */);
```

So: lock a two-file selection, let the remote replace it with a one-file one, and an already-issued request for index 1 fails — even though the lock is precisely the promise that the remote still holds that data. `handle_format_list`'s own comment says this is not the intent:

```rust
// Transition active locks to Expired — but do NOT send Unlock PDUs yet.
// Active downloads from the previous clipboard may still be using these locks.
```

With a server embedder that surfaces the error, that failure reaches the session.

## Change

Mirrors `locked_file_lists` for the receiving direction:

- `locked_remote_file_lists: HashMap<u32, PackedFileList>` — the remote list snapshotted under whatever lock was active when it arrived, at the existing `on_remote_file_list` call site.
- Bounds and range validation consult that snapshot when the request carries a `data_id`, falling back to `remote_file_list` otherwise.
- The snapshot is released with its lock, in both the `drive_timeouts` sweep and `release_outgoing_locks`.
- Bounded by the same `MAX_LOCKED_FILE_LISTS` as the sender side.

`handle_format_list` is untouched: it still clears `remote_file_list`, and the snapshots deliberately outlive it.

## Tests

In `crates/ironrdp-testsuite-core/tests/clipboard/lock_lifecycle.rs`:

- `a_locked_request_validates_against_the_list_its_lock_covers` — drives two selections through the real protocol flow (`FormatList` → `initiate_paste` → `FormatDataResponse`), then requests A's second file under A's `clipDataId`. **Fails on `master`** with "file index out of bounds for remote file list"; passes here. The same index without a lock is still rejected against the current selection.
- `a_snapshot_is_released_with_the_lock_it_belongs_to` — after the sweep sends `Unlock`, the snapshot must stop validating requests, since the remote has released the data.

All 213 clipboard tests pass.


glamberson:
Thanks for this, it's careful work. I like that it grounds the change in the spec's clipDataId rule, mirrors locked_file_lists on the receiving side, and files the snapshot under the same id that on_remote_file_list hands the backend, which makes the two consistent by construction.

I ran the two new tests and the fmt and lints gates on the branch and they pass. The main test fails when I take the change out, as the description says, and the release test fails when I stop removing the snapshot on Unlock, so both are doing real work. I also looked for two ways the keying could go wrong and didn't find either. A late FormatDataResponse can't be filed under a newer lock, because a new Format List clears pending_format_data_request and remote_file_list_format_id. And at the 100-lock cap send_lock returns early, but expire_all_locks has already cleared current_lock_id, so no snapshot is taken.

This matters for ChunkedFetch (#1742) too. It sends the clip_data_id from on_remote_file_list on every request so a transfer can outlive a clipboard change, and on master those requests are still validated against the current list. This is the piece that makes that design work.

Two small suggestions. The Validation section of the request_file_contents docs only says the index must come from a prior file list exchange. A sentence saying that a request carrying a clipDataId is checked against the list its lock covers, and falls back to the current list when there's no snapshot, would tell callers what to expect and record that the fallback is deliberate. And a small citation note. In the current spec the sentence about servicing a request from the locked File Stream data is in 3.1.5.4.6, Processing a File Contents Request PDU, while 3.1.5.4.5 is the requesting side, where the rule that the index comes from a File List lives. Citing both would be the most accurate.

pierophp:
Thank you for the review — especially for re-running it both ways. Knowing that the main test fails without the change and the release test fails without the snapshot cleanup is worth more than my own word for it, and the two keying paths you checked are exactly the ones I'd want a second pair of eyes on. Good point about ChunkedFetch (#1742); I hadn't connected that it depends on this to work as designed.

Both suggestions applied in `78eb0666`.

**Validation docs.** The `## Validation` section of `request_file_contents` now says which list each check runs against:

> Which file list those checks run against depends on `request.data_id`. [MS-RDPECLIP] 3.1.5.4.6 has the remote service a request carrying a `clipDataId` from the locked File Stream data, so such a request is validated against the list that lock covers — not the current remote clipboard, which a new Format List may already have replaced. Without a `clipDataId`, or when the lock has since been released and its snapshot dropped, validation deliberately falls back to the current list: that is the only list the remote can still serve from.

I added the reason for the fallback as well as the fact of it, so it reads as a decision rather than a gap.

**Citation.** You're right, and I confirmed it against the published TOC: 3.1.5.4.5 is *Sending* a File Contents Request PDU, where "This index can be obtained through a File List" and the range-within-file-size rule live; 3.1.5.4.6 is *Processing* a File Contents Request PDU, which is where "If the **clipDataId** field is present, then the locked File Stream data associated with the ID MUST be used to service the request" actually appears. Worse, the repo already had it right — the sender-side path at `lib.rs:1834` cites 3.1.5.4.6 for that same sentence, so this change had gone and contradicted its own neighbour.

Fixed in the four places where the quoted rule is the processing-side one: the `locked_remote_file_lists` field doc, the list-selection comment in `request_file_contents`, the snapshot comment at the `on_remote_file_list` call site, and the test doc. The existing 3.1.5.4.5 citations on the index-from-File-List and range-within-bounds checks are untouched, since those are correct as they stand.

In the two places that carry the full argument — the field doc and the test doc — I cite both, as you suggested, because the argument genuinely needs both halves: 3.1.5.4.6 says the remote services the request from the locked data, 3.1.5.4.5 says the index comes from a File List, therefore validation has to use the File List that locked data arrived with. The commit message is rewritten the same way; the PR description above still has the old citation and I'll fix it on the rebase onto `master` once #1980 lands.

fmt, clippy on `ironrdp-cliprdr`, and all 213 clipboard tests pass.

